Issue #18, opened 2026-05-29
Path/package-scoped visibility — make "private" a property of a subtree, not a whole repo


beardthelion:
## The pitch

The README's roadmap already names whole-repo private-read enforcement:

> "Implement private-read enforcement or remove private repo affordances until it exists."
> "Implement repo write authorization: owner checks, protected branches, and UCAN capability checks."

I want to aim higher than the binary. "Private repo" is what every forge already does. The thing gitlawb is uniquely positioned to do — because identity is a DID and authorization is meant to be capability-based — is make **visibility a per-path property**: one repo, mostly public, with specific subtrees readable only by specific DIDs. Open source stops meaning all-or-nothing.

If you build whole-repo private as a one-off boolean now, path-scoping later is a data-model migration. If the visibility model is scoped from day one, whole-repo private is just the trivial case where the scope is `/`. Same work, strictly more capable, no migration. That's the argument for designing the ambitious version even if the first shipping slice is small.

## What this unlocks

- **Monorepo, one private package.** Public packages discoverable and forkable; a proprietary or in-progress package gated to a DID set — without splitting the repo and shattering shared history.
- **Hidden in-flight work.** Peers mirror a push within ~30s today. A path-scoped visibility rule is the mechanism for "this change is real but not network-visible yet."
- **Coordinated security disclosure.** A fix lands in a gated path, opens to the world on disclosure day — same repo, same history, no out-of-band private fork.

This is the same all-or-nothing critique of git that's been going around lately (Theo's May 2026 source-control thread, which openly invited people to run with it) — but gitlawb has the identity and delegation substrate to actually express it, where git can't.

## Honest read on what exists today

I've read the README; not pretending the pieces are done:

> "Private repository read enforcement is not wired yet."
> "UCAN chain validation and revocation are not complete."
> "HTTP signatures prove identity, not full authorization policy."

So the substrate is **identity that works** (signed requests prove the requesting DID) and a **token format that exists** (`gl ucan`), with enforcement and chain-validation still ahead. That's fine — it means the design isn't boxed in yet, which is exactly when to get the visibility model right. I also see private-read (short-term #6) is sequenced after write authorization (#5); this proposal respects that ordering — read enforcement builds on the same authorization checks.

## Approach

1. **Visibility as a first-class, path-scoped concept** in the repo's authorization model — a rule binds a path glob to a reader set (DIDs, later UCAN capabilities). Whole-repo private = scope `/`.
[...]

PR #24, opened 2026-06-01
Demo (not for merge): is_public enforcement on the git read path — grounds the #18 design decisions

beardthelion:
**Not a merge candidate — a working demonstration to make the open design decisions in #18 concrete instead of theoretical.** It wires the `is_public` flag the node already stores but never checks on the git read path, and runs end to end against a local node. Please read the "What this does NOT do" section before reviewing as a feature.

## What works

`authorize_read` gates the git smart-HTTP read path (`info/refs?service=git-upload-pack` and `git-upload-pack`) on the DID that signed the request (verified upstream by `optional_signature`). Unauthorized reads return **404, not 403**, byte-identical to a missing repo.

| Caller | Repo | Result |
|--------|------|--------|
| anonymous | private | `404` (identical to missing repo) |
| owner DID | private | `200` |
| wrong DID | private | `404` |
| anyone | public | `200` (unchanged) |

A 6-case truth-table unit test pins this, including an assertion that the private-denial payload is indistinguishable from a missing repo. (`cargo fmt`/`clippy` clean, rebased on `main`, no conflicts.)

## What this does NOT do (why it is a demo, not a feature)

These are deliberate, and several are genuinely maintainer decisions, not things I should pick unilaterally:

1. **Other read surfaces are still open.** `authorize_read` is wired into the two HTTP handlers only. The IPFS read path (`api/ipfs.rs`, `ipfs_pin.rs`) and gossip replication still serve the objects ungated. So a repo marked private here is **still readable** through those paths. This is the single biggest reason it must not be mistaken for real privacy, and closing those surfaces is the same decision as the replication question below.
2. **Owner-only, no reader sets.** Authorization is `caller_did == owner_did`. There is no way to grant another DID read access, so this expresses "private to me alone," not the capability-based reader sets #18 is actually about.
3. **Whole-repo boolean, not the scoped model.** This enforces the existing `is_public` boolean; it does not add a path-scoped data model. #18 argues the data model should carry path scope from day one to avoid a later migration — this demo deliberately does not, pending your call on that.
4. **No way to create/flip a private repo via `gl`.** The create API accepts `is_public:false`, but every `gl` path hard-codes `true` and there is no endpoint to change visibility after creation. The demo was exercised by hand-crafting the create call.

## The decision that unblocks the real feature

@kevincodex1 the fork that gates everything (path-scoping, and surfaces #1 above) is replication of gated content: should a peer lacking authorization (a) never receive the objects in gossip/IPFS at all, or (b) receive hashes/metadata and get capability-required on content fetch? (a) is stricter; (b) is friendlier to the durability goal but leaks existence. The byte-identical-404 in this demo is one expression of (a) on the HTTP path; if you want (b), even this behavior changes. Pick a direction and I will build the path-scoping layer on top.

Also: is private-read / authorization enforcement already claimed or in progress? I would rather build on it than collide.

### DIFF
--- a/crates/git-remote-gitlawb/src/main.rs
+++ b/crates/git-remote-gitlawb/src/main.rs
@@ -127,11 +127,22 @@ fn handle_connect(
     let refs_url = format!("{}/info/refs?service={}", repo_base, service);
     tracing::debug!("GET {refs_url}");
 
-    let refs_resp = client
+    let mut refs_req = client
         .get(&refs_url)
-        .header("User-Agent", "git/2.0 git-remote-gitlawb/0.1.0")
-        .send()
-        .with_context(|| format!("GET {refs_url}"))?;
+        .header("User-Agent", "git/2.0 git-remote-gitlawb/0.1.0");
+
+    // Sign the read so the node can authorize private-repo fetches (RFC 9421).
+    // GET has no body, so content-digest covers the empty body. Harmless for
+    // public repos — the node verifies only when the headers are present.
+    if let Some(kp) = keypair {
+        let signed = sign_request(kp, "GET", &url_path(&refs_url), &[]);
+        refs_req = refs_req
+            .header("Content-Digest", signed.content_digest)
+            .header("Signature-Input", signed.signature_input)
+            .header("Signature", signed.signature);
+    }
+
+    let refs_resp = refs_req.send().with_context(|| format!("GET {refs_url}"))?;
 
     if !refs_resp.status().is_success() {
         bail!(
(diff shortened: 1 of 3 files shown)


kevincodex1:
copy

beardthelion:
No rush on the build-out. Just the one fork when you have a sec: for gated content, should a peer lacking authorization (a) never receive the objects in gossip/IPFS at all, or (b) receive hashes/metadata and get capability-required on content fetch?

Even a one-letter answer unblocks the path-scoping work. Also fine to hear if it is already claimed.

beardthelion:
Superseded by #25, which is the real, mergeable Phase 1 of the path-scoped visibility feature.

This demo did its job: it grounded the design decisions by proving the HTTP read-path enforcement end to end (anon and wrong-DID get a 404 byte-identical to a missing repo, owner gets 200). #25 builds the actual path-scoped model on top (visibility_rules with path globs, reader-DID allow-lists, owner-only management API, gl visibility CLI), with whole-repo rules fully enforced.

Happy to close this in favor of #25 whenever you like. The one open question from the thread (the a/b replication semantics) still gates the gossip/IPFS phase, but Phase 1 stands on its own regardless.

beardthelion:
Closing this out. Phase 1 landed in #25 (merged into main as 6abaf1d), so the whole-repo enforcement slice this PR demoed is now the real thing. Thanks for the look, @kevincodex1.

PR #25, opened 2026-06-03
feat: path-scoped repository visibility (Phase 1) for #18

beardthelion:
Implements **path-scoped repository visibility** (Phase 1), the feature proposed in #18: a repo's "private" status becomes a property of a path subtree rather than a whole-repo boolean, enforced on the git read path using the node's existing DID identity and RFC 9421 HTTP signatures.

This supersedes the demo in #24 (which deliberately enforced only the existing `is_public` boolean to make the design concrete). This PR is the real, mergeable Phase 1.

## What this does

- **`visibility_rules` data model** (`path_glob`, `mode`, `reader_dids`) with DB access methods, modeled on `protected_branches`.
- **`visibility_check`** pure decision function: owner is always allowed, most-specific matching rule wins, reader-DID allow-list, falls back to the legacy `is_public` flag. Exhaustively unit-tested (truth table).
- **Read-path enforcement** on `git_info_refs` and `git_upload_pack`: an unauthorized read returns `404` byte-identical to a missing repo, so a private repo's existence does not leak.
- **Owner-only management API**: `PUT`/`DELETE`/`GET /api/v1/repos/{owner}/{repo}/visibility` behind `require_signature`.
- **CLI**: `gl visibility set/remove/list`.

## Phasing and a deliberate constraint

`mode A` (hide existence) is **whole-repo (`/`) only** by construction: hiding a subtree's existence would rewrite every ancestor tree hash and diverge history (the Merkle DAG hashes child names+hashes). Subtrees therefore use `mode B` (content withheld, SHAs intact). The data model carries `mode` from day one so later phases need no migration.

Two pieces are intentionally **not** in this PR and are flagged as such in the CLI/API output:
- **Subtree mode-B clone filtering** (the node shells out to real `git upload-pack`, which won't impose server-side authz; this needs a cached filtered view). Whole-repo rules are fully enforced today.
- **IPFS / gossip enforcement.** This depends on the one open replication-semantics question from #18 / #24, still unanswered: for gated content, should a peer lacking authorization **(a)** never receive the objects in gossip/IPFS at all, or **(b)** receive hashes/metadata and be capability-gated on content fetch? Phase 1 stands on its own regardless of that answer; Phase 2 needs it.

## Tests

260 tests pass (69 `gitlawb-node` + 191 `gl`); `cargo fmt --check` and `cargo clippy --all-targets` clean.

Closes nothing on its own; advances #18.

## Summary by CodeRabbit

* **New Features**
  * Path-scoped visibility rules for granular repo access (set/remove/list) and owner-only API endpoints
  * Visibility-based enforcement for git read/clone/fetch operations
  * CLI commands to set, remove and list visibility rules

* **Tests**
  * Unit and integration tests covering visibility rules, matching behavior, and CLI flows

### DIFF
--- a/crates/gitlawb-node/src/api/mod.rs
+++ b/crates/gitlawb-node/src/api/mod.rs
@@ -17,4 +17,5 @@ pub mod repos;
 pub mod resolve;
 pub mod stars;
 pub mod tasks;
+pub mod visibility;
 pub mod webhooks;
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -14,6 +14,7 @@ use crate::cert;
 use crate::error::{AppError, Result};
 use crate::git::{smart_http, store};
 use crate::state::AppState;
+use crate::visibility::{visibility_check, Decision};
 use crate::webhooks;
 
 // ── Request / Response types ───────────────────────────────────────────────
@@ -308,6 +309,7 @@ pub async fn git_info_refs(
     State(state): State<AppState>,
     Path((owner, repo)): Path<(String, String)>,
     Query(query): Query<InfoRefsQuery>,
+    auth: Option<Extension<AuthenticatedDid>>,
 ) -> Result<Response> {
     let name = repo.trim_end_matches(".git");
     tracing::info!(owner = %owner, repo = %name, "info/refs request");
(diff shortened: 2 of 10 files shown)


kevincodex1:
hello bro @beardthelion kindly rebase to main and please fix conflicts

beardthelion:
Done, rebased onto current main and pushed. The only real conflict was in server.rs: the visibility route now goes through the new add_auth_layers helper alongside protect (so it gets HTTP Signature + UCAN chain validation, with UCAN passthrough when no X-Ucan header is sent, matching the owner-direct flow). The visibility_rules table auto-merged into the v1 migration list next to protected_branches.

281 tests pass (86 gitlawb-node + 195 gl), fmt and clippy clean. Shows mergeable now.

beardthelion:
Thanks @coderabbitai, both addressed in 5a59064:

1. **Migration v1 (critical):** moved the `visibility_rules` table + index out of v1 into a new migration v3. You're right that appending to v1 would be skipped on any node that already recorded v1, so the table would never exist after upgrade. v3 follows the same pattern #23 used for v2 (agent_profiles) and the "never append to v1" rule documented above the catalogue.

2. **path_glob validation:** `set_visibility` now rejects malformed globs before storing (must start with `/`, no trailing `/`, only a trailing `/**` wildcard, and `/` rather than `/**` for whole-repo scope). Scoped the rules to exactly what `visibility_check` understands. Added unit tests for the accepted and rejected forms.

283 tests pass (88 gitlawb-node + 195 gl), fmt and clippy clean.

kevincodex1 (approved):
LGTM!

PR #28, opened 2026-06-06
feat(node): subtree content withholding (Phase 3) for #18

beardthelion:
Phase 1 (#25) made "private" a property of a path subtree and enforced whole-repo reads. This adds the piece Phase 1 deferred: a mode-B subtree rule now actually withholds that subtree's file content on the git read path, while the directory structure and blob SHAs stay visible.

## How it works

`withheld_blob_oids` walks each ref's tree and, reusing the existing `visibility_check`, returns the blob OIDs a caller may not read. A blob is withheld only if it is denied at every path it appears at, so a blob that is also reachable through an allowed path is still sent.

`upload_pack_excluding` builds the response pack from all reachable objects minus those blob OIDs (commits and trees are always included, so SHAs stay intact) and frames it as a protocol v0 upload-pack response, matching what `info_refs` advertises.

`git_upload_pack` branches to the filtered serve only when the caller has at least one withheld blob. Public and fully authorized clones take the unchanged fast path and stay byte-identical to before.

## Client behavior worth knowing

A pack that omits a blob still referenced by a sent tree is not closed under reachability, so a stock full `git clone` is refused at fetch time with "remote did not send all necessary objects". A partial clone (`git clone --filter=...`) accepts it with the private blob absent, and the tree entry and SHA remain visible. This is a git constraint the server cannot override: the client has to opt into partial-clone semantics. A plain `git clone` working without `--filter` is a separate client-side follow-up (git-remote-gitlawb).

The security guarantee, that private bytes never enter the served pack, holds for every client and is what the tests assert.

## Tests

- `withheld_blob_oids` truth table (owner / listed reader / non-reader / no rules) against real git repos.
- `build_filtered_pack` excludes the withheld OID and keeps the public one.
- A real `git` partial clone through the actual `info_refs` + `upload_pack_excluding`: the withheld blob's bytes never arrive, its tree entry and SHA remain.
- A real incremental `git fetch` after a partial clone: new objects arrive, the withheld blob stays absent.

Full suite green, fmt and clippy clean.

## Out of scope (follow-ups)

- git-remote-gitlawb partial-clone UX so a non-reader can clone without `--filter`.
- Thin-pack negotiation: the filtered serve ignores have/want and re-sends the full object set minus the withheld blobs. Correct, but not bandwidth-optimal.
- `info/refs` is intentionally not gated on subtree rules (refs expose commit tips only); noted in code.
- Replication-path enforcement (Phase 2) is unrelated and still waiting on the A/B replication-semantics call in #18.

Refs #18.

## Summary by CodeRabbit

* **New Features**
  * Added privacy controls for Git repositories. Blobs and files can now be restricted based on user permissions, allowing authorized users to access full repository content while unauthorized users receive filtered views that preserve structure but exclude private content.


docs/superpowers/plans/2026-06-05-phase3-subtree-content-withholding.md
```diff
@@ -0,0 +1,709 @@
...
+This is the one genuinely uncertain piece: how to make `git upload-pack` (or `git pack-objects`) produce a clone/fetch response that omits a specific set of blob OIDs while still sending the trees that reference them, and how to frame that as a valid `application/x-git-upload-pack-result` body. Everything downstream depends on a single function signature, not on the mechanism, so this task nails the mechanism by experiment and records the result. No production code is committed in this task.
+
+**Files:**
+- Scratch only (a throwaway shell script and a temp repo). Findings are written back into this plan's "Task 0 Findings" block below.
+
+- [ ] **Step 1: Build a fixture repo with a public and a private file**
+
+Run:
+```bash
+cd "$(mktemp -d)" && export FIX=$PWD
```
kevincodex1: can we remove this docs? I think we should not commit this

kevincodex1 (changes requested):
please remove the plans under doc. i think its not necessary to be committed

beardthelion:
Good catch, removed it.

kevincodex1 (approved):

PR #34, opened 2026-06-08
feat(node): replication enforcement (Phase 2) for #18

beardthelion:
Phase 2 of path-scoped visibility (#18): stop withheld content from leaving the origin node through replication, and stop fully-private repos from being announced to the network. Phase 1 (#25) gates the git read path and Phase 3 (#28) withholds blobs from served packs, but after a push three paths still copied objects off the node ignoring visibility: local IPFS pinning, Pinata pinning, and the gossip/peer-notify/Arweave announcements.

The whole thing reduces to one decision computed once per push in `git_receive_pack`: can an anonymous caller read the repo root, and which blob OIDs are denied to the public. A `withheld: Option<HashSet<String>>` drives both pin sites (`None` means the repo is private, so nothing replicates, not even commit and tree objects), and an `announce` bool gates the network-facing announcements.

What changes:
- IPFS and Pinata pinning skip the withheld blob OIDs (via a small pure `replicable_objects` filter). For a private repo they pin nothing at all, so file names in tree objects and history in commit objects no longer reach public IPFS.
- Gossip ref-update publish, the HTTP peer-notify fallback, and Arweave anchoring are suppressed for repos the public cannot read. Mode B repos (public with a private subtree) still announce, since their commit and tree SHAs are public.
- Fail closed: if visibility can't be determined, the push replicates nothing.
- The in-process GraphQL subscription broadcast and the local branch->CID write are left alone; they are owner-facing/local, not network leaks.

Deferred on purpose, each cheap to add later off the same seam: peer partial-mirrors (peers currently fail closed on repos with withheld content), UCAN-delegated reader sets, and encrypted-at-rest replication of private blobs.

Depends on #28: `withheld_blob_oids` lives on that branch. This PR is stacked on it, so until #28 merges the diff here will also show #28's commits. Rebase onto main once #28 lands.

## Test plan
- [x] `cargo test -p gitlawb-node` (100 pass), `cargo clippy --all-targets -D warnings` clean, `cargo fmt --check` clean
- [x] Unit coverage: `replicable_objects` filter, anonymous-caller contract of `withheld_blob_oids`, and the announce gate across public / legacy-private / mode A / mode B
- [ ] Manual: push to a node with a mode B `/secret/**` rule, confirm the secret blob is absent from IPFS/Pinata while public files and the commit/tree are present
- [ ] Manual: push to a fully-private repo, confirm no objects pinned and no gossip/peer-notify/Arweave anchor

## Summary by CodeRabbit

* **New Features**
  * Repository visibility rules now control which objects are withheld from replication and peer synchronization.
  * IPFS and Pinata pinning operations now respect visibility rules and only pin allowed objects.
  * Replication decisions are enforced at the push level, with downstream dissemination and permanent anchoring gated by visibility.

* **Bug Fixes**
  * Git protocol errors are now properly remapped to appropriate HTTP status codes.

* **Tests**
  * Added test coverage for visibility rule enforcement during replication announcements.


kevincodex1:
bro @beardthelion please send me dm in X.

kevincodex1:
hello bro @beardthelion please rebase and fix conflicts.

PR #52, opened 2026-06-13
fix(node): enforce path-scoped visibility on the REST read API

### LINKED ISSUE #51: REST read API bypasses path-scoped visibility: private repo/subtree content served unauthenticated
## Summary

The repo-scoped REST read endpoints do not enforce path-scoped visibility. Private repositories, and private subtrees of public repositories, have their content and metadata served to unauthenticated callers. This bypasses the entire withholding mechanism that the git smart-HTTP and replication paths enforce.

The git clone/fetch path (`git_upload_pack` / `info_refs`) and the IPFS/Pinata replication path both gate on `visibility_check` and withhold private blobs correctly. The REST API does not, so the same private bytes are reachable through a different door.

## Impact

Highest severity is content disclosure. `GET /api/v1/repos/{owner}/{repo}/blob/{*path}` returns the raw bytes of any file, including files in a withheld subtree and every file in a fully-private repo, with no auth and no visibility check:

```
GET /api/v1/repos/alice/website/blob/secret/.env      -> 200, cleartext
```

`GET .../pulls/{n}/diff` likewise returns diffs (file content) ungated.

## Affected endpoints (confirmed, all with zero visibility/is_public checks)

Content:
- `GET /blob/{*path}` (`repos::get_blob`) — raw file bytes
- `GET /pulls/{n}/diff` (`pulls::get_pr_diff`) — diffs

Structure / metadata:
- `GET /tree`, `GET /tree/{*path}` (`repos::get_tree_root`, `get_tree`)
- `GET /commits` (`repos::list_commits`) — commit messages
- `GET /refs` (`repos::list_refs`)
- `GET /changelog` (`changelog::get_changelog`)
- `GET /pulls`, `GET /pulls/{n}` (`pulls::list_prs`, `get_pr`)
- `GET /repos/{owner}/{repo}` (`repos::get_repo`) — repo metadata

These are mounted in the `read_routes` router ("open for public repos") and merged with no gating layer, so they are fully public.

## Root cause

Path-scoped visibility was retrofitted onto the git protocol and replication paths but never applied to the REST read surface. The handlers call `get_repo` and read straight from disk; none consult `list_visibility_rules` or `visibility_check`.

## Scope

Pre-existing. `get_blob` dates to the initial public release and is present on `main` today, so this is a live hole independent of the in-flight visibility work. It is called out here because the visibility/withholding feature is meaningless while this bypass exists.

## Suggested fix

Two tiers:

1. Repo-level read gate for every repo-scoped read endpoint: resolve the repo, load its rules, and `visibility_check(rules, is_public, owner_did, caller, "/")`; return 404 on `Deny`. This closes fully-private repos across the whole surface and is best done so new routes cannot silently skip it.
2. Path-level gate for endpoints that expose subtree content: `get_blob` must additionally `visibility_check` at the requested file path, because a public repo with a private subtree passes the `/` gate but the specific blob must still be withheld. `get_pr_diff` needs equivalent treatment.

To authenticate authorized readers over REST (so the gate allows them, not just the owner/public), the gated routes need the `optional_signature` middleware, mirroring the git read routes.

beardthelion:
The git smart-HTTP and replication paths enforce path-scoped visibility, but the REST read endpoints never did. They look up the repo and read straight from disk with no visibility check, so a private repository, and a private subtree of a public repository, are served to anyone.

The clearest case:

```
GET /api/v1/repos/{owner}/{repo}/blob/secret/.env   ->  200, cleartext
```

`get_pr_diff` is the same story for diffs. This predates the visibility work (`get_blob` goes back to the initial release), so it is a live hole on `main` today, not something introduced by the in-flight subtree work. It does, however, make that work moot, which is why it is worth fixing on its own.

## What this does

Adds a single `authorize_repo_read` gate (resolve repo, load rules, `visibility_check`, 404 on deny) and routes every repo-scoped read handler through it: blob, tree, commits, refs, changelog, pulls, PR diff, and repo metadata. The 404 matches what the git path returns on a denied read, so existence is not confirmed either.

- `get_blob` gates at the requested file path, so a withheld subtree file and any file in a fully-private repo both return 404, while the owner and listed readers still get them.
- `get_pr_diff` withholds the whole diff if it touches any path the caller cannot read, via a new `store::branch_diff_names`. Whole-diff denial rather than per-hunk redaction keeps it simple and safe; finer redaction can come later if wanted.
- Listings (tree, commits, refs) are gated at the repo root only, which keeps them consistent with mode B: the git protocol already exposes tree and commit SHAs, only blob content is withheld.
- The read routes get the `optional_signature` layer so an authorized reader can authenticate over REST, the same way the git read routes already do. Anonymous reads of public repos are unaffected.

## Not covered here

`list_repos` / `list_federated_repos` enumerate repos and need per-row filtering, a different mechanism. They leak private-repo existence and should be a separate change.

## Testing

`cargo build` and `cargo test -p gitlawb-node --bin gitlawb-node` pass. `store::branch_diff_names` has a unit test; the gate decision itself is covered by the existing `visibility` unit tests. There is no HTTP-level test harness in this crate, so endpoint-level integration tests would be a follow-up.

Closes #51

## Summary by CodeRabbit

* **New Features**
  * Optional HTTP-signature authentication for public repository read endpoints.
  * Branch-diff path detection to determine which file changes are visible.

* **Security**
  * Repository read operations now enforce per-path visibility before returning data.
  * Pull request diffs and changelogs hide changes that include inaccessible paths; forking checks source repo readability.

* **Tests**
  * Added unit test for branch-diff path detection.

### DIFF
--- a/crates/gitlawb-node/src/api/changelog.rs
+++ b/crates/gitlawb-node/src/api/changelog.rs
@@ -4,8 +4,8 @@ use axum::extract::{Extension, Path, Query, State};
 use axum::Json;
 use serde::Deserialize;
 
-use crate::error::{AppError, Result};
 use crate::auth::AuthenticatedDid;
+use crate::error::{AppError, Result};
 use crate::git::store;
 use crate::state::AppState;
 
--- a/crates/gitlawb-node/src/api/changelog.rs
+++ b/crates/gitlawb-node/src/api/changelog.rs
@@ -1,10 +1,11 @@
 //! Changelog endpoint — unified timeline of commits, merged PRs, and closed issues.
 
-use axum::extract::{Path, Query, State};
+use axum::extract::{Extension, Path, Query, State};
 use axum::Json;
 use serde::Deserialize;
 
 use crate::error::{AppError, Result};
+use crate::auth::AuthenticatedDid;
 use crate::git::store;
 use crate::state::AppState;
 
@@ -27,12 +28,11 @@ pub async fn get_changelog(
     State(state): State<AppState>,
     Path((owner, repo)): Path<(String, String)>,
     Query(query): Query<ChangelogQuery>,
+    auth: Option<Extension<AuthenticatedDid>>,
 ) -> Result<Json<serde_json::Value>> {
-    let record = state
-        .db
-        .get_repo(&owner, &repo)
-        .await?
-        .ok_or_else(|| AppError::RepoNotFound(format!("{owner}/{repo}")))?;
+    let caller = auth.as_ref().map(|e| e.0 .0.as_str());
+    let (record, _rules) =
+        crate::api::authorize_repo_read(&state, &owner, &repo, caller, "/").await?;
 
     let limit = query.limit.min(100);
 
--- a/crates/gitlawb-node/src/api/mod.rs
+++ b/crates/gitlawb-node/src/api/mod.rs
@@ -1,3 +1,8 @@
+use crate::db::{RepoRecord, VisibilityRule};
+use crate::error::{AppError, Result};
+use crate::state::AppState;
+use crate::visibility::{visibility_check, Decision};
+
 pub mod agents;
 pub mod arweave;
 pub mod bounties;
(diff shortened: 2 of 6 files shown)


kevincodex1:
please rebase and fix conflicts

kevincodex1 (approved):

PR #30, opened 2026-06-07
feat(git-remote-gitlawb): add --version and --help flags

FrozenRaspberry:
## What

Adds `--version`/`-V` and `--help`/`-h` to `git-remote-gitlawb`.

## Why

`git-remote-gitlawb` is the only workspace binary without `--version`, which is called out in:

- **`docs/MAINTAINER-ROADMAP.md`** — *"Add `--version` support to `git-remote-gitlawb` so release smoke tests can validate it directly."*
- **`docs/OSS-READINESS-AUDIT.md`** — *"`git-remote-gitlawb` has no `--version` flag; release smoke tests should use a helper-specific invocation."*

## How

- `--version`/`-V` prints `git-remote-gitlawb <version>`, matching the `<bin> <version>` format the clap-based `gl` and `gitlawb-node` emit, so release tooling can treat all three binaries uniformly. The version comes from `CARGO_PKG_VERSION`, so it tracks the workspace version automatically.
- `--help`/`-h` prints concise usage (URL form, env vars, flags).
- Flags are recognized **only as the first argument**, so git's `<remote-name> <url>` remote-helper invocation path is completely untouched. A remote literally named `--version` is not mistaken for the flag (covered by a test).
- Informational flags are handled before `tracing` init so the output stays clean for scripts/smoke tests.

No protocol or API changes. Purely additive, no breaking changes.

## Testing

- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test -p git-remote-gitlawb` all pass.
- 8 new unit tests cover argument classification (including the no-args and misplaced-flag edge cases) and the version/help strings.
- Manually verified on a release build:
  - `git-remote-gitlawb --version` → `git-remote-gitlawb 0.3.9`
  - `--help` renders the usage text
  - no-args still exits `1` with a usage hint

## Follow-up (separate PR)

This unblocks per-binary release smoke tests. I'm happy to send a small follow-up wiring `--version` checks for all three binaries into `release.yml` (the audit also notes "release binaries are not smoke-tested after packaging") — kept out of this PR to keep it focused on one change.

## Summary by CodeRabbit

* **New Features**
  * Added `--version` flag to display application version information
  * Added `--help` flag to display usage documentation
  * Improved error messages to guide users to help resources

PR #59, opened 2026-06-19
ci(release): smoke-test packaged binaries' --version before attaching

FrozenRaspberry:
## What

Smoke-tests the packaged release binaries (`gl`, `git-remote-gitlawb`, `gitlawb-node`) by running each one's `--version` in the `release-binaries` job before they're attached to the release.

## Why

This is the follow-up I flagged in #30 ("wiring `--version` checks for all three binaries into `release.yml`"), now unblocked by `git-remote-gitlawb` gaining `--version` in that PR.

`docs/OSS-READINESS-AUDIT.md` notes:

> *"release binaries are not smoke-tested after packaging."*

The `release-binaries` job builds and tars up all three binaries but never executes them, so a broken artifact (bad build, missing runtime symbol) would ship to a release unnoticed. The `docker` job only smoke-tests the image's `gitlawb-node --version`, not the standalone binaries in the tarball.

## How

- After `Build`, a new `Smoke test binaries` step runs `--version` on each of the three binaries and asserts the released version string is present in the output (they print `<name> <version>`, matching the convention #30 aligned `git-remote-gitlawb` to).
- A version mismatch or non-zero exit fails the release with a clear `::error::` annotation.
- Skipped for `aarch64-unknown-linux-musl`, which is cross-compiled and can't execute on the x86_64 runner. The other three targets (`x86_64-linux-musl`, `x86_64-apple-darwin`, `aarch64-apple-darwin`) run natively on their matrix runners.

No protocol, API, or release-artifact changes — purely an added verification gate.

## Testing

- `release.yml` validated as well-formed YAML.
- Verified locally against a release build of the current `main`:
  - `gl --version` → `gl 0.3.9`
  - `git-remote-gitlawb --version` → `git-remote-gitlawb 0.3.9`
  - `gitlawb-node --version` → `gitlawb-node 0.3.9`
  - The exact loop from the workflow passes for all three, and correctly **fails** when the expected version doesn't match (confirming the gate isn't a no-op).

## Summary by CodeRabbit

* **Chores**
  * Enhanced release pipeline with automated verification of compiled binaries to ensure version consistency across all target platforms.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -146,6 +146,27 @@ jobs:
       - name: Build
         run: cargo build --release --target ${{ matrix.target }} -p gl -p git-remote-gitlawb -p gitlawb-node
 
+      # Run each packaged binary's --version so a broken release artifact fails the
+      # build instead of shipping. Skipped for cross-compiled targets that can't run
+      # on the host runner (aarch64 linux on an x86_64 runner).
+      - name: Smoke test binaries
+        if: ${{ matrix.target != 'aarch64-unknown-linux-musl' }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          VERSION="${{ needs.release-please.outputs.version }}"
+          BIN_DIR="target/${{ matrix.target }}/release"
+          for bin in gl git-remote-gitlawb gitlawb-node; do
+            echo "== $bin --version =="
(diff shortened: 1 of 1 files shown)


kevincodex1 (approved):
looks great!

Issue #42, opened 2026-06-12
Phase 3 withholding: incomplete withheld set can leak blobs in cleartext (ref namespace scope + fail-open ref walk)


beardthelion:
## Summary

Two issues in `crates/gitlawb-node/src/git/visibility_pack.rs` (`blob_paths`, the basis of the withheld-blob set for Phase 3 subtree withholding) can produce an **incomplete `withheld` set**, which downstream code then treats as public — a potential cleartext disclosure of withheld content.

### 1. Withholding only walks `refs/heads/*` and `refs/tags/*`

```rust
if !refname.starts_with("refs/heads/") && !refname.starts_with("refs/tags/") {
    continue;
}
```

`blob_paths` enumerates blob→path pairs from heads and tags only. But the consumers of the resulting `withheld` set operate over a broader object graph:
- `crates/gitlawb-node/src/git/smart_http.rs` uses `rev-list --objects --all`
- `crates/gitlawb-node/src/ipfs_pin.rs` filters all repo objects through it

A blob reachable only from another ref namespace (e.g. `refs/notes/*`, `refs/merge-requests/*`, hidden refs) never enters `withheld`, so it can be packed or pinned in cleartext even though its path is denied by visibility rules.

### 2. Ref walk failures fail open

```rust
if !listing.status.success() {
    continue;
}
```

A non-zero `git ls-tree` result is skipped, turning a traversal error into a partial `withheld` set. Callers then treat the missing blobs as public. This should fail closed (return an error so the fetch/pack/pin aborts) rather than silently leak.

## Suggested direction

- Enumerate the object graph for withholding with the **same** ref/object scope the pack and pin paths use (align on one shared enumerator over `--all` / the full reachable set), so nothing the consumers see can escape the withheld computation.
- On any `git ls-tree` / ref-walk error, return `Err` so the caller aborts instead of producing a partial set.

## Notes

This is Phase 3 (#28, issue #18) code. Surfaced by CodeRabbit during review of #40 (it appears in that PR only because the cross-fork diff is cumulative over the unmerged stack). Filing here so it's tracked against the withholding work rather than the recipient-set-blinding PR.

PR #60, opened 2026-06-19
fix(node): skip withheld-walk when no path-scoped rule can withhold

### LINKED ISSUE #50: Serve path: short-circuit withheld walk when no path-scoped rule can withhold
Surfaced by CodeRabbit on #46.

The clone/fetch serve path in `api/repos.rs` runs the full `visibility_pack::withheld_blob_oids` ref walk on every request before deciding whether any blob is actually withheld, then falls back to a plain `upload_pack` when the set comes back empty. For repos with no path-scoped rules (the common public case) that walk can never withhold anything, so it is pure overhead per clone.

This got more expensive with #42: `blob_paths` now walks every ref via `git ls-tree -r` rather than just heads/tags, so the wasted scan is broader. Short-circuit before spawning the walk when the rule set is empty or root-only (no rule with a non-"/" path glob), and only run `withheld_blob_oids` when a path-scoped rule exists.

Owning PR: #28/#33 (serve path); cost amplified by #42.

beardthelion:
## Summary

Public clones and fetches were paying for an authorization walk they could never benefit from. On every clone/fetch the serve path enumerated every blob path across every ref (a blocking `git ls-tree -r` per ref) to compute which blobs to withhold, then threw the result away and served a normal pack whenever nothing was withheld, which is always the case for a repo with no path-scoped visibility rules, i.e. the common public repo. The cost got worse with #42, which widened the walk from heads/tags to every ref.

This skips the walk entirely when no rule can withhold anything.

## What changed

A shared predicate, `has_path_scoped_rule`, now gates all three consumers of the walk:

1. The serve path (`git_upload_pack`) skips the blocking walk and serves the pack directly.
2. The post-push replication path's existing empty-rules short-circuit is generalized to also cover root-only rule sets.
3. The post-push encrypt-then-pin path in `git_receive_pack` is gated on the same predicate, replacing a `!rules.is_empty()` check that still ran the per-ref walk (via `withheld_blob_recipients`) for root-only rule sets.

Routing all three through one predicate keeps the call sites from drifting.

## Why it's safe

The short-circuit is only sound because the whole-repo `/` access gate already runs before every call site: a denying root rule 404s the caller on the serve path, and suppresses announcement on the replication and encrypt-then-pin paths, before the predicate is ever consulted. Any caller that reaches the predicate has therefore already cleared repo-level access, and with no path-scoped rule left there is nothing to withhold. Served-pack contents are unchanged on the success path.

The encrypt-then-pin gate is additionally behavior-preserving: for a root-only rule set `withheld_blob_recipients` returns an empty recipient map, so `encrypt_and_pin` was already a no-op; the gate just skips the wasted walk that produced it.

The predicate's doc comment records this precondition explicitly, plus the dependency on `validate_path_glob` rejecting `/**` (the only non-`/` glob that would match every path).

One intended behavioural difference, on the error path only: a corrupt public/root-only repo whose ref walk would have thrown now skips that walk and goes straight to `upload_pack`, so a spurious 500 originating in the walk no longer surfaces there. This is the wasted-walk overhead going away, not a change to what content is served.

## Tests

- Predicate unit tests (empty / single-root / path-scoped / mixed / multi-root).
- A gate-aware safety invariant test: for any caller that has passed the `/` gate, an empty or root-only rule set yields an empty withheld set, pinning the invariant the skip depends on against future changes to the matching logic.
- A serve-decision test over a real bare repo covering all three predicate branches: root-only skips the walk and serves the full pack, a reader allowed at `/` but denied `/secret/**` is not served the secret blob, and the owner bypasses the rule and is served everything. It composes the same functions `git_upload_pack` composes; the pack-byte filtering itself is covered by `filtered_serve_excludes_withheld_blob` in `smart_http.rs`.
- Full `gitlawb-node` suite: 199 passing; `cargo fmt --check` and `cargo clippy` clean.

The serve/replication HTTP endpoints have no end-to-end git smart-http harness in this crate, so the call-site wiring is covered at the predicate and serve-decision level rather than via a full request roundtrip.

The adjacent fork-path concern (a fork clones a full mirror after only a root `/` check) is pre-existing and out of scope here; it is tracked separately in #98.

Closes #50.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected visibility rule handling for repository replication and adjusted withholding behavior for configurations using only whole-repository rules.

* **Performance Improvements**
  * Improved pack-serving efficiency by avoiding unnecessary per-object withholding checks when no path-scoped visibility rules are present.

* **Tests**
  * Added unit tests for detecting path-scoped visibility rules, covering empty, root-only (allow/deny), mixed, and duplicate rule scenarios, plus safety checks ensuring withheld results remain empty when appropriate.

### DIFF
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -56,7 +56,7 @@ async fn replication_withheld_set(
         // root-only-rules cases), so skip the full withheld_blob_oids walk and
         // withhold nothing. The predicate's safety-invariant test guards that
         // this short-circuit matches what the walk would have returned.
-        Some(rules) if !crate::git::visibility_pack::has_path_scoped_rule(&rules) => {
+        Some(rules) if !visibility_pack::has_path_scoped_rule(&rules) => {
             Some(std::collections::HashSet::new())
         }
         // withheld_blob_oids walks every ref with blocking `git ls-tree`; keep
@@ -838,7 +838,11 @@ pub async fn git_receive_pack(
 
             // Option B1: encrypt-then-pin the withheld blobs so authorized
             // readers can recover them when the origin cannot serve them.
-            if let Some(rules) = rules_for_enc.filter(|r| !r.is_empty()) {
+            // No path-scoped rule can withhold a blob, so withheld_blob_recipients
+            // would return an empty map after a full per-ref walk; skip it. Mirrors
+            // the has_path_scoped_rule gate on the other two withheld-walk sites.
+            if let Some(rules) = rules_for_enc.filter(|r| visibility_pack::has_path_scoped_rule(r))
+            {
                 let p = repo_path_clone.clone();
                 let owner = owner_did.clone();
                 let recip = tokio::task::spawn_blocking(move || {
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -52,7 +52,13 @@ async fn replication_withheld_set(
         return (false, None);
     }
     let withheld = match rules {
-        Some(rules) if rules.is_empty() => Some(std::collections::HashSet::new()),
+        // No path-scoped rule can withhold anything (covers the empty-rules and
+        // root-only-rules cases), so skip the full withheld_blob_oids walk and
+        // withhold nothing. The predicate's safety-invariant test guards that
+        // this short-circuit matches what the walk would have returned.
+        Some(rules) if !crate::git::visibility_pack::has_path_scoped_rule(&rules) => {
+            Some(std::collections::HashSet::new())
+        }
         // withheld_blob_oids walks every ref with blocking `git ls-tree`; keep
         // that off the async worker thread.
         Some(rules) => {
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
# Summary
This PR adds a safe short-circuit that skips the expensive per-ref withheld-blob walk when visibility rules contain only whole-repo `/` entries, and wires that predicate into both fetch serving and post-push replication.

# Findings
No findings.

Please rebase on main and clear merge conflict.

jatmn (changes requested):
## Findings

### 1. Encrypt-then-pin branch still runs the ref-wide walk for root-only rules

**Location:** `crates/gitlawb-node/src/api/repos.rs`, lines 839-846

The post-push encrypt-then-pin path gates the `withheld_blob_recipients` call with `rules_for_enc.filter(|r| !r.is_empty())`. That condition is true for *any* non-empty rule set, including a root-only `/` rule. `withheld_blob_recipients` then calls `withheld_blob_oids`, which performs the same `git ls-tree -r` per-ref walk this PR is trying to avoid.

This is the third consumer of the ref-wide walk, and it was missed by the PR's short-circuit. The same predicate used for the other two call sites should be applied here:

```rust
if let Some(rules) = rules_for_enc
    .filter(|r| crate::git::visibility_pack::has_path_scoped_rule(r))
{
    // ... spawn withheld_blob_recipients ...
}
```

**Impact:** Root-only repos continue to pay the per-push blocking walk cost on every push that triggers encrypt-then-pin, partially defeating the PR's performance goal.

**Suggested action:** Gate the `withheld_blob_recipients` block with `has_path_scoped_rule`, mirroring the other two call sites.

---

### 2. Fork endpoint clones a full mirror after only a root visibility check

**Location:** `crates/gitlawb-node/src/api/repos.rs`, lines 1216-1273

`fork_repo` authorizes the caller only at the whole-repo `/` path and then runs `git clone --mirror` of the source repository. A caller who is allowed at `/` but denied a path-scoped subtree (e.g., `/secret/**`) can therefore fork the entire repo and obtain blobs that the filtered `git_upload_pack` path would have withheld.

This is a pre-existing issue, but it is directly adjacent to the visibility logic this PR is optimizing. CodeRabbit flagged it as a critical security concern in the most recent review.

**Impact:** Non-owner callers with root access but without subtree access can bypass path-scoped visibility by forking.

**Suggested action:** Either fail closed for non-owners when the source has path-scoped rules, or build the fork from the same filtered projection used by `git_upload_pack` rather than from a full mirror. Preserve the existing not-found-style behavior for protected-data failures and use `did_matches(...)` for any owner-only bypass.

beardthelion:
Both addressed.

**#1 — encrypt-then-pin ran the ref-wide walk for root-only rules.** Fixed in cf81708: the block now gates on `has_path_scoped_rule`, matching the serve and replication sites. It's behavior-preserving for the root-only case (the walk would return an empty recipient map, so `encrypt_and_pin` was already a no-op) and unchanged for path-scoped repos.

**#2 — fork clones a full mirror after only a root check.** Pre-existing and a separate authz concern, so I split it to #98 rather than fold it in here. Tracked, not dropped.

Also added a serve-decision test (1d5589b) over a real bare repo covering all three predicate branches: root-only skips the walk and serves the full pack, a reader denied `/secret/**` is not served the secret blob, and the owner bypasses the rule and gets everything. One coverage boundary worth naming: it composes the same functions `git_upload_pack` composes rather than driving the handler through a git smart-http roundtrip (no such harness on this branch); the pack-byte filtering itself is already covered by `filtered_serve_excludes_withheld_blob` in `smart_http.rs`.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM!

PR #64, opened 2026-06-20
fix(core): route seed access through the zeroizing wrapper (#41)

### LINKED ISSUE #41: Harden Keypair::seed_bytes: route seed access through the zeroizing wrapper
## Summary

`Keypair::seed_bytes()` (`crates/gitlawb-core/src/identity.rs`) returns the raw 32-byte Ed25519 seed as a plain `[u8; 32]`, so the private key material is copied out without zeroization and lives in caller stack/locals until the values happen to be overwritten.

A safe sibling already exists in the same type:

```rust
pub fn seed_bytes(&self) -> [u8; 32] { self.signing_key.to_bytes() }          // raw, not zeroized
pub fn to_seed(&self) -> Zeroizing<[u8; 32]> { Zeroizing::new(self.signing_key.to_bytes()) }  // zeroized on drop
```

`zeroize` is already a dependency, so this is purely about routing seed access through the zeroizing wrapper.

## Proposed change

Drop `seed_bytes()` and migrate callers to `to_seed()` (or, if the name is preferred, change `seed_bytes()` to return `Zeroizing<[u8; 32]>`). Either way the plain-array return goes away. `Zeroizing<[u8; 32]>` derefs to `[u8; 32]`, so call sites that need `&[u8; 32]` use `&*seed`.

Call sites to update:
- `crates/gitlawb-core/src/encrypt.rs:126` — `open_blob`, `x25519_secret_from_seed(&keypair.seed_bytes())`
- `crates/gitlawb-node/src/api/repos.rs:672` — node seed for the encrypted-blob recipients tag
- `crates/gitlawb-core/src/encrypt.rs:188` — a test

## Notes

Defense-in-depth hardening, not a known live exploit; the seed is already held in memory by the keypair. Surfaced during review of #40 (which added a `seed_bytes()` caller). Deferred from that PR to keep its scope to recipient-set blinding, since changing the return type touches pre-existing callers.

beardthelion:
## Summary

`Keypair::seed_bytes()` handed out a bare `[u8; 32]` copy of the Ed25519 private seed. Because `[u8; 32]` is `Copy` with no `Drop`, each caller's copy was released without being scrubbed, leaving secret key material in freed memory. This removes that accessor and routes seed access through the existing zeroizing wrapper.

## Motivation & context

Closes #41

`to_seed() -> Zeroizing<[u8; 32]>` already existed right beside `seed_bytes()` and scrubs the copy on drop. There was no reason to keep an unzeroized accessor: it was duplicated, and the only two callers were both in `gitlawb-core`. Removing it makes `to_seed()` the single seed accessor and closes the exposure window to core dumps, swap, and memory-disclosure bugs.

## Kind of change

- [ ] Bug fix
- [ ] Feature
- [x] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

Crate touched: `gitlawb-core`.

- Removed `Keypair::seed_bytes()` (raw, unzeroized seed accessor) in `identity.rs`.
- Migrated the two callers to `to_seed()`: the envelope-decryption path in `encrypt.rs` (`open_blob`) and a crypto agreement test. The migration is mechanical: `Zeroizing<[u8; 32]>` deref-coerces to `&[u8; 32]`, which is what `x25519_secret_from_seed` already takes.
- Added `to_seed_is_sole_seed_accessor` documenting and guarding the single-accessor invariant.

No behavior change to envelope encryption/decryption: the X25519 secret derived from the seed is byte-identical, covered by the existing round-trip tests.

## How a reviewer can verify

```sh
cargo test --workspace
# Confirm no raw accessor remains:
git grep -n '\.seed_bytes()' crates/   # only the unrelated p2p local var, no Keypair method
```

## Before you request review

## Protocol & signing impact

This is in-memory hygiene only. It does not change the signature scheme, key format, DID derivation, or any wire format. The bytes produced and consumed are identical; only the container type of a transient changes.

## Notes for reviewers

There is a related, deliberately out-of-scope leak one hop downstream: `x25519_secret_from_seed` returns the derived X25519 secret as a bare `[u8; 32]`. It will be filed as its own follow-up issue to keep this PR scoped to exactly #41.

## Summary by CodeRabbit

* **Security**
  * Strengthened cryptographic seed handling by switching to a secure, automatically-cleared seed accessor for envelope decryption.
* **Breaking Changes**
  * Removed the previous raw seed accessor; seed retrieval is now provided only through the secure accessor.
* **Tests**
  * Updated and added unit tests to confirm deterministic seed derivation and consistent key reconstruction from the derived seed.

### DIFF
--- a/crates/gitlawb-core/src/encrypt.rs
+++ b/crates/gitlawb-core/src/encrypt.rs
@@ -123,7 +123,7 @@ pub fn open_blob(envelope: &[u8], keypair: &Keypair) -> Result<Vec<u8>> {
             .context("decode header")?;
     let body = &envelope[p + hlen..];
 
-    let my_x = XSecret::from(x25519_secret_from_seed(&keypair.seed_bytes()));
+    let my_x = XSecret::from(x25519_secret_from_seed(&keypair.to_seed()));
 
     // Identities are blinded: no entry says which recipient it belongs to, so
     // try each one. The ChaChaBox AEAD tag authenticates, so exactly the
@@ -185,7 +185,7 @@ mod tests {
         // The X25519 public derived from the Ed25519 public must equal the
         // X25519 public of the X25519 secret derived from the same seed.
         let kp = Keypair::generate();
-        let seed = kp.seed_bytes();
+        let seed = kp.to_seed();
         let xpub_from_public = x25519_public(&kp.verifying_key()).unwrap();
         let xsec = x25519_secret_from_seed(&seed);
         let xpub_from_secret = crypto_box::SecretKey::from(xsec).public_key().to_bytes();
--- a/crates/gitlawb-core/src/identity.rs
+++ b/crates/gitlawb-core/src/identity.rs
@@ -52,13 +52,9 @@ impl Keypair {
         URL_SAFE_NO_PAD.encode(sig.to_bytes())
     }
 
-    /// The raw 32-byte Ed25519 seed. Used to derive the X25519 secret for
-    /// envelope decryption (see `crate::encrypt`).
-    pub fn seed_bytes(&self) -> [u8; 32] {
-        self.signing_key.to_bytes()
-    }
-
-    /// Export the signing key as raw 32-byte seed (wrapped in Zeroizing).
+    /// The raw 32-byte Ed25519 seed, wrapped in `Zeroizing` so the copy is
+    /// scrubbed on drop. This is the only seed accessor; it is used to derive
+    /// the X25519 secret for envelope decryption (see `crate::encrypt`).
     pub fn to_seed(&self) -> Zeroizing<[u8; 32]> {
         Zeroizing::new(self.signing_key.to_bytes())
     }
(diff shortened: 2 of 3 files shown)


beardthelion:
@kevincodex1 #64 is ready when you have a minute. It removes the unzeroized `seed_bytes()` accessor and routes seed access through `to_seed() -> Zeroizing`, so the seed copy scrubs on drop.

One thing to flag: `main` grew a third caller since I branched (the encrypted-pin path in `repos.rs`, from #40), so I merged `main` in and migrated that one too. It moves the seed into a long-lived spawn task, so wrapping it in `Zeroizing` is the same hardening rather than just a compile fix. Behavior is byte-identical, covered by the existing round-trip tests; full CI is green and CodeRabbit had no actionable comments.

It's also the base of #66 (the X25519 follow-up), so merging this unblocks that one.

kevincodex1 (approved):
LGTM-

PR #66, opened 2026-06-20
fix(core): zeroize the derived X25519 secret (#65)

### LINKED ISSUE #65: Harden x25519_secret_from_seed: zeroize the derived X25519 secret (follow-up to #41)
## Background

#41 removed `Keypair::seed_bytes()` and routed Ed25519 seed access through the zeroizing `to_seed()`. While closing that, the review surfaced a sibling leak one hop downstream that #41 deliberately left out of scope to keep that PR to a single hardening slice.

## Current behavior

`x25519_secret_from_seed` (`crates/gitlawb-core/src/encrypt.rs`) derives the X25519 private scalar from the Ed25519 seed and returns it as a bare `[u8; 32]`:

```rust
fn x25519_secret_from_seed(seed: &[u8; 32]) -> [u8; 32] {
    let h = Sha512::digest(seed);     // 64 bytes; lower 32 are the secret
    let mut s = [0u8; 32];
    s.copy_from_slice(&h[..32]);
    // clamp ...
    s                                  // [u8; 32]: Copy, no Drop
}
```

The returned value is the actual X25519 private key (fed into `crypto_box::SecretKey` in `open_blob` to unwrap content keys). Because `[u8; 32]` is `Copy` with no `Drop`, and `sha2`'s digest output (`h`) likewise has no zeroize, the secret-bearing bytes in `s`, `h`, and the returned temporary are released without being scrubbed. This is the same `Copy`-no-`Drop` memory-hygiene class fixed in #41, on the derived secret rather than the raw seed. It sits in the `open_blob` path, which runs on every withheld-blob read.

`crypto_box::SecretKey` already zeroizes its own internal copy on drop; this issue is only about the intermediate copies that precede it.

## Proposal

- Return `Zeroizing<[u8; 32]>` from `x25519_secret_from_seed` and explicitly zeroize the `h` digest before it drops.
- Migrate the two callers (`open_blob` and the crypto test) accordingly.
- One wrinkle to settle in the fix: `crypto_box::SecretKey::from` takes `[u8; 32]` by value, so feeding it from a `Zeroizing` still materializes one transient copy. That's acceptable (the `SecretKey` scrubs its own copy), but the fix should confirm no extra bare copy is left behind.

## Scope

Memory hygiene only. No change to the encryption scheme, key derivation math, or wire format; behavior is byte-identical and covered by the existing `seal_open_round_trip_for_recipients` / `ed25519_to_x25519_keypair_agrees` tests. Not remotely exploitable; defense-in-depth against core dumps, swap, and memory-disclosure bugs.

beardthelion:
> Stacked on #64 (#41). Based on `fix/zeroize-seed-bytes`; the diff here is only this change. Retarget/rebase onto `main` once #64 merges. Please merge #64 first.

## Summary

`x25519_secret_from_seed` derived the X25519 private scalar from the Ed25519 seed and returned it as a bare `[u8; 32]`. Because `[u8; 32]` is `Copy` with no `Drop` (and `sha2`'s digest output likewise has no zeroize), the scalar, the SHA-512 digest, and the returned temporary were released without being scrubbed. This routes the derived secret through `Zeroizing` and explicitly wipes the digest.

## Motivation & context

Closes #65

Direct sibling of #41/#64: same `Copy`-no-`Drop` memory-hygiene class, on the *derived* X25519 secret rather than the raw seed. #41 was deliberately scoped to the seed accessor and flagged this downstream leak for its own follow-up. The derived scalar is the actual private key fed into `crypto_box::SecretKey` in `open_blob`, which runs on every withheld-blob read.

## Kind of change

- [ ] Bug fix
- [ ] Feature
- [x] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

Crate touched: `gitlawb-core`.

- `x25519_secret_from_seed` now returns `Zeroizing<[u8; 32]>` and builds the scalar directly into a zeroizing buffer, so no bare `[u8; 32]` secret local persists.
- The intermediate SHA-512 digest is explicitly wiped (`h.as_mut_slice().zeroize()`, a volatile non-elidable write) before it drops.
- The two callers (`open_blob` and a crypto test) deref the result into `crypto_box::SecretKey`, which already scrubs its own copy.

No behavior change: the derived scalar and all decryption output are byte-identical, covered by the existing `ed25519_to_x25519_keypair_agrees` and `seal_open_round_trip_for_recipients` tests.

## How a reviewer can verify

```sh
cargo test --workspace
```

## Before you request review

- [x] New behavior is covered by tests (behavior is byte-identical; existing crypto round-trip tests guard it)

## Protocol & signing impact

In-memory hygiene only. No change to the encryption scheme, key derivation math, or wire format; output is byte-identical.

### DIFF
--- a/crates/gitlawb-core/src/encrypt.rs
+++ b/crates/gitlawb-core/src/encrypt.rs
@@ -6,6 +6,7 @@
 use crate::identity::Keypair;
 use anyhow::{Context, Result};
 use ed25519_dalek::VerifyingKey;
+use zeroize::Zeroizing;
 
 /// X25519 public key (Montgomery u) for an Ed25519 verifying key.
 fn x25519_public(vk: &VerifyingKey) -> Result<[u8; 32]> {
@@ -18,14 +19,18 @@ fn x25519_public(vk: &VerifyingKey) -> Result<[u8; 32]> {
 }
 
 /// X25519 secret scalar for an Ed25519 seed (SHA-512 of seed, lower 32, clamped).
-fn x25519_secret_from_seed(seed: &[u8; 32]) -> [u8; 32] {
+/// Returns the scalar wrapped in `Zeroizing`, and scrubs the intermediate
+/// SHA-512 digest, so no copy of this secret material lingers in freed memory.
+fn x25519_secret_from_seed(seed: &[u8; 32]) -> Zeroizing<[u8; 32]> {
     use sha2::{Digest, Sha512};
-    let h = Sha512::digest(seed);
-    let mut s = [0u8; 32];
+    use zeroize::Zeroize;
+    let mut h = Sha512::digest(seed);
+    let mut s = Zeroizing::new([0u8; 32]);
     s.copy_from_slice(&h[..32]);
     s[0] &= 248;
     s[31] &= 127;
     s[31] |= 64;
+    h.as_mut_slice().zeroize();
     s
 }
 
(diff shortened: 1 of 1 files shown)


kevincodex1 (approved):
LGTM

PR #67, opened 2026-06-20
fix(node): fail closed when a recipient DID can't be resolved (#47)

### LINKED ISSUE #47: encrypted_pin: sealing to a partial recipient set leaves authorized readers unrecoverable
Surfaced by CodeRabbit on #46 (the review spans the whole unmerged stack, so it lands on B1 code from #36 rather than #46's delta).

`encrypted_pin::encrypt_and_pin` resolves recipient DIDs with:

```rust
let keys: Vec<VerifyingKey> = dids.iter().filter_map(|d| did_to_key(d)).collect();
if keys.is_empty() {
    tracing::warn!(oid = %oid, "no resolvable recipient keys; skipping encrypted pin");
    continue;
}
```

`filter_map` silently drops any DID that fails resolution and seals to whatever remains, as long as at least one key resolves. The blob is then recorded as successfully encrypted and pinned. An authorized reader whose key could not be resolved at seal time is now permanently unable to recover that blob, with no signal that it happened. For a confidentiality/recovery system this is a correctness bug, not a soft-fail.

Treat any unresolved DID as a per-blob failure: resolve all DIDs, and if any fail, log with the oid and the unresolved set and `continue` rather than sealing to a partial set.

Secondary, same function: the `read_object` and `pin_git_object` failure arms collapse to `_ => continue` with no log, which hurts observability during partial outages. Worth adding warn logs while in here.

Owning PR: #36.

beardthelion:
## Summary

`encrypted_pin::encrypt_and_pin` sealed a withheld blob to whatever subset of its recipient DIDs resolved locally and recorded the full intended set as covered, permanently locking out any reader whose DID could not be resolved at seal time. This makes sealing fail closed: resolve all recipients or skip the blob.

## Motivation & context

Closes #47

`did_to_key` only resolves `did:key`; `did:web`/`did:gitlawb` (the canonical identity actors migrate to once DHT anchoring is live) and any malformed `did:key` return `None`. The old `filter_map` dropped those silently and sealed to the rest as long as one key resolved, then recorded the full DID set. Because the dedup compares the recorded set against the desired set, it matched on the next push and never re-sealed, so the dropped reader stayed locked out forever with no signal.

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

gitlawb-node (`crates/gitlawb-node/src/encrypted_pin.rs`):

- Add `resolve_all_recipients`: resolves every recipient DID all-or-nothing, returning `Err(unresolved)` listing the DIDs that failed. `encrypt_and_pin` now skips the blob on any failure instead of sealing to a partial set. Nothing partial is recorded, so the dedup stays honest and the blob re-seals on a later push once resolution is available (no automatic backfill).
- Bounded per-blob warn (the unresolved DIDs are user-controlled rule data; log a count plus a small sample, not an unbounded dump) with wording that stays neutral about the cause (a malformed `did:key` is not DHT-pending).
- One aggregate coverage warn per call, so a fully-migrated `did:gitlawb` org dropping to zero encrypted-pin coverage is a single greppable line rather than a per-oid scrape.
- Give the previously silent `read_object` and `pin_git_object` failure arms warn logs.

Fail-closed is deliberate: a blob with an unresolvable reader gets no encrypted recovery envelope until resolution exists, but it is never leaked in plaintext (it was already pinned as withheld upstream). This trades recovery coverage during a transition for a clean audit invariant: a recorded `encrypted_blobs` row always means every authorized reader can recover the blob.

## How a reviewer can verify

```sh
cargo test -p gitlawb-node encrypted_pin
cargo test --workspace
```

The core regression is `mixed_set_with_did_gitlawb_fails_closed`: a resolvable `did:key` subset plus one `did:gitlawb` must return `Err` naming only the unresolvable DID, never a sealable key set. `malformed_did_key_fails_closed` covers the corrupt-`did:key` case.

## Before you request review

## Protocol & signing impact

This reads DID strings to resolve recipient keys but changes no signature, wire-format, or identity logic; existing envelopes and recorded recipient sets are untouched.

## Notes for reviewers

The peer-sync path (`sync.rs` records an empty recipient set for replicated envelopes via an overwriting upsert) interacts with the dedup but self-heals on the next push; it is pre-existing and out of scope here. A typed resolution-failure reason (distinguishing "method not resolvable yet" from "malformed `did:key`") is a possible follow-up; fail-closed behavior is identical for both today.

## Summary by CodeRabbit

* **Improvements**
  * Strengthened recipient handling to ensure all recipients are resolvable before proceeding, preventing partial processing when any recipient is invalid or unresolvable.
  * Improved behavior and messaging around skipped items, distinguishing between empty recipient sets, unresolvable recipients, and already-up-to-date recipients.
  * Refined pinning outcomes to better differentiate “no result” vs actual failures.

* **Tests**
  * Expanded coverage for recipient validation edge cases and decision paths to confirm fail-closed behavior.

### DIFF
--- a/crates/gitlawb-node/src/encrypted_pin.rs
+++ b/crates/gitlawb-node/src/encrypted_pin.rs
@@ -75,7 +75,10 @@ enum SealPlan {
     /// partial set. Carries the unresolvable DIDs for logging.
     SkipUnresolvable(Vec<String>),
     /// Seal to `keys` and record coverage under `tag`.
-    Seal { keys: Vec<VerifyingKey>, tag: String },
+    Seal {
+        keys: Vec<VerifyingKey>,
+        tag: String,
+    },
 }
 
 /// Decide what to do with one blob given its desired recipient set and the tag
@@ -304,7 +307,11 @@ mod tests {
         match plan_seal(&SEED, &dids, None) {
             SealPlan::Seal { keys, tag } => {
                 assert_eq!(keys.len(), 2, "must seal to the full recipient set");
-                assert_eq!(tag, recipients_tag(&SEED, &dids), "records the full-set tag");
+                assert_eq!(
+                    tag,
+                    recipients_tag(&SEED, &dids),
+                    "records the full-set tag"
+                );
             }
             other => panic!("expected Seal, got {other:?}"),
         }
--- a/crates/gitlawb-node/src/encrypted_pin.rs
+++ b/crates/gitlawb-node/src/encrypted_pin.rs
@@ -63,6 +63,42 @@ fn resolve_all_recipients(dids: &BTreeSet<String>) -> Result<Vec<VerifyingKey>, 
     }
 }
 
+/// What to do with a single withheld blob, decided without any DB or IO so the
+/// fail-closed invariant (#47) is unit-testable in isolation.
+#[derive(Debug)]
+enum SealPlan {
+    /// An existing envelope already covers exactly this recipient set; nothing to do.
+    SkipUnchanged,
+    /// No recipient DID resolved to a key, so there is nothing to seal to.
+    SkipNoRecipients,
+    /// At least one recipient DID is unresolvable. Fail closed: never seal to a
+    /// partial set. Carries the unresolvable DIDs for logging.
+    SkipUnresolvable(Vec<String>),
+    /// Seal to `keys` and record coverage under `tag`.
+    Seal { keys: Vec<VerifyingKey>, tag: String },
+}
+
+/// Decide what to do with one blob given its desired recipient set and the tag
+/// already stored for it (if any). Pure: no DB, no IO.
+///
+/// This is the #47 fail-closed gate in isolation: it returns `Seal` only when
+/// EVERY recipient DID resolves, so no caller can seal to a partial set. A
+/// changed recipient set (different tag) re-seals so a newly added reader can
+/// recover the blob; reader removal is not retroactive (the old envelope is
+/// already public). The comparison is on the opaque node-keyed tag, never the
+/// DID list.
+fn plan_seal(node_seed: &[u8; 32], dids: &BTreeSet<String>, stored_tag: Option<&str>) -> SealPlan {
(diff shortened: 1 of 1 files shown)


kevincodex1:
fix failing tests

beardthelion:
Pushed ca53afd: that was the fmt gate, not the tests (stable + beta were already green). `cargo fmt --all` on encrypted_pin.rs, two cosmetic reflows. fmt + clippy and cargo audit both pass now. @kevincodex1

jatmn (commented):
# Validation notes
- I checked the repo docs (`README.md`, `docs/RUN-A-NODE.md`) and CI config (`.github/workflows/pr-checks.yml`) for a stated Git requirement and did not find one.
- The test job runs on `ubuntu-latest` and executes `cargo test --workspace`, but it does not pin or document a Git version.
- In my review environment (Windows, `git version 2.53.0.windows.2`), `cargo test -p gitlawb-node sync -- --nocapture` failed because this Git build rejected `git clone --mirror --filter=blob:limit=10g` from `crates/gitlawb-node/src/sync.rs:445-454`.
- Since the repo does not document a supported Git version and CI runs on a different OS, I do not have enough evidence to call that a PR bug rather than an environment mismatch.

beardthelion:
Good catch on the missing Git-version floor. That `blob:limit` filter has bitten portability before and the repo really should document a minimum. It's outside this PR's diff so I'll spin it into a separate issue rather than scope-creep here. Thanks for the thorough pass.

beardthelion:
@kevincodex1 this is ready. CI is green across the board (stable/beta, fmt+clippy, build, audit, MSRV, CodeQL, CodeRabbit clean) and mergeable. jatmn's note is a portability observation about sync.rs, which this PR doesn't touch; I'm tracking it as a separate issue.

jatmn (approved):
No findings.

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM!

PR #68, opened 2026-06-20
feat(node): owner-only push enforcement behind GITLAWB_ENFORCE_OWNER_PUSH (#31)

### LINKED ISSUE #31: Repo write authorization (Phase 1): owner-only push behind an opt-in flag
## Background

The README ("Known limitations") and `docs/OSS-READINESS-AUDIT.md` both note that push is authenticated but not authorized:

> *"Push authorization is still not capability-complete. A valid DID signature is authentication, not authorization; unprotected repo branches do not yet enforce owner/UCAN capability checks."*

`docs/MAINTAINER-ROADMAP.md` lists this under Security hardening as the first item. This issue proposes a concrete, non-breaking first slice and offers to implement it.

## Current behavior

For `POST /{owner}/{repo}/git-receive-pack`:

- `require_signature` (`auth/mod.rs`) verifies an RFC 9421 Ed25519 signature and injects `AuthenticatedDid`.
- In `git_receive_pack` (`api/repos.rs`), an owner check only runs **when the target branch is protected** (`is_branch_protected`). Pushes to **unprotected** branches are accepted from any authenticated signer.

Because `did:key` is self-certifying, "authenticated" only proves the caller controls the key for the DID they present — any party can generate a keypair, derive its `did:key`, sign, and push. There is no allow-list and no registration requirement on this path. Net effect: a non-owner can write to unprotected branches of an arbitrary repo.

`protect` and `visibility` endpoints already gate on the owner via a `require_owner`-style check, so the idiom exists; it's just not applied to the general push path.

## Proposal — Phase 1 (this work)

Follow the project's established "opt-in hardening flag before mandatory" pattern (cf. `GITLAWB_REQUIRE_SIGNED_PEER_WRITES`):

- Add `GITLAWB_ENFORCE_OWNER_PUSH` (default `false` -> current behavior, no break to live/test nodes).
- When `true`: in `git_receive_pack`, require the authenticated DID to equal the repo `owner_did` (matching the existing full-DID-or-short-form comparison used in branch protection/visibility) before any ref update is applied; reject unsigned/non-owner pushes with a clear error.
- Factor the repeated owner-match logic into one small shared helper (currently duplicated across branch protection, `protect`, and `visibility`).
- Tests: first coverage for the push authorization path (owner allowed, non-owner rejected, flag off = legacy behavior).
- Docs: `.env.example` + a line in `docs/RUN-A-NODE.md`.

## Out of scope (proposed Phase 2)

- Collaborator write lists and **UCAN `git/push` capability delegation** (the `caps::GIT_PUSH` type already exists; the UCAN chain is validated structurally in `require_ucan_chain` but capabilities aren't yet checked against the operation). This would let non-owners be granted scoped push rights — the natural escape hatch once owner-only is the default.
- Making enforcement the default after a compatibility window.

## Related (not asserted, worth a look)

`pulls/{n}/merge` and `hooks` are write endpoints on the same auth layer; they may warrant the same owner/capability gating. Happy to audit separately to keep this PR focused.

## Offer

I'd like to implement Phase 1 if the approach (opt-in flag, owner-only, helper extraction) looks right to you. Open to a different flag name / default / error semantics — flagging the design here first per CONTRIBUTING before sending a PR.

FrozenRaspberry:
Closes #31.

## What

Adds an opt-in flag **`GITLAWB_ENFORCE_OWNER_PUSH`** (default `false`) that requires the authenticated pusher to be the repo owner on **every** branch of `git-receive-pack`, not just protected ones.

## Why

A valid `did:key` RFC 9421 signature is **authentication, not authorization**: `did:key` is self-certifying, so anyone can generate a key, derive its DID, sign, and push. Today the owner check only runs when the target branch is `is_branch_protected`, so a non-owner can write to **unprotected** branches of an arbitrary repo. This is the first item under *Security hardening* in `docs/MAINTAINER-ROADMAP.md` and the Phase 1 slice agreed on in #31.

Default `false` (mirroring `GITLAWB_REQUIRE_SIGNED_PEER_WRITES`) means live nodes are unaffected until an operator opts in.

## How — addressing the review on #31

- **Authorize on the verified identity.** The handler now takes `Extension<AuthenticatedDid>` (the canonical DID injected by `require_signature`) instead of re-scraping `keyid` from the headers. The legacy branch-protection check is left as-is per your note.
- **Fail closed.** The new gate rejects on an absent identity and on a non-owner caller — no `unwrap_or(false)` fall-through. The decision is a pure fn (`owner_push_rejection`) so it's unit-testable without a DB or git backend.
- **No "reject unsigned" logic.** `require_signature` already rejects unsigned requests upstream; the gate only compares the authenticated DID against the owner.
- **Helper scoped to genuine owner-match sites.** Factored the duplicated owner-match into `auth::is_repo_owner(record, caller)`, reused by receive-pack, `protect.rs`, and `visibility.rs`. Each call site keeps its own error message. `events.rs` and `peers.rs` are intentionally left out — they aren't owner checks.
- **Named by intent.** The push-side gate is `caller_authorized_to_push`, so the Phase 2 UCAN `git/push` capability becomes a pure addition (`is_repo_owner(..) || ucan_grants_push(..)`) rather than a rewrite.
- **Ordering.** The gate runs **before** branch protection, so when the flag is on a non-owner is rejected with a single error body regardless of whether the branch is also protected.

## Out of scope (per your note)

`pulls/{n}/merge` and `hooks` warrant the same gating — kept out of this PR so it stays scoped to owner-push; happy to take the broader write-path review on its own track.

## Testing

- `cargo fmt --all -- --check` clean; `cargo clippy -p gitlawb-node --all-targets -- -D warnings` clean.
- New unit tests for the push authorization path (all passing): owner allowed (full `did:key:…` **and** bare suffix form), non-owner rejected, **missing-DID rejected** (fail-closed), and flag-off legacy behavior.
- Full `gitlawb-node` bin suite: 113 passing. The only failures are 3 pre-existing `sync::tests` cases that fail with `invalid filter-spec 'blob:limit=10g'` from the local git binary — reproduced identically on a clean tree without this change, so unrelated.

## Docs

- `.env.example`: documented `GITLAWB_ENFORCE_OWNER_PUSH=false`.
- `docs/RUN-A-NODE.md`: new "Hardening: owner-only push" section (default off, when to enable, effect).

## Summary by CodeRabbit

* **New Features**
  * Added optional owner-only push enforcement. When enabled via configuration, only the repository owner can push changes; non-owner pushes are rejected with HTTP 403 before any ref updates.

* **Documentation**
  * Added hardening guide explaining the new owner-only push feature, including configuration instructions and current limitations with delegated push workflows.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,38 @@
+# cargo-audit configuration.
+#
+# cargo-audit reads its config from .cargo/audit.toml (relative to the working
+# directory) or ~/.cargo/audit.toml. It does NOT read a repo-root audit.toml,
+# and there is no CLI flag to point at a config file (`-f` is the lockfile path).
+# This file lives at .cargo/audit.toml for that reason.
+#
+# The audit job stays hard-fail: only the advisories explicitly listed below are
+# suppressed, so any NEW advisory still breaks CI. Each entry below has no
+# available upstream fix; everything fixable is fixed in Cargo.lock instead.
+#
+# Tracking: the hickory ignores are reachable accepted risk. A scheduled re-check
+# (.github/workflows/audit-schedule.yml) runs without these ignores and fails if
+# the lockfile moves off the pinned vulnerable versions while the ignores remain.
+
+[advisories]
+ignore = [
+    # hickory-proto 0.25.2 (pulled by libp2p-dns 0.44 -> hickory-resolver 0.25).
+    # REACHABLE: the node wraps its QUIC transport in the system DNS resolver
+    # (crates/gitlawb-node/src/p2p/mod.rs:238), so a malicious DNS response can
+    # trigger these DoS paths. No fix is available: the latest libp2p-dns on
+    # crates.io is 0.44 and pins hickory 0.25; the fix requires hickory >=0.26.1.
+    # REMOVE both once libp2p-dns ships a release using hickory 0.26 (verified
+    # 2026-06-21: cargo update -p hickory-proto locks 0 packages; still 0.25.2).
+    # Tracking: https://github.com/Gitlawb/node/issues/76
+    "RUSTSEC-2026-0118", # hickory NSEC3 closest-encloser unbounded loop
+    "RUSTSEC-2026-0119", # hickory O(n^2) name-compression CPU exhaustion
+
+    # rsa 0.9.10 (Marvin timing sidechannel). Present in Cargo.lock because sqlx
+    # resolves sqlx-mysql unconditionally, and sqlx-mysql depends on rsa. cargo
+    # audit scans the lockfile, so it flags rsa even though it is NOT linked into
+    # our binary: sqlx is built postgres-only (crates/gitlawb-node/Cargo.toml),
+    # so the MySQL RSA auth path is never compiled and there is no RSA oracle to
+    # attack. No upstream fix exists regardless. REMOVE this ignore if sqlx is
+    # ever built with the `mysql` feature, or any other consumer of rsa enters
+    # the build, at which point it becomes a real reachable advisory.
+    "RUSTSEC-2023-0071", # rsa Marvin attack (no fix; not linked in our build)
+]
--- a/.env.example
+++ b/.env.example
@@ -72,6 +72,12 @@ GITLAWB_BOOTSTRAP_DISABLE_SEEDS=false
 # rolling upgrades so existing live nodes can still communicate.
 GITLAWB_REQUIRE_SIGNED_PEER_WRITES=false
 
+# Require the authenticated pusher to be the repo owner on git-receive-pack.
+# A valid did:key signature is authentication, not authorization: anyone can
+# sign as their own DID. When true, pushes from a non-owner DID are rejected.
+# Keep false until the repo owner is ready for owner-only writes.
+GITLAWB_ENFORCE_OWNER_PUSH=false
+
 # Comma-separated libp2p multiaddrs.
 # Example: /ip4/1.2.3.4/udp/7546/quic-v1/p2p/12D3KooW...
 GITLAWB_P2P_BOOTSTRAP=
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,11 @@
+# Ownership for CI, release, and security-sensitive infrastructure.
+# A PR that edits these paths requires review from a code owner once
+# branch protection's "Require review from Code Owners" is enabled.
+
+/.github/                      @beardthelion
+/Dockerfile                    @beardthelion
+/Dockerfile.bins               @beardthelion
+/docker-compose.yml            @beardthelion
+/release-please-config.json    @beardthelion
+/.release-please-manifest.json @beardthelion
+/SECURITY.md                   @beardthelion
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,66 @@
+name: Bug report
+description: Report a defect in a Gitlawb crate or node behavior
+labels: ["bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for the report. Please fill in the fields below so we can reproduce the issue.
+        For security vulnerabilities, do NOT open an issue — follow SECURITY.md instead.
+  - type: textarea
+    id: what-happened
+    attributes:
+      label: What happened?
+      description: A clear description of the bug.
+    validations:
+      required: true
+  - type: textarea
+    id: expected
+    attributes:
+      label: Expected behavior
+    validations:
+      required: true
+  - type: textarea
+    id: repro
+    attributes:
+      label: Steps to reproduce
+      placeholder: |
+        1. ...
+        2. ...
+        3. ...
+    validations:
+      required: true
+  - type: dropdown
+    id: crate
+    attributes:
+      label: Affected crate
+      options:
+        - gitlawb-node
+        - gl
+        - git-remote-gitlawb
+        - gitlawb-core
+        - not sure / other
+    validations:
+      required: true
+  - type: input
+    id: version
+    attributes:
+      label: Version or commit
+      description: Output of the binary's `--version`, or the git commit you built.
+    validations:
+      required: true
+  - type: textarea
+    id: environment
+    attributes:
+      label: Environment
+      description: OS, Rust version (`rustc --version`), and anything else relevant.
+    validations:
+      required: false
+  - type: textarea
+    id: logs
+    attributes:
+      label: Relevant logs
+      description: Paste logs or errors. This will be rendered as code, no need for backticks.
+      render: shell
+    validations:
+      required: false
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,8 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Security vulnerability
+    url: https://github.com/Gitlawb/node/blob/main/SECURITY.md
+    about: Please report security issues privately per our security policy, not as a public issue.
+  - name: Questions & discussion
+    url: https://github.com/Gitlawb/node/discussions
+    about: For usage questions and open-ended discussion.
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -1,0 +1,34 @@
+name: Feature request
+description: Suggest an improvement or new capability
+labels: ["enhancement"]
+body:
+  - type: textarea
+    id: problem
+    attributes:
+      label: Problem / use case
+      description: What are you trying to do that's hard or impossible today?
+    validations:
+      required: true
+  - type: textarea
+    id: proposal
+    attributes:
+      label: Proposed solution
+    validations:
+      required: true
+  - type: textarea
+    id: alternatives
+    attributes:
+      label: Alternatives considered
+    validations:
+      required: false
+  - type: dropdown
+    id: protocol
+    attributes:
+      label: Does this touch the protocol?
+      description: Identity, signatures, UCAN, ref certs, or P2P wire formats. Protocol changes need an issue discussion before any PR.
+      options:
+        - "No"
+        - "Yes"
+        - "Not sure"
+    validations:
+      required: true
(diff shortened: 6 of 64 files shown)


beardthelion (changes requested):
Took a full pass over the owner-push gate. The core is solid: it fails closed, runs ahead of branch protection, the `did:key` suffix match isn't spoofable since the authenticated caller is always the full `did:key:` form, and the new `Extension(auth)` extractor is safe because the route sits behind `require_signature`. `owner_push_rejection` being a pure, unit-tested function is the right call. A few things to tighten before it's ready to merge:

## Findings

- [P2] Return 403 for the authorization denial, not 400
  `crates/gitlawb-node/src/api/repos.rs:484`
  `owner_push_rejection` rejects with `AppError::BadRequest`, but `AppError::Forbidden` already exists and maps to 403, the correct status for "authenticated but not the owner." 400 reads as a malformed request and some git/CI clients retry it. The same applies to the branch-protection denial and the owner checks in `protect.rs` / `visibility.rs`, so this is cleanest as one small sweep.

- [P2] Fold the remaining inline owner-matches into `is_repo_owner`
  `crates/gitlawb-node/src/api/repos.rs:549`
  This PR adds `is_repo_owner` and routes protect/visibility through it, but two hand-rolled `split(':').next_back()` checks remain: the branch-protection check here (which also reads the pusher DID from `extract_did_from_auth(&headers)` rather than the verified `auth.0`), and the list filter at `repos.rs:165`. Routing both through the helper finishes the consolidation and drops the second identity source on the push path.

- [P2] Assert the error variant in the rejection tests
  `crates/gitlawb-node/src/api/repos.rs:1315`
  The rejection tests assert `.is_some()`, so the 400 to 403 change above would pass silently. Match the variant instead. Worth adding a handler-level test that runs `git_receive_pack` with the flag on so a wiring regression on `Extension(auth)` is caught, plus direct tests for `is_repo_owner` / `caller_authorized_to_push`.

- [P3] Warn that enabling the flag blocks delegated and CI agents
  `docs/RUN-A-NODE.md:478`
  The route verifies a UCAN `git/push` chain in `require_ucan_chain` then discards it; `caller_authorized_to_push` only accepts the literal owner. So turning this on today hard-blocks every delegated or CI agent that isn't the owner, and the `gl ucan_delegate` path for `git/push` is a dead end until Phase 2. One line of caution in the hardening section would save someone locking out their own automation.

The `path_glob` validation issue is pre-existing (identical on main, untouched here) and filed separately as #74, so it's out of scope here.

FrozenRaspberry:
Thanks for the thorough pass. All addressed in `f9df6eb`:

**[P2] 403 not 400** — Done as one sweep: the owner-push gate, the branch-protection denial, and the owner checks in `protect.rs` / `visibility.rs` now return `AppError::Forbidden` (403).

**[P2] Finish the `is_repo_owner` consolidation** — The branch-protection check and the list filter at `repos.rs:165` now both route through `is_repo_owner`. The branch-protection check uses the verified `auth.0` rather than `extract_did_from_auth(&headers)`. I went one step further and also switched the post-push event-recording site to `auth.0`, which let me delete `extract_did_from_auth` (and the handler's `HeaderMap` param + import) outright — so there's a single identity source on the push path now. Shout if you'd rather I'd left the event-recording site alone.

**[P2] Assert the variant** — The rejection tests now match `Some(AppError::Forbidden(_))` (via an `assert_forbidden` helper), so the 400→403 change would have failed loudly. Added direct tests for `is_repo_owner` and `caller_authorized_to_push` (owner full + short DID, non-owner).

On the **handler-level `git_receive_pack` test**: I looked at wiring one up but it isn't runnable in CI as-is. `git_receive_pack` calls `state.db.get_repo(...)` before reaching the gate, and `pr-checks.yml` runs `cargo test --workspace` with no Postgres service (the test harness uses lazy placeholder pools), so a handler test would fail at the DB call rather than exercise the gate. The `Extension(auth)` wiring is at least compile-enforced — the gate consumes `auth.0`, so removing the extractor won't build. If you'd like a real end-to-end handler test, I'm happy to add a Postgres service to `pr-checks.yml` and write it on its own track — just didn't want to bundle a CI-infra change into this PR. Let me know your preference.

**[P3] Doc caution** — Added to the hardening section: enabling the flag blocks every non-owner pusher including delegated/CI agents (UCAN `git/push` is verified but not yet honored for authz), with a "don't enable until every pusher is the owner" warning.

Re #74 (the `path_glob` validation bypass) — agreed it's pre-existing and out of scope here; thanks for filing it separately.

beardthelion (dismissed):
Approving. I verified the security core against the code rather than the PR summary, and confirmed the prior round's four findings are fully resolved.

What I checked:

- Identity on the push path is signature-bound: the verifying key is derived from the keyid, the signature is checked against it, and `AuthenticatedDid` is set only after every check passes (`crates/gitlawb-node/src/auth/mod.rs:257`). The route sits behind `require_signature` via `add_auth_layers` (`crates/gitlawb-node/src/server.rs:159`), so `auth.0` cannot be spoofed or absent at the handler.
- The gate fails closed and runs before branch protection, `acquire_write`, and the receive-pack spawn.
- `is_repo_owner` short/full matching has no forgeable collision: the `z6Mk…` form is a 1:1 encoding of the Ed25519 key and the push caller always arrives as the full verified DID.
- All four items from the prior round are fully resolved: 403 at every owner-mismatch site, the consolidation routes through `is_repo_owner` on the verified `auth.0` (`crates/gitlawb-node/src/api/repos.rs:165` and `:547`), `assert_forbidden` pins the variant so a regression to 400 fails loudly, and the docs caution is in place. `extract_did_from_auth` is gone with no dangling references. Build is green, clippy is clean, and the 7 new tests pass.

None of the following block the merge. Noting them so they are tracked.

## Findings (non-blocking)

- [P3] Extend the owner-match consolidation one hop further. `crates/gitlawb-node/src/visibility.rs:17` still carries its own `is_owner` with the same `split(':').next_back()` idiom. It predates this PR and is behavior-identical, but a shared `is_did_owner(owner_did, caller)` that both it and `is_repo_owner` delegate to would close the last copy.
- [P3] Document that `GITLAWB_ENFORCE_OWNER_PUSH` takes effect on restart only. It is read once into `Config` at startup (`crates/gitlawb-node/src/config.rs:58`), so flipping it on a live node is a no-op until restart. A one-line note in `RUN-A-NODE.md` and `.env.example` would set the right expectation.
- [P3] The rejection `warn!` fires on every blocked push (`crates/gitlawb-node/src/api/repos.rs:523`). When the flag is intentionally on, rejections are expected traffic, so a probing or misconfigured client can inflate log volume. `info!` or a per-DID rate-limited line would be calmer.
- [P3, pre-existing] The git remote helper bails on a non-2xx without reading the body (`crates/git-remote-gitlawb/src/main.rs:297`), so the descriptive rejection message never reaches the pushing user. Filed as #77.

Phase 2 (tracked on #31, not this PR): once UCAN `git/push` is honored, add `owner_did` and a structured error code to the 403 body so agent clients can act on the rejection rather than just log it, and plan the capability-injection path from `require_ucan_chain` into the push authorization check.

@kevincodex1 this is ready when you are. The security core is sound, the prior round is fully resolved, and everything above is a non-blocking follow-up.

kevincodex1:
hi @FrozenRaspberry thank you for this contribution. kindly rebase to main and please fix conflicts.

FrozenRaspberry:
Resolved the merge conflict with `main` (8ab7e50).

`main` landed #87 (per-route authorization), which introduced the canonical `crate::api::did_matches` owner comparison — DID-safe on **both** sides (full `did:key:` vs bare suffix), and cross-method-safe — and routed `protect.rs` / `visibility.rs` through it. My branch had added a weaker `auth::is_repo_owner` (owner side only) for the same purpose.

Reconciled onto `main`'s helper rather than keeping mine:
- Removed `auth::is_repo_owner`; `caller_authorized_to_push` now delegates to `did_matches`, keeping the intent-named Phase 2 seam (`did_matches(..) || ucan_grants_push(..)`).
- Branch-protection check and the repos list filter now use `did_matches` too.
- Dropped my now-redundant `is_repo_owner` unit test (the canonical helper has its own tests in `api`); kept the `caller_authorized_to_push` + `owner_push_rejection` (Forbidden-variant) tests.

This also closes the loop on the prior P3 "fold the last owner-match into a shared helper" — there's now a single owner-comparison (`did_matches`) across receive-pack, branch protection, protect, and visibility. The `git_receive_pack` owner-push gate itself is unchanged.

`cargo fmt`/`clippy -D warnings` clean. The owner-push, `caller_authorized_to_push`, and `did_matches` tests pass locally; the remaining local failures are the Postgres-backed `#[sqlx::test]` suite (no local DB) and the pre-existing `blob:limit=10g` sync tests — both environmental and green under CI's Postgres service.

beardthelion (approved):
Re-approving on the merge head `8ab7e50a`. The reconciliation onto `did_matches` (removing `is_repo_owner`) is substantive, so I re-verified the security core against the merged code rather than the diff summary.

What I checked:

- `did_matches` (`crates/gitlawb-node/src/api/mod.rs:63`) is sound on both sides: exact-match short-circuit, strips `did:key:` from each operand, and the `!ka.contains(':') && !kb.contains(':')` guard blocks cross-method collisions (`did:gitlawb:X`/`did:web:X` keep their colon after the no-op strip, so they cannot match a bare `X`). It is symmetric, so the differing argument order between the push gate and the branch-protection check is harmless.
- The owner-push gate fails closed and runs before branch protection, `acquire_write`, and the receive-pack spawn (`repos.rs:531` < `557` < `577`). `auth.0` is the verified `sig.key_id` from `require_signature`, and `git_receive_pack` is wired through `add_auth_layers`, so the non-`Option` `Extension(auth)` extractor cannot be absent or header-spoofed.
- `extract_did_from_auth` is fully removed with no remaining callers — this closes the prior header-reparse divergence. The cert / p2p / webhook-payload paths now use the verified DID directly; nothing depended on the old `"unknown"`/empty sentinels.
- The dropped `is_repo_owner` unit test is covered by `did_matches`'s own tests; the retained owner-push and Forbidden-variant tests pass. Build green, clippy clean.

Two non-blocking follow-ups, both cheap:

## Findings

- [P2] Enroll `git_receive_pack` in the owner-gate guard test. `crates/gitlawb-node/src/api/mod.rs:171` — the new opt-in push gate is the one in-scope mutation handler with no row in `authz_guard::every_in_scope_mutation_has_its_gate`. That table is the mechanical guard against a gate being silently removed, so the new gate should be enrolled: add `(repos, "git_receive_pack", "owner_push_rejection(")`. The marker is real code in the handler body (`repos.rs:531`), so it satisfies the comment-stripped `fn_body` check.
- [P3] Unit-test the branch-protection rejection arm. `crates/gitlawb-node/src/api/repos.rs:557` — the `did_matches` substitution and the 400→403 change on the protected-branch path have no test. Correct by inspection, but worth pinning: protected + non-owner → Forbidden, protected + owner → passes, unprotected + non-owner → passes.

Neither blocks. @kevincodex1 this is ready when you are.

kevincodex1 (approved):
LGTM

PR #69, opened 2026-06-20
fix(node): preserve promisor mirror mode on unknown withheld-paths lookup (#48)

### LINKED ISSUE #48: Peer sync downgrades still-withheld mirrors to Plain on a transient withheld-paths error
Surfaced by CodeRabbit on #46 (review spans the full stack; this is peer-sync code from #35).

In `sync.rs` the per-repo sync does:

```rust
let withheld = fetch_withheld(client, &origin_url, owner_short, repo_name).await;
let mode = classify_mirror(withheld);
```

`fetch_withheld` collapses 404, 5xx, network errors, and parse failures all into `None`, and `classify_mirror(None)` yields `MirrorMode::Plain`. For an existing promisor mirror that flows into `fetch_repo(.., Plain)`, which clears `remote.origin.promisor` and `remote.origin.partialclonefilter` before refetching. So a transient `withheld-paths` outage against a still-withheld repo downgrades the mirror to a full clone and starts failing syncs until the endpoint recovers.

Only downgrade to `Plain` on a definite `Some(empty)`. When the lookup is unknown (`None` from an error), preserve the repo's existing promisor/partial-clone mode instead of reclassifying.

Owning PR: #35.

beardthelion:
## Summary

Peer sync downgraded a still-withheld promisor mirror to a full clone on a transient `withheld-paths` outage. This makes the mirror-mode decision distinguish "unknown lookup" from "genuinely public" and preserve the existing promisor mode when the lookup is unknown.

## Motivation & context

Closes #48

`process_batch` decided how to mirror a repo from the origin's anonymous `withheld-paths` answer. `fetch_withheld` returns `Some(globs)` on a 2xx and `None` on every failure (404, 5xx, network, parse). `classify_mirror` then mapped `Some(non-empty) => Promisor` but `_ => Plain`, collapsing both `Some(empty)` (genuinely public) and `None` (unknown) into `Plain`. For an existing promisor mirror that ran `fetch_repo(.., Plain)`, which unset `remote.origin.promisor` + `remote.origin.partialclonefilter` and ran `git fetch --refetch`. So a transient outage of the `withheld-paths` endpoint against a still-withheld repo tore down the partial-clone config and forced a full refetch against an origin that still withholds blobs for anonymous callers, breaking syncs until the endpoint recovered. Sibling of #47, same fail-open-vs-fail-closed family.

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

gitlawb-node (`crates/gitlawb-node/src/sync.rs`):

- Replace `classify_mirror(withheld)` with `resolve_mirror_mode(withheld, exists, promisor)`. `Some(non-empty) => Promisor` and `Some(empty) => Plain` are unchanged. On `None`, an existing mirror preserves its promisor mode instead of downgrading; a fresh clone stays `Plain` (the existing fail-closed-at-the-git-layer behavior). A genuine public transition always returns `Some(empty)`, never `None`, so preserving on `None` cannot mask one; it only suppresses the transient-error false positive.
- Add a three-valued `PromisorProbe` (`Promisor` / `NotPromisor` / `Unknown`) read from `git config --get remote.origin.promisor` exit codes. `git config --get` collapses "key absent" (exit 1) and "git errored" (exit ≥2 / spawn failure) into the same non-zero exit otherwise; distinguishing them means a transient probe failure (`Unknown`) is not read as a definitive `NotPromisor` and cannot itself trigger the downgrade.
- Probe only when the lookup is unknown and the mirror already exists (the sole case the answer changes the mode). One `warn!` marks the preserve path, derived from the resolved mode so it cannot drift from the decision.

A persistently unreachable `withheld-paths` route pins an existing promisor mirror until the endpoint recovers (the per-cycle warn is the operator-visible signal). This is an intentional confidentiality-over-liveness bias: a still-withheld repo's anonymous git read endpoint 404s, so a preserved mirror fetches nothing harmful, and only a public-but-route-permanently-dropped repo loses backfill liveness.

## How a reviewer can verify

```sh
cargo test -p gitlawb-node sync
cargo test --workspace
```

The core regression is `resolve_preserves_promisor_on_unknown_lookup_for_existing_mirror`: `None` + existing + on-disk promisor must stay `Promisor`, never downgrade. `resolve_preserves_promisor_on_indeterminate_probe` covers the probe-failure (`Unknown`) case. `probe_reports_promisor_for_promisor_mirror` / `probe_reports_not_promisor_when_key_absent` / `probe_reports_unknown_on_git_error` exercise the three probe outcomes against real git repos. `plain_fetch_clears_promisor_config_on_transition` is unchanged and still passes, proving a genuine `Some(empty)` public transition still clears promisor config.

## Before you request review

## Notes for reviewers

Out of scope (follow-ups): typed error reasons in `fetch_withheld` (404 vs 5xx vs parse) and retry/backoff on a failed lookup; both deferred deliberately since the empty-vs-unknown distinction the fix needs already survives in the `Option` return. The `withheld-paths` endpoint is unauthenticated (pre-existing); node-level auth for it is worth a separate look.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed unsafe mirror mode downgrade behavior when origin lookup fails, preventing potential data synchronization issues (resolves issue `#48`).
  * Improved robustness of mirror mode detection with better handling of uncertain lookup states.

* **Improvements**
  * Enhanced existing mirror detection logic to preserve promisor mode when lookups are inconclusive, ensuring safer fallback behavior.

### DIFF
--- a/crates/gitlawb-node/src/sync.rs
+++ b/crates/gitlawb-node/src/sync.rs
@@ -32,19 +32,50 @@ enum MirrorMode {
     Promisor,
 }
 
-/// Decide the mirror mode from the origin's `withheld-paths` response.
+/// The on-disk promisor state of an existing mirror, read from
+/// `remote.origin.promisor`. Three-valued so a git error is not mistaken for a
+/// definitive "not a promisor": `git config --get` collapses "key absent" and
+/// "git failed" into the same non-zero exit otherwise, and treating an errored
+/// probe as `NotPromisor` would let a transient failure downgrade a still-withheld
+/// mirror (issue #48).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum PromisorProbe {
+    /// `remote.origin.promisor` is `"true"`.
+    Promisor,
+    /// The key is absent (git exit 1) or set to a non-`true` value.
+    NotPromisor,
+    /// The probe itself failed (git spawn error or other non-zero exit).
+    Unknown,
+}
+
+/// Decide the mirror mode from the origin's `withheld-paths` response plus, when
+/// the response is unknown, the existing mirror's on-disk promisor state.
 ///
 /// `Some(non-empty)` → the repo has a private subtree → `Promisor`.
 /// `Some(empty)`     → fully public → `Plain`.
-/// `None`            → the lookup 404'd or failed. Attempt a `Plain` mirror; a
-///                     mode-A repo also 404s the git read endpoint, so the clone
-///                     fails and nothing is mirrored (fail-closed at the git
-///                     layer), while a public repo on a peer that predates the
-///                     `withheld-paths` route still gets mirrored.
-fn classify_mirror(withheld: Option<Vec<String>>) -> MirrorMode {
+/// `None`            → the lookup 404'd or failed; the answer is *unknown*, which
+///                     is not the same as "public". For a fresh clone this stays
+///                     `Plain` (a mode-A repo also 404s the git read endpoint, so
+///                     the clone fails and nothing is mirrored — fail-closed at the
+///                     git layer — while a public repo on a peer that predates the
+///                     `withheld-paths` route still gets mirrored). For an existing
+///                     mirror it *biases toward preserving* the promisor state: a
+///                     genuine public transition returns `Some(empty)`, so on
+///                     `None` we cannot distinguish "still withheld, unreachable"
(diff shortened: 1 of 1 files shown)


kevincodex1 (approved):
LGTM

PR #70, opened 2026-06-20
fix(gl): paginate gl-clone Arweave recovery and make /encrypted-blobs parsing schema-strict (#49)

### LINKED ISSUE #49: gl clone Arweave recovery: paginate manifest discovery and make /encrypted-blobs parsing schema-strict
Surfaced by CodeRabbit on #46 (review spans the full stack; this is B3 Arweave-recovery code from #36).

Two issues in the `gl clone` Arweave fallback recovery path (`crates/gl/src/clone.rs`):

1. **Discovery truncates at 100 transactions.** The GraphQL manifest query hard-codes `first:100` and stops. Once a repo has more than 100 encrypted-manifest anchors (one per push), any private blob whose most recent seal falls outside that window drops out of recovery even though it is still live at HEAD. Paginate (cursor/`after` + `pageInfo`) until coverage is complete before merging per-push manifests.

2. **`/encrypted-blobs` parsing is fail-open on schema drift.** `recover_encrypted_blobs` parses into `serde_json::Value` and defaults `blobs` to `[]` via `unwrap_or_default()`. `/withheld-paths` by contrast deserializes a typed struct and errors on missing/wrong fields. Schema drift on `/encrypted-blobs` would silently yield "nothing to recover" rather than an error. Switch it to a strict typed response.

Owning PR: #36.

beardthelion:
## Summary

`gl clone`'s Arweave fallback recovery dropped private blobs whose most recent seal fell outside the first 100 manifest anchors, and treated an `/encrypted-blobs` schema regression as "nothing to recover". This paginates discovery and makes the parse fail closed.

## Motivation & context

Closes #49

Surfaced by CodeRabbit on #46; this is the B3 Arweave-recovery code from #36. Two defects in `crates/gl/src/clone.rs`:

1. Discovery hard-coded a single `first:100` GraphQL query with no pagination. One manifest is anchored per push, so a repo with more than 100 pushes silently lost recovery coverage for any blob whose latest seal fell outside that window, even though the blob is live at HEAD.
2. `/encrypted-blobs` parsed the 200 body into a `serde_json::Value` and defaulted `blobs` to `[]` via `unwrap_or_default()`, so a missing or renamed field yielded a silent empty recovery instead of an error. The sibling `/withheld-paths` path already deserializes a typed struct and errors on drift.

## Kind of change

- [x] Bug fix
- [ ] Feature
- [x] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

Crate touched: `gl`.

- Strict typed parse of the node `/encrypted-blobs` 200 body (`EncryptedBlobsResponse`) so schema drift is a hard error instead of a silent empty result, mirroring the existing `WithheldPathsResponse` convention. The sole caller surfaces the error rather than swallowing it into `unwrap_or_default()`.
- Cursor-paginate Arweave manifest discovery (`pageInfo` + `after`) until coverage is complete, replacing the single `first:100` query.
- Bound discovery against a hostile or buggy gateway: 1000-page and 10k-transaction caps, a per-page edge bound, cross-page dedup, and a non-advancing-cursor guard. Incomplete discovery warns rather than silently truncating.
- Resolve equal-timestamp merge ties by Arweave block height (defensively parsed: a not-yet-mined anchor counts as newest, an unreadable height ranks lowest), preserving the pre-pagination "newest wins" behavior independent of page order.

## How a reviewer can verify

```sh
cargo test -p gl --bin gl clone::
```

30 clone tests cover the strict-parse matrix, multi-page discovery, cross-page dedup (fetched once), non-advancing-cursor termination, the per-page bound, and the block-height tie-break cases (distinct height, pending anchor, same-block, empty timestamp).

## Before you request review

## Notes for reviewers

- Node-side anchoring and the `/encrypted-blobs` endpoint are unchanged; this is read-path only.
- Block height is gateway-reported, not independently verified. It removes page-order dependence and honest-node clock skew between nodes, not trust in the gateway (discovery already assumes an honest gateway). Ranking on block height as the primary signal instead of the node-reported timestamp is recorded as a deliberate follow-up.
- Known follow-up: `clone.rs` was already over 1000 lines before this change; extracting the discovery loop into a helper to bring it back down was raised in review and deferred to keep this PR focused.

## Release Notes

* **Bug Fixes**
  * Enhanced recovery validation to catch missing or incorrect data earlier, preventing silent failures.
  * Improved fallback recovery robustness with safeguards against infinite loops and transaction ID deduplication.
  * Strengthened error handling in recovery processes.

* **Tests**
  * Expanded test coverage for recovery mechanisms and edge case scenarios.

### DIFF
--- a/crates/gl/src/clone.rs
+++ b/crates/gl/src/clone.rs
@@ -373,6 +373,11 @@ struct TxRef {
     height: Option<u64>,
 }
 
+/// Edges requested per Arweave GraphQL page (`first:` in the discovery query) and
+/// the per-page bound enforced when parsing a response. The `first:N` literal in
+/// the query string MUST equal this constant.
+const ARWEAVE_PAGE_SIZE: usize = 100;
+
 /// A parsed page of an Arweave GraphQL `transactions` response: the edge refs
 /// plus pagination state. `has_next` is `None` when the response omitted
 /// `pageInfo` entirely (an older or partial gateway), which the discovery loop
@@ -387,6 +392,11 @@ struct TxPage {
 /// height, plus `pageInfo` for cursor pagination. Hand-walks the `Value` (the
 /// envelope nests `data.transactions.{edges,pageInfo}`) rather than deriving a
 /// wrapper struct, matching the surrounding style.
+///
+/// At most `ARWEAVE_PAGE_SIZE` edges are taken from a single page: the query asks
+/// for `first:100`, so a response with more edges is a misbehaving (or hostile)
+/// gateway. Bounding here caps per-page allocation and the downstream per-id fetch
+/// loop so the `MAX_TX_IDS` budget can't be defeated by one oversized page.
 fn parse_tx_page(v: &serde_json::Value) -> TxPage {
     let txns = v.get("data").and_then(|d| d.get("transactions"));
     let refs: Vec<TxRef> = txns
@@ -395,6 +405,7 @@ fn parse_tx_page(v: &serde_json::Value) -> TxPage {
         .map(|edges| {
             edges
                 .iter()
+                .take(ARWEAVE_PAGE_SIZE)
                 .filter_map(|edge| {
                     let node = edge.get("node")?;
                     let id = node.get("id").and_then(|i| i.as_str())?.to_string();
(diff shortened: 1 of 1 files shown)


kevincodex1 (approved):
LGTM!

PR #73, opened 2026-06-20
fix(node): dedupe mirror and canonical repo rows on list surfaces (#6)

### LINKED ISSUE #6: Deduplicate mirrored repo rows with canonical did:key owner on profile/list surfaces
## Summary

`node.gitlawb.com` can show the same logical repo twice for one agent when a short-owner peer mirror row and a canonical `did:key:` repo row both exist.

This is visible on the public profile for nipmod:

https://gitlawb.com/z6MkwbuduCUUwy8fp78CZ2pnhLyRSibkSjcCGexT355xNw5R

The profile currently renders two `z6Mkwbud / nipmod` cards, both linking to the same repo path:

https://gitlawb.com/node/repos/z6Mkwbud/nipmod

## Repro

```sh
curl -fsSL 'https://node.gitlawb.com/api/v1/repos?owner=z6MkwbuduCUUwy8fp78CZ2pnhLyRSibkSjcCGexT355xNw5R'
```

The response currently contains two records for `nipmod`:

```json
{
  "id": "z6MkwbuduCUUwy8fp78CZ2pnhLyRSibkSjcCGexT355xNw5R/nipmod",
  "name": "nipmod",
  "owner_did": "z6MkwbuduCUUwy8fp78CZ2pnhLyRSibkSjcCGexT355xNw5R",
  "description": "mirrored from peer"
}
```

```json
{
  "id": "9d92186a-c233-4e64-ac82-3dadf1de1eb1",
  "name": "nipmod",
  "owner_did": "did:key:z6MkwbuduCUUwy8fp78CZ2pnhLyRSibkSjcCGexT355xNw5R",
  "description": "Decentralized npm for agents on Gitlawb"
}
```

Direct canonical lookup resolves the canonical row:

```sh
curl -fsSL 'https://node.gitlawb.com/api/v1/repos/z6Mkwbud/nipmod'
```

## Likely cause

The mirror import path appears to create a local repo row using the short DID owner:

- `upsert_mirror_repo(...)` sets `id = "{owner_short}/{name}"`, `owner_did = owner_short`, and `description = "mirrored from peer"`.
- `list_repos` returns all matching rows after accepting both full and short owner forms.
- The profile/list surface then renders both rows as separate repos.

Source references from current `gitlawb/node` main:

- `crates/gitlawb-node/src/db/mod.rs`: `upsert_mirror_repo`
- `crates/gitlawb-node/src/api/repos.rs`: `list_repos`
- `crates/gitlawb-node/src/server.rs`: no public repo delete route
- `crates/gl/src/repo.rs`: no `repo delete`/`repo unpublish` CLI command

## Expected behavior

Profile and repo list surfaces should show one logical repo per normalized owner/repo.

If both records exist:

- canonical `did:key:z.../repo` should win for title, description, star count and canonical metadata;
- short-owner `z.../repo` mirror evidence should remain accessible as mirror/replica metadata, not as a second repo card.

## Why this matters

For projects using Gitlawb as canonical public source, the duplicate looks like an accidental fork or duplicate publication, even though both cards resolve to the same repo path. It is especially confusing for founder review, package registry tooling and public launch links.

## Safe fix direction

I would avoid client-side cleanup because there is no public repo delete API and `ReplicaUnregister` only removes replica table rows, not local mirror repo rows.

Safer options:

1. Normalize owner IDs during list/profile rendering and dedupe by `(normalized_owner_did, name)`.
2. Prefer canonical `did:key:` rows over short-owner mirror rows when both exist.
3. Optionally add an operator migration to merge/delete short-owner mirror rows only when the canonical row exists and resolves.

beardthelion:
## Summary

Profile and repo-list surfaces rendered the same logical repo twice when a short-owner peer mirror row and the canonical `did:key:` row both existed. This collapses them to one card on the surfaces that were missing the dedup.

## Motivation & context

Closes #6

`node.gitlawb.com` showed two `nipmod` cards on a profile: one from the peer mirror row (`owner_did = "z6Mk…"`, description "mirrored from peer") and one from the canonical row (`owner_did = "did:key:z6Mk…"`). The paged repo list already deduped these in SQL, but the non-paged `GET /api/v1/repos` legacy path and `list_federated_repos` returned every matching row, so both showed up.

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

Crate touched: `gitlawb-node`.

- Added `dedupe_canonical_repos` in `api/repos.rs`: groups rows by `(normalized owner, name)` (the key segment after the last `:`, so `did:key:z6Mk…` and the bare `z6Mk…` mirror row collapse together), keeps the canonical row (non-mirror beats "mirrored from peer", ties broken by earliest `created_at`), and carries the group's most recent `updated_at` onto the survivor so a gossip push that only touched the mirror row still floats the repo to the top. This matches the existing SQL dedup in `Db::list_all_repos_paged`.
- Applied it at the two non-paged surfaces: the legacy `list_repos` fallback and `list_federated_repos`. As a side effect the legacy path's `X-Total-Count` now counts logical repos rather than raw rows, consistent with the paged path.
- Added a `repos::tests` module covering canonical-wins, distinct-repos-preserved, same-owner-different-repo, and the mirror tie-break.

## How a reviewer can verify

```sh
cargo test --bin gitlawb-node repos::tests
# Against a node that has both a mirror and a canonical row for one repo:
curl -fsSL 'http://<node>/api/v1/repos?owner=z6Mkwbud...'  # one record, owner_did = did:key:..., real description
```

## Before you request review

## Notes for reviewers

The dedup logic now lives in two places: the SQL `DISTINCT ON` in `Db::list_all_repos_paged` and this Rust helper for the non-paged surfaces. They use the same preference rules and the helper's doc comment flags that they must stay in sync. Consolidating both behind one path is possible later but would change the legacy "return all rows" contract that peer/CLI callers rely on, so I kept it out of scope.

## Summary by CodeRabbit

* **Bug Fixes**
  * Repository listings now properly deduplicate canonical and mirrored entries, so each logical repository shows once.
  * Total counts and “most recent activity” ordering now reflect the deduplicated view (with deterministic tie-breaking).
* **API Updates**
  * GraphQL repository results use the deduplicated dataset.
  * The `/api/v1/stats` `repos` metric now counts logical repositories consistently with deduplication.
* **Tests**
  * Added coverage for canonical-vs-mirror selection, deterministic ordering, did-key-aware grouping, and empty-table behavior.

### DIFF
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -6,7 +6,8 @@ use bytes::Bytes;
 use std::sync::Arc;
 
 use crate::auth::{caller_authorized_to_push, AuthenticatedDid};
-use chrono::Utc;
+use crate::db::RepoRecord;
+use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
@@ -227,8 +228,8 @@ pub async fn list_repos(
     }
 
     let repos = state.db.list_all_repos_with_stars().await?;
-    let filtered: Vec<_> = repos
-        .iter()
+    let filtered: Vec<(RepoRecord, i64)> = repos
+        .into_iter()
         .filter(|(r, _)| {
             if let Some(owner) = &query.owner {
                 crate::api::did_matches(owner.as_str(), &r.owner_did)
@@ -237,10 +238,11 @@ pub async fn list_repos(
             }
         })
         .collect();
-    let total = filtered.len() as i64;
-    let resp: Vec<_> = filtered
+    let deduped = dedupe_canonical_repos(filtered);
+    let total = deduped.len() as i64;
+    let resp: Vec<_> = deduped
         .into_iter()
-        .map(|(r, stars)| to_response(r, &state, *stars))
+        .map(|(r, stars)| to_response(&r, &state, stars))
         .collect();
     let mut response = Json(resp).into_response();
     response.headers_mut().insert(
@@ -1121,7 +1123,7 @@ pub async fn list_refs(
 pub async fn list_federated_repos(
     State(state): State<AppState>,
 ) -> Result<Json<serde_json::Value>> {
-    let local_repos = state.db.list_all_repos_with_stars().await?;
+    let local_repos = dedupe_canonical_repos(state.db.list_all_repos_with_stars().await?);
     let local_node_url = state
         .config
         .public_url
@@ -1379,11 +1381,90 @@ fn to_response(record: &crate::db::RepoRecord, state: &AppState, star_count: i64
     }
 }
 
+/// Description marker that `upsert_mirror_repo` stamps on short-owner peer mirror rows.
+const MIRROR_DESCRIPTION: &str = "mirrored from peer";
+
+/// Collapse short-owner mirror rows and canonical `did:key:` rows that point at the
+/// same logical repo into a single entry, so profile/list surfaces don't render the
+/// same repo twice (issue #6).
+///
+/// Rows are grouped by `(normalized owner, name)`, where the normalized owner is the
+/// key segment after the last `:` (so `did:key:z6Mk…` and the bare `z6Mk…` mirror row
+/// collapse together). Within a group the canonical row wins: a non-mirror row is
+/// preferred over a `"mirrored from peer"` row, ties broken by earliest `created_at`.
+/// The survivor inherits the group's most recent `updated_at` so a gossip push that
+/// only touches the mirror row still floats the repo to the top.
+///
+/// This mirrors the SQL dedup already applied on the paged path in
+/// `Db::list_all_repos_paged`; the two must stay in sync.
+fn dedupe_canonical_repos(rows: Vec<(RepoRecord, i64)>) -> Vec<(RepoRecord, i64)> {
+    use std::collections::HashMap;
+
+    fn is_mirror(r: &RepoRecord) -> bool {
+        r.description.as_deref() == Some(MIRROR_DESCRIPTION)
+    }
+
+    // Strictly more canonical: non-mirror beats mirror; on a tie the earlier row wins.
+    fn outranks(candidate: &RepoRecord, current: &RepoRecord) -> bool {
+        match (is_mirror(candidate), is_mirror(current)) {
+            (false, true) => true,
+            (true, false) => false,
+            _ => candidate.created_at < current.created_at,
+        }
+    }
+
+    // Preserve first-seen group order so output ordering stays deterministic.
+    let mut order: Vec<(String, String)> = Vec::new();
+    let mut winners: HashMap<(String, String), (RepoRecord, i64)> = HashMap::new();
+    let mut latest: HashMap<(String, String), DateTime<Utc>> = HashMap::new();
+
+    for (rec, stars) in rows {
+        let short = rec
+            .owner_did
+            .rsplit(':')
+            .next()
+            .unwrap_or(&rec.owner_did)
+            .to_string();
+        let key = (short, rec.name.clone());
+
+        latest
+            .entry(key.clone())
+            .and_modify(|u| {
+                if rec.updated_at > *u {
+                    *u = rec.updated_at;
+                }
+            })
+            .or_insert(rec.updated_at);
+
+        match winners.get(&key) {
+            None => {
+                order.push(key.clone());
+                winners.insert(key, (rec, stars));
+            }
+            Some((current, _)) if outranks(&rec, current) => {
+                winners.insert(key, (rec, stars));
+            }
+            Some(_) => {}
+        }
+    }
+
+    order
+        .into_iter()
+        .filter_map(|key| {
+            let max_updated = latest.get(&key).copied();
+            winners.remove(&key).map(|(mut rec, stars)| {
+                if let Some(u) = max_updated {
+                    rec.updated_at = u;
+                }
+                (rec, stars)
+            })
+        })
+        .collect()
+}
+
 #[cfg(test)]
 mod tests {
-    use super::owner_push_rejection;
-    use crate::auth::caller_authorized_to_push;
-    use crate::error::AppError;
+    use super::*;
 
     const OWNER_DID: &str = "did:key:z6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH";
     const OWNER_SHORT: &str = "z6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH";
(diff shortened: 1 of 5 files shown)


beardthelion:
@coderabbitai review

beardthelion:
@coderabbitai full review

jatmn (changes requested):
## Findings

### 1. GraphQL `repos` query still returns duplicate mirror/canonical rows

The PR dedups `GET /api/v1/repos` and `/api/v1/repos/federated`, but the GraphQL `repos` query at `crates/gitlawb-node/src/graphql/query.rs:12-28` still calls `db.list_all_repos()` without any dedup. Since that method returns every raw row, a client using `/graphql` will still see the same logical repo twice when both a mirror row and a canonical row exist.

Evidence from the checkout:

- `graphql/query.rs:15-17` calls `db.list_all_repos()` and maps the result directly to `RepoType`.
- `list_all_repos()` in `db/mod.rs:826-836` selects all rows from `repos` with no dedup.
- The REST list path now calls `dedupe_canonical_repos` on the output of `list_all_repos_with_stars()` (`repos.rs:173` and `:1005`).

Because the linked issue (#6) asks for "profile and repo-list surfaces" to show one logical repo, and the GraphQL endpoint is a repo-list surface, the PR does not fully close the issue as claimed.

**Recommended action:** Apply the same dedup to the GraphQL `repos` resolver, or add a dedicated `Db::list_all_repos_deduped()` / `list_all_repos_with_stars_deduped()` method and use it consistently across REST, GraphQL, and stats.

### 2. `stats` endpoint inflates repo count when mirror rows exist

`/api/v1/stats` (`server.rs:435-450`) uses `db.list_all_repos().await` and returns `r.len() as i64`. Because this path is not deduped, a node with both a canonical and a mirror row for the same repo counts them as two repos. This value is displayed by `gl sync` (`crates/gl/src/sync.rs:49-55`), so it is user-visible.

This is the same underlying issue as the list-surface bug, but applied to a count. The PR changed the legacy `X-Total-Count` to count logical repos, which makes the lack of consistency with `/api/v1/stats` more noticeable.

**Recommended action:** Either reuse the dedup helper for the stats count, or move the canonical-dedup logic into the DB layer so all callers get consistent counts automatically.

### 3. Mirror detection relies on a user-settable description string

`dedupe_canonical_repos` treats any row whose `description == "mirrored from peer"` as a mirror. The same string is used in the SQL `list_all_repos_paged` query. Because `description` is user-provided at repo creation, a canonical repo created with that exact description would be deprioritized against another canonical row, and a mirror row with a different description would be treated as canonical. This is a pre-existing fragility, but the PR duplicates it in Rust code rather than using a dedicated marker (e.g., a boolean column, the `id` format, or the `machine_id`/`disk_path` pattern used by `upsert_mirror_repo`).

**Recommended action:** Consider adding an explicit `is_mirror` column or other non-user-visible marker and update both the SQL and Rust dedup paths to use it.

beardthelion:
Thanks, all three are addressed. Rebased onto main first (the conflict is gone; `mergeable` now).

**1 & 2 — GraphQL `repos` and `/api/v1/stats` no longer show raw rows.** Both now go through new DB methods, `list_all_repos_deduped` and `count_repos_deduped`, that reuse the same `DISTINCT ON (split_part(owner_did,':',-1), name)` selection as `list_all_repos_paged` (factored into a shared `DEDUP_CTE` const so the three can't drift). `count_repos_deduped` uses the `COUNT(DISTINCT (split_part(...), name))` idiom already in the paged empty-page path. `#[sqlx::test]`s cover both surfaces, plus mirror-only, empty-table, and a count-equals-list-length guard.

**3 — mirror detection no longer keys on the description.** Both the SQL paths and `dedupe_canonical_repos` now classify a mirror by its slash-form `id` (`{owner_short}/{name}`), which `upsert_mirror_repo` is the only writer of; canonical rows use UUID ids and repo names are sanitized, so no other row can carry a slash. A test seeds a canonical row whose description is literally `"mirrored from peer"` and confirms it still wins. I went with the structural id marker rather than an `is_mirror` column to keep this migration-free; happy to add the column as a follow-up if you'd prefer the explicit schema signal.

While there I made the dedup survivor deterministic on a full `(mirror-status, created_at)` tie via an `id ASC` backstop, in both the SQL and Rust paths.

One thing I left out of scope: GraphQL `repos` and `/api/v1/stats` don't filter on `is_public` (the previous `list_all_repos` they called didn't either, so this PR doesn't change that). If those should be public-only surfaces it's a separate visibility fix; say the word and I'll open an issue.

The legacy non-paged list and federated paths keep their existing `did_matches` owner filter in Rust rather than moving to SQL: `did_matches` is DID-method-scoped (it won't match `did:key:z6X` against `did:gitlawb:z6X`), and the SQL group key is method-blind, so deduping in SQL before that filter could drop a repo from its own owner's listing. Leaving those paths as-is avoids that.

jatmn (changes requested):
## Findings

### 1. Required: `cargo fmt` is not clean
**Severity:** Required  
**Location:** `crates/gitlawb-node/src/api/repos.rs`, `crates/gitlawb-node/src/db/mod.rs`, `crates/gitlawb-node/src/test_support.rs`  

`cargo fmt --all -- --check` reports diffs in the PR’s changed code (long single-line `record(...)` calls, over-length `rec(...)` signature, `assert_eq!` macro invocations that should be multi-line, and a few comment alignments). The PR checklist claims `cargo fmt --all` is clean, but it is not. This must be fixed before merge.

---

### 2. Positive: Logic is consistent across SQL and Rust dedup paths
**Severity:** No issue  

I verified the deduplication rules line up between the new Rust helper and the shared SQL CTE:

- Grouping key: `split_part(owner_did, ':', -1)` / `owner_did.rsplit(':').next()`.
- Mirror marker: `position('/' in id) > 0` / `r.id.contains('/')`.
- Survivor preference: canonical (no slash) over mirror.
- Tie-break: `created_at ASC, id ASC` in both SQL and Rust.
- Activity timestamp: the survivor inherits the group’s max `updated_at` (SQL window function in `DEDUP_CTE`, `latest` map in Rust).

The GraphQL and stats surfaces now call the deduped methods, and the only remaining raw consumer is the IPFS object scan, which is intentionally documented as the non-listing path.

---

### 3. Positive: Tests cover the important edge cases
**Severity:** No issue  

The new tests exercise:
- Canonical wins over mirror.
- Distinct repos are preserved.
- Same owner, different repo does not collapse.
- Tie-breaking by earliest `created_at` and then `id ASC`.
- Mirror description on a canonical row does not misclassify it.
- Mirror-only group survives.
- Empty table returns empty / zero.
- `count_repos_deduped` matches `list_all_repos_deduped` length.
- Real `upsert_mirror_repo` row shape is classified correctly.
- GraphQL and REST stats integration.

Unit tests in `api::repos::tests` pass.

---

### 4. Optional / pre-existing: paged owner filter still only handles short-form owner
**Severity:** Optional observation (not introduced by this PR)  

In `list_all_repos_paged`, the owner filter is:

```sql
WHERE ($1::text IS NULL OR owner_did = $1 OR owner_did LIKE '%:' || $1)
```

If a caller passes the full `did:key:z6Mk…` form, the `LIKE` pattern becomes `%:did:key:z6Mk…`, which will not match the bare `z6Mk…` mirror row. The non-paged legacy path filters in Rust using `did_matches`, which handles both forms correctly. Fixing the paged SQL filter to handle full DID is out of scope here, but worth a follow-up to keep the two list paths consistent.

---

beardthelion:
Fixed in 5721791. Ran `cargo fmt --all` over the three files (repos.rs, db/mod.rs, test_support.rs) — the multi-line `record`/`rec` calls, single-line `assert_eq!`s, and the `create_repo(seed_repo(...))` reflows. `cargo fmt --all -- --check` is clean now and clippy is warning-free. Finding 4 (paged owner filter on full DID) is pre-existing and out of scope here; I'll leave it for a follow-up.

jatmn (changes requested):
## Findings

### 1. Deduplication key is method-blind and can collapse distinct DID methods (Major)

**Severity:** Major / correctness

**Location:**

- `crates/gitlawb-node/src/db/mod.rs` — `DEDUP_CTE` (lines 905–925) and `count_repos_deduped` (lines 1011–1018)
- `crates/gitlawb-node/src/api/repos.rs` — `dedupe_canonical_repos` (lines 1384–1466)

**Issue:** The dedup grouping key is the last `:` segment of `owner_did`:

```sql
split_part(owner_did, ':', -1)
```

```rust
rec.owner_did.rsplit(':').next().unwrap_or(&rec.owner_did)
```

This is method-blind. A repo owned by `did:key:z6MkExample` and a repo owned by `did:gitlawb:z6MkExample` (or any other DID method with the same trailing segment) and having the same name will be collapsed into one logical repo. The codebase already recognizes this risk and explicitly guards against it in `crates/gitlawb-node/src/api/mod.rs`:

```rust
/// Match a presented DID against a stored DID ... never let a bare id match across methods —
/// `did:web` / `did:gitlawb` share the base58 space with `did:key`, so a
/// trailing-segment compare would treat `did:key:X` and `did:gitlawb:X` as equal.
pub(crate) fn did_matches(a: &str, b: &str) -> bool { ... }
```

The new dedup logic does exactly the trailing-segment comparison that `did_matches` warns against. It is true that repo creation currently only accepts `did:key` owners, but the database schema does not enforce that, and the project already treats cross-method collision as a real concern. The dedup key should match the project's own DID-matching semantics.

**Recommended action:** Make the dedup key did:key-aware (and bare-id-as-did:key). For example:

- Treat `did:key:<id>` as `<id>`.
- Treat a bare `<id>` (no colon) as `<id>`.
- Leave any other `did:<method>:<id>` as the full string.

Apply the same normalization in `DEDUP_CTE`, `count_repos_deduped`, and `dedupe_canonical_repos` so the list, count, and legacy paths stay consistent.

---

### 2. Paged owner filter is inconsistent with the legacy path for full-DID owner queries (Optional / pre-existing)

**Severity:** Optional / observation

**Location:**

- `crates/gitlawb-node/src/db/mod.rs` — `list_all_repos_paged` WHERE clause (line 917 and empty-page fallback line 972)
- `crates/gitlawb-node/src/api/repos.rs` — legacy filter (lines 233–239)
- `crates/gitlawb-node/src/api/mod.rs` — `did_matches` (lines 63–74)

**Issue:** The paged path filters in SQL:

```sql
WHERE ($1::text IS NULL OR owner_did = $1 OR owner_did LIKE '%:' || $1)
```

If a caller passes the full owner `did:key:z6MkExample`, the `LIKE` pattern becomes `%:did:key:z6MkExample`, which will not match the bare mirror row `z6MkExample`. The legacy path uses `did_matches` in Rust, which correctly matches both forms. This means a full-DID owner filter returns different results on the paged and legacy surfaces.

This is not introduced by the PR, but the PR touches the paged query and leaves the inconsistency in place. It is worth a follow-up to keep the two list paths aligned.

**Recommended action:** In the paged SQL, normalize the owner filter the same way the grouping key is normalized (ideally after fixing finding #1), or add an explicit OR branch for the bare mirror form when the filter is a `did:key:` value.

beardthelion:
Both addressed in `9b9b120`.

**1 (Major) - method-blind dedup key.** Fixed. Replaced the `split_part(owner_did, ':', -1)` / `rsplit(':')` last-segment key with a did:key-aware key that strips a `did:key:` prefix only when the remainder has no `:`, and keeps the full DID otherwise:

- SQL: `CASE WHEN owner_did LIKE 'did:key:%' AND position(':' in substr(owner_did, 9)) = 0 THEN substr(owner_did, 9) ELSE owner_did END`
- Rust: `owner_did.strip_prefix("did:key:").filter(|rest| !rest.contains(':'))`

That reproduces `did_matches`/`key_id` as an equivalence relation, so `did:key:z6Mk…` and the bare mirror `z6Mk…` still collapse while `did:key:z6Mk…` and `did:gitlawb:z6Mk…` stay distinct. The residual-colon guard matches `key_id`'s `!ka.contains(':')` check, so even a malformed `did:key:did:gitlawb:X` keeps its full form rather than collapsing onto the bare method DID.

Applied in every spot the key is read: `DEDUP_CTE` (DISTINCT ON / PARTITION BY / ORDER BY), `count_repos_deduped`, the empty-page `COUNT(DISTINCT …)` fallback in `list_all_repos_paged`, and `dedupe_canonical_repos`. The backing index lived in the already-released migration v1, so I left v1 untouched and added migration v7 to drop `idx_repos_owner_short_name` and build `idx_repos_owner_key_name` on the matching expression; the CASE is byte-identical across all of them so the planner still uses the index.

Tests cover cross-method distinctness and bare/`did:key` collapse on both the in-memory and SQL paths, plus the `did:key:`-wrapped-full-DID and empty-residual boundaries.

**2 (Optional / pre-existing) - paged owner filter.** Left as-is for now. The `owner_did LIKE '%:' || $1` branch still trailing-matches, so a full `did:key:` owner filter won't hit the bare mirror row, and it now diverges from the grouping key on cross-method ids the same way you noted. It predates this PR and is out of scope for the under-withholding intent, so I'd rather not widen the diff here. Happy to file a follow-up to align the filter with the new key (and with `did_matches` on the legacy path) if you'd prefer that tracked.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #75, opened 2026-06-21
fix(node): reject malformed path globs in visibility validation (#74)

### LINKED ISSUE #74: validate_path_glob accepts malformed globs with non-trailing wildcards
## Summary

`validate_path_glob` in `crates/gitlawb-node/src/api/visibility.rs` is meant to reject malformed globs before they reach the store, but its wildcard check lets through any pattern that contains a `*` outside the trailing `/**`, as long as it still ends with `/**`. Values like `/a*b/**` and `/a/**/**` pass validation when they should be rejected.

## The bug

```rust
if path_glob.contains('*') && !path_glob.ends_with("/**") {
    return Err(AppError::BadRequest(
        "the only supported wildcard is a trailing '/**'".into(),
    ));
}
```

For `/a*b/**`, `contains('*')` is true but `ends_with("/**")` is also true, so the guard is false and the value is accepted. The check only catches a non-trailing `*` when the string does not also end in `/**`.

## Impact

This is a correctness / silent-misconfiguration bug, not an access-escalation one. Tracing the consumer:

- `glob_prefix("/a*b/**")` strips the trailing `**` and `/`, yielding the prefix `/a*b` (the `*` is left in place as a literal character).
- `glob_matches` then compares literally: `path == "/a*b"` or `path.starts_with("/a*b/")`.

So a malformed rule matches no real repo path and silently grants nobody. A non-trailing `*` only ever makes the prefix more specific, never broader, so there is no over-grant or privilege escalation; the rule just quietly does nothing. That is exactly the failure mode `validate_path_glob`'s own doc comment says it exists to prevent ("malformed globs ... would silently misconfigure access"). An operator who typos `/a*b/**` instead of `/ab/**` gets a visibility rule that appears accepted but never matches.

## Fix

Strip the trailing `/**` first, then reject if a `*` remains in the prefix:

```rust
if let Some(prefix) = path_glob.strip_suffix("/**") {
    if prefix.contains('*') {
        return Err(AppError::BadRequest(
            "the only supported wildcard is a single trailing '/**'".into(),
        ));
    }
} else if path_glob.contains('*') {
    return Err(AppError::BadRequest(
        "the only supported wildcard is a single trailing '/**'".into(),
    ));
}
```

Add `/a*b/**` and `/a/**/**` to the `rejects_malformed_globs` test.

## Notes

Pre-existing on `main`; surfaced during review of #68 but not introduced by it (the file is byte-identical there). Filing separately to keep #68 scoped.

beardthelion:
## Summary

`validate_path_glob` accepted malformed visibility globs that silently misconfigure access. It now rejects a wildcard anywhere but the single trailing `/**`, and rejects empty path segments.

## Motivation & context

Closes #74

The wildcard guard checked `path_glob.ends_with("/**")` without first stripping it, so any glob that still ended in `/**` slipped through even with a `*` earlier in the prefix (e.g. `/a*b/**`, `/a/**/**`). `glob_prefix`/`glob_matches` treat that prefix as a literal, so the rule matched no real path and granted nobody, the exact silent misconfiguration this validator exists to prevent.

Reviewing the fix surfaced a second hole in the same function: `glob_prefix` greedily trims all trailing `**` and `/`, so `//**` collapsed to `/` (whole repo) and bypassed the explicit `/**` guard, and `//secret/**` became a dead rule whose prefix never matches a real single-slash path. Both are fail-closed and owner-only (only the repo owner can set rules, and a bad rule grants nobody), so this is hardening, not a live exploit, but the `//**` case is a one-character bypass of the whole-repo guard and worth closing alongside.

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

Crate: `gitlawb-node`

- Strip the one permitted trailing `/**` before checking for stray `*`, so a non-trailing wildcard is rejected instead of slipping through.
- Reject path globs containing `//` (empty path segments), keeping the validator and `glob_prefix` agreed on the canonical form and closing the `//**` whole-repo bypass.
- Added `/a*b/**`, `/a/**/**`, `/**/**`, `//**`, `//secret/**` to the `rejects_malformed_globs` test.

## How a reviewer can verify

```sh
cargo test -p gitlawb-node glob
```

Both `accepts_supported_globs` (`/`, `/secret`, `/secret/**`, `/a/b/c`, `/a/b/**` still pass) and `rejects_malformed_globs` (the five added cases now rejected) cover the change.

## Before you request review

## Notes for reviewers

`remove_visibility` still does not run `validate_path_glob`, but deleting a never-matching rule is a no-op, so that asymmetry is harmless and left out of scope.

## Summary by CodeRabbit

* **Bug Fixes**
  * Path pattern validation now strictly rejects malformed patterns containing consecutive slashes, preventing invalid inputs from bypassing existing validation rules and security checks. This improves error handling, protects against problematic patterns, and ensures more robust validation across the system. The change maintains full backward compatibility with all valid glob patterns currently in use.

### DIFF
--- a/crates/gitlawb-node/src/api/visibility.rs
+++ b/crates/gitlawb-node/src/api/visibility.rs
@@ -57,9 +57,23 @@ fn validate_path_glob(path_glob: &str) -> Result<()> {
             "path_glob must not end with '/'".into(),
         ));
     }
-    if path_glob.contains('*') && !path_glob.ends_with("/**") {
+    // Reject empty path segments. `glob_prefix` collapses "//**" to "/" (whole
+    // repo), so without this an operator could bypass the "/**" guard above and
+    // grant repo-wide read with "//**", or set a dead rule like "//secret/**"
+    // whose prefix never matches a real single-slash path and silently leaves
+    // the subtree readable.
+    if path_glob.contains("//") {
         return Err(AppError::BadRequest(
-            "the only supported wildcard is a trailing '/**'".into(),
+            "path_glob must not contain empty path segments".into(),
+        ));
+    }
+    // Strip the one permitted trailing "/**" before checking for stray
+    // wildcards, otherwise a "*" elsewhere in the prefix (e.g. "/a*b/**" or
+    // "/a/**/**") slips through because the string still ends with "/**".
+    let prefix = path_glob.strip_suffix("/**").unwrap_or(path_glob);
+    if prefix.contains('*') {
+        return Err(AppError::BadRequest(
+            "the only supported wildcard is a single trailing '/**'".into(),
         ));
     }
     Ok(())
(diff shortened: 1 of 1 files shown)


kevincodex1 (approved):
LGTM

PR #79, opened 2026-06-22
feat(storage): storage-agnostic BlobStore layer + push-latency quick wins

### LINKED ISSUE #89: feat(storage): backend-agnostic BlobStore layer (S3 / filesystem / IPFS)
### Problem / use case

  Summary — Replace the hardcoded Tigris integration with a backend-agnostic BlobStore interface (get/put/head/delete/list) over S3-compatible, local filesystem, and IPFS backends, plus push-latency quick wins. Storage becomes a deployment config concern, not a code
  dependency.

  Acceptance criteria — no behavior change for existing Tigris deployments; selectable backend; atomic concurrency-safe extraction preserved (the #80 hardening); fmt/clippy/tests green; unit tests for storage::fs + repo_store.

Tracked by: PR #79 · Labels: kind:feature, crate:node, subsystem:storage, sev:medium

### Proposed solution

Proposed changes — BlobStore trait + 3 backends; RepoArchive replacing Option<TigrisClient>; backward-compatible env config (legacy GITLAWB_TIGRIS_BUCKET honored); skip redundant pre-write download on etag match; write-back ack before durable upload.

### Alternatives considered

_No response_

### Does this touch the protocol?

No

kevincodex1:
Replace the hardcoded Tigris client with a backend-agnostic object store.

BlobStore trait (get/put/head/delete/list) with backends:
- s3:   any S3-compatible service (Tigris, Cloudflare R2, AWS S3, MinIO, B2)
- fs:   local filesystem (atomic writes; unit-tested)
- ipfs: Kubo node via MFS (etag = content CID)

RepoArchive composes a bare repo into one repos/v1/{slug}/{repo}.tar.zst object on top of any backend (tar+zstd moved here; git/tigris.rs removed). RepoStore rewired from Option<TigrisClient> to Option<RepoArchive>.

Config (backward compatible): GITLAWB_STORAGE_BACKEND, GITLAWB_S3_BUCKET, GITLAWB_S3_ENDPOINT, GITLAWB_S3_FORCE_PATH_STYLE, GITLAWB_STORAGE_FS_DIR, GITLAWB_ASYNC_UPLOAD. GITLAWB_TIGRIS_BUCKET kept as a legacy alias.

Push-latency quick wins:
- Skip the redundant pre-write download when the cached etag already matches storage (previously up to two full-repo downloads per push).
- Write-back ack: ack the client before the durable upload completes; the advisory lock is held until upload finishes so cross-node consistency holds.

cargo check clean; storage::fs + repo_store unit tests pass.

Closes #89

## Release Notes
- **New Features**
  - Added pluggable durable repo-archive storage backends (S3-compatible, filesystem, IPFS) with new configuration, backend selection rules, and object-storage controls (including async upload behavior).
  - Added an adjustable advisory-lock pool size for durable write coordination.
- **Bug Fixes**
  - Improved durability handling and error reporting across push/merge/issue/fork flows; durable failures now surface more reliably.
  - Improved archive sync to avoid unnecessary downloads using stored metadata and to fail safely on stale/corrupt data.
- **Tests**
  - Expanded filesystem-backed coverage for archive skipping, marker/retry behavior, divergence detection, and concurrent write safety.
- **Chores**
  - Added startup cleanup for orphaned swap artifacts and removed the legacy storage client.


kevincodex1:
please have a look into this @beardthelion

beardthelion (dismissed):
Reviewed the merged state against current main (clean merge), built it, ran the storage and repo_store unit tests, and traced the write and consistency paths. The BlobStore abstraction and the Tigris-to-archive migration are clean and complete, and the fs backend's atomic swap and per-method key validation hold up. My concerns are all in the push-latency changes: the write-back default has a data-loss path, and it leans on an advisory-lock pattern that isn't sound over a connection pool. Highest first.

## Findings

- **[P1] Default `async_upload` to false; ack-before-upload can lose an acked push on a rolling deploy.** `crates/gitlawb-node/src/config.rs` (`GITLAWB_ASYNC_UPLOAD` is `default_value_t = true`), `api/repos.rs:561`. With it on, `git_receive_pack` acks the client and uploads in a detached task. If the process stops in that window (a SIGTERM during a rolling deploy is the routine trigger, not just a crash), storage keeps the old archive. On restart the in-process `versions` cache is empty, so the write-path `sync_down_if_stale` (`require_fresh=true`) finds no cached etag, downloads the stale remote, and overwrites the newer local copy at `repo_store.rs:102-113`. Lazy migration does not save it: `acquire` only uploads when storage has no object (`:154-175`), so an already-present stale archive is never re-synced. The "local copy survives; lazy migration re-syncs" comment does not hold. Default this to false, or add a durable local-newer marker plus a visible signal on upload failure and a shutdown drain before making it the default.

- **[P2, pre-existing] The advisory lock is acquired and released on different pooled connections.** `repo_store.rs:233` locks via `fetch_one(&self.pool)`; `:260`/`:494` unlock via `execute(&self.pool)`. Session-scoped Postgres advisory locks only release on the connection that took them, and sqlx hands out an arbitrary pooled connection per query, so the unlock can land on a different session and silently no-op (the boolean result is not checked) while the lock persists. A later same-session acquire is re-entrant and succeeds spuriously; a different session blocks 60s then 500s. This is already on main, so the PR does not introduce it, but the write-back change leans on the lock being held through the upload and now runs the unlock from a detached task that is dropped on crash, so it depends on a guarantee that is not actually there. Worth fixing as part of leaning on it: hold one connection (`pool.acquire()`) for the lock's lifetime and run lock+unlock on it, or use a transaction-scoped `pg_advisory_xact_lock`.

- **[P2] The S3 client sets no request timeout.** `storage/s3.rs:31`. `S3BlobStore::new` builds the client with no timeout config, and the AWS Rust SDK sets none by default, so a hung S3 endpoint blocks the calling task unbounded. Under `async_upload` that task holds the advisory lock, so one stuck upload blocks every later push to the repo. The IPFS client already sets connect and total timeouts; S3 should match, via `aws_sdk_s3::config::timeout::TimeoutConfig` on the client builder or a `tokio::time::timeout` wrapper.

- **[P2] "Fail closed when the backend is unreachable" only covers missing config, not connectivity.** `storage/mod.rs:69`. The doc says a configured-but-unavailable backend returns Err, but `S3BlobStore::new` only loads credentials and builds the client with no network call, so an unreachable bucket boots clean and fails on the first live request. Scope the comment to config validation, or add a cheap probe against the bucket at boot if real fail-closed is intended.

- **[P2] The risky new paths have no tests.** `compress_repo`/`decompress_repo` are pure filesystem functions but there is no round-trip test and nothing covering the "corrupt archive leaves the original intact" atomicity claim, which is the whole download-durability story. `sync_down_if_stale`'s `require_fresh=true` (the lost-update guard) and the etag-skip are also untested (`for_testing` forces `archive=None`). Both are cheap to add with the fs backend and a test constructor that accepts an archive.

P3s, briefly: IPFS `list()` doesn't reject traversal in its prefix the way the other four methods do (`ipfs.rs:159`), latent since it has no caller (a traversal-segment check that still allows the empty list-all prefix, or just drop `list()`); absent-vs-error detection keys on Kubo error-string substrings (`ipfs.rs:68`), and since Kubo's error `Code` is always 0 there's nothing structured to switch on, so a pinned comment is the realistic fix; IPFS `put()` returns Err after a successful write when the follow-up stat fails (`ipfs.rs:104`); and several doc comments still say "Tigris" after the rename (`repo_store.rs:131-133,212,282`, `archive.rs:4`).

The abstraction is the right call and the migration is complete. Once the `async_upload` default and the advisory-lock soundness are sorted I'm glad to take another pass. @kevincodex1

kevincodex1:
please check again @beardthelion

beardthelion (dismissed):
Thanks, this addresses the prior round cleanly. I re-reviewed the merged state on the new head: `async_upload` now defaults false, the advisory lock is acquired and released on a single pinned connection (sound on every exit path I traced), the S3 client has operation/attempt timeouts, `ipfs` `put` no longer fails on a post-write stat, and the new compress/decompress and `sync_down_if_stale` tests exercise real behavior. All five prior findings are resolved.

One regression the lock fix introduces, which is why I'm holding for another round.

## Findings

- **[P2] Bound concurrent pushes per DID before pinning a pool connection per push.** `git/repo_store.rs:236` (the `pool.acquire()` now held for the whole write), `server.rs:178` (push routes). `acquire_write` holds one connection from the shared pool across the write, the 60s lock-retry, and the upload, and `git_write_routes` has no per-DID rate limit (unlike `creation_routes` at `server.rs:96`). Because identities are permissionless, a single actor can fire enough concurrent or slow pushes to exhaust the default 10-connection pool (`db/mod.rs:269`, no `PgPoolOptions`), after which every DB-touching handler node-wide blocks on `pool.acquire()` for the ~30s acquire timeout and then fails. It's bounded (30s acquire timeout, 300s S3 op timeout) and self-recovers, but it's repeatable and reachable, and the previous connection-per-query design didn't have it. Bound concurrent writes per DID on `git_write_routes` (the `rate_limit_by_did` layer creation already uses, or a semaphore), and/or give lock connections a dedicated, sized pool. Not `pg_advisory_xact_lock` here: the lock has to outlive the transaction (it is held across the post-commit upload), so a transaction-scoped lock would release before the upload and reopen the stale-archive window the lock exists to close.

- **[P3] Add `#[must_use]` to `RepoWriteGuard`.** `git/repo_store.rs:456`. The connection pinning is correct, but a guard dropped without `release()` returns its connection to the pool with the session advisory lock still held (sqlx does not reset session state on check-in), so it lingers until that backend connection is evicted. Every current caller releases on all paths, so this is not reachable today; `#[must_use]` is cheap insurance against a future caller that forgets.

The lock-soundness approach is right and the rest is solid. Once the push path is bounded I'll approve. @kevincodex1

kevincodex1:
@coderabbitai please do full review of this PR

jatmn (changes requested):
## Findings

- [P2] Clean up the local fork directory when fork creation fails after the clone
  `crates/gitlawb-node/src/api/repos.rs:1187-1218`
  After `git clone --mirror` succeeds, `release_after_write` is called without any cleanup on failure. If the durable upload fails, the handler returns an error but leaves the cloned mirror at `disk_path`. A later fork with the same name (or a retry of this one) will then fail because the directory already exists. The same leak happens if `state.db.create_repo` fails after the upload. Remove `disk_path` before returning from either error path. CodeRabbit's latest review explicitly requested this cleanup.

- [P2] Don't block the Tokio runtime in `FsBlobStore::head` and `::put`
  `crates/gitlawb-node/src/storage/fs.rs:39-51` and `crates/gitlawb-node/src/storage/fs.rs:69-107`
  `head` and `put` both call `Self::meta_of`, which uses `std::fs::metadata` directly inside an async trait method. That blocks the async executor thread while stat'ing local files. For `put`, move the stat into the existing `spawn_blocking` closure; for `head`, use `tokio::fs::metadata` (or a `spawn_blocking` call) instead of the synchronous helper.

- [P3] Advisory-lock acquisition timeout is shorter than the backend operation timeout
  `crates/gitlawb-node/src/git/repo_store.rs:234-250`
  `acquire_write` retries `pg_try_advisory_lock` for 60 seconds, but the S3 backend sets a 300-second operation timeout (`crates/gitlawb-node/src/storage/s3.rs:44-47`) and the IPFS backend sets the same 300-second total timeout (`crates/gitlawb-node/src/storage/ipfs.rs:30-31`). A single large-repo upload can therefore hold the advisory lock long enough for concurrent pushes to the same repo to be rejected after only 60 seconds. Align the lock-acquisition timeout with the configured storage operation timeout, or use a blocking advisory-lock call with a matching timeout.

beardthelion (dismissed):
Re-reviewed the current head against merged-state. The three findings from the last round are genuinely addressed: the advisory-lock acquire is now bounded with an explicit timeout, archive temp dirs are cleaned on every error path, durability errors propagate to callers, and the `#[must_use]` guard plus isolated lock pool are all in place and behave as described. The blocking-`stat` item is mostly handled too (the multi-MB write moved into `spawn_blocking`); what's left is narrow and noted below. Owner gating on the four touched routes is intact and the blob-key/traversal defenses check out.

One thing should change before merge: the storage auto-detect quietly repurposes an existing IPFS pinning config as the repo-archive backend on upgrade. The rest are polish.

## Findings

- [P2] Require an explicit opt-in before selecting the IPFS repo-archive backend. `crates/gitlawb-node/src/storage/mod.rs:96` — auto-detection falls through to `ipfs` whenever `ipfs_api` is set and no s3/fs backend is configured. But `ipfs_api` already exists on main and today only drives the per-object encrypted pinning path (`sync.rs`, `encrypted_pin.rs`, `ipfs_pin.rs`); repo storage on main is Tigris-or-local, never IPFS. So a node that set `GITLAWB_IPFS_API` purely for pinning, with local-only repo storage, silently starts routing whole-repo archives into IPFS MFS after this upgrade, with no operator action. Gate the ipfs backend behind an explicit `GITLAWB_STORAGE_BACKEND=ipfs` rather than auto-selecting it from `ipfs_api`, so the pinning config keeps its existing meaning.

- [P2] Don't strand the advisory lock when the write-back upload is cancelled. `crates/gitlawb-node/src/git/repo_store.rs:577` and the spawn at `crates/gitlawb-node/src/api/repos.rs:560` — in `async_upload` mode `release()` runs in a detached, untracked `tokio::spawn`, and the `pg_advisory_unlock` is the last step of its linear control flow rather than a `Drop`. If the runtime shuts down (or the task is otherwise cancelled) mid-upload, the `PoolConnection` returns to the pool with the session lock still held, and subsequent pushes to that repo burn the full retry window before failing. The root weakness (unlock not tied to `Drop`) predates this PR, but the detached write-back makes it materially easier to hit. `async_upload` defaults off, so this isn't a default-path break. Direction: tie the unlock to a guard that runs on drop, and/or track in-flight write-backs and drain them on shutdown. Flagging the direction rather than asserting a fix, since an async unlock in `Drop` is not a one-liner.

- [P3] Move the remaining `stat` off the runtime in the fs backend. `crates/gitlawb-node/src/storage/fs.rs:102` (`head`) and `:95` (`put`'s trailing `meta_of`) still call `std::fs::metadata` synchronously on the async runtime, while the rest of the backend uses `spawn_blocking`/`tokio::fs`. A single `stat` is cheap, so low impact, but it's the residual of the earlier blocking-IO item and worth finishing for consistency.

- [P3] Reconcile or sweep the `.bak` orphan left by a crash mid-swap. `crates/gitlawb-node/src/storage/archive.rs:185` — the move-aside/rename swap is correct on its error path, but a hard kill between the backup rename and the new rename leaves `local_path` absent with an orphaned `.bak-UUID` sibling. It self-heals on the next `acquire()` (the swap is on the download path, so storage still holds the copy and re-downloads), so this is litter plus a narrow transient, not data loss. A startup sweep of stale `.bak-`/`.tmp-extract.` dirs would close it.

- [P3] Fold the duplicated advisory-lock retry loop into one helper. `crates/gitlawb-node/src/git/repo_store.rs:235` and `:351` — `acquire_write` (60 attempts) and `upload_under_lock` (30) carry the same `pg_try_advisory_lock` loop verbatim; they differ only in attempt count and error text. Extract a private helper taking both as params.

Advisory, not blocking: the new fail-closed `storage::build()` branches, the `release()` upload-failure path, and the `async_upload` write-back have no test coverage; the IPFS backend has none at all and detects not-found by string-matching Kubo error bodies, which will misclassify silently if Kubo changes its wording.

The structural change is solid and the storage trait is the right shape. Once the IPFS auto-detect is gated I'm happy to take another pass.

beardthelion:
@kevincodex1 this one's gone stale but the work is worth finishing. Nothing on main has replaced it, repo storage is still Tigris-only, and the BlobStore trait is the right shape.

Two things stand between it and merge:

1. It's conflicting and about 72 commits behind main, so it needs a rebase before anything else. Some of the overlapping files are auth/storage paths, so I'll want a fresh pass on the rebased head to be sure nothing that landed in the meantime gets dropped in the merge.
2. The change requests from jatmn and me are still open on the current head. The substantive ones:
   - Gate the IPFS backend behind an explicit `GITLAWB_STORAGE_BACKEND=ipfs`. Right now auto-detect repurposes an existing `GITLAWB_IPFS_API` (today only pinning) as the repo-archive backend on upgrade, silently.
   - Clean up the local fork directory when fork creation fails after the clone.
   - Move the blocking `std::fs::metadata` calls off the async runtime in the fs backend.
   - Don't strand the advisory lock if the async write-back task is cancelled mid-upload.

   The rest are polish.

Do you want to rebase and take it the rest of the way, or shelve it? Your call either way.

kevincodex1:
@coderabbitai please do an intensive review on this

kevincodex1:
@coderabbitai please make an intensive review again on our changes

kevincodex1:
hey @coderabbitai please make an intensive review on this again

kevincodex1:
hey guys please review again @jatmn @beardthelion

beardthelion (dismissed):
Re-reviewed the moved head (`d30af0d2`) against merged state. The storage abstraction is well-built and the authz surface is untouched: owner-push, branch protection, quarantine, path-scoped read gates, and the `git_upload_pack` 404-deny all still sit above the new storage layer, and `pool()`'s removal has no stranded callers. One data-loss defect in the new write-back path blocks it, plus two should-fix issues. Findings below, highest first.

## Findings

- [P1] Don't let a failed write-back upload roll the local copy back over an acked push
  `crates/gitlawb-node/src/git/repo_store.rs:85`
  With `GITLAWB_ASYNC_UPLOAD=true`, `git_receive_pack` acks the client, then `guard.release(true)` uploads in a detached task (`api/repos.rs:962`). If that upload fails (a transient S3/IPFS error, or the node restarts first), `release` invalidates the in-memory etag and storage still holds the pre-push archive. The next access runs `sync_down_if_stale(require_fresh=true)`, whose logic assumes an etag mismatch means "remote is newer" and downloads it over local -- but here remote is older, so the acked commit is silently overwritten. The comment at `api/repos.rs:957` ("the local copy survives... until the next successful push re-uploads") is inverted: the next push pulls stale storage down before it can re-upload. Recovery needs a persisted local generation/marker so "local is ahead of storage" is distinguishable from "storage is ahead of local," rather than treating any mismatch as storage-wins. The default strict path is safe; this is the opt-in write-back path only.

- [P2] Bound the whole-archive in-memory buffering
  `crates/gitlawb-node/src/storage/mod.rs:47`
  `get`/`put` move the entire `.tar.zst` through RAM as one `Bytes`, and `compress_repo` builds a full `Vec<u8>`, with no per-op size cap or concurrency guard. An unauthenticated cold-cache clone of a large public repo forces the whole archive resident, and N concurrent archive ops is N x archive-size toward OOM. `GITLAWB_MAX_PACK_BYTES` bounds the protocol body, not this buffer. The buffering carried over from `tigris.rs`, but the fresh trait signature is where to fix it (stream get/put through a temp file) or to defer explicitly with a tracking issue rather than bake the limit in.

- [P2] Detach the lock connection when `pg_advisory_unlock` itself fails
  `crates/gitlawb-node/src/git/repo_store.rs:517`
  On an unlock-query error, `unlock` only logs and lets the pinned connection return to the pool, while `Drop` and the acquire poll-error path both `detach()` to close the socket and free a possibly-held lock. A transient error there repools a connection that may still hold the advisory lock, wedging that repo's writes until the pool recycles the connection (the lock pool sets no `max_lifetime`). Mirror `Drop`: `detach()` the connection on unlock failure instead of repooling it.

- [P3] Clean up local/storage state on the two orphan paths
  `crates/gitlawb-node/src/git/repo_store.rs:246`
  `init` leaves the bare repo dir on disk when its upload fails, so a retry of the same name hits an existing destination; and `fork_repo` (`api/repos.rs:1640`) removes the local mirror on a post-upload DB failure but leaves the already-uploaded archive orphaned in storage with no owning row. Neither is data loss, but both leave litter that can block or confuse a retry.

- [P3] Add the new storage knobs to `.env.example` and the README config table
  `crates/gitlawb-node/src/config.rs:143`
  `GITLAWB_STORAGE_BACKEND`, `GITLAWB_S3_BUCKET`/`_ENDPOINT`/`_FORCE_PATH_STYLE`, `GITLAWB_STORAGE_FS_DIR`, and `GITLAWB_ASYNC_UPLOAD` are new operator-facing settings but appear in neither `.env.example` nor the README env table (which still lists only `GITLAWB_TIGRIS_BUCKET`). The clap `--help` docs are good; this is just the operator-doc surface the repo already maintains.

- [P3] Minor cleanups
  `crates/gitlawb-node/src/storage/mod.rs:65`
  `list()` ships fully implemented across all four backends but is `#[allow(dead_code)]` with no production caller (only tests) -- wire a consumer or drop it until one exists. Two receive-pack test strings still say "Tigris" after the rename (`api/repos.rs:2648`). And the PR body's "cross-node consistency holds" overstates it: the advisory lock serializes writers only, so another node can still serve a stale archive mid-upload -- worth narrowing the claim.

Core design is sound and the gates are clean; the P1 is contained to the opt-in write-back path and fixable without a redesign.

jatmn (changes requested):

## Findings

- [P1] Persist the write-back marker before acknowledging the push
  `crates/gitlawb-node/src/api/repos.rs:949`
  `tokio::spawn` returns before `RepoWriteGuard::release()` is first polled, but that method is where the pending-upload marker is written.  If the process stops in the interval after the successful receive-pack response is sent and before the task runs, the acknowledged local commit has neither an uploaded archive nor a marker.  On restart `sync_down_if_stale` treats the old archive as newer and replaces the local repository with it.  Create the durable intent marker before spawning/acknowledging (and retain it if the task is cancelled), then add coverage for that shutdown window.

- [P1] Do not let a pending-upload marker bypass every later freshness check
  `crates/gitlawb-node/src/git/repo_store.rs:96`
  After a failed upload, node A retains this marker and returns its local copy before checking storage.  Node B can subsequently acquire the advisory lock, upload a newer archive, and release it; A's next write still skips the download and overwrites B's archive with A's divergent copy.  The marker needs to record/reconcile the version it was based on, or otherwise retry safely under the lock, rather than making local state authoritative indefinitely.

- [P1] Fail lazy migration when the storage existence check is unknown
  `crates/gitlawb-node/src/git/repo_store.rs:367`
  `archive.exists(...).await.unwrap_or(false)` converts a failed HEAD request into "archive absent" and uploads the node's local cache.  A restarted node can therefore overwrite a newer shared S3/IPFS/FS archive whenever HEAD transiently fails but PUT succeeds; the advisory lock only serializes writers and cannot establish freshness.  Propagate or retry the HEAD error and upload only after a successful absence check.

- [P1] Scope fork cleanup to the attempt that created the archive
  `crates/gitlawb-node/src/api/repos.rs:1667`
  Two same-owner/name fork requests on different nodes can both pass the preflight lookup and upload to the same archive key.  One DB insert wins; the loser unconditionally removes both its `disk_path` and that shared archive after its uniqueness error, deleting the winner's durable repository and leaving its live DB row unable to hydrate on another node.  The same danger applies to an ambiguous DB error after a committed insert.  Serialize/claim creation and never delete a key that may belong to an existing repo.

- [P1] Do not report irreversible Git mutations as failed requests
  `crates/gitlawb-node/src/api/issues.rs:71`
  `create_issue`, `close_issue`, and `merge_pr` now return a 5xx when the post-mutation archive upload fails, even though the Git mutation is retained locally behind the pending marker.  A caller that retries issue creation receives a new UUID and later publishes both the supposedly failed issue and the retry; for a merge, the target ref is already merged while the PR database status remains open.  Use a recoverable accepted-operation state/compensation, or make the mutation and durability outcome externally idempotent instead of returning an ordinary request failure after commit.

- [P2] Use a collision-resistant filesystem archive revision
  `crates/gitlawb-node/src/storage/fs.rs:39`
  The fs backend calls `size-mtime` an etag, and RepoStore treats equality as proof that its local copy is current.  The configured fs backend explicitly supports mounted storage, where timestamp precision is not guaranteed to be nanoseconds: two different same-size archive writes in one timestamp tick can have the same token.  A stale node then skips the required download under its advisory lock and uploads over the newer archive.  Use a content hash or reliably changing persisted generation, or disable the etag fast path for this backend.

- [P2] Clean up a partially created mirror when `git clone --mirror` fails
  `crates/gitlawb-node/src/api/repos.rs:1620`
  The new cleanup closure is defined only after a successful clone.  `git clone` can create its destination before returning a non-zero status (for example during a fetch/write failure), and this branch returns without removing it.  A later fork retry then fails because the destination exists.  Define and invoke the cleanup on the clone-failure path as well.

- [P2] Compensate failed repository creation after the synchronous archive upload
  `crates/gitlawb-node/src/api/repos.rs:230`
  `RepoStore::init` now durably uploads the empty archive before this `create_repo` insert.  If the insert loses a concurrent-create race or fails transiently, the handler returns with the new `repos/v1/...` archive orphaned; the local bare directory also blocks a retry.  The local-directory leak existed before this PR, but making the archive upload synchronous makes the durable orphan deterministic.  Give create-repo the fork flow's compensation, without deleting an already-successful concurrent creation.

beardthelion (dismissed):
Re-reviewed the merged state at 75060ea7. The two new commits genuinely close both prior rounds: the marker is written before the ack, the versioned base distinguishes local-ahead from cross-node divergence and fails the write path closed, exists() errors propagate instead of reading as "absent", create/fork are claim-first with compensation keyed strictly by their own row id (and undo_claim never touches a storage key, so a racing winner's archive is safe under any topology), unlock failure now detaches the pinned connection, the fs etag is a per-write sidecar UUID, and the operator docs cover every new knob. I built and checked the merged state and traced each of those to the fixing lines. Two new problems in the marker machinery itself keep this at request-changes; both are contained and the fixes are small.

## Findings

- [P1] Preserve the pending marker's base etag instead of overwriting it on re-mark
  `crates/gitlawb-node/src/git/repo_store.rs:764`
  Every upload-failure arm removes the repo's `versions` entry (repo_store.rs:364, 720, 733), and the serve-local-under-marker branch of `sync_down_if_stale` (96-150) never repopulates it, so the next write re-marks the marker with an empty base. I reproduced this with a throwaway test: fail one upload (marker holds the true base), write again with the upload failing too, and the marker now contains "" while storage never moved; the third write then dies on the "storage advanced ... reconcile manually" bail, and since issues, closes, and merges all go through `acquire_write`, the repo's whole write surface is wedged until someone hand-deletes the marker file. Two consecutive upload failures is just "storage was down for a minute spanning two writes", and this hits the strict path as well as async_upload. Direction: have `mark_pending_upload` return early when a marker already exists. The first marker's base is by construction the last storage etag this node confirmed, and it stays correct for every write stacked on the same undiverged local copy, so divergence detection is unharmed. Add the two-consecutive-failures regression test; it goes red on the current code.

- [P1] Run the fork's post-clone upload under the advisory lock
  `crates/gitlawb-node/src/api/repos.rs:1694`
  `release_after_write` (repo_store.rs:339) calls `archive.upload` with no `LockedConn`, and claim-first ordering makes the fork visible to `git_receive_pack` before the upload finishes. A push to the fresh fork uploads under the lock; the fork's unlocked upload can compress the pre-push (or mid-receive, torn) tree and PUT after it. The push's upload succeeded, so no marker protects it: the next require_fresh sync sees the fork's etag, downloads, and rolls back an acked push. This is a race, not a certainty (fork must compress before the push lands and PUT after), but it is exactly the window `upload_under_lock`'s own comment describes, and `init()` already routes through it. Direction: acquire the lock inside `release_after_write` (or call `upload_under_lock` with `skip_if_exists=false`).

- [P2] Make the marker wiring's tests exercise the production writers
  `crates/gitlawb-node/src/git/repo_store.rs:1037`
  All three new marker tests plant the marker by calling `mark_pending_upload` directly; nothing drives `guard.mark_pending()` (repos.rs:976) or `release()`'s own marker writes, so deleting the load-bearing before-ack line leaves the suite green. jatmn's round explicitly asked for coverage of the ack-to-first-poll window and it is still uncovered. Add a test that acquires a guard over the fs-backed store, calls `mark_pending()`, drops the guard without release (the crash), and asserts the next sync serves local; plus one asserting `release(true)` with a failing archive leaves the marker.

- [P2] Fall back to strict upload-before-ack when the marker write itself fails
  `crates/gitlawb-node/src/git/repo_store.rs:764`
  `mark_pending_upload` warns and continues on a write error, and repos.rs:976 acks unconditionally, so the "marker must be on disk BEFORE the ack" comment is not enforced: a full disk at the wrong moment silently reopens the crash window this mechanism exists to close. Direction: return Result from `mark_pending` and take the synchronous release path when it fails.

- [P2] Write the fs etag sidecar before renaming the blob into place
  `crates/gitlawb-node/src/storage/fs.rs:97`
  The blob is renamed into place first and the sidecar written after, so a crash between them leaves new content under the old etag. Etag-equality consumers then skip the download and serve stale, which is the same failure the sidecar was built to remove, relocated into a crash window. Writing the sidecar (its own tmp+rename) before the blob rename inverts the failure into a harmless redundant download.

- [P2] Use try_exists() for the pending marker probe
  `crates/gitlawb-node/src/git/repo_store.rs:97`
  `Path::exists()` returns false on EACCES/EIO, not just NotFound, so a transient stat error skips the whole marker branch and the write path falls through to etag-compare-and-download, rolling back the pending local write. This delta fixed the identical two-valued collapse for `archive.exists()` in `upload_under_lock`; the same treatment belongs here, with Err reading as marker-present on the write path.

- [P2] Clear the stale marker on repo recreation
  `crates/gitlawb-node/src/api/repos.rs:1650`
  `undo_claim` removes the clone dir and the row but not the sibling marker that the failed `release_after_write` just wrote. Recreating the same owner/name re-creates the local dir, and once the new repo's archive exists, the leftover empty-base marker trips the divergence bail and wedges the recreated repo's writes. Clearing any pre-existing marker in `init()` after `init_bare` covers every recreate path; removing it in `undo_claim` too keeps the litter down.

- [P2] Close out the whole-archive buffering thread explicitly
  `crates/gitlawb-node/src/storage/mod.rs:47`
  `get`/`put` still move whole archives through RAM and `compress_repo` still builds a full Vec; unchanged at this head and the thread is still unresolved. Streaming is a real design change and deferring it past this PR is a reasonable call, but make it explicit: a tracking issue referenced from the PR, and resolve the thread.

- [P2] Add a bounded re-upload retry behind the 2xx-on-failed-upload paths
  `crates/gitlawb-node/src/api/issues.rs:82`
  Returning 2xx when the git mutation committed but the upload failed is the right call and I consider that settled: a 5xx invites retries that mint duplicate issue UUIDs or contradict an already-merged ref, and the marker keeps the local copy authoritative. What is missing is the second half: nothing ever retries the upload, so a repo that gets no further writes stays divergent indefinitely, visible only as one log line at failure time. A bounded background retry (or a startup/periodic sweep of pending markers) plus a counter for outstanding markers closes the loop.

- [P2] Add a failing-store test double for the new error branches
  `crates/gitlawb-node/src/git/repo_store.rs:394`
  No mock or failing BlobStore exists in the crate, so the `head_etag` Err branches (marker and main flow, both require_fresh values), `upload_under_lock`'s exists-Err propagation (the fix for the `unwrap_or(false)` finding; reverting it stays green today), and `init`'s cleanup-on-failed-upload are all structurally untestable. One small injectable failing store unlocks all of them.

- [P3] Skip the marker when no storage backend is configured: `mark_pending` (repo_store.rs:672) writes markers even with `archive: None`; they are inert until a backend with pre-existing archives is enabled, then wedge those repos. Early-return when there is no archive.
- [P3] Harden the fs etag test against revert: back-to-back writes get distinct mtimes, so `assert_ne` on consecutive put etags also passes with the legacy size-mtime etag. Also assert the sidecar file's content equals the returned etag.
- [P3] Fix the claim-first comments' scope: with per-node Postgres (three fly apps, per-node DATABASE_URL), "a concurrent same-name create loses at the insert" holds within one node only; the cross-node safety argument is that compensation never deletes a storage key. Worth one comment tweak so the next reader builds the right model.

Two candidates from this pass were checked and dropped: init's unconditional upload cannot clobber a racing first push (`upload_under_lock` compresses under the lock, so it snapshots post-push state), and the disk_path-UNIQUE-as-claim concern does not apply given per-node databases.

The remaining fixes are small and local; with the two P1s and the wiring tests in place this is close.

jatmn (changes requested):
## Findings

### P1: Do not acknowledge write-back pushes when the pending marker cannot be persisted

`mark_pending` awaits a method that suppresses its `std::fs::write` error (`crates/gitlawb-node/src/git/repo_store.rs:673-677`, `764-773`). In the current `git_receive_pack` flow (`crates/gitlawb-node/src/api/repos.rs:318-343`), an asynchronous upload can therefore be acknowledged before the node stops or the upload fails, without a pending marker surviving. A later normal freshness download can then overwrite the acknowledged local repository. Make marker persistence fallible and atomic, and do not acknowledge the push or spawn its release unless that persistence succeeds.

### P1: Serialize the fork archive publication with writes to the new fork

The database row makes a fork addressable before its clone and archive publication complete (`crates/gitlawb-node/src/api/repos.rs:1624-1701`). `release_after_write` calls `archive.upload` without the per-repository lock (`crates/gitlawb-node/src/git/repo_store.rs:335-367`), while normal writers use `acquire_write`. A concurrent push can upload a newer snapshot and then have the delayed fork snapshot overwrite it in the remote archive. Hold the same write lock across publication, including the necessary clone phase, or route fork publication through the locked write protocol.

### P2: Reject a zero-sized advisory-lock pool at configuration parsing

The unrestricted `u32` configuration value reaches `max_connections` (`crates/gitlawb-node/src/config.rs:174-179`, `main.rs:284-291`). A value of zero permits startup, but every `acquire_write` waits the configured 300 seconds and then fails, effectively disabling writes. Require a Clap range of `1..` and add a parsing test.

kevincodex1:
@coderabbitai make an intensive review on this

kevincodex1:
@coderabbitai please do an intensive review again

beardthelion (dismissed):
Re-reviewed the merged state at 3cbcb2bf, including the gauge and stripe commit that landed mid-review. The points raised literally in the prior rounds are addressed: `mark_pending` propagates its write error and `git_receive_pack` falls back to strict upload-before-ack instead of acking (`repos.rs:984-999`), the zero-sized lock pool is rejected at parse by a guard I confirmed goes red when removed, and the marker gauge is now maintained by deltas at the create and clear sites rather than only at startup. Moving the divergence check under the advisory lock is the right shape. Two things still block: the test meant to prove the ack-window fix does not exercise it, and there is a crash window that permanently wedges a repo whose content already matches storage.

## Findings

- [P1] Clear the etag cache in the ack-window test so it exercises the marker
  `crates/gitlawb-node/src/git/repo_store.rs:1595`
  `mark_pending_alone_protects_the_ack_window` still passes with the entire body of `mark_pending` replaced by `Ok(())`. The test never clears `store.versions`, so `sync_down_if_stale` returns on the etag-cache hit before it consults the marker, which a real restart would not have. Adding `store.versions.lock().await.clear();` after `drop(guard)` makes it fail with the marker gutted and pass with it restored. I ran all three states, on this head.

- [P1] Do not treat a crash between a successful upload and the marker clear as divergence
  `crates/gitlawb-node/src/git/repo_store.rs:549`
  `upload_locked_with_marker` advances storage, then rewrites and unlinks the marker. If the process dies in between, the marker still names the pre-upload base while local and storage are byte-identical, and every later freshness check reads remote != base as divergence. I reproduced it on this head: the startup sweep reports 0 reuploaded and 1 still pending, then `acquire_write` fails with "advanced while local changes were pending upload ... reconcile manually" on a fully consistent repo. Recovery requires deleting the marker on the node's disk. The doc on `clear_pending_upload_after_success` (`repo_store.rs:1016`) covers the rewrite-to-unlink window but not the upload-to-rewrite one. Recording the intended next etag in the marker before the upload, so a marker whose in-flight value matches the current remote is recognised as this node's own completed upload, would close it. I have not compiled that, so treat it as a direction rather than a specified fix.

- [P2] Restore the previous etag when the blob rename fails
  `crates/gitlawb-node/src/storage/fs.rs:131`
  `put` publishes the sidecar before the blob so a crash yields new-etag/old-content, which the comment argues is the safe direction. That held before; it does not now. Under the new rule a remote etag not matching the marker base is terminal divergence, so a failed blob rename leaves a new etag over stale content and hands the next write the same permanent wedge as the finding above. Reading the old sidecar inside the publish lock and restoring it on the error path keeps the claimed property. The stripe change does not affect this.

- [P2] The fork publication window is documented but still open
  `crates/gitlawb-node/src/api/repos.rs:1671`
  Routing `release_after_write` through the locked upload closes the literal complaint. The ask was to hold the lock across publication including the clone, and the row still makes the fork pushable before clone and upload finish with no lock held, as the comment itself states. The window is confined to the forker's own namespace, so I am fine deferring it, but it should be a tracked issue rather than a comment.

- [P3] Log the sweep's directory-read failures
  `crates/gitlawb-node/src/git/repo_store.rs:439`
  A failed `read_dir` on `repos_dir` returns `(0, 0)` and a failed per-owner read continues silently, so a sweep that scanned nothing looks identical in the logs to a node with no pending markers. With the gauge now seeded from that same scan, a silent failure also publishes a zero.

Minor: `.pending-upload.tmp-<uuid>` files are unlinked only on the rename error path, so a crash mid-rename leaves litter nothing collects.

Checked and clean: the sweep's reconstruction of owner and repo from directory names cannot produce another tenant's storage key, since the on-disk slug is the same string `local_path` and the archive key are built from and the did:key alphabet excludes the substituted characters; the advisory lock key change has no persisted dependency on the old value and the 8-byte truncation is not a practical collision risk, though the one-deploy mixed-binary gap the comment notes is real; the guard in `git_receive_pack` releases exactly once on all four branches; the new gauge deltas are balanced, since the increment sits after the marker is actually created and both decrements are conditional on the unlink succeeding.

On the open bot thread asking to promote `list` onto `BlobStore`: nothing here needs it and implementing it across every backend is scope with no caller. I would close that as won't-do rather than grow the trait.

jatmn (changes requested):

## Findings

- [P1] Make the filesystem blob publication recoverably atomic
  `crates/gitlawb-node/src/storage/fs.rs:128`
  The sidecar ETag is renamed into place before the matching blob. If the process dies between the two renames, storage contains the old archive bytes but the new archive MD5. A pending marker has already recorded that MD5, so recovery treats the stale bytes as this node's completed upload, clears the marker, and caches/serves the old repository as the new generation. Publish the bytes and generation as one recoverable unit (or ensure recovery can reject an ETag whose bytes do not match) before using the sidecar as the freshness authority.

- [P1] Do not base crash recovery on a client MD5 for arbitrary backends
  `crates/gitlawb-node/src/storage/archive.rs:62`
  The marker records the archive MD5 before `put`, then identifies a completed upload by comparing it with the backend ETag. That is not a valid contract for the advertised IPFS backend (which returns a CID), or for standard S3 configurations such as SSE-KMS, where an ETag is not the object MD5. A crash after a successful PUT but before marker removal is consequently classified as divergence, leaving the marker in place and blocking every later write even though local and remote contain the same archive. Persist and compare a backend-valid generation/content token, or provide a recovery path that validates the actual object rather than assuming its ETag.

- [P1] Fail the upload before PUT when in-flight provenance cannot be persisted
  `crates/gitlawb-node/src/git/repo_store.rs:1107`
  `record_inflight_upload` drops both write and rename failures, but its callers still execute the PUT. If that rewrite fails and the process stops after the PUT succeeds, only the old-base marker remains; recovery sees the new remote generation as external divergence and wedges the repo. The intent rewrite needs to return an error and prevent the PUT/early acknowledgement unless it has been atomically recorded.

- [P1] Do not acknowledge issue and merge mutations with no recoverable durable state
  `crates/gitlawb-node/src/api/issues.rs:73`
  `RepoWriteGuard::release` logs a marker-write failure and continues; if the subsequent archive upload also fails, it removes the version cache but there is no marker for startup or the next access to protect the local mutation. These handlers deliberately log and ignore that release error after the Git mutation, so a later fresh write downloads the stale archive over an acknowledged issue, close, or merge. Either propagate this combination as a failed request or retain durable retry/provenance state before reporting success.

- [P1] Keep a new repository non-addressable until creation/fork publication is committed
  `crates/gitlawb-node/src/api/repos.rs:210`
  Both creation paths insert a visible DB row before initialization/publication, then delete that row and its directory when their own later upload fails. A client can push through the visible claim in that window and successfully upload its update; the original create/fork request can then run its compensation and delete the pushed repository's local directory and DB row. The fork code even documents a narrower version of this race. Hold the repository lock across the claim-to-publication lifecycle or expose the row only after it can no longer be compensated away.

- [P1] Do not offer the IPFS MFS backend as shared archive storage
  `crates/gitlawb-node/src/storage/ipfs.rs:40`
  Every operation uses Kubo's local MFS path (`/gitlawb/<key>`). No shared mutable pointer, CID registry, IPNS publication, or peer lookup is written, so two nodes using separate Kubo daemons read and write different namespaces. A push accepted on one node is invisible to another, violating the cross-node archive/lock assumptions behind this feature. Either restrict this backend to a single shared Kubo/MFS instance or publish and resolve a shared versioned mapping.

- [P1] Preserve exclusion while migrating the advisory-lock key
  `crates/gitlawb-node/src/git/repo_store.rs:1155`
  Replacing the base branch's `DefaultHasher` key with a SHA-256 key means old and new nodes acquire different Postgres advisory locks for the same repository during the required rolling deployment. They can both sync, mutate, and PUT competing snapshots, so the later upload loses an acknowledged push. The comment acknowledges this gap but leaves it live. Acquire both generations of the lock (or otherwise perform a compatible lock-key migration) until no old nodes remain.

jatmn (changes requested):

## Findings

- [P1] Publish filesystem archive bytes and their freshness token atomically
  `crates/gitlawb-node/src/storage/fs.rs:128`
  The filesystem backend publishes the new `.etag` sidecar before it renames the new archive body.  A crash in that interval leaves the old bytes paired with the new MD5.  Another node can download those old bytes and cache the new token; when the intended bytes are later retried, they have the same MD5, so that node will keep treating its stale copy as current and can overwrite the recovered push on its next write.  Use a single atomically-published generation/manifest (or otherwise make body and version recover as one unit) rather than allowing the token to lead the content.

- [P1] Do not use per-node PostgreSQL locks as the cross-node write serializer
  `crates/gitlawb-node/src/git/repo_store.rs:347`
  The new comments explicitly describe Fly apps as having separate Postgres instances, but these advisory locks are acquired from each node's local `DATABASE_URL`.  Two nodes sharing an archive backend can therefore both acquire their local lock, read the same archive, accept independent pushes, and PUT competing snapshots; the later upload silently loses the first acknowledged push.  This needs a serializer/CAS at a shared boundary (or an enforced shared-Postgres deployment), not a session lock local to each node database.

- [P2] Only clean up a repository directory that this creation attempt created
  `crates/gitlawb-node/src/git/repo_store.rs:440`
  `create_published` removes `local_path` and its pending marker after every failed build/publish without establishing that this invocation created either.  If a create/fork collides with a pre-existing local repo (for example a DB-missing legacy repo, or a shared-mount race), `init_bare`/`git clone --mirror` fails and this error path recursively deletes that existing repository; it also removes its recovery marker if directory deletion fails.  Track successful ownership of the destination and clean only a directory/marker created by this attempt.

- [P2] Restrict the startup orphan sweep to the exact temporary-directory format
  `crates/gitlawb-node/src/storage/archive.rs:178`
  The new startup cleanup recursively deletes any dot-prefixed directory containing `.tmp-extract.` or `.bak-`.  Peer mirroring intentionally accepts names outside the API validator, so a legitimate mirrored repo such as `.project.bak-2026` is stored as `.project.bak-2026.git` and matches this broad predicate.  It will be deleted on every restart.  Parse and require the precise generated `.{repo}.git.(tmp-extract|bak)-{UUID}` form before removing a directory.

beardthelion (dismissed):
Re-reviewed the merged state at `1d5b479a` (clean merge, base current with main). CI is green across all 15 checks. The storage abstraction itself is good work and I want it in: the trait, the three backends, the archive layer and the key/traversal validation are all sound, and the authorization surface is untouched (no gate added, removed or reordered anywhere in the diff).

The problem is no longer any single defect. Eleven maintainer review rounds and fourteen commits in, the shape of the loop is the finding: the original commit was +1003, the rounds since have added +2457 on top of it, and `repo_store.rs` alone went from +186 in the original change to +1648 across the fixes. Three of the four issues raised in the latest round are defects in remedies from earlier rounds, not in the original design. The sidecar-before-blob ordering exists because I asked for it; the unconditional creation cleanup exists because jatmn asked for clone-failure cleanup; the startup sweep exists because I asked for one. Each round has been individually correct and the aggregate is still going backwards.

So I'm settling the scope rather than running another round. Details below, findings first.

## Findings

- [P1] Split the crash-recovery protocol out and land the storage layer on its own
  `crates/gitlawb-node/src/git/repo_store.rs:96`
  The BlobStore trait, the three backends, `RepoArchive` and the config surface are ready. The pending-marker machinery (marker lifecycle, versioned base, in-flight provenance, content-validated recovery, startup sweep, dual lock keys) was designed incrementally in review comments, and it is where every remaining defect lives. Land this PR with `async_upload` fixed off and the strict upload-before-ack path only, and drop the marker protocol entirely from it. Then bring the write-back path back as its own change with a written design for the freshness contract, so we can review one protocol instead of six accumulated patches to one. That is a scope call, not a criticism of the fixes: the individual rounds landed what was asked for each time.

- [P1] Publish the filesystem archive body and its freshness token as one unit
  `crates/gitlawb-node/src/storage/fs.rs:140`
  Confirming jatmn's finding, and the deciding detail is on the reader side rather than the writer side. A crash between the two renames leaves new-token/old-bytes; a second node then takes the plain freshness path, reads the token at `repo_store.rs:187`, downloads the old bytes, and caches that token at `repo_store.rs:217` without ever validating content. When the intended bytes are retried they carry the same MD5, so that node's cached token still matches and `repo_store.rs:203` keeps skipping the download indefinitely. The content validation added last round only protects the node holding the marker, which is not the node that loses the push. Reversing the two renames does not fix it either; the body and its token have to publish as one recoverable generation.

- [P2] Reserve the lazy-migration slot before spawning, not after it succeeds
  `crates/gitlawb-node/src/git/repo_store.rs:283`
  `acquire()` spawns a migration task whenever the repo is not in the `migrated` set, and the set is only populated in the success arm at `:295`. A repo whose upload keeps failing therefore spawns a fresh task on every read, and each task now takes the per-repo advisory lock from the bounded pool that `main.rs:284` sizes for pushes. Eight sequential reads produce seven upload attempts on a failing backend; I ran that. The reads are anonymous on a public repo (`list_commits` in `api/repos.rs:360` takes `Option<Extension<AuthenticatedDid>>`), so this hands an unauthenticated caller a lever on push availability that the base commit did not have, since background migration there took no lock and no dedicated pool. Claim the key under the same mutex guard as the check before spawning, and give the opportunistic path a short lock budget instead of the 300s write budget.

- [P2] Only remove a repository directory this attempt actually created
  `crates/gitlawb-node/src/git/repo_store.rs:440`
  `init_bare` bails when the destination exists (`git/store.rs:9`), and the failure path then recursively deletes it. A create against a name whose directory is already on disk without a DB row (a legacy repo, or one whose row was compensated away) destroys that repository. I ran it: `create_published` on a pre-existing `legacy.git` returns "initializing bare repo" and leaves the directory gone. `clear_pending_upload` at `:447` also runs when the directory removal failed, so a partial failure takes the recovery marker with it. Capture whether the destination existed before `build`, and gate both the removal and the marker clear on having created it.

- [P2] Match the exact generated swap-directory name before deleting anything
  `crates/gitlawb-node/src/storage/archive.rs:178`
  The predicate is a leading dot plus a substring, so it deletes any repo directory whose name happens to contain `.bak-` or `.tmp-extract.`. Peer mirroring builds its path straight from the peer-supplied string at `sync.rs:239` without going through `validate_repo_name`, so a mirrored `.project.bak-2026` lands as a directory this sweep removes on every restart. I ran both cases against the real sweep: `.project.bak-2026.git` and `.notes.tmp-extract.draft.git` are each removed. The existing test only offers a live `repo.git` and a `.keep` file as negatives, so neither shape is covered. Parse the generated form and require a trailing UUID rather than matching a substring.

- [P2] Take the publish lock before reading the etag being rolled back
  `crates/gitlawb-node/src/storage/fs.rs:125`
  `prev_etag` is captured at the top of the blocking task, but the stripe lock is not taken until `:136`. If another put to the same key completes in between and this put's blob rename then fails, the rollback at `:152` writes the pre-existing token back over the other put's content. That is the one direction the sidecar design exists to prevent: a token that says "unchanged" over content that changed. Move the capture and the rollback inside the guard. Narrow (it needs a rename failure racing a same-key put), and I confirmed the ordering by reading rather than by running it.

- [P3] Fix two claims in the PR description, and state the lock's precondition in the code
  `crates/gitlawb-node/src/git/repo_store.rs:348`
  The description still advertises `list` on the trait; it was dropped and now survives only as a test helper on the fs backend. It also says the lock being held until the upload finishes means "cross-node consistency holds", which overstates it in the same way `acquire_write`'s doc comment does: the lock excludes writers that share one Postgres, and during the window other nodes still serve the stale archive, which `config.rs:171` correctly documents. On jatmn's related point, machines of one Fly app share a `DATABASE_URL`, so they do exclude each other, and I could not find the code comment that finding cites about separate Postgres instances. The gap worth closing is documentary: say plainly that a fleet sharing one archive backend must share one database, and stop calling this cross-node consistency.

The whole-archive-in-memory thread from the earlier round is still open, and last round's recovery check made it heavier by fetching the entire remote object to hash it (`repo_store.rs:254`). If the scope split above happens that thread goes with the recovery protocol; if it doesn't, it needs an answer here.

beardthelion:
Follow-up to my last round, settling something that spans this PR and #215 rather than sitting inside either.

**The advisory-lock layer belongs to #215.** Both PRs replace `DefaultHasher` and both cite #210, but the derivations differ: this branch hashes `format!("{owner}/{repo}")` and folds big-endian, #215 hashes `owner` + `":"` + `repo` and folds little-endian. Same repo, different key. Whichever merges second re-keys every repository a second time and makes operators pay the exclusion-gap window twice. #215 is the focused change that exists only to close #210 and it got there first, so it owns the key. This branch should drop `advisory_lock_key`, `legacy_advisory_lock_key` and `lock_keys_for` and take #215's once it lands.

Worth being explicit about one trap: do not rebase the dual-key transition mechanically. Its legacy half is main's `DefaultHasher` key, which stops being main's key the moment #215 merges, so carrying it forward would leave a transition that covers a key nobody runs while missing the one everybody does.

**The dedicated lock pool should move with it.** `GITLAWB_ADVISORY_LOCK_POOL_SIZE` and the separate bounded pool are the right design and #215 does not have them yet; jatmn asked for exactly that there, and #215 already carries the connection-pinning half it pairs with. Sending it there keeps one implementation rather than two. This branch needs it more than #215 does, since holding the lock across a full archive upload is what makes a shared pool untenable, so say so on #215 if you want the sizing argued from your use case.

To be clear about what that decision is and is not: it settles where the code lands, not whose mechanics are better. Yours are better in three places I want carried across rather than lost. `LockedConn::acquire` wraps the connection in the guard before the first attempt, so cancellation anywhere in the retry loop closes the session instead of repooling a possibly-granted lock; #215 holds a bare connection there and I raised it. It detaches on a poll error, on the same reasoning. And `unlock` detaches when the unlock query errors instead of returning the connection to the pool, where #215 discards that result entirely. Porting those three into #215 is the point of the move.

## Findings

- [P1] Keep the connection inside `LockedConn` across the unlock awaits
  `crates/gitlawb-node/src/git/repo_store.rs:937`
  `self.conn.take()` moves the connection out before the loop, so from that point `Drop` sees `None` and the close-on-failure path you wrote can only fire if the query returns an error. Cancellation during either unlock, or between the two keys, leaves a bare `PoolConnection` that returns to the pool with a lock still held. That matters more than a single stuck lock because advisory locks are counted per session: I checked against Postgres directly, and two acquires followed by one unlock leave the session still holding it. So the next writer handed that pooled connection re-acquires reentrantly and its own unlock only decrements the count, leaving the lock held until sqlx recycles the connection, which on the 0.8.6 defaults is 10 minutes idle or 30 minutes lifetime. Borrow through `self.conn` across the awaits and take it only once every key is confirmed released. Note also that `pg_advisory_unlock` reports "this session did not hold it" as a `false` return with a WARNING rather than an error, so `execute()` cannot see that case at all; `fetch_one::<(bool,)>` can. I have raised the same shape on #215.

- [P2] Do not pin a lock-pool connection for the whole acquire wait
  `crates/gitlawb-node/src/git/repo_store.rs:858`
  The connection is checked out before the retry loop and held across every sleep, so a writer waiting on a contended repo occupies one of the 16 lock-pool connections for up to the full 300-second budget without doing work. The dedicated pool contains the damage, which is the main reason it is worth having, but the containment is to the write path rather than to nothing: 16 waiters block every subsequent push, issue, close and merge on the node. Reachability is lower than it looks from the push route, since `create_issue` gates on `authorize_repo_read` and nothing stronger (`api/issues.rs:46`) before reaching the lock at `:66`, so a read-level identity can occupy those connections. Only the winning attempt needs its connection pinned, so acquiring inside the loop and dropping before each sleep keeps the wait cheap. That change interacts with the guard-wrapping above, so whoever lands it should keep the wrap for the attempt that is actually in flight.

Nothing else changes from the last round: the scope split and the four findings there still stand.

jatmn (changes requested):

## Merge readiness

- [P1] Resolve the failing CodeQL check before merge
  `crates/gitlawb-node/src/storage/archive.rs:292`
  The current head's CodeQL check fails with nine new high-severity alerts in code changed by this PR. GitHub reports the PR as mergeable but blocked; rerun and resolve the alerts on the final head before merge.

## Findings

- [P1] Keep sidecar rollback inside the filesystem publish critical section
  `crates/gitlawb-node/src/storage/fs.rs:125`
  `prev_etag` is read before the striped `put_publish_lock` is acquired. The lock is then dropped when `write_and_swap` returns, but the error handler restores that earlier snapshot afterward. A concrete interleaving is: writer A reads the old sidecar, publishes A's new sidecar, and fails renaming A's blob; writer B then takes the stripe and successfully publishes B's blob plus B's sidecar; finally A resumes its unlocked cleanup and overwrites or removes B's sidecar using A's stale snapshot. `head()` now reports an etag that does not describe B's bytes, and `sync_down_if_stale` can mistake a stale cached etag for current storage, skip the download, and later upload its stale repository over B's archive.

  The root cause is treating rollback as outside the atomic publish transaction. Make the per-key critical section cover the prior-sidecar snapshot, sidecar/blob publication, and every rollback action; do not let a failed operation mutate sidecar state after another same-key operation has committed. Add a deterministic interleaving test where A fails after publishing its sidecar and B succeeds before A's cleanup, then assert that `head()` reports B's etag and bytes.

- [P1] Retain the in-process repo lease through async write-back
  `crates/gitlawb-node/src/api/repos.rs:2363`
  The async path spawns `guard.release(true)`, which intentionally retains the PostgreSQL advisory lock until the storage upload finishes, then drops `lease` before returning the response. A later push for that repo therefore passes the in-process lease, acquires the per-source and global write permits, and blocks in `acquire_write` behind the background task's advisory lock. Enough requests from different source keys can occupy the shared write pool for the storage timeout window and shed unrelated repositories—the exact global-admission starvation that the lease-before-admission ordering at lines 2172–2181 is intended to avoid.

  The root cause is that the two serializers no longer cover the same write lifetime: the lease protects only the request until acknowledgement, while the advisory lock protects the still-running write-back operation. Transfer a clone/ownership of the existing lease into the write-back task and release it only after `guard.release(true)` completes (including its error path), or restructure admission so a same-repo request cannot obtain shared write permits until it can actually proceed past the prior advisory lock. Preserve the opt-in early acknowledgement and the existing advisory-lock serialization; add a test with async upload stalled that proves a second same-repo push remains parked at the lease without reducing available global write permits.

beardthelion (changes requested):
Re-reviewed the merged state at `9ce694b0`. This round is current main merged in plus one commit fixing the two CI failures that merge caused. Neither the scope call from my last round nor any of its seven findings is answered, and the CodeQL half of this round's own fix did not land: the check is still failing on this head, with nine open `rust/path-injection` alerts all stamped `9ce694b0`. The F4 restore is right, and its gate's region markers each occur exactly once in the file, so it is scoped where the docstring says it is.

## Findings

- [P1] Land the storage layer without the crash-recovery protocol
  `crates/gitlawb-node/src/git/repo_store.rs:96`
  Repeating last round's call unchanged, because nothing here addresses it and the new defect below is again a defect in the marker machinery rather than in the abstraction. The BlobStore trait, the three backends, `RepoArchive` and the config surface are ready. Ship them with `async_upload` fixed off and the strict upload-before-ack path, and drop the pending-marker protocol from this PR. Bring the write-back path back as its own change with a written freshness contract. Until that happens I am not going to approve this PR, so further rounds on the marker code inside it are not worth your time.

- [P2] Cover the four `decompress_repo` sinks CodeQL is still flagging, or dismiss the alerts with the reason written down
  `crates/gitlawb-node/src/storage/archive.rs:300`
  The new barriers cover the sibling paths (`tmp_dir` at `:262`, the backup at `:296`), but the three renames and the `create_dir_all` still consume `local_path` as it arrived, and that parameter is what `rust/path-injection` traces. I checked the reachable callers (`repo_store.rs:293` and `:399`) and each passes a validated path, so I do not believe there is a live traversal. The problem is that the commit message asserts the component walk is a barrier "static analysers (CodeQL `rust/path-injection`) recognise", and the check disproves that on this head. Either re-establish the barrier on `local_path` inside `decompress_repo` the way `sync_down_if_stale` does, or dismiss the nine alerts with the caller argument attached. A required check that stays red is not a merge path.

- [P2] Reject a traversal component, not the `..` substring, when deriving sibling paths
  `crates/gitlawb-node/src/git/repo_store.rs:1612`
  `validate_repo_name` accepts a trailing dot, so a repo named `a.` lives at `a..git`, and every sibling this round derives from that directory (`.a..git.pending-upload`) contains `..` and is rejected by `validated_sibling_path`. `sync_down_if_stale` propagates that with `?` at `:174`, and both `acquire_fresh` (`:424`) and `acquire_write` (`:481`) run through it, so with storage enabled that repo is neither readable nor writable. I ran it: `validate_repo_name("a.")` returns `Ok(())` while `pending_upload_marker` on the resulting directory returns `Err(sibling file name contains a parent-directory reference)`, with `plain.git` as the passing control. `POST /repos` (`api/repos.rs:238`) and the fork route (`:3118`) both reject dots outright, so the reach is the mirror path: `notify_sync` gates on `validate_repo_slug`, which accepts `a.`, and a known peer can enqueue it. Check the derived name for a `ParentDir` component, or normalise the directory name before building siblings from it.

Not asks, for the record. My own P2 from last round about the swap-directory sweep was wrong about reachability and I withdraw that half: I said a peer-mirrored `.project.bak-2026` reaches the substring predicate at `archive.rs:179` because `sync.rs` built the path without validating the name. The #272 fix arrived with this merge, `sync.rs:231` validates the slug now, and `validate_repo_name` rejects a leading dot. No creation surface I can find admits one. The predicate is still looser than the names it generates, but that is hardening, not a live path. The rest of that round (filesystem publish atomicity, the migration slot reservation, the create-failure directory removal, the `prev_etag` capture ordering) is unchanged at this head.

PR #92, opened 2026-06-24
fix(security): gate webhook creation through the public-host validator (#81)

### LINKED ISSUE #81: Webhook delivery is an unvalidated SSRF sink with no repo-ownership check
Webhook creation accepts any URL behind only a scheme check, and the create handler does not verify the caller owns the repo. On repo events the node POSTs to the stored URL, so this is an SSRF sink reachable by any authenticated DID.

### Where
- `crates/gitlawb-node/src/api/webhooks.rs:43` validates the URL with only `starts_with("http://") || starts_with("https://")`. No private/loopback/link-local check.
- The handler calls `get_repo` only to confirm the repo exists; it never checks ownership. The route's auth layers verify identity but not per-repo ownership.
- `crates/gitlawb-node/src/webhooks.rs:85` does `client.post(&hook.url)` on events (e.g. fired from `api/pulls.rs:84`), with no re-validation.

### Impact
Any DID that can authenticate can register a webhook on any repo with `url=http://127.0.0.1:<port>/...` or `http://169.254.169.254/latest/meta-data/...` and make the node POST to internal/loopback/metadata endpoints. The shared client also follows redirects (see the related outbound-redirect finding on #78), so even a public URL can be bounced inward.

### Suggested fix
- Gate the URL through the same public-host validator used for peers, but note that validator still needs the IPv4-mapped/CGNAT/trailing-dot tightening tracked on #78, otherwise the webhook path inherits those same bypasses.
- Add an ownership check using the existing `require_owner` helper (`api/visibility.rs`), which matches the caller against both the full DID and its short form. A bare `auth.0 == record.owner_did` equality would 403 a legitimate owner whose stored DID is the full form.
- Harden the delivery client (redirect policy + connect-time guard), same as the peer clients.

beardthelion:
Webhook creation accepted any http(s) URL behind a scheme check and never re-validated the host, so any authenticated repo owner could register a webhook at `http://127.0.0.1:<port>/` or `http://169.254.169.254/latest/meta-data/` and turn the node's on-event POST into an SSRF probe against loopback, private, or cloud-metadata endpoints. This routes the registration URL through `is_public_http_url`, the hardened validator the peer announce path already uses, so the two SSRF surfaces share one definition of "public".

The other two prongs the issue raised are already closed on main: the owner check is present (`require_repo_owner`, full-and-short DID match), and delivery runs on the shared no-redirect client from #78, so a public URL that 302s inward is not followed.

## The validator gap this surfaced

Wiring the validator into webhooks exposed a real bypass in it: it folded IPv4-mapped and IPv4-compatible IPv6 but not 6to4 or NAT64, so `http://[2002:7f00:1::]/` (which decodes to 127.0.0.1) sailed past. The webhook gate is only as strong as the validator behind it, so this hardens the shared function rather than the webhook path alone:

- Fold 6to4 (`2002::/16`) and the NAT64 well-known prefix (`64:ff9b::/96`) down to their embedded v4 and re-run the range checks.
- Reject the NAT64 local-use prefix (`64:ff9b:1::/48`, RFC 8215) and the rest of `64:ff9b::/32` outright. The embedded v4 sits at a prefix-length-dependent offset (RFC 6052 §2.2), so attempting to decode every variant is error-prone; refusing the block is the safe call, and a NAT64 literal is never a legitimate webhook or peer target.
- Reject the RFC 1122 `0.0.0.0/8` "this host" block.

The change is tightening-only (it never accepts an address it previously rejected), so all four callers of the validator move in the same direction: the webhook gate, the peer announce handler, the DB-boundary `upsert_peer` writer, and the boot-time prune. A 6to4 or NAT64 literal that wraps a genuinely public v4 still passes, so no legitimate peer or webhook is rejected or pruned.

## Tests

Extended the peer validator's reject and accept sets for the 6to4 / NAT64 / local-use / `0.0.0.0/8` cases, and added a webhook-site regression test pinning issue #81's exact exploit URLs so the gate cannot regress to the old scheme-only check. `cargo test -p gitlawb-node` is green (172 tests), clippy clean.

## Out of scope

Noted for separate follow-up, not addressed here: DNS rebinding between registration and delivery (accepted residual, same posture as the peer path), and three pre-existing issues this review surfaced: the gossip task building a redirect-following client, unauthenticated `list_webhooks` exposing webhook target URLs, and the validator's lack of a port restriction.

Closes #81.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced URL validation for peers and webhooks to prevent access to private/loopback addresses and internal services.
  * Improved handling of IPv6 addresses to better detect and reject unsafe URL targets.

* **Tests**
  * Added comprehensive tests validating URL safety against known attack patterns.

### DIFF
--- a/crates/gitlawb-node/src/api/peers.rs
+++ b/crates/gitlawb-node/src/api/peers.rs
@@ -40,6 +40,38 @@ pub struct PeerResponse {
     pub reachable: bool,
 }
 
+/// Extract an IPv4 address embedded in an IPv6 literal across the transition
+/// formats that carry one: IPv4-mapped (`::ffff:a.b.c.d`), IPv4-compatible
+/// (`::a.b.c.d`), 6to4 (`2002:WWXX:YYZZ::/16`), and the NAT64 well-known prefix
+/// (`64:ff9b::/96`). Returns `None` for native IPv6. Callers fold the result
+/// back through the v4 range checks so loopback/private addresses smuggled in
+/// via any of these encodings are rejected, not just the mapped/compatible pair.
+///
+/// INCOMPLETE-NAT64 CONTRACT: non-well-known NAT64 prefixes (e.g. the RFC 8215
+/// local-use `64:ff9b:1::/48`) are deliberately **not** decoded here — their v4
+/// sits at a prefix-length-dependent offset (RFC 6052 §2.2) — so they return
+/// `None`. Any caller that needs them blocked must reject the wider
+/// `64:ff9b::/32` itself; `is_public_http_url` does this in its native-v6 arm.
+fn embedded_ipv4(v6: std::net::Ipv6Addr) -> Option<std::net::Ipv4Addr> {
+    if let Some(v4) = v6.to_ipv4_mapped().or_else(|| v6.to_ipv4()) {
+        return Some(v4);
+    }
+    let s = v6.segments();
+    // 6to4 (2002::/16) embeds W.X.Y.Z in the next two segments.
+    if s[0] == 0x2002 {
+        let [a, b] = s[1].to_be_bytes();
+        let [c, d] = s[2].to_be_bytes();
+        return Some(std::net::Ipv4Addr::new(a, b, c, d));
+    }
+    // NAT64 well-known prefix (64:ff9b::/96) embeds the IPv4 in the low 32 bits.
+    if s[0] == 0x0064 && s[1] == 0xff9b && s[2..6] == [0, 0, 0, 0] {
+        let [a, b] = s[6].to_be_bytes();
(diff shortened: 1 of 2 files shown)


jatmn (approved):
### Findings

No findings.

I completed the no-findings challenge gate. The one potential bypass I investigated — IPv6 literals with zone IDs such as `http://[::1%25eth0]/` — is blocked at `reqwest::Url::parse` (returns `InvalidIpv6Address`), so it cannot reach the validator. Other checked classes (handler wiring regression, boot-time prune over-deletion, NAT64 well-known prefix misidentification) also held up.

@kevincodex1 LGTM

kevincodex1 (approved):
nice one! thank you

Issue #76, opened 2026-06-22
Drop hickory advisory ignores once libp2p-dns ships hickory 0.26


beardthelion:
Tracking issue for the `cargo audit` suppressions added in #57.

`.cargo/audit.toml` ignores two reachable advisories in `hickory-proto 0.25.2`:

- `RUSTSEC-2026-0118` — NSEC3 closest-encloser unbounded loop
- `RUSTSEC-2026-0119` — O(n^2) name-compression CPU exhaustion

These are reachable: the node wraps its QUIC transport in the system DNS resolver (`crates/gitlawb-node/src/p2p/mod.rs:238`), so a malicious DNS response can hit these DoS paths. They are suppressed only because there is no available fix — `hickory-proto >=0.26.1` is required, and the latest `libp2p-dns` on crates.io (0.44) still pins hickory 0.25.

Action: when a `libp2p-dns` (or libp2p major) release using `hickory >=0.26.1` is available, bump it, then remove both `RUSTSEC-2026-0118` and `RUSTSEC-2026-0119` from `.cargo/audit.toml`.

Enforcement: the scheduled audit workflow (`.github/workflows/audit-schedule.yml`) re-runs `cargo audit` without the ignores weekly and hard-fails if `hickory-proto` moves off 0.25.x while the ignores remain, so this will not silently outlive the fix.

beardthelion:
Worth knowing before this lands: the weekly drift guard turns into a permanent false red the moment the ignores come out.

`audit-schedule.yml:74-75` extracts the version under `set -euo pipefail`:

```
set -euo pipefail
version="$(grep -A1 'name = "hickory-proto"' Cargo.lock | grep '^version' | head -1 | cut -d'"' -f2)"
```

Once hickory-proto leaves the lockfile the inner grep exits 1, pipefail carries that to the assignment, and the step dies right there. It never reaches its own `echo "hickory-proto in Cargo.lock: ${version:-not present}"` on the next line, and it never reaches the branch written for exactly this case at `:84`, whose comment says "OR was removed entirely (dead ignore entries left behind). Both must fail."

Four fixtures through the extracted pipeline:

```
(i)   0.25.x present, ignores present    pass, exit 0
(ii)  absent, ignores present            exit 1, no ::error:: annotation
(iii) 0.26 present, ignores present      exit 1, WITH the designed annotation
(iv)  absent, ignores REMOVED            exit 1   <- the post-#76 state
```

[...]

PR #105, opened 2026-06-25
chore(deps): bump tar to 0.4.46 and rand to 0.9.3 to clear advisories

beardthelion:
Two lockfile-only security bumps, no manifest or source changes.

- `tar` 0.4.45 -> 0.4.46 clears the PAX header desynchronization issue (GHSA-3pv8-6f4r-ffg2, medium). We use tar in `crates/gitlawb-node/src/git/tigris.rs` for archive read/write, which is the affected read path; the fix is behavior-preserving.
- `rand` 0.9.2 -> 0.9.3 clears the `rand::rng()` custom-logger unsoundness (GHSA-cq8v-f236-94qc, low). Our own code only uses `OsRng`/`RngCore`, never `rand::rng()`, so this is hygiene for transitive users rather than a path we were exposed on. `rand` 0.8.6 stays, it is outside the vulnerable range.

These bumps carry no Cargo.toml version-requirement changes. The `cargo update` did move four packages onto older `windows-sys` selections already present in the tree, a normal resolver side-effect with no Windows target in our build:

- `colored`: `windows-sys` 0.61.2 -> 0.48.0
- `errno`: `windows-sys` 0.61.2 -> 0.52.0
- `rustix`: `windows-sys` 0.61.2 -> 0.52.0
- `tempfile`: `windows-sys` 0.61.2 -> 0.52.0

Left alone, with reason:
- `lru` 0.12.5 (GHSA-rhfx-m35p-ff5j, low) is pinned by aws-sdk-s3 (`lru ^0.12.2`) and cannot move to the 0.16.3 fix until aws-sdk-s3 bumps it.
- The two hickory-proto advisories remain accepted risk per `.cargo/audit.toml`, tracked in #76 (waiting on libp2p-dns to ship hickory 0.26).

Verification: `cargo build -p gitlawb-node`, `cargo test -p gitlawb-node` (220 passed), and `cargo audit` all pass; the tar and rand advisories no longer appear.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -329,7 +329,7 @@ dependencies = [
  "keccak-asm",
  "paste",
  "proptest",
- "rand 0.9.2",
+ "rand 0.9.3",
  "rapidhash",
  "ruint",
  "rustc-hash",
@@ -2107,7 +2107,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -2262,7 +2262,7 @@ checksum = "6ddc2d09feefeee8bd78101665bd8645637828fa9317f9f292496dbbd8c65ff3"
 dependencies = [
  "crc",
  "digest 0.10.7",
- "rand 0.9.2",
+ "rand 0.9.3",
  "regex",
  "rustversion",
 ]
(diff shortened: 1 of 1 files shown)


jatmn (changes requested):
This PR bumps `tar` and `rand` to patched versions that clear two advisories. I found one minor scope/accuracy issue to address before merge.

## Findings

- [P3] Document the transitive `windows-sys` lockfile changes
  `Cargo.lock`
  The PR description calls this a "precise lockfile bump" for `tar` and `rand`, but the diff also changes selected `windows-sys` versions: `dynarray` now resolves to `windows-sys 0.48.0` (was `0.61.2`), and `getrandom`, `rustix`, and `tempfile` now resolve to `windows-sys 0.52.0` (was `0.61.2`). These are normal resolver side-effects of a `cargo update`, but they are not mentioned in the PR body or commit message. Please update the description to list the full lockfile delta so reviewers do not mistake the `windows-sys` version changes for an unintended change.

beardthelion:
Good catch, updated the description with the full lockfile delta. Walking the diff, the `windows-sys` movers are `colored` 0.61.2 -> 0.48.0, plus `errno`, `rustix`, and `tempfile` 0.61.2 -> 0.52.0. All four are resolver side-effects with no Windows target in our build.

jatmn (approved):
No actionable findings.

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

Issue #100, opened 2026-06-25
Filter out bots from intelligent agents


kevincodex1:
### Problem / use case

We need to implement a captcha for AI agents, its a smart gate filter so we can prevent bot abuse. This could be a separate service on top of the node

### Proposed solution

Create a separate service for agent captcha 

### Does this touch the protocol?

No

Issue #98, opened 2026-06-24
fork_repo clones a full mirror after only a root (/) visibility check, bypassing path-scoped withholding


beardthelion:
`fork_repo` authorizes the caller only at the whole-repo `/` path, then `git clone --mirror`s the entire source. A caller allowed at `/` but denied a path-scoped subtree can fork the full mirror and obtain blobs the filtered `git_upload_pack` path would have withheld. This bypasses path-scoped visibility.

### Where

`crates/gitlawb-node/src/api/repos.rs`

- `fork_repo` authorizes at `/` only:
  ```rust
  let (source, _rules) =
      crate::api::authorize_repo_read(&state, &owner, &name, Some(auth.0.as_str()), "/").await?;
  ```
- then clones the whole repo as a mirror:
  ```rust
  let output = std::process::Command::new("git")
      .args(["clone", "--mirror", source_path..., disk_path...])
  ```

The `_rules` returned by `authorize_repo_read` are discarded, so any path-scoped withholding is ignored for the fork path.

### Impact

A non-owner caller with root (`/`) read access but without access to a path-scoped subtree (e.g. `/secret/**`) can fork the repo and read every blob in that subtree from their own copy. The read path (`git_upload_pack`) withholds those blobs via `withheld_blob_oids`; the fork path does not, so fork is a clean bypass of path-scoped visibility.

This is pre-existing (not introduced by #60). Surfaced while reviewing #60, which optimizes the same withheld-walk on the read path; the fork path defeats that invariant wholesale.

### Suggested remediation

Either:

1. Fail closed for non-owners when the source has any path-scoped rule (`has_path_scoped_rule`), using `did_matches(...)` for the owner-only bypass, and preserve the existing not-found-style response for protected-data failures; or
2. Build the fork from the same filtered projection used by `git_upload_pack` rather than from a full `--mirror`, so withheld blobs are never copied.

Option 1 is the smaller, safer fix; option 2 preserves fork-for-readers semantics but is more involved. Owner forks (and forks of repos with no path-scoped rules) must continue to work unchanged.

PR #108, opened 2026-06-27
feat(node): iCaptcha proof-of-intelligence gate on create_repo + register

kevincodex1:
## Summary
Adds an optional iCaptcha "proof-of-intelligence" gate on the two spam-prone
endpoints (`create_repo`, `register`). Callers present an Ed25519-signed proof
(earned by solving an escalating challenge at icaptcha.gitlawb.com); the node
verifies it OFFLINE with ed25519-dalek against the published public key and binds
it to the authenticated agent DID so proofs cannot be shared. Motivated by the
did_key spam flood that filled a node's disk.

## Impact / compatibility
- Backward-compatible by default. `ICAPTCHA_MODE` is `off` unless set, so merging
  and deploying changes nothing observable. No schema changes, no response-shape
  changes, and `git push` (receive-pack) is intentionally NOT gated.
- `shadow`: verifies and logs would-be rejections, always allows (safe to tune).
- `enforce`: rejects create_repo/register without a valid, strong-enough proof.
  This is the only breaking mode and is opt-in; roll out client proof support and
  verify via shadow first.
- Fail safe: if the public key cannot be loaded at startup (iCaptcha unreachable),
  the gate stays inert rather than rejecting everyone.

## Config (env)
- `ICAPTCHA_MODE` = off | shadow | enforce (default off)
- `ICAPTCHA_URL` (default https://icaptcha.gitlawb.com)
- `ICAPTCHA_PUBKEY` (optional base64url override; else fetched from /v1/pubkey)
- `ICAPTCHA_REQUIRED_LEVEL` (default 3)

Clients send the proof in the `X-ICaptcha-Proof` header.

## Testing
- [x] `cargo test -p gitlawb-node icaptcha` — 11 tests pass, including offline
      verification of a REAL captured proof (JS-signed, Rust-verified),
      tamper/expiry/wrong-DID/insufficient-level/missing-header rejections, and
      the off/shadow/enforce/fail-safe mode matrix.
- [x] `cargo fmt`, no new warnings. `Cargo.lock` diff is a single `base64` line.

## Notes / follow-ups
- This raises the cost of abuse but is not the hard cap. Recommend adding
  per-DID repo/creation quotas in the node as the absolute ceiling.
- Rollout: deploy with mode unset (inert) -> shadow (observe) -> enforce.

Address #100 

## Summary by CodeRabbit

* **New Features**
  * Added an optional iCaptcha “proof-of-intelligence” gate for **registration**, **repo creation**, and **repo forking** with `off`, `shadow`, and `enforce` modes plus DID-level validation.
  * In `enforce` mode, proofs are required and cannot be replayed.
* **Configuration**
  * Updated `.env.example` with `ICAPTCHA_MODE`, `ICAPTCHA_URL`, `ICAPTCHA_REQUIRED_LEVEL`, and optional `ICAPTCHA_PUBKEY`.
* **Bug Fixes**
  * Proof verification now happens earlier, and proof consumption occurs only after admissibility checks pass.
* **Tests**
  * Added unit tests for mode handling, validation failures, and one-time proof consumption.
* **Chores**
  * Added an iCaptcha proof ledger table and periodic cleanup of expired entries.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -93,3 +93,19 @@ GITLAWB_MAX_PACK_BYTES=2147483648
 # ── Sync ─────────────────────────────────────────────────────────────────
 # Enable automatic background sync from known peers
 GITLAWB_AUTO_SYNC=false
+
+# ── iCaptcha proof-of-intelligence gate ───────────────────────────────────
+# Optional gate on create_repo + register: require callers to present an
+# iCaptcha proof (X-ICaptcha-Proof header) earned at icaptcha.gitlawb.com.
+# Default off = fully inert (no behavior change).
+#   off     - gate disabled (default)
+#   shadow  - verify + log would-be rejections, but always allow
+#   enforce - reject requests without a valid, strong-enough proof
+ICAPTCHA_MODE=off
+# Base URL of the iCaptcha service.
+ICAPTCHA_URL=https://icaptcha.gitlawb.com
+# Optional base64url Ed25519 public key override; if unset it is fetched from
+# ICAPTCHA_URL/v1/pubkey at startup.
+ICAPTCHA_PUBKEY=
+# Minimum proof difficulty level required to pass.
+ICAPTCHA_REQUIRED_LEVEL=3
--- a/.env.example
+++ b/.env.example
@@ -103,7 +103,9 @@ GITLAWB_AUTO_SYNC=false
 #   enforce - reject requests without a valid, strong-enough proof
 ICAPTCHA_MODE=off
 # Optional base64url Ed25519 public key override; if unset it is fetched from
-# ICAPTCHA_URL/v1/pubkey at startup.
+# ICAPTCHA_URL/v1/pubkey at startup. Recommended in production: pinning the key
+# removes the startup dependency on the iCaptcha service being reachable and
+# avoids silently running inert if that one-time fetch fails.
 ICAPTCHA_PUBKEY=
 # Minimum proof difficulty level required to pass.
 ICAPTCHA_REQUIRED_LEVEL=3
--- a/.env.example
+++ b/.env.example
@@ -102,10 +102,10 @@ GITLAWB_AUTO_SYNC=false
 #   shadow  - verify + log would-be rejections, but always allow
 #   enforce - reject requests without a valid, strong-enough proof
 ICAPTCHA_MODE=off
-# Base URL of the iCaptcha service.
-ICAPTCHA_URL=https://icaptcha.gitlawb.com
 # Optional base64url Ed25519 public key override; if unset it is fetched from
 # ICAPTCHA_URL/v1/pubkey at startup.
 ICAPTCHA_PUBKEY=
 # Minimum proof difficulty level required to pass.
 ICAPTCHA_REQUIRED_LEVEL=3
+# Base URL of the iCaptcha service.
+ICAPTCHA_URL=https://icaptcha.gitlawb.com
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3365,6 +3365,7 @@ dependencies = [
  "aws-config",
  "aws-sdk-s3",
  "axum",
+ "base64",
  "bytes",
  "chrono",
  "cid",
--- a/crates/gitlawb-node/Cargo.toml
+++ b/crates/gitlawb-node/Cargo.toml
@@ -13,6 +13,7 @@ path = "src/main.rs"
 [dependencies]
 gitlawb-core = { path = "../gitlawb-core" }
 ed25519-dalek = { workspace = true }
+base64 = { workspace = true }
 tokio = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
--- a/crates/gitlawb-node/src/api/register.rs
+++ b/crates/gitlawb-node/src/api/register.rs
@@ -39,8 +39,10 @@ pub async fn register(
     headers: axum::http::HeaderMap,
     Json(req): Json<RegisterRequest>,
 ) -> Result<(StatusCode, Json<RegisterResponse>)> {
-    // iCaptcha proof-of-intelligence gate (inert unless ICAPTCHA_MODE is set).
-    crate::icaptcha::check(&state.db, &headers, &auth.0).await?;
+    // iCaptcha gate (inert unless ICAPTCHA_MODE is set). Verify up front; spend
+    // the proof only once the request is admissible, just before the write, so a
+    // rejected request never burns a valid proof.
+    let proof = crate::icaptcha::verify_request(&headers, &auth.0)?;
 
     // Parse and validate the DID
     let agent_did: Did = req
@@ -55,6 +57,9 @@ pub async fn register(
             "did must be the authenticated signer".into(),
         ));
     }
+
+    // Request is admissible — spend the proof now, immediately before the write.
+    proof.consume(&state.db).await?;
 
     // Store the agent in the local index
     state
--- a/crates/gitlawb-node/src/api/register.rs
+++ b/crates/gitlawb-node/src/api/register.rs
@@ -40,7 +40,7 @@ pub async fn register(
     Json(req): Json<RegisterRequest>,
 ) -> Result<(StatusCode, Json<RegisterResponse>)> {
     // iCaptcha proof-of-intelligence gate (inert unless ICAPTCHA_MODE is set).
-    crate::icaptcha::check(&headers, &auth.0)?;
+    crate::icaptcha::check(&state.db, &headers, &auth.0).await?;
 
     // Parse and validate the DID
     let agent_did: Did = req
--- a/crates/gitlawb-node/src/api/register.rs
+++ b/crates/gitlawb-node/src/api/register.rs
@@ -36,8 +36,12 @@ pub struct RegisterResponse {
 pub async fn register(
     State(state): State<AppState>,
     axum::Extension(auth): axum::Extension<crate::auth::AuthenticatedDid>,
+    headers: axum::http::HeaderMap,
     Json(req): Json<RegisterRequest>,
 ) -> Result<(StatusCode, Json<RegisterResponse>)> {
+    // iCaptcha proof-of-intelligence gate (inert unless ICAPTCHA_MODE is set).
+    crate::icaptcha::check(&headers, &auth.0)?;
+
     // Parse and validate the DID
     let agent_did: Did = req
         .did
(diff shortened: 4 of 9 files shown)


jatmn (changes requested):
## Findings

- [P2] Complete CodeRabbit's request to surface invalid iCaptcha modes
  `crates/gitlawb-node/src/icaptcha.rs:51`
  CodeRabbit's inline request is still valid: `parse_mode` maps every unrecognized `ICAPTCHA_MODE` value to `Mode::Off`, so an operator typo such as `ICAPTCHA_MODE=enforced` silently disables the protection they were trying to enable. Please complete that review request by accepting only the documented values and logging or otherwise surfacing invalid input before falling back.

- [P2] Complete CodeRabbit's request to bound startup key fetches
  `crates/gitlawb-node/src/icaptcha.rs:98`
  CodeRabbit's startup-availability request is still valid: `main` awaits `icaptcha::init()`, and `fetch_key` uses `reqwest::get()` without a timeout. With `ICAPTCHA_MODE=shadow` or `enforce`, a hung `/v1/pubkey` endpoint can leave the node stuck during startup instead of reaching the intended fail-safe inert path. Please complete that review request by adding a bounded request timeout while preserving the current fallback behavior.

- [P2] Surface invalid required-level configuration instead of lowering to the default
  `crates/gitlawb-node/src/icaptcha.rs:110`
  `ICAPTCHA_REQUIRED_LEVEL` is parsed with `.and_then(|v| v.parse().ok()).unwrap_or(3)`, so an operator trying to raise the gate with a typo such as `ICAPTCHA_REQUIRED_LEVEL=5 ` or `ICAPTCHA_REQUIRED_LEVEL=five` silently runs at level `3`. In `enforce` mode that weakens the configured challenge threshold with no warning. Please validate and warn or reject invalid level values instead of defaulting silently.

- [P3] Complete CodeRabbit's dotenv-linter request for the new env block
  `.env.example:101`
  CodeRabbit's later lint comment is still current: the new block leaves `ICAPTCHA_URL` before `ICAPTCHA_PUBKEY` and `ICAPTCHA_REQUIRED_LEVEL`, and its dotenv-linter output flags both as `UnorderedKey`. Please complete that review request by reordering the new `ICAPTCHA_*` entries so the example stops carrying the lint warning.

jatmn (changes requested):

## Findings

- [P2] Consume proof IDs so one solved challenge cannot be replayed
  `crates/gitlawb-node/src/icaptcha.rs:230`
  The verifier decodes the signed claims and checks `exp`, `level`, and `sub`, but it does not read or store the proof's `jti`/nonce, even though the captured proof fixture includes one. That means the same `X-ICaptcha-Proof` can be reused repeatedly by the same authenticated DID until it expires, including across the spam-prone `create_repo` and `register` routes this PR is trying to gate. With the existing per-DID throttle, one solved challenge can still cover every allowed creation request in the window, and new-DID registration remains a replayable bearer proof for that DID. Please consume a unique proof ID, at least until its `exp`, and reject repeats so `enforce` actually requires a fresh earned proof for each gated creation action.

kevincodex1:
@coderabbitai can you do a full review on this ?

kevincodex1:
hey @coderabbitai please do a full review again and approve if its all good

kevincodex1:
@coderabbitai please make a full review again . approve if all good

beardthelion (changes requested):
Reviewed the merged state at `66e7a34` (CI green, merge clean). The core gate is sound and I confirmed the load-bearing parts by execution: Ed25519 `verify_strict`, the proof-to-DID binding via the shared `did_matches` normalizer (handles short/full `did:key`, not a bare `==`), atomic single-use through `INSERT ... ON CONFLICT DO NOTHING`, the verify-then-consume split, and that the gate is additive to the existing owner/auth checks (mutation owner-gating intact). Migration 8 is the correct next version, additive and idempotent. The CodeRabbit threads and the two earlier change requests are all addressed on this head, including the shadow read-only replay path.

Two things to close before merge, then a few notes.

## Findings

- [P2] Gate `fork_repo`, or state why it is exempt
  `crates/gitlawb-node/src/api/repos.rs:1420`
  `fork_repo` creates a repo (`db.create_repo` after a `git clone --mirror`) and never calls `icaptcha::verify_request`; its handler signature does not even take `HeaderMap`. Under `enforce` an agent bypasses the `create_repo`/`register` gate entirely by forking. It is the third repo-creation entrypoint and the one that previously drifted on visibility (#98). `create_repo` and `register` are gated and GraphQL exposes no creation mutation, so fork is the only open path. Either thread the gate through it (verify before the clone, consume just before `db.create_repo`) or document the exemption with the reasoning.

- [P2] Add DB-level tests for the single-use ledger
  `crates/gitlawb-node/src/db/mod.rs:1119`
  `consume_proof_jti`, `check_proof_jti`, `sweep_expired_proofs`, and `ProofGuard::consume` have no execution coverage, although a `#[sqlx::test]` harness already exists in this same file. Single use is the feature's security value, and right now nothing runs it. The path itself is correct (I checked: first consume succeeds, replay returns false, `check_proof_jti` does not mutate, sweep keeps still-valid rows and drops expired), so this is just landing those as committed tests.

## Notes

- [P3] Document `ICAPTCHA_PUBKEY` as the supported `enforce` configuration
  `crates/gitlawb-node/src/icaptcha.rs:246`
  `decide()` returns `Allow` when no key loaded, so an unreachable `/v1/pubkey` at startup degrades `enforce` to inert with only a warn. That is a reasonable availability choice, but it means `enforce` fails open on a startup network blip. Pinning `ICAPTCHA_PUBKEY` removes the startup fetch dependency; worth calling it out as the way to run `enforce`.

- [P3] Consider 403 rather than 401 for `IcaptchaProofRequired`
  `crates/gitlawb-node/src/error.rs:60`
  The caller is already HTTP-signature authenticated; what is missing is an anti-abuse proof, which is closer to 403 Forbidden than 401. The distinct `icaptcha_proof_required` code is right; only the status is debatable. Confirm no client keys on the numeric 401 before changing it.

- [P3] Confirm the challenge has a headless solve path before flipping `enforce`
  The verification side is agent-clean (header-carried, offline, DID-bound). But if `icaptcha.gitlawb.com` requires interactive human solving, enabling `enforce` locks autonomous agents out of repo creation and registration, which are the two core write paths. Worth stating the programmatic acquisition flow in the PR or a linked doc.

- [P3] Pure-shadow replay detection is effectively inert
  `crates/gitlawb-node/src/icaptcha.rs`
  Shadow now routes to the read-only `check_proof_jti`, but shadow never writes the ledger, so in a shadow-only deployment nothing is ever recorded and the replay warning cannot fire. It only does anything in a mixed shadow+enforce setup. Minor, just flagging the limited value of the new path on its own.

Core is solid; the only real gap is the ungated fork path. Once `fork_repo` is covered (or consciously exempted) and the ledger has tests, this is good to go.

jatmn (changes requested):
[P2] Gate fork_repo, or explicitly document why it is exempt
crates/gitlawb-node/src/api/repos.rs:1420
fork_repo still creates a repository with state.db.create_repo(&record).await?, but unlike create_repo and register its handler does not take HeaderMap and never calls crate::icaptcha::verify_request. In ICAPTCHA_MODE=enforce, an authenticated agent can therefore bypass the new anti-abuse gate by forking any readable repo instead of calling POST /api/v1/repos, even though this route is in the same rate-limited creation route group and produces a new repo row. Please either thread the iCaptcha proof through this path as well, consuming it immediately before db.create_repo, or document the intentional exemption and why fork creation is not part of this abuse boundary.

[P2] Add execution coverage for the replay ledger
crates/gitlawb-node/src/db/mod.rs:1119
The security value of the gate now depends on the new DB-backed single-use ledger, but the committed tests only exercise the pure verifier/decision logic. consume_proof_jti, check_proof_jti, sweep_expired_proofs, and the ProofGuard::consume path are not executed by any committed test even though this file already has a #[sqlx::test] harness. Please add DB-level coverage that proves first consume succeeds, replay returns false, the shadow read-only check does not mutate state, and sweeping removes only expired proof rows.

kevincodex1:
@coderabbitai please do a full review again. approve only if all is good

jatmn (changes requested):
[P2] Complete the request to gate fork_repo or document its exemption
crates/gitlawb-node/src/api/repos.rs:1420
Jatmn's change request is still valid on the current head: fork_repo is in the same rate-limited creation route group as create_repo and register, and it still creates a new repo row with state.db.create_repo(&record).await?, but the handler does not take HeaderMap and never calls crate::icaptcha::verify_request. In ICAPTCHA_MODE=enforce, an authenticated agent can bypass the anti-abuse gate by forking any readable repo instead of calling POST /api/v1/repos. Please complete that review request by threading the iCaptcha proof through this path and consuming it immediately before db.create_repo, or explicitly document why fork creation is outside this abuse boundary.

[P2] Add committed execution coverage for the replay ledger
crates/gitlawb-node/src/db/mod.rs:1119
Jatmn's replay-ledger coverage request is also still open. The committed tests exercise the pure verifier/decision logic, but consume_proof_jti, check_proof_jti, sweep_expired_proofs, and the ProofGuard::consume path are not executed by any committed test even though the file already has a #[sqlx::test] harness. Because the single-use DB ledger is the security boundary that makes proofs non-replayable, please add DB-level tests proving first consume succeeds, replay returns false, the shadow observation path behaves as intended, and sweeping removes only expired proof rows.

[P2] Complete CodeRabbit's request to make shadow replay detection record first-seen proofs
crates/gitlawb-node/src/icaptcha.rs:218
CodeRabbit's current shadow-mode replay request is still valid: the shadow path calls db.check_proof_jti(&job.jti).await?, but that helper is read-only, so a shadow-only deployment never records the first use of a proof. A repeated proof therefore keeps appearing fresh and never produces the "would reject replay" signal that shadow mode is supposed to provide before operators flip to enforce. Please complete that request by recording first-seen proof IDs in the shadow observation path while still allowing the request through.

[P2] Complete CodeRabbit's request to keep shadow replay-check errors non-fatal
crates/gitlawb-node/src/icaptcha.rs:218
The same shadow branch still propagates db.check_proof_jti(...) with ?. Since both gated handlers call proof.consume(&state.db).await?, a transient DB error in shadow mode rejects register or create_repo, even though shadow mode is documented as an observe/log mode that always allows. Please handle replay-check failures locally in the shadow branch, log the jti and error, and return Ok(()) so only enforce mode fails closed on ledger errors.

kevincodex1:
@coderabbitai make a full review please

PR #109, opened 2026-06-27
fix(node): gate fork_repo on per-caller path-scoped visibility (#98)

### LINKED ISSUE #98: fork_repo clones a full mirror after only a root (/) visibility check, bypassing path-scoped withholding

beardthelion:
Closes #98.

`fork_repo` authorized the caller only at the repo root `/` and then `git clone --mirror`d the whole source, discarding the visibility rules it had just fetched. A non-owner with root read but a denied path-scoped subtree could fork the full mirror and obtain blobs the filtered `git_upload_pack` path withholds.

## Fix

Refuse the fork when the caller has any withheld path in the source, the same per-caller decision the read path makes:

- New pure helper `fork_withheld_blocks(rules, is_public, owner_did, caller)` returns true (refuse) iff `withheld_globs(...)` is non-empty for that caller. It delegates to the existing `visibility_check`/`withheld_globs` seam, so the owner bypass (full and short DID) and `reader_dids` grants are inherited from the read path rather than reimplemented, and the two cannot drift.
- The gate runs before `repo_store.acquire`, so no withheld object is materialized on disk, and fails closed with `RepoNotFound` (mirroring `authorize_repo_read`s Deny) so a withheld subtrees existence is not leaked.

Owners, fully-authorized readers, and forks of repos with no path-scoped withholding for the caller are unaffected.

The predicate is a conservative over-approximation of the read paths object-level withholding: never weaker (so fork cannot leak), and marginally stricter only in the duplicate/co-located-blob case. A filtered fork projection (serve readers a fork minus the withheld subtree) is the larger follow-up and is intentionally out of scope.

## Tests

- 8 unit tests on the pure predicate: owner full/short DID, a non-owner denied a subtree (the core regression), a non-owner who is a listed `reader_did` of that subtree (must still be allowed to fork), root-only rule, no rules, mixed root+subtree, and the partial-reader case.
- One `#[sqlx::test]` integration test that drives the handler: a public repo with a `/secret/**` rule excluding the caller returns 404 and creates no fork row. Because the gate precedes `acquire`, it needs no on-disk source.

## Scope note

While mapping every content-egress path during review, I found a separate, pre-existing leak in `GET /ipfs/{cid}` (`api/ipfs.rs` `get_by_cid`): it returns any git object by raw SHA-256 across all repos with no visibility check, which is the same bypass class on a different surface. It is out of scope here and filed separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Forking a repository now correctly refuses requests when any path-scoped content would be hidden from the caller.
  * Improved privacy handling so restricted subtrees do not reveal whether they exist, returning a generic not-found response instead.
  * Added regression coverage to verify fork requests are denied for partially restricted repositories and no fork is created after a failed request.

### DIFF
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -584,6 +584,30 @@ fn owner_push_rejection(
     }
 }
 
+/// Decide whether the fork gate refuses a `fork_repo` request (#98).
+///
+/// Returns `true` when the fork must be refused: the source carries at least one
+/// path-scoped subtree that `caller` may not read, so a full `git clone --mirror`
+/// would copy out content the filtered read path (`git_upload_pack`) withholds.
+/// Pure function so the policy is unit-testable without a database or git backend.
+///
+/// Delegates the per-caller decision to [`withheld_globs`](crate::visibility::withheld_globs)
+/// / `visibility_check`, so the owner bypass (full and short DID) and `reader_dids`
+/// grants are inherited from the read path and the two cannot drift on who may read
+/// what. The predicate is a conservative (fail-closed) over-approximation of the
+/// read path's object-level withholding: never weaker (so the fork cannot leak
+/// content the read path withholds), and stricter only in the narrow
+/// duplicate/co-located-blob case. Only called after `authorize_repo_read("/")`
+/// has already granted the caller root read.
+fn fork_withheld_blocks(
+    rules: &[crate::db::VisibilityRule],
+    is_public: bool,
+    owner_did: &str,
+    caller: &str,
+) -> bool {
+    !crate::visibility::withheld_globs(rules, is_public, owner_did, Some(caller)).is_empty()
+}
+
 /// Path of the peer sync-notify endpoint. Used both to build the target URL
 /// and as the signing path, so they can never drift apart.
 const SYNC_NOTIFY_PATH: &str = "/api/v1/sync/notify";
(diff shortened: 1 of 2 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

Issue #118, opened 2026-06-29
Owner-push enforcement is opt-in (off by default): decide the default write posture


beardthelion:
## Summary

This is a **default-posture decision**, not a code defect: in the default configuration, owner-push enforcement is off, so any authenticated (signed) non-owner DID can push to any repo. `git_receive_pack` rejects a non-owner push only when `config.enforce_owner_push` is set, and that flag defaults to `false` (`GITLAWB_ENFORCE_OWNER_PUSH`). The git-receive-pack POST requires *some* valid `did:key` signature (`require_signature` via `add_auth_layers`), but does not bind that signature to the repo owner unless an operator opts in.

This is deliberate, documented behavior (introduced opt-in by #68 / commit `0a15e76`; the config comment says to keep it off during rolling upgrades and flip it on when owners are ready; there is a committed test asserting the default-off allow). Filing it to get an explicit maintainer decision on the default, not to report broken code.

## Where

- `crates/gitlawb-node/src/api/repos.rs` — `git_receive_pack` -> `owner_push_rejection(config.enforce_owner_push, &repo, caller_did)`; `owner_push_rejection` returns `None` (no rejection) whenever the flag is off, regardless of caller.
- `crates/gitlawb-node/src/config.rs` — `enforce_owner_push` defaults to `false`.

## Confirmed by execution

`owner_push_rejection(false, &repo, Some(STRANGER_DID))` returns `None` (a non-owner is not rejected when the flag is off). Verified by a throwaway unit test (reverted); a pre-existing committed test already documents the same default-off behavior.

## Impact

A default-config node accepts pushes from any authenticated identity, not just the repo owner. Write integrity reduces to "holds a valid signature of any DID" rather than "is the repo owner." Branch protection still runs, but it only covers branches an owner explicitly protected and does not restore owner identity. With the receive-pack advertisement now read-parity (a granted reader can see the advertisement), this default surface is more visible.

## Options

1. Default `enforce_owner_push` to `true` and let operators opt out for rolling upgrades (rather than opt in).
2. Keep opt-in, but surface the default-off state in deployment docs and the startup log so operators consciously choose it.
3. Gate the receive-pack POST on owner (or explicit writer grant) directly, independent of the flag.

Pre-existing; independent of the receive-pack advertisement fix (which only touches reads).

beardthelion:
[...]

Issue #117, opened 2026-06-29
git-remote-gitlawb deadlocks on incremental (multi-round) fetch


beardthelion:
## Summary

The `git-remote-gitlawb` helper bridges git's stateful upload-pack `connect` stream onto a single stateless HTTP POST, so it deadlocks on an **incremental (multi-round) fetch**. `read_upload_pack_request` treats a flush pkt (`0000`) as "keep reading" and only breaks on the `done\n` pkt-line, and `handle_connect` issues exactly one POST after `done`. In an incremental fetch git sends `want`s, then a `have`-batch terminated by a flush, and **blocks waiting for the server's ACK/NAK before sending `done`**. The helper is blocked reading stdin for a `done` that never comes; git is blocked reading the helper's stdout for a response that is never sent. The fetch hangs until the request timeout.

A fresh clone (the client sends `want`s + `done` in one shot) works, which is why this was not caught earlier.

## Where

- `crates/gitlawb-node/../git-remote-gitlawb/src/main.rs` — `read_upload_pack_request` (breaks only on `done\n` or EOF; `continue`s past a flush) and the single-POST flow in `handle_connect`.

## Reproduced by execution

Live repro against a local node: built a ~50-commit shared history, cloned via `gitlawb://`, advanced the server by a few commits, then `git fetch origin`.

- Fresh clone (control): exit 0, ~1s.
- Incremental fetch, few haves (git short-circuits, sends `want` + haves + `done` in one shot): exit 0, ~0s.
- Incremental fetch, >32 haves (forces multi-round): **hung the full 40s, killed by `timeout` (exit 124)**.

`GIT_TRACE_PACKET` confirmed git sent 1 `want`, 32 `have` lines, a flush `0000`, and **zero `done`**, then blocked. Only the initial ref advertisement ever came back.

## Impact

Any fetch where the client has more than ~32 commits of overlapping history with the server (the common "pull updates into an existing clone") hangs indefinitely. Not a security issue; a functional break in the helper's fetch path.

## Options

1. Implement git's stateless-RPC multi-round negotiation (a POST per have-batch, streaming ACK/NAK back), or
2. advertise/implement the `stateless-connect` capability instead of `connect`.

Pre-existing; surfaced while verifying the receive-pack advertisement fix end to end.

PR #119, opened 2026-06-29
fix(node,git-remote): gate receive-pack advertisement, sign client fetch/push

### LINKED ISSUE #116: git-receive-pack info/refs advertisement skips the visibility gate, leaking private repo ref metadata
## Summary

`git_info_refs` (`crates/gitlawb-node/src/api/repos.rs`) serves both the `git-upload-pack` and `git-receive-pack` ref advertisement off one anonymous route, but the read-visibility gate was wrapped in `if service == "git-upload-pack"`. An anonymous

```
GET /{owner}/{repo}.git/info/refs?service=git-receive-pack
```

against a **private** repo returned its branch/tag names and commit tips, bypassing the read gate the upload-pack path enforces. This is the git-transport sibling of #112 (GraphQL `refUpdates`) and #114 (REST events feed): the same private ref-metadata, a different egress.

## Where

- `crates/gitlawb-node/src/api/repos.rs` — `git_info_refs`, the visibility check scoped to a single service.
- The route is on the `optional_signature` group, so it is reachable unauthenticated.

## Impact

For any private repo whose owner-DID slug an attacker can guess or enumerate, an unauthenticated request discloses its branch/tag names, commit SHAs, and existence, over plain git smart-HTTP. The in-code comment claimed the receive-pack path "is authorized separately on the POST" — true for the push itself, but the *advertisement* discloses the same metadata the read path withholds.

## Fix

Run `visibility_check(.., "/")` for the advertisement regardless of service (deny -> 404, existence-hiding). Because the advertisement is then gated for both services, the `git-remote-gitlawb` helper must authenticate its own requests, so it now signs the Phase-1 advertisement GET and the Phase-2 pack POST for both services over `path_and_query` when an identity keypair is present (public repos stay anonymous). This also repaired a pre-existing break where a repo's owner could not fetch their own private repo, because the upload-pack POST was sent unsigned.

A fix is implemented and verified in branch `fix/gate-receive-pack-advertisement` (PR to follow). Verified by execution: the anonymous receive-pack advertisement now returns 404 (was a 200/served leak); a live `git clone gitlawb://` of a private repo succeeds for the owner and is denied (404, no content) for anonymous and non-owner callers; node-side and seam tests cover both services.

beardthelion:
Closes #116.

The `info/refs` ref advertisement was gated only for `service=git-upload-pack` (the visibility check sat inside `if service == "git-upload-pack"`), so an anonymous `GET /{owner}/{repo}.git/info/refs?service=git-receive-pack` returned a private repo's branch names and commit tips. This runs the visibility `"/"` check for both services, so a denied caller gets a 404 with nothing leaked.

Gating both services means `git-remote-gitlawb` has to authenticate its own requests, or a private repo's owner can neither fetch nor push it. The helper now signs the Phase-1 advertisement GET and the Phase-2 pack POST for both services over `path_and_query` when an identity keypair is present; public repos still work anonymously. That also repairs a pre-existing break where an owner could not fetch their own private repo, because the upload-pack POST was sent unsigned.

Verification:
- Node-side gate: anonymous advertisement denied (404) for both services, owner-signed clears the gate, public stays anonymous.
- Real-signature seam: the signature `git-remote-gitlawb` actually emits verifies under the node's own verification primitives, over the path it transmits (a tampered path fails).
- Real served content: with an on-disk repo, the owner gets the ref bytes and a denied caller's 404 omits them.
- Live `git clone gitlawb://` against a running node: the owner succeeds with content; anonymous and a non-owner are denied (404, no content).

Two pre-existing issues surfaced while verifying this end to end, both independent of this change: #117 (incremental-fetch deadlock in the helper) and #118 (owner-push enforcement default posture).

## Summary by CodeRabbit

* **New Features**
  * Extended RFC-9421 signing to cover both phases (ref advertisements and pack exchanges) for both `git-upload-pack` and `git-receive-pack` when an identity keypair is available.
  * Updated protocol/help text to describe “signed fetch/push” behavior.

* **Bug Fixes**
  * Improved `info/refs` read-visibility enforcement so private repositories are hidden for both supported services.
  * Requests missing required signatures (including unsigned `git-receive-pack` POST) are rejected.

* **Tests**
  * Expanded integration/e2e coverage for public vs. private served refs, signed vs. unsigned behavior, and correct signature/path handling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3304,6 +3304,7 @@ version = "0.4.0"
 dependencies = [
  "anyhow",
  "gitlawb-core",
+ "mockito",
  "reqwest",
  "tracing",
  "tracing-subscriber",
--- a/crates/git-remote-gitlawb/Cargo.toml
+++ b/crates/git-remote-gitlawb/Cargo.toml
@@ -16,3 +16,6 @@ anyhow = { workspace = true }
 reqwest = { workspace = true }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
+
+[dev-dependencies]
+mockito = "1"
--- a/crates/git-remote-gitlawb/src/main.rs
+++ b/crates/git-remote-gitlawb/src/main.rs
@@ -14,7 +14,10 @@
 //! # Protocol flow (connect capability)
 //!   capabilities → "connect\n\n"
 //!   connect git-upload-pack  → GET /info/refs | POST /git-upload-pack
-//!   connect git-receive-pack → GET /info/refs | POST /git-receive-pack (+ auth header)
+//!   connect git-receive-pack → GET /info/refs | POST /git-receive-pack
+//! When an identity keypair is present, both the advertisement GET and the pack
+//! POST are RFC-9421 signed for BOTH services, so the node authorizes the owner on
+//! visibility-gated private repos (fetch and push); public repos work anonymously.
 
 use anyhow::{bail, Context, Result};
 use gitlawb_core::http_sig::sign_request;
@@ -63,7 +66,9 @@ fn main() -> Result<()> {
     let repo_base = format!("{}/{}/{}", node_base, short_owner, repo_name);
     tracing::debug!("repo_base: {repo_base}");
 
-    // Load keypair for signing push requests (optional — push still works unsigned in v0.1)
+    // Load keypair for signing requests (optional). When present, the helper signs
+    // both fetch and push so a private repo's owner is authorized; absent, public
+    // repos still work and push falls back to unsigned (v0.1 local alpha only).
     let keypair = load_keypair();
 
     run_helper(&repo_base, keypair.as_ref())
@@ -115,7 +120,7 @@ fn help_text() -> String {
          \n\
          ENVIRONMENT:\n\
          \x20   GITLAWB_NODE   Node base URL (default: http://127.0.0.1:7545)\n\
-         \x20   GITLAWB_KEY    Identity PEM path for signed pushes (default: ~/.gitlawb/identity.pem)\n\
+         \x20   GITLAWB_KEY    Identity PEM path for signed fetch/push (default: ~/.gitlawb/identity.pem)\n\
          \x20   GITLAWB_LOG    Log filter (default: warn)\n\
          \n\
          FLAGS:\n\
@@ -201,9 +206,7 @@ fn handle_connect<R: Read>(
     let refs_url = format!("{}/info/refs?service={}", repo_base, service);
     tracing::debug!("GET {refs_url}");
 
-    let refs_resp = client
-        .get(&refs_url)
-        .header("User-Agent", "git/2.0 git-remote-gitlawb/0.1.0")
+    let refs_resp = build_advertisement_request(&client, &refs_url, keypair)
         .send()
         .with_context(|| format!("GET {refs_url}"))?;
 
@@ -274,29 +277,7 @@ fn handle_connect<R: Read>(
     let post_url = format!("{}/{}", repo_base, service);
     tracing::debug!("POST {post_url} ({} bytes)", request_body.len());
 
-    // Extract the URL path for signing (e.g., "/z6Mk.../my-repo/git-receive-pack")
-    let path_for_sig = url_path(&post_url);
-
-    let mut req = client
-        .post(&post_url)
-        .header("Content-Type", format!("application/x-{}-request", service))
-        .header("User-Agent", "git/2.0 git-remote-gitlawb/0.1.0");
-
-    // Add RFC 9421 HTTP Signature auth on push operations
-    if service == "git-receive-pack" {
-        if let Some(kp) = keypair {
-            let signed = sign_request(kp, "POST", &path_for_sig, &request_body);
-            req = req
-                .header("Content-Digest", signed.content_digest)
-                .header("Signature-Input", signed.signature_input)
-                .header("Signature", signed.signature);
-            tracing::debug!("attached RFC 9421 HTTP Signature (DID: {})", kp.did());
-        } else {
-            tracing::warn!(
-                "no identity keypair found — push will be unsigned (v0.1 local alpha only)"
-            );
-        }
-    }
+    let req = build_pack_post_request(&client, &post_url, service, &request_body, keypair);
 
     // Attach the body after signing so the pack bytes are moved, not cloned —
     // packs can be large and the clone doubled peak memory on push.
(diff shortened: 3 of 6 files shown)


jatmn (changes requested):
I found one issue that should be cleaned up before this is ready.

## Findings

- [P3] Complete CodeRabbit's request to correct the auth-contract comments
  `crates/git-remote-gitlawb/src/main.rs:323`
  CodeRabbit's two comment requests are still valid in the current patch. The helper comment says the node read-gates both `git-upload-pack` and `git-receive-pack` POSTs with `visibility_check("/")`, but the actual receive-pack route is signature-required and separately owner-gated; similarly, the node comment says read access is a strict subset of push access, which reverses the intended invariant that push implies read. These comments sit directly on the security-sensitive gate/signing code this PR is changing, so please complete the CodeRabbit request and make both comments match the implemented auth contract.

beardthelion:
Addressed both auth-contract comments in 8ea7281.

- `git-remote-gitlawb/src/main.rs` (build_pack_post_request): now states the node read-gates the git-upload-pack POST with `visibility_check` at "/", and *separately owner-gates* the git-receive-pack POST (signature required, enforced by middleware). Verified against the code: `git_receive_pack` (repos.rs:716) carries no `visibility_check`; it requires `AuthenticatedDid` from `require_signature` and runs `owner_push_rejection` + branch protection; the route sits under `add_auth_layers` (server.rs:178).
- `gitlawb-node/src/api/repos.rs` (git_info_refs): "read access is a strict subset of push access" → "push access implies read access here".

While in there I added a test (47fd5d0) pinning the signature-gating half: an unsigned git-receive-pack POST returns 401 from `require_signature` before reaching the handler (401, not the handler's 404/500, is the discriminator).

jatmn (changes requested):
This PR fixes the private-repo `info/refs` advertisement leak by running `visibility_check` for both `git-upload-pack` and `git-receive-pack`, and updates `git-remote-gitlawb` to RFC-9421 sign both the Phase-1 advertisement GET and the Phase-2 pack POST for both services when an identity keypair is present.

I found one small test-isolation cleanup item; the core auth fix and the cross-crate seam tests look solid.

The review covered the gate change, the new request builders, the prior jatmn/CodeRabbit auth-contract comments, test isolation, and duplicate/superseded PR checks.

## Findings

- [P3] Clean up the served-content test's temp directories on failure
  `crates/gitlawb-node/src/test_support.rs:1083`
  `advertisement_serves_real_refs_only_to_authorized_callers` creates real git repos under `/tmp/{slug}/{name}.git` and removes them only after the last assertion, so a failed assertion leaves those directories behind. The names are randomized, so this does not cause cross-run collisions, but it does leak state on the test runner. Please use a RAII guard (a wrapper that calls `remove_dir_all` in `Drop`, or `tempfile::TempDir` where the path layout allows) so cleanup runs even when the test panics.

beardthelion:
Addressed the temp-dir cleanup in 8b8a4ea.

`advertisement_serves_real_refs_only_to_authorized_callers` now wraps each created path (`src`, `served-pub.git`, `served-priv.git`) in a small `DirGuard` whose `Drop` calls `remove_dir_all`, bound right after each dir is created, and the three trailing manual removes are gone. `tempfile::TempDir` doesn't fit here because `repo_store::for_testing` fixes the on-disk layout, so the guard wraps the known path instead.

Verified the panic path by execution: injected a panic before the end and confirmed all three dirs were removed during unwind; on the success path the dirs are removed too, leaving only the empty `/tmp/<slug>` parent (unchanged from before). Suite green.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found one repository-state blocker that needs to be handled before this is ready.

## Findings

- [P1] Rebase on main and resolve the test-support conflict first
  `crates/gitlawb-node/src/test_support.rs`
  GitHub currently reports this branch as `CONFLICTING` / `DIRTY`, and a merge-tree check against the current `main` reports a content conflict in `crates/gitlawb-node/src/test_support.rs`. Please rebase on the latest `main` and resolve that conflict first, so the auth/test cleanup changes are applied to the code that will actually be merged.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found one remaining item to address.

## Findings

- [P3] Complete CodeRabbit's request to recurse through nested API modules
  `crates/gitlawb-node/src/api/mod.rs:379`
  CodeRabbit's latest request is still valid in the current patch. The completeness scan only uses `std::fs::read_dir(api_dir)`, so it checks direct `src/api/*.rs` files and skips subdirectories entirely. That means a future `src/api/<module>/mod.rs` or nested `.rs` file could add a repo-scoped handler without being included in `sources` and without tripping this guard, even though the guard comment says a whole new module is covered. Please complete the CodeRabbit request by walking nested API source files too, so the assertion applies to every discovered API source rather than only immediate children.

beardthelion:
@jatmn fixed in ef61b57. The completeness scan now walks `src/api` recursively (new `collect_rs_files` helper) instead of the flat `read_dir`, so a nested `api/<module>/mod.rs` that declares a repo-scoped handler trips the guard too, not just the immediate children. Verified a nested repo-scoped handler now fails the guard where the flat scan passed it. Ready for another look.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found a repository-state blocker plus a couple of remaining review items.

## Findings

- [P1] Rebase on main and resolve the merge conflict first
  `crates/git-remote-gitlawb/src/main.rs`
  GitHub currently reports this branch as `CONFLICTING` / `DIRTY`, and a merge-tree check against the current `main` reports a content conflict in `crates/git-remote-gitlawb/src/main.rs`. Please rebase on the latest `main` and resolve the conflict before addressing the review items below, so the signing/helper changes are applied to the code that will actually be merged.

- [P2] Keep public fetches anonymous when an identity exists
  `crates/git-remote-gitlawb/src/main.rs:209`
  `git-remote-gitlawb` now calls `build_advertisement_request(..., keypair)` for every `git-upload-pack` connection, and that builder attaches `Signature-Input` / `Signature` whenever `~/.gitlawb/identity.pem` exists. That means a normal user who has created an identity now discloses their DID to any node they fetch a public repo from, even though public repos do not need authentication and the PR summary says public repos still work anonymously. Please keep public clone/fetch anonymous unless authentication is actually needed, or make the identity disclosure explicit/opt-in so this does not become a silent privacy regression for public reads.

- [P3] Complete CodeRabbit's request with a path-aware source skip
  `crates/gitlawb-node/src/api/mod.rs:430`
  CodeRabbit's latest request is still valid in the current patch. The recursive completeness scan walks nested files, but it still skips files by basename with `listed.contains(fname.as_str())`; if a future nested module adds `src/api/<module>/repos.rs`, it will be skipped because the top-level `repos.rs` is already listed, even though the nested file is a distinct unscanned source file. Please complete the request by comparing paths relative to `api_root` (and making `sources` / `listed` path-aware as needed), so only the exact already-covered source files are skipped.

beardthelion:
Rebased on latest main and pushed. The only conflict was in the `git-remote-gitlawb` test module (your #137 error-body tests vs this PR's tests); both sets are additive, so I kept all of them. All three items:

**P1 (rebase).** Done, on current main; conflict resolved.

**P2 (public fetches stay anonymous).** `git-remote-gitlawb` no longer signs `git-upload-pack` unconditionally. Fetch sends the advertisement GET anonymously and retries signed only if the node answers 404 (how it withholds a private repo from an unauthenticated reader); when the retry is needed the upload-pack POST is signed too. Push (`git-receive-pack`) is unchanged and signs from the first request, since a push can never be anonymous. A public clone no longer discloses the caller's DID.

One fact the design keys on: an unsigned private fetch returns 404, not 401, so the retry triggers on 404. Verified end to end against a running node: a public fetch with an identity present stays anonymous, the owner's private fetch retries signed and clones, a valid signature from a non-reader still gets 404, and a keyless fetch surfaces the 404 without retrying.

**P3 (nested source skip).** The egress-gate completeness scan now skips already-covered files by their path relative to `api_root` rather than by basename, so a nested `api/<module>/repos.rs` is still scanned instead of being shadowed by the top-level `repos.rs`.

Ready for another look.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

Issue #121, opened 2026-06-29
Unauthenticated metadata indexes leak private-repo data: /ipfs/pins and /arweave/anchors


beardthelion:
## Summary

Two node-wide metadata indexes are served with no authentication and no visibility filtering, exposing metadata about private-repo objects and ref history.

## Where (verified by reading the handlers; both on routers with no auth layer)

- `crates/gitlawb-node/src/api/ipfs.rs` — `list_pins` (:103), route `GET /api/v1/ipfs/pins` (`server.rs:192`). Returns the `(sha256, cid, pinned_at)` index for every pinned object on the node, with no repo scope and no caller. This is the targeting oracle for #110: it hands an anonymous caller the exact CIDs to fetch private-repo blob content through the ungated `GET /ipfs/{cid}`.
- `crates/gitlawb-node/src/api/arweave.rs` — `list_anchors` (:24), route `GET /api/v1/arweave/anchors` (`server.rs:195`). Returns ref-transition anchors (owner DID, ref names, old/new SHAs, CIDs, tx ids) filtered only by `?repo=`, with no visibility check. Today it is bounded by the write-side announce invariant (anchoring only fires for announce-able repos), but it trusts that invariant rather than enforcing it, and anchors are permanent: a repo anchored while public and later made private leaks its ref graph here forever.

## Impact

- `/ipfs/pins` + #110 together let an unauthenticated caller enumerate and then fetch private-repo blob content.
- `/arweave/anchors` discloses ref-transition metadata (ref names, commit SHAs, pusher/owner DIDs) for anchored repos, and survives a later visibility change.

## Fix

- `/ipfs/pins`: restrict to the node operator, or filter the index to objects the caller may read; at minimum stop exposing a node-wide CID index anonymously. Best handled together with #110.
- `/arweave/anchors`: gate per `?repo=` on the caller's read visibility (deny -> 404), rather than relying on the write-side invariant. Derive any count from the post-filter set.

Found during a read-only egress audit; related to #110 (content-addressed blob egress).

Issue #120, opened 2026-06-29
Repo-scoped read surfaces gated by existence only, not visibility (certs, issues, labels, bounties, stars)


beardthelion:
## Summary

Several repo-scoped read handlers verify the repo *exists* (`get_repo`) but never run the visibility gate (`authorize_repo_read` / `visibility_check`), so they return a private repo's metadata to an anonymous caller. This is the same class as #94 (hooks) and its PR #113 siblings (events, replicas, protected branches); these are the surfaces that batch did not cover. The mutation handlers on the same files do gate correctly; only the reads were missed.

## Where (all verified by reading the handler bodies; none take an `AuthenticatedDid` caller)

- `crates/gitlawb-node/src/api/certs.rs` — `list_certs` (:10), `get_cert` (:42). Returns `ref_name`, `old_sha`, `new_sha`, `pusher_did`, `node_did` for the repo. **This is the same private ref metadata as #114 / #116** (branch names, commit SHAs, pusher identity).
- `crates/gitlawb-node/src/api/issues.rs` — `list_issues` (:90), `get_issue` (:120), `list_issue_comments` (:191). Returns the repo's issue titles/bodies/comments from its git store.
- `crates/gitlawb-node/src/api/labels.rs` — `list_labels` (:74). Returns the repo's label set.
- `crates/gitlawb-node/src/api/bounties.rs` — `list_repo_bounties` (:114). Returns the repo's bounties (no repo lookup at all).
- `crates/gitlawb-node/src/api/stars.rs` — `get_star_status` (:78). Existence + star count oracle for a private repo.

All sit on the anonymous read group (`optional_signature`).

## Impact

For any private repo whose owner-DID slug an attacker can guess or enumerate, an unauthenticated request discloses its ref history and pusher identity (certs), issue content, labels, bounties, and existence. `certs` is the most severe (ref names + commit SHAs + pusher DID, identical to the #114/#116 leak class); the others are lower but still leak private-repo metadata that the read path otherwise withholds.

## Fix

Add `authorize_repo_read(&state, &owner, &repo, caller, "/").await?` to each read handler (deny -> 404 RepoNotFound, existence-hiding), threading the `Option<Extension<AuthenticatedDid>>` caller, exactly as #113 did for events/replicas/protected. The matching `gl`/MCP read callers must then sign (the #115 client-signing class).

Found during a read-only egress audit of every repo-data surface.

beardthelion:
[...]

Issue #114, opened 2026-06-28
Unauthenticated GET /api/v1/events/ref-updates leaks private-repo ref metadata (REST analog of #112)


beardthelion:
## Summary

`GET /api/v1/events/ref-updates` (`list_ref_updates`, `crates/gitlawb-node/src/api/events.rs:12`, mounted at `crates/gitlawb-node/src/server.rs:325`) takes no auth, no repo argument, and no visibility check. It returns `ref_name`, `old_sha`, `new_sha`, `pusher_did`, `node_did`, and the repo slug for every row in `received_ref_updates` across all repos, including private ones, to any anonymous caller.

This is the REST analog of #112 (GraphQL `refUpdates`). PR #113 closed the per-repo `list_repo_events` leak for locally-hosted private repos, but it deliberately left this global feed alone to keep that PR scoped to the per-repo surfaces. The per-repo fix does not close this path: an attacker who knows (or guesses) a repo's full owner-DID slug can still read its push history from the global feed.

### Where

- `crates/gitlawb-node/src/api/events.rs:12` — `list_ref_updates` handler, no auth/visibility gate
- `crates/gitlawb-node/src/server.rs:325` — route on the `optional_signature` read group
- the query is `list_ref_updates(limit)` with no per-caller filter

### Options

1. Filter the feed to rows for repos the caller may read (a visibility join on the feed query), deriving any count from the post-filter set.
2. Restrict the endpoint to the node operator (signature + node-owner check).
3. Decide explicitly that the global gossip feed is intentionally public (federation requires it) and document the threat-model decision plus what data minimization applies.

The same load-bearing assumption sits under the `list_repo_events` gossip-only (no-local-row) branch: it is only safe to serve publicly if origin nodes' announce-gate keeps private-at-origin repos out of gossip. A misbehaving peer would leak private ref metadata through both this feed and that branch.

### Related

- #112 (GraphQL `refUpdates`, same leak class)
- #94 / PR #113 (the per-repo `list_repo_events` gate that this complements)

Issue #112, opened 2026-06-28
GraphQL refUpdates serves private-repo ref metadata to anonymous callers (visibility bypass)


beardthelion:
The GraphQL `refUpdates` query is not visibility-gated, so an anonymous caller can read ref-level metadata for a private repo by passing its name. This is the same withholding-bypass class as #97 (repo listing) and #110 (blob content), on the GraphQL ref-history surface. The #97 visibility gate was applied to the sibling `repos` resolver but not to `refUpdates`.

### Where

`crates/gitlawb-node/src/graphql/query.rs`, the `ref_updates` resolver (`async fn ref_updates`, around line 48).

- The resolver calls `db.list_ref_updates_filtered(repo.as_deref(), limit)` and maps every row straight to `RefUpdateType`. It never reads `ctx.data::<crate::auth::AuthenticatedDid>()` and never consults visibility rules. The `repos` resolver directly above it threads the caller DID through `crate::visibility::listable_at_root`; `ref_updates` does not.
- `Db::list_ref_updates_filtered` (`crates/gitlawb-node/src/db/mod.rs`, ~line 1975) runs `SELECT ... FROM received_ref_updates WHERE repo=$1 ORDER BY timestamp DESC LIMIT $2` with no `is_public`, no visibility join, and no reader-DID predicate. The unfiltered branch (no `repo` arg) is equally open.
- `/graphql` is wrapped only by `auth::optional_signature`, not `require_signature` (`crates/gitlawb-node/src/server.rs:66`), so the query is reachable anonymously.

### Impact

For any private repo whose name an attacker can guess or enumerate, an unauthenticated GraphQL query returns its `refName`, `oldSha`, `newSha`, `pusherDid`, `nodeDid`, and `timestamp`. That discloses the repo's existence, its branch/tag names, commit SHAs, and the identity of who pushed, for a repo the read path otherwise withholds.

Verified by execution: a test that seeds a private repo (`is_public=false`), records a received ref update for it, then runs an anonymous `{ refUpdates(repo: "secret-repo") { refName newSha pusherDid } }` returns the row:

```json
{"refUpdates":[{"newSha":"deadbeef...","pusherDid":"did:key:zPUSHER...","refName":"refs/heads/main"}]}
```

### Suggested remediation

[...]

PR #122, opened 2026-06-29
test(node): guard that every repo-scoped handler is visibility-gated

beardthelion:
Every PR that gates one read surface keeps revealing two more (A1/#116, #110, #112, #114, #120, #121). The root cause is structural: the visibility invariant is re-applied by hand on each egress path, and nothing fails when a new path forgets it. The codebase already proves the fix shape works for mutations (`authz_guard::every_in_scope_mutation_has_its_gate`); this adds the read/egress equivalent.

`every_repo_scoped_handler_is_gated` scrapes every repo-scoped handler (any `pub`/`pub(crate) async fn` taking `Path<(String, String..)>`) and asserts it carries an authz marker, either a read gate (`authorize_repo_read` / `visibility_check`) or a write/owner gate (`require_repo_owner` / `require_owner` / `did_matches` / a `&auth.0` self-binding), or is listed in `KNOWN_UNGATED` with its tracking issue. It is verb-agnostic on purpose, so a repo-scoped read named outside `list_*`/`get_*` (the git transport handlers, `replicate_encrypted_blobs`, `withheld_paths`) is still checked. A completeness scan over `src/api/` trips for a whole new module that adds a repo-scoped handler without being wired in. So a new ungated egress of any name fails CI instead of shipping as the next leak.

`KNOWN_UNGATED` is the live gap list, currently the `certs`/`issues`/`labels`/`bounties`/`stars` reads (#120) plus the `events`/`webhooks`/`replicas`/`protected` reads being closed in #113. A staleness assert forces each entry out the moment its gate lands, so the allowlist cannot outlive the bug it tracks.

Scope: this is a source scrape (paired with a runtime route guard for behaviour). It proves a gate is called, not that it runs on the requested subtree, and it sees the owner+repo tuple `Path` shape. Those limits are documented in the test and pinned by the helper unit tests.

Verified by execution: it covers ~54 handlers today, and removing an allowlist entry, listing a gated handler, dropping a source file, breaking the parser, and confirming a non-list/get handler is scraped each trip the guard.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened authorization enforcement verification to ensure repo-scoped API handlers are always protected by the proper access control gate.
  * Added safeguards to flag newly added or reorganized API modules that introduce repo-scoped handlers without required protection.
* **Tests**
  * Expanded automated coverage for detecting routed async handlers, identifying repo-scoped signatures, and confirming gate application behavior—reducing the chance of missed authorization gaps.

### DIFF
--- a/crates/gitlawb-node/src/api/mod.rs
+++ b/crates/gitlawb-node/src/api/mod.rs
@@ -266,6 +266,48 @@ mod authz_guard {
         header.contains("Path<(String, String")
     }
 
+    /// True when at least one gate marker runs for EVERY service — i.e. some
+    /// marker sits outside any `if service == ...` discriminator block. A gate
+    /// that appears ONLY inside such a block (as the info/refs advertisement gate
+    /// did before #119: `visibility_check` ran under `if service ==
+    /// "git-upload-pack"`, leaving `git-receive-pack` ungated) covers a subset of
+    /// services and must NOT count as a full gate. Other handlers carry no
+    /// `service ==` discriminator, so for them this matches the plain
+    /// substring check.
+    fn gate_runs_unconditionally(body: &str, markers: &[&str]) -> bool {
+        // Brace-matched spans of each `if service == ...` block.
+        let mut cond_spans: Vec<(usize, usize)> = Vec::new();
+        let mut search = 0;
+        while let Some(rel) = body[search..].find("if service ==") {
+            let cond_start = search + rel;
+            let Some(brace_rel) = body[cond_start..].find('{') else {
+                break;
+            };
+            let open = cond_start + brace_rel;
+            let mut depth = 0i32;
+            let mut end = body.len();
+            for (i, c) in body[open..].char_indices() {
+                match c {
+                    '{' => depth += 1,
+                    '}' => {
+                        depth -= 1;
+                        if depth == 0 {
+                            end = open + i;
+                            break;
+                        }
+                    }
+                    _ => {}
+                }
+            }
+            cond_spans.push((open, end));
+            search = end + 1;
+        }
+        markers.iter().any(|m| {
+            body.match_indices(m)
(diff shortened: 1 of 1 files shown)


jatmn (changes requested):

## Findings

- [P2] Do not let a partial gate satisfy the egress guard
  `crates/gitlawb-node/src/api/mod.rs:388`
  The new guard treats a handler as gated when any marker string appears anywhere in its body, so `git_info_refs` currently passes because it contains `visibility_check(`. In the actual handler, though, that check only runs under `if service == "git-upload-pack"`; `service=git-receive-pack` skips it and is the live private-ref advertisement gap being fixed separately in #119. That means this PR would merge a guard that says every repo-scoped handler is gated while the known `git_info_refs` egress is neither fully gated nor in `KNOWN_UNGATED`, and the same pattern would miss future partially-gated handlers. Please make the guard distinguish fully-applied gates from conditional/partial ones, or explicitly track `git_info_refs` as a known ungated gap until the #119 fix lands.

beardthelion:
Addressed in cbde4fd, with a follow-up hardening in 93b842b.

The guard counted a handler as gated on bare substring presence, so `git_info_refs` false-passed on `visibility_check(` even though that call sits under `if service == "git-upload-pack"` and leaves the git-receive-pack advertisement ungated. Added `gate_runs_unconditionally`: a marker that appears only inside an `if service ==` block no longer counts as a full gate. `git_info_refs` is now tracked in `KNOWN_UNGATED` against #119, and the staleness assert forces its removal the moment #119 makes the gate unconditional (verified by simulating that change: the assert trips its `now gated` staleness check).

`git_info_refs` is the only handler with an `if service ==` discriminator, so no other handler's classification changes. 93b842b clamps the span scan against an unclosed block (no panic; fail-safe to over-flag) and documents that `match service` is not detected.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

Issue #126, opened 2026-06-30
GET /ipfs/{cid} serves withheld DANGLING (unreachable) blobs to anon: gate uses fail-open deny-set, not fail-closed allowed-set


beardthelion:
The `GET /ipfs/{cid}` visibility gate added in #110 closes the reachable case but still serves a **withheld blob in cleartext to an anonymous caller when that blob is unreachable** (a dangling / force-push-orphan object). The gate filters on the deny-set (`withheld_blob_oids`), which enumerates only commit-reachable blobs, while `store::read_object` serves any object by oid. This is the #99 dangling-blob class on the content-addressed egress surface.

Found during the #110 code review; independent of and not closed by #110.

### Where

- `crates/gitlawb-node/src/api/ipfs.rs` — `get_by_cid`, the per-repo gate (skips an oid only if it is in `withheld_blob_oids`)
- `crates/gitlawb-node/src/git/visibility_pack.rs` — `withheld_blob_oids` / `blob_paths` (reachable-only enumeration)
- `crates/gitlawb-node/src/git/store.rs:275` — `read_object` (`git cat-file`, no reachability check)

### Root cause

`withheld_blob_oids` builds its set from `blob_paths`, which walks `git rev-list --all` + `git ls-tree` — **reachable objects only**. A dangling blob (written via `git hash-object -w`, referenced by no tree or commit) is absent from that set. `get_by_cid`, for a path-scoped repo, serves the requested object unless its oid is in the withheld set, and `read_object` resolves any object in the DB by oid via `git cat-file` regardless of reachability. So the gate is **fail-open** for unreachable objects: deny-set membership says "this reachable blob is withheld," not "this caller may read this object."

This is the inverse of the serve/replication path, which is **fail-closed**: it intersects against the reachable allowed-set (`replicable_objects_fail_closed` / `replicable_blob_set` over `all_blob_oids`), so an unreachable object is never in the allowed set and is dropped. The IPFS handler consults the deny-set instead of the allowed-set, which is the wrong direction for an egress filter.

### Reproduction (observed by execution against the committed #110 handler)

A throwaway `#[sqlx::test]` (reverted, not committed): seed a SHA-256 repo, public with a `/secret/**` deny rule and one reachable secret blob; write a second secret blob straight into the bare repo's object DB with `git hash-object -w` (dangling); build each blob's CID via `Cid::from_sha256_bytes`; anonymous `GET /ipfs/{cid}` for each.

```
[...]

Issue #124, opened 2026-06-30
Replicated mirror rows are public with no rules: unauthenticated read of withheld repo content via bare short-DID owner


beardthelion:
Replicated mirror rows are stored `is_public=true` with no visibility rules, and `get_repo` can resolve a request to the mirror row instead of the canonical one. The result is that an unauthenticated caller can read a private/withheld repo's refs (and, on the pack path, its objects) through the bare short-DID owner, bypassing the canonical row's visibility rules.

Found as a follow-up while planning the `GET /ipfs/{cid}` visibility gate (#110). It is independent of #110: the leak is already live on the smart-HTTP pack path today.

### Where

- `upsert_mirror_repo` — `crates/gitlawb-node/src/db/mod.rs:813`
- `get_repo` — `crates/gitlawb-node/src/db/mod.rs:843`
- `git_info_refs` / `git_upload_pack` gate — `crates/gitlawb-node/src/api/repos.rs:486`
- `sync.rs` replication ingest (no rule replication)

### Root cause

Three facts combine:

1. **Mirror rows are born public with no rules.** When a peer's repo is replicated here, `upsert_mirror_repo` inserts a row with `id = "{owner_short}/{name}"`, `owner_did = <bare short DID>`, and a hardcoded `is_public = true`. It inserts no visibility rules, and `sync.rs` never replicates the owner's rules. Visibility rules attach per physical `repo.id` (`visibility_rules.repo_id`; `list_visibility_rules(&repo.id)`), so the mirror row has an empty rule set regardless of how the canonical repo is gated.

2. **`get_repo` fuzzy-matches both rows and picks one nondeterministically.** Its query is `WHERE (owner_did = $1 OR owner_did LIKE '%:' || $1 || '%') AND name = $2` with `fetch_optional` and no `ORDER BY`. For a request to the bare short DID, both the mirror row (`owner_did = <short>`, exact match) and the canonical row (`owner_did = did:key:<short>`, LIKE match) qualify; the DB returns one with no defined preference.

3. **The read gate trusts whichever row it got.** `git_info_refs` / `git_upload_pack` call `list_visibility_rules(&record.id)` + `visibility_check(.., "/")`. When the record is the mirror row, the rule set is empty and `is_public = true`, so the check returns `Allow` and the advertisement/pack is served, even though the canonical row carries a `/secret/**` deny (or is fully private).

### Reproduction (observed)

[...]

PR #128, opened 2026-06-30
fix(node): gate GET /ipfs/{cid} on per-caller path-scoped visibility (#110)

### LINKED ISSUE #110: GET /ipfs/{cid} serves any git object by raw hash with no visibility check, leaking withheld blobs
The content-addressed retrieval endpoint bypasses path-scoped visibility entirely. This is the same withholding-bypass class as #98 (fork) on a different egress surface.

### Where

`crates/gitlawb-node/src/api/ipfs.rs`, `get_by_cid` (route `GET /ipfs/{cid}` registered at `server.rs:191`).

- The handler takes no `AuthenticatedDid` extension and the route sits outside `add_auth_layers`, so it is unauthenticated.
- It decodes the CID to a SHA-256, calls `list_all_repos()`, and for every repo does `repo_store.acquire(...)` + `store::read_object(repo_path, sha)`, returning the first match as raw bytes.
- There is no `authorize_repo_read`, no `visibility_check`, and no `withheld_blob_oids` filtering anywhere on the path.

### Impact

For a public repo with a path-scoped deny rule (e.g. `/secret/**`), `git_upload_pack` serves a filtered pack: the withheld blob content is omitted, but the tree objects that name those blobs and their SHAs are kept intact (mode B). An unauthenticated caller can therefore:

1. Clone the repo and read the tree under the withheld path to recover the withheld blob SHAs and filenames.
2. Compute each blob CID and fetch the cleartext content via `GET /ipfs/{cid}`.

So every blob the read path is careful to withhold is retrievable in cleartext by anyone who can reach the node. The fork gate (#98) and the upload-pack filter do not help, because the leak is on a separate route.

### Suggested remediation

`get_by_cid` must not serve a blob the requesting caller may not read. Options:

1. Resolve the object to its owning repo(s) and apply the same per-caller withheld check the serve path uses (`withheld_blob_oids` / `visibility_check`) before returning content; fail closed (404) when the object is in any repos withheld set for the caller. Requires threading caller identity onto the route.
2. If the endpoint is meant only for already-public content, restrict it to objects reachable from a repo+path the caller passes `authorize_repo_read` for, rather than a blind cross-repo SHA scan.

Found while mapping egress paths during the #98 review. Pre-existing, independent of #98.

beardthelion:
Closes #110.

`GET /ipfs/{cid}` served any git object by its raw SHA-256 with no authentication and no visibility check. For a public repo with a path-scoped deny rule (e.g. `/secret/**`), an unauthenticated caller could recover a withheld blob's oid from the served trees, derive its CID, and fetch the cleartext here, bypassing the pack-path withholding. This is the withholding-bypass class of #98 and #116 on the content-addressed egress surface.

## What changed

- Bind optional caller identity: `/ipfs/{cid}` gets its own sub-router under `optional_signature`; `/api/v1/ipfs/pins` stays unsigned (see residuals).
- Gate each iterated repo row against its own rules: `visibility_check` at `/` denies the row (continue), and for rows with path-scoped rules a blob in the caller's `withheld_blob_oids` set is skipped. The object is served only from a row the caller passes; denial and genuine not-found both return an opaque 404.
- The row is gated directly rather than through `authorize_repo_read`, whose fuzzy `get_repo` re-resolve could authorize a different physical row than the one read. Visibility rules are fetched in one batched query; the per-repo withheld set is memoized within the request; the history walk runs on `spawn_blocking` and fails closed (skips the repo) on a walk error or task panic.
- `get_by_cid` is added to the `authz_guard` drift table so the gate cannot be silently removed.

## Tests

New `#[sqlx::test]` coverage, executed RED-before / GREEN-after: anon and signed-non-reader denied (404, no leak), owner and listed reader served, public blob and withheld-subtree tree CID served (trees are not withheld), multi-repo availability from a public copy when withheld elsewhere, repo-level private deny, and the fail-closed walk-error path (verified load-bearing by mutation). Full `gitlawb-node` suite green; fmt and clippy clean.

## Scope and known residuals

This closes the direct unauthenticated CID scan. It does not close the whole withholding bypass on the IPFS surface, and should not be read as doing so:

- **#124** — a stale-public mirror row serves withheld content on both this path and the pack path (pre-existing; mirror rows are born `is_public=true` with no replicated rules).
- **#126** — a withheld but unreachable (dangling/force-push-orphan) blob is still served, because this gate filters on the reachable deny-set rather than the fail-closed allowed-set the serve/pin path uses.
- **#121** — `/api/v1/ipfs/pins` still exposes withheld CIDs to anonymous callers; until it is gated, the opaque-404 here is undermined by that oracle.

Note: `store::init_bare` creates repos with `--object-format=sha1`, where a sha256 CID digest never matches a git oid, so this endpoint is dormant for those repos and the gate is defense-in-depth there. The tests seed sha256 repos to exercise the gate.

## Summary by CodeRabbit

* **New Features**
  * `/ipfs/{cid}` now respects caller identity when serving content, enabling per-repository and path-based access control.
  * Content that should be hidden is now skipped automatically, even when the object exists in a repository.
* **Bug Fixes**
  * Anonymous or unauthorized requests now receive a consistent not-found response for protected content.
  * Public copies of shared content can still be served when a hidden copy exists elsewhere.
  * Private repositories now correctly deny access unless the caller is allowed.

### DIFF
--- a/crates/gitlawb-node/src/api/ipfs.rs
+++ b/crates/gitlawb-node/src/api/ipfs.rs
@@ -9,6 +9,10 @@
 //! `git cat-file <type> <sha256>` (i.e. without the git framing header).
 //! This is consistent with how `gitlawb_core::cid::Cid::from_git_object_bytes`
 //! computes CIDs when objects are pushed.
+//!
+//! Serving is access-controlled: an object is returned only from a repo row the
+//! requesting caller is permitted to read (per-caller path-scoped visibility,
+//! see `get_by_cid`).
 
 use axum::{
     extract::{Path, State},
@@ -71,6 +75,16 @@ pub async fn get_by_cid(
         .await
         .map_err(AppError::Internal)?;
 
+    // Fetch every repo's visibility rules in one query rather than one per row
+    // (the gate runs each row against its OWN rules — KTD2a). A row absent from
+    // the map has no rules.
+    let repo_ids: Vec<String> = repos.iter().map(|r| r.id.clone()).collect();
+    let rules_by_repo = state
+        .db
+        .list_visibility_rules_for_repos(&repo_ids)
+        .await
+        .map_err(AppError::Internal)?;
+
     // Request-scoped memo of the per-repo withheld set (KTD1). The caller is
     // constant for one request, so `repo.id` alone is a safe, sufficient key —
     // never a coarse caller "class", which `visibility_check`'s exact full-DID
@@ -79,13 +93,11 @@ pub async fn get_by_cid(
 
     for repo in &repos {
         // Repo-level read gate against THIS row's own rules (KTD2a).
-        let rules = state
-            .db
-            .list_visibility_rules(&repo.id)
-            .await
-            .map_err(AppError::Internal)?;
-        if visibility_check(&rules, repo.is_public, &repo.owner_did, caller, "/") == Decision::Deny
-        {
+        let rules: &[crate::db::VisibilityRule] = rules_by_repo
+            .get(&repo.id)
+            .map(Vec::as_slice)
+            .unwrap_or(&[]);
+        if visibility_check(rules, repo.is_public, &repo.owner_did, caller, "/") == Decision::Deny {
             continue;
         }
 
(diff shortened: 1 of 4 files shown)


kevincodex1 (approved):
LGTM

PR #133, opened 2026-06-30
fix(node): gate GET /ipfs/{cid} on reachable allowed-set, not deny-set (#126)

### LINKED ISSUE #126: GET /ipfs/{cid} serves withheld DANGLING (unreachable) blobs to anon: gate uses fail-open deny-set, not fail-closed allowed-set

Gravirei:
Closes #126.

The `GET /ipfs/{cid}` visibility gate used `withheld_blob_oids` (a deny-set enumerating only reachable blobs), so a dangling/unreachable blob — absent from the set — was served in cleartext to anonymous callers. Flipped to an allowed-set (`allowed_blob_set_for_caller`) that enumerates reachable blobs the caller may read: a dangling blob has no path, is never in the set, and 404s.

**Changes:**
- `visibility_pack.rs`: added `allowed_blob_set_for_caller`, the caller-aware allowed-set. `replicable_blob_set` now delegates to it.
- `ipfs.rs`: `get_by_cid` gates on the allowed-set instead of the deny-set; a blob passes only under path-scoped rules and if its oid is in the reachable allowed-set.
- `test_support.rs + visibility_pack.rs`: unit and integration tests cover anon, listed reader, and owner against dangling blobs.

All 32 unit tests and 5 integration tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened `GET /ipfs/{cid}` so blob content is served only when it’s reachable and explicitly permitted by repository visibility rules, particularly under path-scoped rules.
  * Unreachable/dangling blobs are now excluded and return an opaque “not found” response (not the underlying content), even when walk/authorization evaluation fails.
* **Tests**
  * Added regression coverage verifying dangling blobs fail closed for anonymous and owner-signed requests under path-based visibility rules.


beardthelion (changes requested):
The allow-set flip is the right fix for #126 and it is fail-closed. I built the merged state and ran it: dangling and walk-error cases 404 for every caller class (anon, listed reader, owner), `replicable_blob_set` is behavior-identical after delegating to the new helper, and the existence-before-walk reorder addresses the earlier perf note. One mechanical blocker before this can merge, plus one divergence worth its own issue (already filed).

## Findings

- [P3] Run `cargo fmt`, `ipfs.rs` fails `rustfmt --check` (blocks the format gate)
  `crates/gitlawb-node/src/api/ipfs.rs:145`
  The `allowed_blob_set_for_caller(...)` call (line ~145) and the `x-content-cid` `HeaderValue::from_str(...)` line (~180) are not rustfmt-formatted, so the format check fails. `cargo fmt` fixes both. This is the only change required to merge.

- [P2] Tree/commit CIDs still disclose a withheld subtree's structure (follow-up, not a blocker here)
  `crates/gitlawb-node/src/api/ipfs.rs:139`
  Under a path-scoped rule the gate applies only to `obj_type == "blob"`, so a tree/commit/tag object is served unconditionally and its raw body hands an anonymous caller every child filename and OID of a withheld subtree. This is pre-existing (the old deny-set never held trees either) and explicitly out of #126's scope, so it should not block this PR. But `get_tree` (`api/repos.rs:444`, the N3 fix) deliberately protects exactly this, so the two read surfaces over the same data disagree. Tracked separately in #135.

- [P3] Remove the now-dead `withheld_blob_oids` or note why it stays
  `crates/gitlawb-node/src/git/visibility_pack.rs:232`
  After this PR nothing in production calls `withheld_blob_oids`; only its own tests reference it. Because it is `pub`, the compiler and clippy will not flag it. Either delete it (and its tests) or add a line on why it is retained.

Notes (non-blocking): the existence-check-before-walk reorder means an existing-but-withheld CID pays the walk while an absent CID returns immediately, so latency weakly signals object existence; both still 404, and this is the perf win from the earlier review, so I would accept it. Test-wise, the KTD3 tree-serve is asserted only by status (not body contents), and there is no `get_by_cid` test for a listed-reader serve or an allowed-at-both-paths blob.

Nice work on the fail-closed tests. The dangling-blob coverage across all three caller classes is exactly right. Once `cargo fmt` is run, this is good to go from my side (note the fork CI still needs a maintainer to approve the workflow run).

jatmn (changes requested):

## Findings

- [P2] Avoid reading unauthorized blob bodies before the allowed-set gate
  `crates/gitlawb-node/src/api/ipfs.rs:125`
  The reordered path now calls `store::read_object` before checking whether a path-scoped blob is in `allowed_blob_set_for_caller`, and `read_object` does both `git cat-file -t` and `git cat-file <type>` before returning. That means a caller who knows a withheld or dangling blob CID can make the node materialize the full unauthorized blob body into memory just to discard it and return 404. Please split the existence/type probe from the content read, run the allowed-set check after confirming the object is a blob, and only call the body-reading path once the blob is authorized to be served.

beardthelion (approved):
The split-probe reorder closes the P2. `get_by_cid` now type-probes with `cat-file -t`, runs the allowed-set check, and only reads the object body once that check passes, so a withheld or dangling blob is never materialized for a caller who can't reach it. I built the merged state and walked the caller matrix: anon, non-reader, listed-reader, and owner all behave; dangling and walk-error fail closed (404) for every class; public copies still serve.

@kevincodex1 LGTM.

jatmn (changes requested):
I found one gate issue that needs to be addressed before this is ready.

## Findings

- [P2] Run rustfmt so the required format check passes
  `crates/gitlawb-node/src/git/store.rs:286`
  The current head still fails `cargo fmt --all -- --check`; rustfmt wants to split the new `object_type` return expression across multiple lines. The PR Checks workflow runs that exact command before clippy, and the contribution guide says CI rejects formatting failures, so this will block the `fmt + clippy` job even though the focused visibility tests pass. Please run `cargo fmt --all` and push the formatted result.

beardthelion (approved):
Re-reviewed on 03ba7149. The rustfmt blocker is resolved: the store.rs change is a pure line-wrap of the same expression, and fmt + clippy is green. The gate is unchanged from the head I approved, get_by_cid still gates on the reachable allowed-set with the object_type existence probe ahead of the walk, and the dangling-blob path still fails closed on the merged state. Merge from main is clean, CI green.

@kevincodex1 LGTM.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
looks good . thanks

PR #137, opened 2026-06-30
fix(remote): include HTTP error response body

### LINKED ISSUE #77: git remote helper drops the HTTP error body on failed push/fetch, hiding the server's rejection reason
### Summary

The git remote helper discards the HTTP response body on a failed push, so the server's descriptive error never reaches the user. `crates/git-remote-gitlawb/src/main.rs:297`:

```rust
if !pack_resp.status().is_success() {
    bail!("POST /{} returned {}", service, pack_resp.status());
}
```

It reports only the status line. The JSON body (which carries the `message` field, e.g. the owner-push rejection `"push rejected — only the repo owner may push to this repository (GITLAWB_ENFORCE_OWNER_PUSH is enabled)"`) is dropped.

### Impact

On any non-2xx receive-pack response the user sees `POST /git-receive-pack returned 403 Forbidden` and nothing about why. This is pre-existing and affects every push error path, not one feature. It became more visible with the owner-push enforcement landed in #68, which returns a 403 whose body explains the rejection and what to do, none of which currently surfaces.

The same pattern is on the fetch path at `main.rs:210-213` (`GET /info/refs`), worth fixing together.

### Suggested fix

Read the body before bailing and include it when present:

```rust
if !pack_resp.status().is_success() {
    let status = pack_resp.status();
    let body = pack_resp.text().await.unwrap_or_default();
    if body.is_empty() {
        bail!("POST /{service} returned {status}");
    }
    bail!("POST /{service} returned {status}: {body}");
}
```

`AppError` serializes as `{"error":"<code>","message":"..."}`, so surfacing the raw body is already useful; extracting `message` for a cleaner line is optional.

### Scope

Pre-existing, not a regression from #68. Localized to the remote helper.

paranoa233:
Closes #77

## Summary
- include non-empty HTTP response bodies in remote-helper GET and POST error messages
- keep the existing empty-body hint for missing `info/refs` responses
- add local HTTP tests for both failing ref advertisement and failing pack exchange paths

## Tests
- `cargo +stable-x86_64-pc-windows-gnu test -p git-remote-gitlawb`
- `cargo +stable-x86_64-pc-windows-gnu clippy -p git-remote-gitlawb --all-targets -- -D warnings`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages for smart-protocol proxy failures by including a sanitized, truncated excerpt of the server response.
  * Enhanced handling of unsuccessful `GET /info/refs` and `POST /{service}` responses, with clearer guidance when the server body is missing or empty.
* **Tests**
  * Updated the local HTTP test harness to capture and assert raw incoming request text.
  * Added/expanded coverage for error-message sanitization and truncation, including request-specific failure scenarios.

### DIFF
--- a/crates/git-remote-gitlawb/src/main.rs
+++ b/crates/git-remote-gitlawb/src/main.rs
@@ -209,7 +209,7 @@ fn handle_connect<R: Read>(
 
     if !refs_resp.status().is_success() {
         let status = refs_resp.status();
-        let body = refs_resp.text().unwrap_or_default();
+        let body = read_error_body(refs_resp);
         bail!(
             "{}",
             http_error_message(
@@ -307,7 +307,7 @@ fn handle_connect<R: Read>(
 
     if !pack_resp.status().is_success() {
         let status = pack_resp.status();
-        let body = pack_resp.text().unwrap_or_default();
+        let body = read_error_body(pack_resp);
         let path = format!("/{service}");
         bail!("{}", http_error_message("POST", &path, status, &body, None));
     }
@@ -444,6 +444,20 @@ fn url_path(url: &str) -> String {
         .unwrap_or_else(|| "/".to_string())
 }
 
+const MAX_ERROR_BODY_BYTES: u64 = 4096;
+const MAX_ERROR_BODY_CHARS: usize = 2000;
+
+fn read_error_body(mut response: reqwest::blocking::Response) -> String {
+    let mut body = Vec::new();
+    let mut limited = (&mut response).take(MAX_ERROR_BODY_BYTES);
+
+    if limited.read_to_end(&mut body).is_err() {
+        return String::new();
+    }
+
+    String::from_utf8_lossy(&body).into_owned()
+}
+
 fn http_error_message(
     method: &str,
     path: &str,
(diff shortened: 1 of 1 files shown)


jatmn (approved):

@kevincodex1 LGTM

beardthelion (dismissed):
The refactor and the core change hold up: the `<R: Read>` generalization preserves buffering (the production caller still passes a `BufReader`), the `bail!("{}", body)` sites don't reinterpret `{}` in the body, and the new tests are sound. The issue is what the new error path echoes.

## Findings

- [P2] Sanitize the node error body before surfacing it to the terminal
  `crates/git-remote-gitlawb/src/main.rs:447` (helper), call sites `:212` and `:310`
  On a non-2xx response the body is read with `.text()` and embedded into the error after only `body.trim()`, then `bail!`-ed to stderr, straight to the user's terminal. The node is untrusted here: `git clone gitlawb://` can target an attacker-run node, and the default transport is plain `http://`, so a MITM also controls the response. A 4xx/5xx body carrying ANSI/OSC escapes (screen clear, cursor moves, a `\x1b]0;...\x07` title rewrite, a faked "clone succeeded" line) is then rendered by the terminal. Before this PR only the status code was echoed, so the exposure is introduced here. Strip or escape C0/C1 control bytes (keep printable text, drop ESC/BEL and friends) inside `http_error_message` so both call sites are covered; trimming whitespace alone is not enough. I ran a control-stripping pass against the existing error tests and the legible node message still surfaces, so the fix doesn't cost the feature this PR adds.

- [P3] Bound the error-body read
  `crates/git-remote-gitlawb/src/main.rs:212,310`
  `.text()` has no size cap, so a hostile node can return an arbitrarily large error body that the helper buffers entirely into memory on the error path (also new here). Read a bounded amount (a few KiB via `Read::take`) before converting, which also limits the escape surface above.

Highest first; the P2 is the one to fix before merge.

kevincodex1 (approved):
nice catch!

kevincodex1:
please address @beardthelion comments

beardthelion (approved):
Re-reviewed on `ef449fd`. Both earlier findings are addressed and I verified the sanitizer against the threat: `safe_error_body_excerpt` drops every `char::is_control()` char and collapses whitespace, and `read_error_body` bounds the read to 4 KiB. Checked the escape introducers directly, ESC (`0x1b`), BEL (`0x07`), DEL (`0x7f`) and the full C1 block including the 8-bit CSI (`0x9b`) and OSC (`0x9d`) are all `is_control()`, so neither the `\x1b[…` form nor the 8-bit forms can start a sequence, and a bare `\r` collapses to a space instead of splicing a fake line. Both non-2xx paths route through it and no other wire body reaches the terminal unsanitized. Closes the terminal escape-sequence exposure.

CodeRabbit's one nitpick (bound the error-body length) is addressed, implemented as the 2000-char excerpt cap and a 4096-byte read cap, both verified.

Two optional, non-blocking notes: the defense rests on `char::is_control()` classifying the C1 range as control, so a future switch to a byte-level `is_ascii_control` check (which stops at `0x7f`) would silently reopen the 8-bit path; a one-line test asserting `U+009B`/`U+009D` are stripped would pin it. And bidi/format chars (e.g. `U+202E`) still pass through, but that is text reordering rather than terminal control, out of scope here.

@kevincodex1 this is good to merge.

PR #138, opened 2026-07-01
feat(gl): sanctioned iCaptcha client flow + secure git lifecycle

kevincodex1:
## Summary

Builds the **client side** of the iCaptcha gate so a legitimate agent completes the full git lifecycle (register → repo create → push → clone) transparently — no manual headers, no `GITLAWB_ICAPTCHA_PROOF` env hack. Server enforcement is unchanged.

## Motivation & context

The node enforces iCaptcha on writes (create/fork/register), but `gl`/`git-remote-gitlawb` had no native support — an agent testing the lifecycle hit a chain of failures (silent version drift → `401`, then `403 iCaptcha proof required`, then `sub` mismatch, plus misleading `repo info` stubs). This implements the *sanctioned* flow: solve the challenge and retry automatically.
Closes #

## Kind of change

- [ ] Bug fix
- [x] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## Summary

Builds the **client side** of the iCaptcha gate so a legitimate agent completes the full git lifecycle (register → repo create → push → clone) transparently — no manual headers, no `GITLAWB_ICAPTCHA_PROOF` env hack. Server enforcement is unchanged.

## Motivation & context

The node enforces iCaptcha on writes (create/fork/register), but `gl`/`git-remote-gitlawb` had no native support — an agent testing the lifecycle hit a chain of failures (silent version drift → `401`, then `403 iCaptcha proof required`, then `sub` mismatch, plus misleading `repo info` stubs). This implements the *sanctioned* flow: solve the challenge and retry automatically.
Closes #

## Kind of change

- [ ] Bug fix
- [x] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

- **New crate `icaptcha-client`** — deterministic solvers (arithmetic/algebra/sequence) + `obtain_proof()` challenge→solve→answer loop; `requesterId` = caller DID (so proof `sub` matches the signer).
- **`gl` (`crates/gl`)** — `http.rs`: unified signed writes that, on a `403` iCaptcha challenge (detected via new `x-icaptcha-url`/`x-icaptcha-level` headers), solve and retry with `x-icaptcha-proof` (bounded retry for the ~5 min TTL); **removes the env hack**; actionable hint on `401 not_an_agent`. `repo.rs`: `repo info`/`clone` return a real **404** instead of a stub with a placeholder owner DID. `doctor.rs`: gl-vs-node version drift + iCaptcha reachability checks.
- **`git-remote-gitlawb`** — same shared client; push stays RFC-9421 signed, with a 403-iCaptcha solve/retry safety net; clone/fetch unsigned; missing-repo 404 kept.
- **`gitlawb-node`** — the `403 icaptcha_proof_required` response now advertises the service URL + required level as JSON fields and `x-icaptcha-url`/`x-icaptcha-level` headers (mirrors the existing `human_detected` pattern). The `sub == authenticated-DID`, expiry, and level checks are **unchanged**.
- **README** — full lifecycle docs (`requesterId == DID` rule, proof TTL).

## How a reviewer can verify

```sh
cargo fmt --all --check
cargo test --workspace            # needs Postgres (DATABASE_URL); node still rejects missing/expired/wrong-sub proofs
cargo test -p icaptcha-client     # deterministic solvers, no DB
# behavioural: repo-not-found is a clean 404, not a stub
cargo run -p gl -- repo info zNoSuchOwner/definitely-not-real --node https://node.gitlawb.com

Before you request review

- [x] cargo test --workspace passes locally (against Postgres)
- [x] New behavior is covered by tests (solvers; node proof-rejection tests retained)
- [x] cargo fmt --all and cargo clippy --workspace --all-targets -- -D warnings are clean
- [x] Docs / .env.example updated if behavior or config changed (README lifecycle)

Protocol & signing impact

- [x] Touches DID / did:key, Ed25519 / RFC 9421 signatures, UCAN, ref certs, or P2P wire formats

The only wire additions are additive, backward-compatible: the x-icaptcha-url/x-icaptcha-level response headers + JSON fields on an existing 403, and the client echoing the existing x-icaptcha-proof header. No change to signing coverage or proof verification; old clients/nodes interoperate unchanged.

Notes for reviewers

- Push is intentionally signed-only (owner signature is the gate) — not per-push iCaptcha-gated; auto-create-on-push was deferred.
- The proof stays an unsigned header for now; binding it into the RFC-9421 signature is noted as future hardening.
- Depends on the node advertising the x-icaptcha-* headers (included here) — deploy node + clients together.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added support for iCaptcha-protected write actions, with automatic challenge handling and retry.
  * Expanded `gl doctor` to check iCaptcha connectivity and surface setup warnings.

* **Bug Fixes**
  * Improved repo cloning and info errors with clearer, more specific messages.
  * Tightened handling of proof reuse and expired challenges during writes.

* **Documentation**
  * Added guidance for using iCaptcha-enabled nodes, including signature requirements and optional setup steps.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3304,7 +3304,7 @@ version = "0.4.0"
 dependencies = [
  "anyhow",
  "gitlawb-core",
- "icaptcha-client",
+ "mockito",
  "reqwest",
  "tracing",
  "tracing-subscriber",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3304,7 +3304,7 @@ version = "0.4.0"
 dependencies = [
  "anyhow",
  "gitlawb-core",
- "mockito",
+ "icaptcha-client",
  "reqwest",
  "tracing",
  "tracing-subscriber",
@@ -3420,6 +3420,7 @@ dependencies = [
  "clap",
  "dirs",
  "gitlawb-core",
+ "icaptcha-client",
  "mockito",
  "reqwest",
  "serde",
@@ -3812,6 +3813,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f31827a206f56af32e590ba56d5d2d085f558508192593743f16b2306495269f"
 dependencies = [
  "cc",
+]
+
+[[package]]
+name = "icaptcha-client"
+version = "0.4.0"
+dependencies = [
+ "anyhow",
+ "gitlawb-core",
+ "reqwest",
+ "serde",
+ "serde_json",
+ "thiserror 2.0.18",
+ "tracing",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
     "crates/gl",
     "crates/git-remote-gitlawb",
     "crates/gitlawb-attest",
+    "crates/icaptcha-client",
 ]
 
 [workspace.package]
--- a/README.md
+++ b/README.md
@@ -218,8 +218,11 @@ Notes:
   `git push` is **signed-only** (owner signature is the gate — no per-push challenge);
   reads (clone / fetch / `repo info`) need no proof. A non-existent repo returns a
   clear `404`, never a placeholder.
-- **API-key iCaptcha deployments:** set `GITLAWB_ICAPTCHA_API_KEY` and the client
-  sends it as a bearer token to the iCaptcha service.
+- **API-key iCaptcha deployments:** set `GITLAWB_ICAPTCHA_URL` to your iCaptcha
+  origin and `GITLAWB_ICAPTCHA_API_KEY` to its key. The client only talks to an
+  `https` origin whose host is allowlisted (that URL or the public default), and
+  sends the bearer token **only** to your configured origin — never to a URL a
+  node advertises — so a hostile node can't capture the key or redirect the solve.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -185,6 +185,42 @@ git clone gitlawb://did:key:z6Mk.../my-repo
 
 For public-network use, make sure `GITLAWB_NODE` points to the node you want. The helper defaults to localhost for local development.
 
+### Full lifecycle against an iCaptcha-enforcing node
+
+Public nodes (e.g. `node.gitlawb.com`) require two things on writes:
+
+1. **RFC 9421 HTTP Signatures** — every write is signed by your identity key. `gl`
+   and the `git-remote-gitlawb` helper do this automatically. An old/unsigned CLI
+   fails with `401 not_an_agent`; `gl` will tell you to upgrade and register.
+2. **An iCaptcha proof** on the spam-gated writes (**repo create, fork, register**).
+   `gl` solves this for you: on the node's `403 icaptcha_proof_required` it reads the
+   `x-icaptcha-url` / `x-icaptcha-level` hints, requests a challenge, solves it
+   locally (arithmetic / algebra / sequence), and **retries the same signed request**
+   with the `x-icaptcha-proof` header — no manual steps, no env vars.
+
+```bash
+gl identity new                                   # create did:key identity
+gl register      --node https://node.gitlawb.com  # signed + auto-solves iCaptcha
+gl repo create memlawb --node https://node.gitlawb.com   # signed + auto-solves iCaptcha
+git push  origin2 main                            # origin2 = gitlawb://<your-did>/memlawb (signed)
+git clone gitlawb://<your-did>/memlawb            # public read, no proof needed
+gl doctor                                         # preflight: identity, node, version, iCaptcha
+```
+
+Notes:
+
+- **`requesterId` is always your DID.** The proof's `sub` claim must equal the
+  authenticated signer; `gl`/helper set this automatically and the node enforces
+  `sub == authenticated DID` (so a proof minted for another identity is rejected).
+- **Proofs are short-lived (~5 min TTL) and single-use.** If one expires between
+  solving and use, the client transparently solves a fresh one and retries.
+- **What needs what:** create / fork / register are signed **and** iCaptcha-gated;
+  `git push` is **signed-only** (owner signature is the gate — no per-push challenge);
+  reads (clone / fetch / `repo info`) need no proof. A non-existent repo returns a
+  clear `404`, never a placeholder.
+- **API-key iCaptcha deployments:** set `GITLAWB_ICAPTCHA_API_KEY` and the client
+  sends it as a bearer token to the iCaptcha service.
+
 ---
 
 ## Architecture
--- a/crates/gitlawb-node/src/error.rs
+++ b/crates/gitlawb-node/src/error.rs
@@ -23,8 +23,14 @@ pub enum AppError {
     #[allow(dead_code)]
     Forbidden(String),
 
-    #[error("icaptcha proof required: {0}")]
-    IcaptchaProofRequired(String),
+    #[error("icaptcha proof required: {message}")]
+    IcaptchaProofRequired {
+        message: String,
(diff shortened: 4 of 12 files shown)


jatmn (changes requested):

## Findings

- [P2] Complete CodeRabbit's request to keep the remote helper non-interactive
  `crates/git-remote-gitlawb/src/main.rs:342`
  CodeRabbit's thread is still unresolved, and the current push retry path calls `icaptcha_client::obtain_proof(&cfg, None)`. Passing `None` lets the client fall back to `interactive_prompt`, which reads from process stdin. In the remote helper, stdin is the Git remote-helper protocol stream, so an unexpected challenge type can block waiting for input or consume protocol bytes instead of returning a clean Git error. Please complete that review request by using a non-interactive proof path in `git-remote-gitlawb` and surfacing an actionable failure when the proof cannot be solved automatically.

- [P2] Complete CodeRabbit's request to preserve the short owner form for bare clone names
  `crates/gl/src/repo.rs:311`
  CodeRabbit's bare-name clone comment is still valid. `cmd_info` and `resolve_owner_repo_pair` resolve owner-less repo names through `resolve_owner_did`, which returns the short key segment, but `cmd_clone` now uses `kp.did().to_string()` directly. That makes `gl repo clone myrepo` build `gitlawb://did:key:z.../myrepo` while the adjacent commands and helper path shape use `gitlawb://z.../myrepo`. Please complete the review request by reusing `resolve_owner_did` or the same short-owner derivation for the bare-name branch.

- [P2] Complete CodeRabbit's request to avoid copying the full push pack
  `crates/git-remote-gitlawb/src/main.rs:332`
  The retry loop clones `request_body` before every send, so even the normal first push attempt duplicates the entire pack in memory. Large pushes are exactly where this helper needs predictable memory use, and the previous code avoided that copy by moving the body once. Please complete CodeRabbit's request by using a shared cloneable body representation, such as `bytes::Bytes`, so retries can reuse the body without duplicating the pack bytes.

- [P3] Complete CodeRabbit's request to align `gl doctor` with header-driven iCaptcha discovery
  `crates/gl/src/doctor.rs:210`
  The write retry path derives the iCaptcha endpoint from the node's `x-icaptcha-url` header and `IcaptchaCfg::new`; the only env var that path reads is `GITLAWB_ICAPTCHA_API_KEY`. `gl doctor` instead probes `GITLAWB_ICAPTCHA_URL` or the default URL and tells users to set that variable, which will not fix a gated-write failure against a node advertising a different iCaptcha service. Please complete the review request by checking or describing the same header-driven endpoint the retry path actually uses.

- [P3] Complete CodeRabbit's request to fix the push/iCaptcha documentation drift
  `README.md:217`
  The README says `git push` has "no per-push challenge", but this PR adds a `git-remote-gitlawb` 403 challenge/retry path for push. Even if push is signed-only by policy today, the absolute wording contradicts the implementation and will mislead users debugging a node that does return a challenge. Please complete the review request by documenting that pushes are signed by default while the helper can retry if the node returns an iCaptcha challenge.

- [P3] Link the protocol discussion required by the contribution guide
  `PR description`
  The PR body marks DID/signing/wire-format impact and notes new `x-icaptcha-*` response headers plus client `x-icaptcha-proof` behavior, but the issue link is still `Closes #` and the protocol-change checklist item for prior discussion is unchecked. `CONTRIBUTING.md` asks contributors to open an issue first for protocol-level changes. Please link the issue or discussion that covers this wire change before this lands, or split/adjust the scope if the protocol direction has not been agreed yet.

beardthelion (changes requested):
The core flow is right: structured iCaptcha discovery mirroring `human_detected`, `sub`-bound proofs, node gate untouched, CI green. Two security issues in the new `icaptcha-client` crate that the earlier reviews did not reach, plus a consolidation of the open push-path threads.

## Findings

- [P1] Sanitize and length-bound the iCaptcha response before it reaches the terminal
  `crates/icaptcha-client/src/lib.rs:133`
  On a non-2xx from the iCaptcha service, `resp.text().unwrap_or_default()` (and `AnswerResult::Failed { reason }` at :113) goes raw into `bail!`, which prints through `main() -> Result` in both `gl` and `git-remote-gitlawb`. The service URL is the node's `x-icaptcha-url`, and nodes are untrusted here (a MITM on the default http transport qualifies), so these are attacker-influenceable bytes reaching the TTY with control chars intact. That is the CWE-150 terminal-escape class #137 fixed in `http_error_message`, reintroduced in the new crate. Driving a hostile endpoint through `obtain_proof` confirmed it: raw ESC bytes land in the error and a 50KB body is echoed unbounded. Strip C0/C1 controls and cap the excerpt (same shape as #137), and bound the read.

- [P2] Do not hand credentials or trust to a node-chosen iCaptcha URL
  `crates/icaptcha-client/src/lib.rs:126`
  `x-icaptcha-url` becomes the base URL with no scheme or host check, then `request_challenge`/`submit_answer` POST the caller DID and, when `GITLAWB_ICAPTCHA_API_KEY` is set, that key as a bearer token to it on the first hop. In the repro a hostile node captured `Authorization: Bearer <key>` directly, no redirect involved (reqwest already strips `Authorization` on cross-host redirects, so that path is closed). It is also a client-side SSRF primitive. Require https and pin or allowlist the host, and do not attach the API key to a node-discovered origin. With a key configured this is effectively P1 exfiltration.

- [P2] Remove the dead iCaptcha retry loop from the push path
  `crates/git-remote-gitlawb/src/main.rs:319`
  This is the root cause under the two open main.rs threads (clone-the-pack and interactive-prompt). `git_receive_pack` (`crates/gitlawb-node/src/api/repos.rs:789`) never calls `verify_request`; only create/fork/register do. So a push 403 is always a plain `Forbidden` with no `x-icaptcha-*` headers, and the loop bails on the first pass without ever solving (`icaptcha_cfg_from_headers` returns `None` on a headerless 403). The loop is dead, yet it reintroduces `request_body.clone()` on every push and carries the latent `interactive_prompt` stdin read. Deleting it and restoring the moved body resolves both threads at once and keeps the helper non-interactive, rather than adding a `bytes::Bytes` body plus a non-interactive solver for a path that cannot fire.

- [P3] Concur on the two open doctor/README threads
  `crates/gl/src/doctor.rs:207`, `README.md:219`
  `gl doctor` probes `GITLAWB_ICAPTCHA_URL`, which nothing in the solve path reads (`IcaptchaCfg::new` reads only `GITLAWB_ICAPTCHA_API_KEY`), and it hits `/v1/pubkey` while the client uses `/v1/challenge` and `/v1/answer`. The README "no per-push challenge" line becomes accurate again once the dead loop above is removed. Both already raised by CodeRabbit and jatmn; the only add is that removing the loop moots the README wording.

On the DID-format thread (CodeRabbit repo.rs:329, seconded as jatmn's J2): I ran this end to end and it is not a break. `git clone gitlawb://did:key:z6Mk.../repo` (full DID) and the short form both dispatch to the helper and produce a byte-identical node request (`GET /z6Mk.../repo/info/refs...` in both), because `parse_gitlawb_url` reduces either authority to the same short owner before any node contact. Full-DID is also what `gl repo create` (repo.rs:261) and `gl repo info` (repo.rs:362) emit and what the README documents, so bare-name clone matches them; the short form is used only for REST API paths, never a `gitlawb://` URL. Reusing `resolve_owner_did` here is harmless but would make the emitted clone URL differ from what `create` prints, so I would leave it. jatmn's protocol-discussion-link ask (J6) still stands on its own.

kevincodex1:
@coderabbitai please do a full review. Approve if its all good

PR #140, opened 2026-07-01
fix(node,gossip): route gossip HTTP through the no-redirect client (#93)

### LINKED ISSUE #93: gossip_task uses a redirect-following client, enabling SSRF on the peer-ping path
`gossip_task` builds its own HTTP client with `reqwest::Client::new()` (`crates/gitlawb-node/src/main.rs:506`), whose default redirect policy follows up to 10 redirects. The periodic ping loop issues `GET {peer.http_url}/health` for every peer row, and peer rows are populated from `POST /api/v1/peers/announce`, which accepts unsigned requests by default (warn-only unless `GITLAWB_REQUIRE_SIGNED_PEER_WRITES=true`).

`is_public_http_url` validates the literal `http_url` at announce and again in `upsert_peer`, so the stored host must be public. But a public host the attacker controls can answer `/health` with `302 Location: http://169.254.169.254/...` (cloud metadata) or any internal address, and this client follows it. This is the same redirect-bounce SSRF the shared `http_client` was hardened against in #78 with `Policy::none()` (see the comment at `main.rs:147-150`); gossip_task is the one outbound path that did not get the same policy.

It is a blind SSRF: only the request status feeds `mark_peer_ping`, so no response body returns to the attacker. Even so, an unauthenticated peer can drive outbound requests from the node to internal hosts and the instance metadata endpoint once per gossip cycle.

### Where
- `crates/gitlawb-node/src/main.rs:506` builds `reqwest::Client::new()`
- `crates/gitlawb-node/src/main.rs:580-588` GETs `{peer.http_url}/health` via that client
- `crates/gitlawb-node/src/main.rs:151-156` the shared client uses `Policy::none()` for exactly this reason

### Fix
Reuse `state.http_client` (already on `AppState`) inside `gossip_task` instead of building a fresh client; or, if a distinct timeout is needed, build it with `.redirect(reqwest::redirect::Policy::none())`.

beardthelion:
Closes #93.

`gossip_task` built its own `reqwest::Client::new()`, which follows up to 10 redirects by default, and used it for the bootstrap announce and the periodic peer `/health` ping. A peer host (public at announce time per `is_public_http_url`) can answer `/health` with a `302` pointing at a link-local or internal address, and the node follows it. So an unauthenticated announce becomes a blind SSRF, one outbound request to an attacker-chosen internal target per gossip cycle.

The fix reuses the shared `state.http_client`, which is already built with `redirect::Policy::none()`, the same guard `sync.rs` and the announce fan-out got in #78. The ping is extracted into `ping_peer_health` so the redirect behavior is testable, and the shared client's construction is pulled into `build_http_client` so the tests exercise the exact client the node runs rather than a hand-rolled copy.

Tests: a `302` to an internal mock (registered `expect(0)`) is not followed, a `200` is healthy, a connection error is unhealthy. The redirect guard is load-bearing against the production builder (flipping `build_http_client` to follow redirects makes it fail). Full crate suite green, clippy `-D warnings` clean.

One thing not covered by a test: that gossip's call site reads `state.http_client` rather than a fresh client. Driving the real gossip loop needs a mock peer, but mockito binds to `127.0.0.1` and both `upsert_peer` and `list_peers` reject non-public hosts, so a local mock never enters the loop. That call site is guarded by a comment against reintroducing `Client::new()`; the redirect policy itself is test-bound via `build_http_client`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved outbound network handling by reusing a shared HTTP client with a 10-second timeout and redirects disabled.
  * Peer health checks are now more strict: only 2xx responses count as healthy; connection errors are treated as unhealthy.
  * Health check behavior avoids any redirect-based bypasses.
* **Tests**
  * Added unit tests covering redirect behavior and connection failure scenarios for peer health checking.

### DIFF
--- a/crates/gitlawb-node/src/main.rs
+++ b/crates/gitlawb-node/src/main.rs
@@ -145,16 +145,8 @@ async fn main() -> Result<()> {
         None
     };
 
-    // Do not follow redirects: this client fans out to peer-supplied URLs
-    // (sync trigger, profile/repo fetches). A peer answering `302 Location:
-    // http://127.0.0.1/` would otherwise bypass the announce-time public-URL
-    // check and turn the redirect into an SSRF.
-    let http_client = Arc::new(
-        reqwest::Client::builder()
-            .timeout(std::time::Duration::from_secs(10))
-            .redirect(reqwest::redirect::Policy::none())
-            .build()?,
-    );
+    // Shared no-redirect HTTP client. See build_http_client for the SSRF rationale.
+    let http_client = Arc::new(build_http_client()?);
 
     let (ref_update_tx, _) = tokio::sync::broadcast::channel::<state::RefUpdateBroadcast>(256);
     let (task_event_tx, _) = tokio::sync::broadcast::channel::<state::TaskEventBroadcast>(256);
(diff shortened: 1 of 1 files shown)


beardthelion:
@coderabbitai full review

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

Issue #144, opened 2026-07-01
Ref-update feeds over-drop distinct-owner remote rows because the wire slug is method-lossy


beardthelion:
The cross-repo ref-update feeds (`refUpdates` GraphQL #112, `GET /api/v1/events/ref-updates` REST #114) drop a genuinely-remote row for anonymous callers when a local private repo shares both the repo name and the owner's trailing DID segment.

Root cause is the wire slug. `publish_ref_update` (`crates/gitlawb-node/src/api/repos.rs`) broadcasts the slug as `{owner_did.split(':').last}/{name}`, discarding the DID method/host. `did:web:host:alice` and `did:gitlawb:alice` both emit `alice/widget`. The feed gate (`ref_update_row_visible`) therefore cannot tell a remote public repo's row from a local private repo's row when their names and trailing segments collide, so it fails closed and drops the row.

This is fail-safe (it never serves a private repo's row) and negligible for full `did:key` ids, but for `did:web` / truncated handles it silently hides legitimate remote activity from the anonymous feed.

## Reproduction (by execution)

Local private `did:web:host:alice/widget`; a remote row for the different owner `did:gitlawb:alice`, same name, is dropped for anon in both slug forms:

```
ref_update_row_visible(private did:web:host:alice/widget, anon, "did:gitlawb:alice/widget") -> false
ref_update_row_visible(private did:web:host:alice/widget, anon, "alice/widget")             -> false
```

## Fix direction

Carry the full owner DID on the ref-update wire slug (and align the mirror-row bare-key storage) so the gate compares methods instead of a lossy trailing segment. The feed gate could then reuse the did:key-aware normalization used elsewhere (`did_matches` / `DEDUP_CTE`) without the fail-open the current short-slug scheme would cause. This is a cross-peer format change, so it needs a compatibility path for older peers.

Context: raised during review of #143 (P1). The gate there is intentionally strict to stay fail-closed given today's lossy slug; this issue tracks the format fix that removes the over-drop without reopening the #112 / #114 leak.

Gravirei:
taking this one....

PR #145, opened 2026-07-02
fix(node): carry full owner DID on ref-update wire event (#144)

Gravirei:
## Summary

Carry the full owner DID on the ref-update wire event so the feed gate can distinguish different DID methods that share the same trailing segment (e.g. `did:web:host:alice` vs `did:gitlawb:alice`).

## Motivation & context

Refs #144

The ref-update wire slug was previously constructed from the last colon-segmented component of the owner DID (`{owner_did.split(":").last}/{name}`), which is lossy. Two owners with different DID methods but the same trailing segment would produce identical slugs, causing the feed gate to over-drop remote rows for anonymous callers.

This PR is the plumbing half of #144: it carries the full owner DID on the wire and stores it in the database so a follow-up can teach the feed gate to use `did_matches` for proper disambiguation. The gate itself (the second half of #144) will land in a separate change on top of this.

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI
- [x] Breaking or protocol change (issue required first)

## What changed

**crates/gitlawb-node (gitlawb-node):**

- `p2p/mod.rs`: Added `owner_did: Option<String>` to `RefUpdateEvent` with `#[serde(default)]` for backward compat with older peers that do not include the field
- `db/mod.rs`: Added `owner_did: Option<String>` to `ReceivedRefUpdate`, added migration v10 to add `owner_did TEXT` column and `idx_ref_updates_owner` index to the `received_ref_updates` table, updated all SELECT/INSERT queries to include the column
- `api/repos.rs`: Set `owner_did: Some(record.owner_did)` when publishing ref-update events over gossip; added `owner_did` to the HTTP sync notify body
- `api/peers.rs`: Added `owner_did` to `NotifyRequest` and wired it through to `ReceivedRefUpdate`

## How a reviewer can verify

```sh
DATABASE_URL=postgresql://gitlawb:changeme@172.19.0.2:5432/gitlawb cargo test -p gitlawb-node
cargo fmt --check
```

## Before you request review

- [x] `cargo test --workspace` passes locally (327/327)
- [x] Commit titles use Conventional Commits (`fix(...)`, `test(...)`, `style(...)`)
- [x] Checked existing PRs so this is not a duplicate

## Protocol & signing impact

This is a backward-compatible wire-format addition: the new `owner_did` field is `#[serde(default)]`, so events from older peers deserialize without it (`owner_did: None`). The `repo` field format is unchanged. Migration v10 adds the column to existing databases; `ADD COLUMN IF NOT EXISTS` is idempotent for re-runs.

## Summary by CodeRabbit

* **New Features**
  * Ref-update notifications and event listings can now include an optional `owner_did` when available.
  * Peer-to-peer sync notifications now propagate the owner DID into published ref-update events.
  * GraphQL ref-updates and subscriptions expose `owner_did` as optional (omittable).

* **Bug Fixes**
  * Owner DID is now stored and returned consistently across sync, peer updates, API event endpoints, and persisted ref-update history.
  * Backward compatibility is preserved when `owner_did` is missing or null in incoming notifications.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3300,7 +3300,7 @@ dependencies = [
 
 [[package]]
 name = "git-remote-gitlawb"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "anyhow",
  "gitlawb-core",
@@ -3312,7 +3312,7 @@ dependencies = [
 
 [[package]]
 name = "gitlawb-attest"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "base64",
  "ed25519-dalek",
@@ -3329,7 +3329,7 @@ dependencies = [
 
 [[package]]
 name = "gitlawb-core"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "anyhow",
  "base64",
@@ -3356,7 +3356,7 @@ dependencies = [
 
 [[package]]
 name = "gitlawb-node"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "alloy",
  "anyhow",
@@ -3412,7 +3412,7 @@ dependencies = [
 
 [[package]]
 name = "gl"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "alloy",
  "anyhow",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3824,6 +3824,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
+ "sha2",
  "thiserror 2.0.18",
  "tracing",
 ]
--- a/crates/gitlawb-node/src/api/events.rs
+++ b/crates/gitlawb-node/src/api/events.rs
@@ -143,18 +143,22 @@ pub async fn list_ref_updates(
     // only when it matches the canonical owner of the local repo the slug
     // names (#P1); legacy None rows are attributed via an exact unique local
     // match (#P3). Both surfaces share this resolver so they cannot drift.
-    let owner_dids: Vec<serde_json::Value> = futures::future::join_all(updates.iter().map(|u| {
-        state
-            .db
-            .resolve_ref_update_owner_did(&u.repo, u.owner_did.as_deref())
-    }))
-    .await
-    .into_iter()
-    .map(|r| {
-        r.map(|o| o.map_or(serde_json::Value::Null, |s| serde_json::json!(s)))
-            .unwrap_or(serde_json::Value::Null)
-    })
-    .collect();
+    // The batch resolver issues at most one query per distinct local repo
+    // rather than one per event row (#P2).
+    let raw_dids: Vec<Option<String>> = state
+        .db
+        .resolve_ref_update_owner_dids(
+            &updates
+                .iter()
+                .map(|u| (u.repo.as_str(), u.owner_did.as_deref()))
+                .collect::<Vec<_>>(),
+        )
+        .await?;
+
+    let owner_dids: Vec<serde_json::Value> = raw_dids
+        .into_iter()
+        .map(|o| o.map_or(serde_json::Value::Null, |s| serde_json::json!(s)))
+        .collect();
 
     let events: Vec<serde_json::Value> = updates
         .iter()
@@ -264,18 +268,19 @@ pub async fn list_repo_events(
     // them.
     let gossip_updates =
         collect_visible_ref_updates(&state.db, Some(&repo_id_str), limit, caller).await?;
-    let gossip_owner_dids: Vec<serde_json::Value> =
-        futures::future::join_all(gossip_updates.iter().map(|u| {
-            state
-                .db
-                .resolve_ref_update_owner_did(&u.repo, u.owner_did.as_deref())
-        }))
-        .await
+    let gossip_raw: Vec<Option<String>> = state
+        .db
+        .resolve_ref_update_owner_dids(
+            &gossip_updates
+                .iter()
+                .map(|u| (u.repo.as_str(), u.owner_did.as_deref()))
+                .collect::<Vec<_>>(),
+        )
+        .await?;
+
+    let gossip_owner_dids: Vec<serde_json::Value> = gossip_raw
         .into_iter()
-        .map(|r| {
-            r.map(|o| o.map_or(serde_json::Value::Null, |s| serde_json::json!(s)))
-                .unwrap_or(serde_json::Value::Null)
-        })
+        .map(|o| o.map_or(serde_json::Value::Null, |s| serde_json::json!(s)))
         .collect();
 
     let gossip_events: Vec<serde_json::Value> = gossip_updates
--- a/crates/gitlawb-node/src/api/events.rs
+++ b/crates/gitlawb-node/src/api/events.rs
@@ -260,7 +260,7 @@ pub async fn list_repo_events(
                     "cert_id":     u.cert_id,
                     "received_at": u.received_at,
                     "from_peer":   u.from_peer,
-                    "owner_did":   u.owner_did,
+                    "owner_did":   serde_json::json!(record.owner_did),
                     "source":      "gossipsub",
                 })
             })
@@ -327,7 +327,6 @@ mod ref_updates_feed_tests {
             new_sha: "a".repeat(40),
             timestamp: Utc::now().to_rfc3339(),
             cert_id: None,
-            owner_did: None,
             received_at: Utc::now().to_rfc3339(),
             from_peer: "peer1".into(),
             owner_did: None,
--- a/crates/gitlawb-node/src/api/events.rs
+++ b/crates/gitlawb-node/src/api/events.rs
@@ -326,6 +326,7 @@ mod ref_updates_feed_tests {
             cert_id: None,
             received_at: Utc::now().to_rfc3339(),
             from_peer: "peer1".into(),
+            owner_did: None,
         }
     }
 
--- a/crates/gitlawb-node/src/api/events.rs
+++ b/crates/gitlawb-node/src/api/events.rs
@@ -143,11 +143,11 @@ pub async fn list_ref_updates(
     // only when it matches the canonical owner of the local repo the slug
     // names (#P1); legacy None rows are attributed via an exact unique local
     // match (#P3). Both surfaces share this resolver so they cannot drift.
-    let owner_dids: Vec<serde_json::Value> = futures::future::join_all(
-        updates
-            .iter()
-            .map(|u| state.db.resolve_ref_update_owner_did(&u.repo, u.owner_did.as_deref())),
-    )
+    let owner_dids: Vec<serde_json::Value> = futures::future::join_all(updates.iter().map(|u| {
+        state
+            .db
+            .resolve_ref_update_owner_did(&u.repo, u.owner_did.as_deref())
+    }))
     .await
     .into_iter()
     .map(|r| {
@@ -264,18 +264,19 @@ pub async fn list_repo_events(
     // them.
     let gossip_updates =
         collect_visible_ref_updates(&state.db, Some(&repo_id_str), limit, caller).await?;
-    let gossip_owner_dids: Vec<serde_json::Value> = futures::future::join_all(
-        gossip_updates
-            .iter()
-            .map(|u| state.db.resolve_ref_update_owner_did(&u.repo, u.owner_did.as_deref())),
-    )
-    .await
-    .into_iter()
-    .map(|r| {
-        r.map(|o| o.map_or(serde_json::Value::Null, |s| serde_json::json!(s)))
-            .unwrap_or(serde_json::Value::Null)
-    })
-    .collect();
+    let gossip_owner_dids: Vec<serde_json::Value> =
+        futures::future::join_all(gossip_updates.iter().map(|u| {
+            state
+                .db
+                .resolve_ref_update_owner_did(&u.repo, u.owner_did.as_deref())
+        }))
+        .await
+        .into_iter()
+        .map(|r| {
+            r.map(|o| o.map_or(serde_json::Value::Null, |s| serde_json::json!(s)))
+                .unwrap_or(serde_json::Value::Null)
+        })
+        .collect();
 
     let gossip_events: Vec<serde_json::Value> = gossip_updates
         .iter()
--- a/crates/gitlawb-node/src/api/events.rs
+++ b/crates/gitlawb-node/src/api/events.rs
@@ -153,6 +153,7 @@ pub async fn list_ref_updates(
                 "cert_id":     u.cert_id,
                 "received_at": u.received_at,
                 "from_peer":   u.from_peer,
+                "owner_did":   u.owner_did,
             })
         })
         .collect();
@@ -225,6 +226,7 @@ pub async fn list_repo_events(
                 "pusher_did": c.pusher_did,
                 "node_did":   c.node_did,
                 "timestamp":  c.issued_at,
+                "owner_did":  record.owner_did,
                 "source":     "local",
             })
         })
@@ -258,6 +260,7 @@ pub async fn list_repo_events(
                     "cert_id":     u.cert_id,
                     "received_at": u.received_at,
                     "from_peer":   u.from_peer,
+                    "owner_did":   u.owner_did,
                     "source":      "gossipsub",
                 })
             })
@@ -324,6 +327,7 @@ mod ref_updates_feed_tests {
             new_sha: "a".repeat(40),
             timestamp: Utc::now().to_rfc3339(),
             cert_id: None,
+            owner_did: None,
             received_at: Utc::now().to_rfc3339(),
             from_peer: "peer1".into(),
             owner_did: None,
--- a/crates/gitlawb-node/src/api/events.rs
+++ b/crates/gitlawb-node/src/api/events.rs
@@ -138,9 +138,25 @@ pub async fn list_ref_updates(
     let caller = auth.as_ref().map(|e| e.0 .0.as_str());
     let updates = collect_visible_ref_updates(&state.db, None, limit, caller).await?;
 
+    // Resolve the local owner_did for each row so locally-hosted repos
+    // display their canonical owner rather than a peer-supplied wire value.
+    let deduped = state.db.list_all_repos_deduped().await?;
+    let resolve_local = |slug: &str| -> Option<&str> {
+        for record in &deduped {
+            if crate::visibility::ref_update_row_names_repo(record, slug) {
+                return Some(&record.owner_did);
+            }
+        }
+        None
+    };
+
     let events: Vec<serde_json::Value> = updates
         .iter()
         .map(|u| {
+            let owner_did = match resolve_local(&u.repo) {
+                Some(local) => serde_json::json!(local),
+                None => serde_json::json!(u.owner_did),
+            };
             serde_json::json!({
                 "id":          u.id,
                 "node_did":    u.node_did,
@@ -153,7 +169,7 @@ pub async fn list_ref_updates(
                 "cert_id":     u.cert_id,
                 "received_at": u.received_at,
                 "from_peer":   u.from_peer,
-                "owner_did":   u.owner_did,
+                "owner_did":   owner_did,
             })
         })
         .collect();
@@ -1109,7 +1125,14 @@ mod ref_updates_feed_tests {
             .await
             .unwrap();
         assert_eq!(resp.status(), StatusCode::OK);
-        assert_eq!(count(&body_json(resp).await), 2);
+        let body = body_json(resp).await;
+        assert_eq!(count(&body), 2);
+        for event in body["events"].as_array().unwrap() {
+            assert_eq!(
+                event["owner_did"], OWNER,
+                "each event must carry the local owner_did"
+            );
+        }
     }
 
     // Anon reads a quarantined mirror → 404 (withheld without disclosing existence
--- a/crates/gitlawb-node/src/api/events.rs
+++ b/crates/gitlawb-node/src/api/events.rs
@@ -142,7 +142,14 @@ pub async fn list_ref_updates(
     // record's trailing-segment match). Fall back to the local record's
     // owner_did only for backward-compat rows (stored as None) that name a
     // repo this node hosts, so legacy rows still display an owner.
-    let deduped = state.db.list_all_repos_deduped().await?;
+    // The deduped set is loaded only when a legacy None row is actually
+    // present, avoiding the scan on every global-feed read.
+    let needs_fallback = updates.iter().any(|u| u.owner_did.is_none());
+    let deduped: Vec<crate::db::RepoRecord> = if needs_fallback {
+        state.db.list_all_repos_deduped().await?
+    } else {
+        Vec::new()
+    };
     let resolve_local = |slug: &str| -> Option<&str> {
         for record in &deduped {
             if crate::visibility::ref_update_row_names_repo(record, slug) {
@@ -281,7 +288,7 @@ pub async fn list_repo_events(
                     "cert_id":     u.cert_id,
                     "received_at": u.received_at,
                     "from_peer":   u.from_peer,
-                    "owner_did":   serde_json::json!(record.owner_did),
+                    "owner_did":   u.owner_did.as_ref().map(|s| serde_json::json!(s)).unwrap_or(serde_json::json!(record.owner_did)),
                     "source":      "gossipsub",
                 })
             })
--- a/crates/gitlawb-node/src/api/events.rs
+++ b/crates/gitlawb-node/src/api/events.rs
@@ -138,8 +138,10 @@ pub async fn list_ref_updates(
     let caller = auth.as_ref().map(|e| e.0 .0.as_str());
     let updates = collect_visible_ref_updates(&state.db, None, limit, caller).await?;
 
-    // Resolve the local owner_did for each row so locally-hosted repos
-    // display their canonical owner rather than a peer-supplied wire value.
+    // Prefer the stored wire owner_did (full DID, may differ from local
+    // record's trailing-segment match). Fall back to the local record's
+    // owner_did only for backward-compat rows (stored as None) that name a
+    // repo this node hosts, so legacy rows still display an owner.
     let deduped = state.db.list_all_repos_deduped().await?;
     let resolve_local = |slug: &str| -> Option<&str> {
         for record in &deduped {
@@ -153,9 +155,12 @@ pub async fn list_ref_updates(
     let events: Vec<serde_json::Value> = updates
         .iter()
         .map(|u| {
-            let owner_did = match resolve_local(&u.repo) {
-                Some(local) => serde_json::json!(local),
-                None => serde_json::json!(u.owner_did),
+            let owner_did = match &u.owner_did {
+                Some(wire) => serde_json::json!(wire),
+                None => match resolve_local(&u.repo) {
+                    Some(local) => serde_json::json!(local),
+                    None => serde_json::Value::Null,
+                },
             };
             serde_json::json!({
                 "id":          u.id,
--- a/crates/gitlawb-node/src/api/events.rs
+++ b/crates/gitlawb-node/src/api/events.rs
@@ -138,37 +138,29 @@ pub async fn list_ref_updates(
     let caller = auth.as_ref().map(|e| e.0 .0.as_str());
     let updates = collect_visible_ref_updates(&state.db, None, limit, caller).await?;
 
-    // Prefer the stored wire owner_did (full DID, may differ from local
-    // record's trailing-segment match). Fall back to the local record's
-    // owner_did only for backward-compat rows (stored as None) that name a
-    // repo this node hosts, so legacy rows still display an owner.
-    // The deduped set is loaded only when a legacy None row is actually
-    // present, avoiding the scan on every global-feed read.
-    let needs_fallback = updates.iter().any(|u| u.owner_did.is_none());
-    let deduped: Vec<crate::db::RepoRecord> = if needs_fallback {
-        state.db.list_all_repos_deduped().await?
-    } else {
-        Vec::new()
-    };
-    let resolve_local = |slug: &str| -> Option<&str> {
-        for record in &deduped {
-            if crate::visibility::ref_update_row_names_repo(record, slug) {
-                return Some(&record.owner_did);
-            }
-        }
-        None
-    };
+    // Resolve the trusted display owner_did per row. The stored wire value is
+    // untrusted (arrives over gossip / unsigned peer notify), so it is echoed
+    // only when it matches the canonical owner of the local repo the slug
+    // names (#P1); legacy None rows are attributed via an exact unique local
+    // match (#P3). Both surfaces share this resolver so they cannot drift.
+    let owner_dids: Vec<serde_json::Value> = futures::future::join_all(
+        updates
+            .iter()
+            .map(|u| state.db.resolve_ref_update_owner_did(&u.repo, u.owner_did.as_deref())),
+    )
+    .await
+    .into_iter()
+    .map(|r| {
+        r.map(|o| o.map_or(serde_json::Value::Null, |s| serde_json::json!(s)))
+            .unwrap_or(serde_json::Value::Null)
+    })
+    .collect();
 
     let events: Vec<serde_json::Value> = updates
         .iter()
-        .map(|u| {
-            let owner_did = match &u.owner_did {
-                Some(wire) => serde_json::json!(wire),
-                None => match resolve_local(&u.repo) {
-                    Some(local) => serde_json::json!(local),
-                    None => serde_json::Value::Null,
-                },
-            };
+        .enumerate()
+        .map(|(i, u)| {
+            let owner_did = owner_dids[i].clone();
             serde_json::json!({
                 "id":          u.id,
                 "node_did":    u.node_did,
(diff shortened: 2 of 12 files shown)


beardthelion (dismissed):
The write-side plumbing is right: the full DID comes from the local repo record, it's threaded through both the gossip and HTTP-notify paths, and the `#[serde(default)]` / nullable column keeps older peers compatible. CodeRabbit's feed-serialization ask is genuinely addressed too. Two things block merge, and the first is a hard one: the schema change breaks every existing node on upgrade, and the PR closes #144 without fixing #144's symptom.

## Findings

- [P1] Ship `owner_did` as a new migration version, not appended to v1
  `crates/gitlawb-node/src/db/mod.rs:542`
  The `ALTER TABLE ... ADD COLUMN owner_did` and `idx_ref_updates_owner` were added inside `Migration { version: 1 }`, but `run_migrations` skips any version already recorded in `schema_migrations`, and the catalogue comment right above the array says new schema must be a new version (v2, v3, ...). Every existing node has v1 recorded, so on upgrade the column is never created; `insert_ref_update` (binds `$12`) and all three `list_*` SELECTs then reference a missing column and error, which stops ref-update ingestion on both the gossip and notify paths and 500s the events feed. Fresh databases pass because they run v1 from scratch, which is why CI is green. I reproduced it against the real migration runner: an existing node (v1..v9 recorded, no `owner_did`) upgraded to this build leaves the column absent and the insert/select fail; moving both statements into a new `Migration { version: 10 }` makes the same upgrade round-trip cleanly.

- [P2] Don't close #144 on this PR; nothing consumes `owner_did` yet
  `crates/gitlawb-node/src/api/events.rs:38`
  #144 has two halves: carry the full owner DID (done here) and make the feed gate use it to stop the over-drop. Nothing in this diff reads `owner_did` for a visibility decision; it's stored and echoed in the feed JSON only, and `idx_ref_updates_owner` is an index no query uses. The gate that over-drops isn't in this branch, so merging this auto-closes #144 while the reported symptom persists. Keep the plumbing, but change "Closes #144" to "Refs #144" and land the gate consumption (the did:key-aware match #144 describes) in the follow-up that sits on top of the gate.

- [P3] Populate `owner_did` on the other read surfaces, or hold the JSON change
  `crates/gitlawb-node/src/graphql/query.rs:61`, `crates/gitlawb-node/src/api/events.rs:117`
  The field lands on the global REST feed but the GraphQL `ref_updates` resolver drops it (`RefUpdateType` has no such field) and the repo-scoped `list_repo_events` gossip block omits it. No consumer reads it yet so there's no impact today, but the asymmetry is the same drift the duplicated four-copy SELECT column list invites, and it becomes a trap once something depends on the field. Either populate all three surfaces or defer the JSON exposure until the gate needs it. While here, add a migration upgrade-path test (seed `schema_migrations` at v1, run migrations, assert the column round-trips) so the v1-append class can't recur.

The direction is right and the receive/publish wiring is sound; once the migration moves to its own version this is close.

beardthelion (dismissed):
Re-reviewed on `72ccb28d`. The P1 migration break is resolved and I verified it by execution: the `owner_did` DDL now lives in a new `Migration{version:10}`, and on a simulated existing node (`schema_migrations` at v1..v9, `received_ref_updates` without the column) `migrate()` adds the column and index and records v10. Putting the DDL back inside the already-applied v1 makes that same check fail (the column is never re-added), so the guard is load-bearing. Wire compat (`#[serde(default)] Option<String>`) and the DB plumbing are sound, and "Refs #144" is correct. Four items to fold in.

## Findings

- **[P2] Drop the premature `idx_ref_updates_owner` index, or defer it to the gate PR**
  `crates/gitlawb-node/src/db/mod.rs:833`
  Nothing reads `received_ref_updates` by `owner_did` yet: every reference is a projection, no WHERE/JOIN/ORDER. Migrations run inside a transaction so `CREATE INDEX CONCURRENTLY` isn't available here (mod.rs:426), which means this index builds under a write-blocking lock at startup on exactly the populated nodes this PR targets, for no current benefit. Ship the column now and add the index in #143 next to the query that reads it.

- **[P2] Expose owner_did on the repo-scoped events feed**
  `crates/gitlawb-node/src/api/events.rs:116`
  `list_ref_updates` returns `owner_did` now, but the `gossip_events` block in `list_repo_events` omits it, so `GET /api/v1/repos/{owner}/{repo}/events` silently drops the field. Add `"owner_did": u.owner_did` to that `json!` block to match line 38.

- **[P2] Carry owner_did through the GraphQL query and subscription**
  `crates/gitlawb-node/src/graphql/types.rs:52`
  `RefUpdateType` and `RefUpdateBroadcast` (`state.rs:12`) have no `owner_did`, so the `ref_updates` query (`query.rs:61`) and the subscription drop it while REST exposes it. Add `owner_did: Option<String>` to both structs and populate it at query.rs, subscription.rs, and the broadcast send in repos.rs.

- **[P2] Add an existing-node upgrade test for v10**
  `crates/gitlawb-node/src/db/mod.rs:3200`
  `migration_v10_creates_owner_did_column` runs only a fresh v1..v10 chain, which can't catch DDL appended to an already-applied version. Seed `schema_migrations` with v1..v9 and a `received_ref_updates` without the column, then `migrate()` and assert the column, index, and v10 row appear. I ran exactly this against the head and it passes; the same body fails when the DDL sits back in v1, so it is the missing regression guard, not a live bug.

One forward note for #143, not a blocker here: the wire `owner_did` is attacker-controlled (the peer signature authenticates the relaying node, not this field) and is stored verbatim. Every authorization path today reads the trusted `record.owner_did` from the local repo row, so nothing consumes the wire value yet, but the gate must cross-check it against a trusted source and must not fall back to the lossy last-segment / `did_matches` matcher the full DID exists to avoid.

@Gravirei migration fix verified, thanks. The above are plumbing-consistency and deploy-window items.

beardthelion (dismissed):
Re-reviewed on `63ab6015`. All four items from my last pass are in, CodeRabbit's local-cert-events fix landed, and I verified the security- and deploy-relevant paths by execution:

- Migration, both ways: on a simulated existing node (v1..v9 recorded, no `owner_did` column), `migrate()` adds the column and records v10; moving that DDL back into the already-applied v1 makes `migration_v10_existing_node_upgrade` fail (`RowNotFound`) while the fresh-DB test still passes, so the upgrade guard is load-bearing. A pre-upgrade row (`owner_did` NULL, since the ALTER has no default) round-trips through the feed as JSON null, no 500. The index is correctly deferred to #143.
- Feed provenance, executed: the repo-scoped feed's local block carries the trusted `record.owner_did`, and the gossip block echoes the stored wire value verbatim, which I confirmed with a spoofed `owner_did` surfacing unchanged (display-only, nothing filters on it). All four `received_ref_updates` SELECTs project `owner_did`, so the shared row mapper never hits a missing column. The global feed and GraphQL (query, subscription, broadcast) carry the field through as a passthrough.

- [P3] The stored `received_ref_updates.owner_did` is attacker-controlled (the peer signature authenticates the relaying node, not this field, and `notify_sync` accepts it unsigned by default) and is echoed verbatim, as the spoof test shows. It is display-only and nothing gates on it, so not a break. A short comment at the two ingest sites (`p2p/mod.rs`, `api/peers.rs:398`) marking the column untrusted would keep #143's DID-aware feed gate from consuming it as trusted; that gate must reconcile it against the repo slug or a verified cert.

@Gravirei the P2s and the migration all check out, thanks. @kevincodex1 this is ready.

kevincodex1 (approved):
looks good to me!

jatmn (approved):
@kevincodex1 LGTM

kevincodex1:
hello @Gravirei kindly rebase to main and fix conflicts

beardthelion (dismissed):
Re-reviewed on `b46005bf`. The owner_did work (#144) is solid: `owner_did` is carried on all four emit paths, the shared `received_ref_updates` mapper and migration v10 line up, and the feed gate never trusts the stored value. But the rebase onto 0.5.0 pulled `api/ipfs.rs` backward relative to main and reintroduces an unauthenticated availability regression, so I can't approve this head.

## Findings

- [P1] Restore the pre-walk object-existence check on `GET /ipfs/{cid}`
  `crates/gitlawb-node/src/api/ipfs.rs` (`get_by_cid`)
  Current main (via #133, `466a550`) runs `store::object_type` and `continue`s before the reachability walk, so a random CID can't trigger a full-history walk on repos that don't hold the object. This branch removes that guard and runs `allowed_blob_set_for_caller` (`git rev-list --all` + a per-commit `ls-tree`) before `read_object`. The route is `optional_signature` and unthrottled, and `get_by_cid` iterates the full `list_all_repos()`, so one anonymous request with an absent CID walks every public path-scoped repo's history. I reproduced it: an anon request for an object not in the repo fires the walk on this head and fires zero on main. Restore main's `object_type` pre-filter, or gate the walk behind an indexed proof the object is present.

- [P2] Drop the `/ipfs/{cid}` change from this PR
  `crates/gitlawb-node/src/api/ipfs.rs`
  `d995c4de` is a ~150-line ipfs gate revision unrelated to #144, and main already carries #133. Rebasing onto current main to drop it (keeping main's `ipfs.rs`) removes both the P1 and the scope creep in one move.

- [P2] Correct the docstring claiming owner_did feeds the feed gate
  `crates/gitlawb-node/src/p2p/mod.rs:42`
  The comment says owner_did was added "so the feed gate can distinguish different DID methods that share the same trailing segment," but `visibility::ref_update_row_names_repo` matches on `row_repo` plus the local record and collapses both via `.split(':').next_back()`, never reading the wire owner_did. The field is additive display/storage today. Either wire owner_did into the matcher or fix the doc so it doesn't imply that collision is closed.

- [P3] Exercise the real v9 to v10 upgrade over populated data
  `crates/gitlawb-node/src/db/mod.rs` (`migration_v10_creates_owner_did_column`)
  The test runs the full v1..v10 chain on a fresh pool; nothing seeds `schema_migrations` at v9 with existing `received_ref_updates` rows before applying v10. The migration is safe (`ADD COLUMN IF NOT EXISTS`, nullable, Option readers), so this is coverage rather than a defect. A test that migrates to v9, inserts a row without owner_did, then applies v10 and asserts `owner_did IS NULL` pins the actual upgrade path.

- [P3] Don't render peer-supplied owner_did as authoritative
  `crates/gitlawb-node/src/api/events.rs:261`
  Gossip and notify rows store owner_did verbatim off the wire with no signature binding it to the owner's key, and the feed echoes it. It isn't an authz break (the gate uses the matched local record, never the stored value) and it's the same spoofable class as `pusher_did`/`node_did`, but rendering the display owner from the local record for locally-known rows would close it.

The owner_did core is ready; the ipfs regression is the only merge blocker. @Gravirei recommend rebasing onto current main to drop `d995c4de` and pick up the `#133` ipfs gate.

jatmn (changes requested):
I found a couple of review-feedback/test issues that need to be addressed before this is ready.

## Findings

- [P2] Make the v10 migration test actually start from a v9 schema
  `crates/gitlawb-node/src/db/mod.rs:3320`
  This test says it simulates an existing v9 node, but it first runs `db.migrate()`, which applies v10 and creates `received_ref_updates.owner_did` before the test deletes and reseeds `schema_migrations` at v9. The later insert without `owner_did` therefore runs against a table that already has the new column, so the test would still pass if the upgrade path being claimed here was not actually exercised. Please build the pre-v10 schema without applying migration 10 first, or explicitly drop the new column/index before the reseed, then apply v10 and assert the old populated row survives with `owner_did IS NULL`.

- [P3] Complete CodeRabbit's owner_did feed assertion
  `crates/gitlawb-node/src/test_support.rs:2432`
  CodeRabbit's request to expose `owner_did` in `/api/v1/events/ref-updates` is marked addressed, and the handler now serializes the field, but the end-to-end test still only asserts the returned `repo`. Since the row inserted above has `owner_did: Some(owner.into())`, this test would pass even if the API regressed to omitting the new field again. Please complete that review item by asserting `events[0]["owner_did"]` matches the inserted owner DID.

jatmn (changes requested):
Thanks for the update. I found one remaining issue that needs to be addressed before this is ready.

## Findings

- [P2] Defer the unused owner_did index until the gate query lands
  `crates/gitlawb-node/src/db/mod.rs:834`
  Migration v10 creates `idx_ref_updates_owner`, but this PR only stores and projects `received_ref_updates.owner_did`; no current query filters, joins, or orders that table by `owner_did`. Since migrations run inside a transaction here, this index cannot be built concurrently, so existing nodes with populated ref-update tables take avoidable write-blocking startup work for no current read path. Please ship just the nullable column in this plumbing PR, then add the index in the follow-up that actually consumes `owner_did` for the feed gate.

jatmn (approved):
@kevincodex1 LGTM

beardthelion (dismissed):
Approving on `093a3c0d`. My earlier request-changes was on a stale base; after the rebase, `ipfs.rs` and the feed gate (`visibility.rs`) are byte-identical to main, and the owner_did work verifies clean by execution.

What I checked:
- The wire `owner_did` is display/storage only. I drove a private repo's ref-update with `owner_did` spoofed to an attacker's own DID and read the feed authenticated as that DID: the row stays denied, because the gate matches on the local record, not the wire field. The owner still sees the row.
- The GraphQL subscription broadcast stays `announce`-gated (private repos do not announce), so a private push does not leak ref metadata to anonymous subscribers, and the broadcast `owner_did` is the local record's.
- Migration v10 is a new highest version, idempotent, nullable, no index. The upgrade test genuinely migrates a v9 node with a pre-existing row and asserts `owner_did IS NULL`. A NULL row reads back through the shared mapper as `None` with no panic, and old peers (missing field or explicit null) deserialize to `None`.

## Findings

- [P3] Show the local owner on the global feed for locally-known repos. `crates/gitlawb-node/src/api/events.rs:155`. The cross-repo feed echoes the peer-supplied `owner_did`, so a locally-hosted repo's row can display whatever owner a peer claimed. The per-repo feed already does the right thing at `events.rs:261` (it emits the matched `record.owner_did`). Since `ref_update_row_visible` already resolves the matching local record, use its `owner_did` here too and keep the wire value only for rows with no local match.
- [P3] Assert the nullable invariant in the v10 migration test. `crates/gitlawb-node/src/db/mod.rs`. `migration_v10_creates_owner_did_column` selects `is_nullable` into the tuple and comments "nullable TEXT" but never asserts it; add `assert_eq!(col.2, "YES")` so a future accidental NOT NULL is caught.
- [P3] Add an `owner_did` echo assertion on the GraphQL `refUpdates` surface (and on `list_repo_events`). The REST cross-repo feed has one; the others do not, so a dropped or nulled field there would pass CI.

Core is sound and this is display/test polish only. @kevincodex1 LGTM.

jatmn (changes requested):
I found one issue to address.

## Findings

- [P2] Do not resolve `owner_did` from the lossy repo slug alone
  `crates/gitlawb-node/src/api/events.rs:145`
  The global REST feed overrides a row's stored `owner_did` whenever `ref_update_row_names_repo(record, u.repo)` matches a local repo, but that matcher intentionally collapses both owners to the trailing DID segment. If this node hosts a readable `did:web:host:alice/widget` and receives a remote row for `did:gitlawb:alice/widget` with the legacy `repo` slug `alice/widget`, the visibility gate keeps the row because the local repo is public, then the response rewrites its `owner_did` to `did:web:host:alice`. That loses the full owner DID this PR is adding for same-slug disambiguation. GraphQL has the opposite drift (`query.rs:86` still returns the stored wire value directly), so the two global feeds can disagree for the same row. Please include the new `owner_did` in the local-record resolution before replacing it with `record.owner_did`, and share or mirror that projection in GraphQL with an assertion for the exact `ownerDid` value.

jatmn (approved):
Thanks for the update. I rechecked the previously discussed `owner_did`, migration, REST, and GraphQL paths and do not see any remaining actionable code issues from my side.

GitHub still reports this PR as blocked even though it is mergeable and the visible checks are passing, so there may be a remaining branch-protection or review-state item for maintainers to clear.


@kevincodex1 LGTM

beardthelion (dismissed):
Re-reviewed the new head 1074aa4d; refreshing my earlier approval, which sat on 093a3c0d.

The change since then resolves the owner_did display divergence: both the REST feed and GraphQL now prefer the stored wire owner_did and fall back to a local lookup only for legacy rows with no stored owner, so the two surfaces agree. I verified the security-relevant behavior by execution against a local Postgres, not by reading:

- The wire owner_did is display-only, never a visibility input. A ref-update row carrying a spoofed owner_did is dropped for anonymous callers and for a caller authenticated as the spoofed DID; only the real owner sees it, with the spoofable value merely displayed. Reverting the gate's fail-closed arm flips that to a leak, so the gate is load-bearing.
- The local fallback cannot surface a private owner. A slug that collides on the loose owner matcher fails closed when the collision partner is a private repo the caller cannot read (verified the cross-method did:gitlawb vs did:key case drops the row for anon), and where the collision is the same key in two DID spellings the dedup resolves to the owner's public repo, so the displayed owner is always one the caller can already list.
- The migration nullability assertion from last round is in.

A few non-blocking follow-ups, none of which need to hold this:

- The feed handlers load list_all_repos_deduped() twice per request (once in the collector, once for the owner_did projection) on both REST and GraphQL. Threading the collector's set through would drop the redundant query on the hot feed path.
- The fallback resolves by first match over that query, whose outer ORDER BY has no id tiebreaker, and REST and GraphQL query it separately. For legacy no-owner rows under a tied updated_at plus a name collision the two surfaces can still pick different owners. A stable tiebreaker (and sharing the set, per the point above) closes it.
- A peer's empty or malformed wire owner_did is shown verbatim, since the fallback only applies to a missing value. Treating an empty string as missing would recover the local owner.

@kevincodex1 good to merge.

kevincodex1:
hello @Gravirei please rebase to main and fix conflicts

beardthelion (approved):
Re-reviewed on dbfb0a1. Approving. I verified the security-critical properties by execution on this head.

The wire owner_did is display-only and the gate does not trust it: I seeded a private local repo and a received_ref_updates row naming it with a forged owner_did, drove the global feed as anon, and the private row stays dropped. A public repo's row with the same forged owner_did is shown with the forged value echoed. So a peer cannot use owner_did to unlock a private row, only to mislabel a public one on the feed. The shared row mapper reads owner_did as Option, so a backward-compat NULL row round-trips as None with no panic (insert_and_list_without_owner_did covers it). The v10 migration is a new highest-version block, nullable and idempotent, and its upgrade-path test genuinely starts from a v9 schema. All four emit paths carry owner_did, and I confirmed the HTTP-notify path stores it end-to-end.

Two non-blocking notes:

- The feed echoes the wire owner_did verbatim even on rows that name a local repo this node hosts, where the true owner is already known from the record. It is display-only and joins the feed's existing untrusted peer fields (pusher_did, node_did, from_peer), so it is not a gate issue, but for local-matching rows preferring record.owner_did would avoid displaying a peer's claim over a fact the node holds. This reflects the prefer-wire resolution from an earlier round, so treat it as a judgment call rather than a required change.
- Thin emit-path coverage: the notify_sync tests do not assert owner_did is stored, so a revert of the binding would pass all of them, and the REST None-fallback branch and the subscription projection are untested. The behavior is correct, but a regression in those paths would not be caught.

@kevincodex1 good to merge.

kevincodex1:
hello @Gravirei please rebase to main and fix conflicts

jatmn (approved):
LGTM

kevincodex1:
hello @Gravirei please rebase to main again and fix conflicts

beardthelion (dismissed):
The `owner_did` wire plumbing is sound: it's display-only (projected after the visibility gate, which keys on the local repo record, never the peer-supplied wire value), the migration is a correct new v10 with a load-bearing upgrade test, and all four fan-out paths carry the field. Two things still gate merge.

The branch conflicts with `main`, and its base predates #149 (`list_ref_certificates` bound + upsert) and #180 (per-IP creation rate-limit), both of which touch the same files this PR does (`events.rs`, `repos.rs`, `db/mod.rs`, `state.rs`, `test_support.rs`). Please rebase onto current main and fix the conflicts; the resolved diff needs another pass, because a careless resolution on those files silently drops one of those merged gates.

## Findings

- [P2] Drop the per-request full-repo dedup on the ref-updates feed
  `crates/gitlawb-node/src/api/events.rs:144`, `crates/gitlawb-node/src/graphql/query.rs:80`
  `list_ref_updates` and the `refUpdates` resolver call `list_all_repos_deduped()` for the display fallback, but `collect_visible_ref_updates` already loaded that exact set, so the full-`repos`-table dedup CTE runs twice per request. The second load is also unconditional: it fires even when every returned row already carries a wire `owner_did` and the fallback is never consulted, which is the steady state once peers upgrade. The endpoint is anonymous, so a permissionless caller drives two full-table scans per hit, scaling with node repo count rather than the (<=200) rows returned. Thread the set the collector already holds through to the handler, and gate the fallback load on `updates.iter().any(|u| u.owner_did.is_none())`.

- [P3] Resolve the display owner with an exact key match, not the fail-safe gate matcher
  `crates/gitlawb-node/src/api/events.rs:147`, `crates/gitlawb-node/src/graphql/query.rs:85`
  The fallback picks the first record matching `ref_update_row_names_repo`, a deliberately loose prefix matcher built for the visibility gate (where over-matching fails safe). For display it mis-attributes: a legacy row naming owner `zX` is shown with the owner of any local repo whose key prefixes `zX`. This is cosmetic only (the private-match case is withheld by the gate rather than disclosed) and affects legacy pre-#144 rows only. Resolve with the canonical `normalize_owner_key` and require a unique match, skipping on ambiguity.

beardthelion (changes requested):
Re-reviewed the current head. The owner_did wire plumbing is sound, and I re-verified the security-critical property by execution on this head rather than reading it: the wire owner_did is display-only and never a visibility-gate input. Driving a forged wire owner_did that names a private repo through the resolver, the row stays dropped for anon and for a caller authenticated as the forged DID, and only the true owner sees it; forcing the gate open flips it to a leak, so the gate is load-bearing. The rebase is clean, #149's `list_ref_certificates` bound and #180's per-IP creation limiter both survive (verified by their tests). Migration is a correct new v11 (nullable, idempotent) with a load-bearing upgrade test, all four emit paths carry owner_did, and the shared mapper round-trips a null owner_did with no panic.

One item from my last pass is still open, plus two small ones.

## Findings

- [P2] Drop the redundant per-request repo dedup on the ref-updates feed. `crates/gitlawb-node/src/api/events.rs:145`, `crates/gitlawb-node/src/graphql/query.rs:80` — `collect_visible_ref_updates` already loads `list_all_repos_deduped()` for the gate (`events.rs:57`), but both handlers load it a second time for the owner_did display fallback, so every anonymous request now runs the full-table dedup CTE twice (one before this PR). The second load is unconditional: it fires even when every returned row already carries a wire owner_did and the fallback is never consulted, which is the steady state once peers upgrade. Thread the set the collector already holds through to the handler, or gate the second load on `updates.iter().any(|u| u.owner_did.is_none())`.

- [P3] Resolve the display owner with an exact key match, not the fail-safe gate matcher. `crates/gitlawb-node/src/visibility.rs:229` — the legacy-None fallback reuses `ref_update_row_names_repo`, whose `record_key.starts_with(row_key)` is always true when an empty-owner-segment slug (`/name`) reduces `row_key` to `""`, so a legacy row can display the wrong owner. Display-only: the same over-match in the gate only ever drops (covered by `feed_empty_owner_slug_matches_and_drops`), and it affects only pre-#144 rows stored with `owner_did = None`. Resolve with `normalize_owner_key` and require a unique match, skipping on ambiguity.

- [P3] Add a regression test for a forged wire owner_did. The existing gate tests forge the slug or set `owner_did` to `None`; none drives a forged wire `owner_did` field, which is the exact must-not (a peer-supplied owner_did must not unlock a private row). The property holds today, but a permanent test guards a future change that starts keying on the wire value. A drop-in that mirrors the existing resolver tests:

  ```rust
  #[sqlx::test]
  async fn forged_wire_owner_did_never_unlocks_private_row(pool: PgPool) {
      let db = db(pool).await;
      db.create_repo(&repo("r1", OWNER, "widget", false)).await.unwrap(); // private
      let mut row = ref_row("u1", "z6MkOwner/widget");
      row.owner_did = Some("did:key:z6MkAttacker".into()); // forged wire owner
      db.insert_ref_update(&row).await.unwrap();
      let schema = schema(db);
      let q = r#"{ refUpdates(repo: "z6MkOwner/widget") { refName ownerDid } }"#;
      assert_eq!(count(&anon(&schema, q).await), 0);
      assert_eq!(count(&authed(&schema, q, "did:key:z6MkAttacker").await), 0);
      assert_eq!(count(&authed(&schema, q, OWNER).await), 1);
  }
  ```

  I ran this against the branch (passes) and against a fail-open gate (fails), so it is load-bearing.

The core is solid and the security properties hold. The P2 is the one I'd want fixed before merge, and it's a small change; the two P3s are polish.

jatmn (changes requested):

## Findings

- [P1] Restore the iCaptcha PoW protocol support before merging
  `crates/icaptcha-client/src/lib.rs:194`
  This head is behind the current PR base and removes the `Challenge.pow` field, its solver, and `powNonce` from the `/v1/answer` request (as well as deleting `pow.rs`).  A PoW-enabled iCaptcha service therefore has its required `pow` extension silently ignored and receives only `{token, answer}`, so it rejects the solve and `gl` cannot mint a proof for register/repository-create/fork writes.  Rebase onto current `main` and preserve the optional PoW parsing, solving, nonce submission, and regression coverage.

- [P1] Keep iCaptcha key trust scoped to the full origin
  `crates/icaptcha-client/src/lib.rs:110`
  `x-icaptcha-url` is controlled by the node, but the new `host_of` comparison accepts any HTTPS port on the configured operator host and marks it `key_trusted`.  For example, an operator configured with `https://icap.example:443` can be redirected to `https://icap.example:8443`; both challenge and answer requests then send `GITLAWB_ICAPTCHA_API_KEY` to that distinct listener.  The base implementation compared scheme, host, and effective port specifically to prevent this.  Restore full-origin matching for both acceptance and bearer-key attachment.

- [P2] Return the received event's owner DID on the repo-scoped feed
  `crates/gitlawb-node/src/api/events.rs:284`
  The new field is stored on `ReceivedRefUpdate`, and the global REST and GraphQL feeds prefer that wire value, but this projection unconditionally substitutes `record.owner_did`.  A received row whose slug collides with a hosted repo is consequently reported as owned by the local record rather than by its stored full DID—the exact identity ambiguity this change is intended to carry.  Project `u.owner_did` when present and use the local value only as the legacy-`None` fallback, with a collision regression test.

- [P2] Do not perform a second full repository-dedup scan for every feed read
  `crates/gitlawb-node/src/api/events.rs:145`
  `collect_visible_ref_updates` already loads `list_all_repos_deduped()` for its visibility gate, but the new REST projection immediately loads it again for the legacy fallback; the GraphQL resolver repeats the same pattern.  This doubles the full-repository scan on every anonymous global-feed request even when all bounded result rows already contain `owner_did`.  Thread the collector's deduped records through, or load them only if a returned legacy row actually needs fallback resolution.

jatmn (changes requested):

## Findings

- [P1] Do not persist peer-supplied `owner_did` as trusted event ownership
  `crates/gitlawb-node/src/api/peers.rs:398`
  The HTTP notify path copies `req.owner_did` straight into `ReceivedRefUpdate`, and the gossipsub receive path does the same with `event.owner_did`. The new REST and GraphQL projections then prefer that stored wire value whenever it is present. In deployments that still accept unsigned peer notify, any caller can set `node_did` to a known peer and make a visible row for `zAlice/widget` report `owner_did: "did:key:zVictim"`; even with signatures enforced, a compromised or buggy known peer can poison persisted event ownership. Because this field is presented as the full owner DID and is intended to become DID-aware feed-gate input, please either validate/bind it to the row's repo owner before storing/exposing it or treat it as untrusted and derive display ownership from trusted local state when possible.

- [P3] Resolve legacy fallback owners with an exact unique match
  `crates/gitlawb-node/src/api/events.rs:153`
  The `owner_did = None` fallback reuses `ref_update_row_names_repo`, which is deliberately loose for the fail-closed visibility gate: it strips both owners to their last colon segment and accepts prefix matches. That is safe for dropping rows, but it is not safe for display attribution. For example, a legacy visible row such as `/widget` or `alice/widget` can match the first local readable `widget` repo whose owner key shares that empty/short/trailing segment, so REST and GraphQL can synthesize the wrong full owner DID for the exact cross-method collision class this PR is trying to disambiguate. Please use an exact/normalized, unique local match for legacy display fallback, or leave `owner_did` null when the owner cannot be proven.

jatmn (changes requested):

## Findings

- [P2] Batch the owner-DID resolution instead of issuing a query per feed row
  `crates/gitlawb-node/src/api/events.rs:146`
  A public `GET /api/v1/events/ref-updates?limit=200` now starts up to 200 `get_repo` queries through `join_all` after the collector has already fetched the repository/rule data; GraphQL performs the same lookups serially. The node's pool defaults to 20 connections, so a few ordinary max-page feed requests can queue or time out the pool and starve other API work. Resolve the bounded page in one batch (or reuse a resolver map built from the collector's repository snapshot) rather than querying once per event.

- [P2] Do not turn owner-resolution database failures into successful incomplete feeds
  `crates/gitlawb-node/src/api/events.rs:153`
  Both REST projections coerce every `resolve_ref_update_owner_did` error to `null`, so a pool timeout or database failure after the visibility collector succeeds returns HTTP 200 with missing ownership that clients cannot distinguish from an unknown owner. The GraphQL projection propagates the identical resolver error, and the surrounding REST code explicitly treats database errors as errors. Propagate this failure (or return an explicit per-row error state) instead of silently emitting partial data.

- [P2] Preserve a full DID when the matching repository is a mirror-only row
  `crates/gitlawb-node/src/db/mod.rs:2380`
  `upsert_mirror_repo` stores a mirror's `owner_did` as the bare normalized key, and `get_repo` intentionally returns that mirror when no canonical record exists. A received event carrying `owner_did: "did:key:z..."` then passes the normalized comparison but the resolver returns the mirror's bare `"z..."`, defeating this PR's full-DID wire/API contract for mirror-only repos. Return the validated full wire DID in this case, or preserve a canonical full DID for mirror records, and cover the mirror-only path.

beardthelion (approved):
Approving on a05f61d3. I re-verified the security-critical properties by execution on this head, not by reading.

jatmn's P1 (peer-supplied `owner_did` persisted as trusted) is resolved: the display resolver echoes a wire `owner_did` only when it normalizes to the local repo's canonical owner, so a forged value drops to null. I drove the real resolver across a forged and method-spoof matrix (`did:key:zVictim`, `did:web:evil.com:zAlice`, and the mirror variants); all dropped. I confirmed the guard is load-bearing by removing the normalize check and watching the forged values leak, and the committed `events_drop_forged_peer_owner_did` test fails when the binding is removed. End to end through `list_ref_updates`, a stranger who forges ownership of a private repo still sees `count=0` with no `new_sha` leak, the owner sees the row, and the same row on a public repo is visible, so the empty result is real gating rather than a vacuous feed. `normalize_owner_key` only collapses `did:key:X` and bare `X`, so no cross-identity or cross-method spoof is possible, and the visibility gate keys on the local repo record and is unchanged by this PR.

The rest of the round is addressed and I verified each: batched owner-DID resolution (one query per distinct local repo), error propagation (a DB failure inside the resolver now surfaces as 500, not a 200 with null ownership), full-DID preservation for mirror-only repos, exact-match legacy fallback, and the iCaptcha PoW and key-origin restoration (`pow.rs` and `lib.rs` are byte-identical to current main). The new `owner_did` column is a v11 migration with an upgrade-path test, it is projected in the shared `received_ref_updates` mapper, and the feed keyset cursor stays internal (no client-facing cursor). CI is green.

jatmn (changes requested):

## Findings

- [P2] Rebase this stale branch onto current `main` before merge
  `crates/gl/src/http.rs:189`
  The head is still based at `3ec8cbf`, before the merged iCaptcha work in #181 and its client retry tests in #168.  As a result, the PR's base-to-head diff includes that unrelated already-merged work and, when compared to current `main`, removes the entire `gl` HTTP/iCaptcha test module.  Git's synthetic merge happens to restore those tests, but that means the tree being merged is not the head CI and review are evaluating.  This also leaves the repeated maintainer rebase request unresolved.  Rebase onto current `main`, preserve the current iCaptcha changes/tests, and rerun the checks so the reviewable diff is only the owner-DID work.

- [P2] Resolve display owners without serial per-slug database queries
  `crates/gitlawb-node/src/db/mod.rs:2403`
  `resolve_ref_update_owner_dids` is invoked by the anonymous REST and GraphQL feeds after they can collect up to 200 rows, but it awaits `get_repo` once for every distinct received slug.  Received notify/gossip rows can contain 200 different remote slugs, including slugs with no local repository, so one public feed request now adds 200 serialized SQL round trips after the collector has already loaded the local repository/rule data.  Repeated reads can therefore turn peer-controlled feed contents into avoidable connection-pool pressure and latency.  Resolve the distinct keys in one set-based query (or reuse/filter the collector's already-loaded local records) instead of issuing one awaited lookup per key.

jatmn (approved):
@beardthelion LGTM

beardthelion (approved):
Refreshing my approval on the current head. My last one sat on `a05f61d3`; the delta since then is exactly your two P2s, and both are clean.

The rebase is real, not just a synthetic merge: `merge-base(083540b5, main)` is current main and the three-dot diff no longer carries the iCaptcha/#181/#168 work, so the reviewable tree is only the owner-DID change. I checked the rebase didn't drop a gate that main added after the old base, the canonical-over-mirror (#124), quarantine-drop (#143), and repo-scoped read gating (#157) fixes are all present and untouched on the head.

The N+1 fix is sound. `resolve_ref_update_owner_dids` now resolves every distinct slug in one batched query instead of a `get_repo` per key. I verified it's equivalent to the loop it replaced: same `WHERE (OWNER_KEY_CASE) = $p AND name = $q` and the same `ORDER BY [canonical-over-mirror], created_at ASC, id ASC`, with first-wins `or_insert`, so each key resolves to the exact row `get_repo`'s `LIMIT 1` returned, canonical preferred over mirror preserved. The SQL is injection-safe (the CASE fragment is a const, every value is bound), the disjuncts hit `idx_repos_owner_key_name` (the CASE is byte-identical to the index expression), and the page is clamped to 200 so the OR-list is bounded. The re-key fails safe: any Rust/SQL normalization drift yields row-not-found → null owner, never a wrong owner.

The security-critical property is intact and I re-verified it by execution on this head, not by reading. Step 3, the forged-owner drop, is byte-identical to my prior approved head, and `events_drop_forged_peer_owner_did` is load-bearing: making the mismatch arm echo the wire owner turns it red, restoring returns it green. The full `gitlawb-node` suite is 501 passed / 0 failed, including the whole feed visibility matrix and the mirror-doesn't-unlock-private-canonical guard. Migration is v11 (new highest block, upgrade-path test green), `owner_did` is projected in the shared feed SELECT, and the keyset cursor stays internal.

kevincodex1 (approved):
LGTM

PR #134, opened 2026-06-30
fix(node): gate /ipfs/pins and /arweave/anchors behind authentication (#121)

Gravirei:
Closes #121 (anonymous-denial half; see "Out of scope" below for the residual class).

## Summary

Two node-wide metadata endpoints accepted unsigned requests, and the `?repo=` variant on the Arweave anchors endpoint also accepted signed non-readers. This PR closes the anonymous-denial path on both endpoints and the authorization hole on the scoped anchor path. A follow-up commit also fixes a false-empty-page failure mode on the scoped anchor path and tightens the regression that pins the fix.

### Changes

**`/api/v1/ipfs/pins` (`list_pins`):**
- Moved `ipfs_pins_routes` onto its own `Router` and applied `optional_signature` to that router before the merge with `ipfs_cid_routes`. The previous shape layered `optional_signature` on the post-merge router, which (per the `axum-layer-vs-merge-pitfall` memory entry) only covers routes added before the `.layer()` call and silently leaves merged routes outside the layer.
- Handler rejects requests without a verified `AuthenticatedDid` with `401`.

**`/api/v1/arweave/anchors` (`list_anchors`):**
- Round 1: added the `optional_signature` middleware to the route — it was missing entirely before. When `?repo=<owner>/<name>` is provided, parse the slug and call the canonical `authorize_repo_read` helper before the SQL query. Missing repos, quarantined mirrors, and signed non-readers all collapse to the standard `404` (`repo_not_found`).
- Without `?repo=`: auth-only. The global anchor listing stays as auth-only — the wider visibility filtering is the `#136` stale-index class (see below).
- Clamped the limit to `[0, 200]` before the SQL query so `?limit=-1` (or any negative) doesn't reach Postgres as `LIMIT -1`.
- Round 2: built the SQL filter from the canonical stored slug (`{normalize_owner_key(record.owner_did)}/{record.name}`) instead of the user-supplied `?repo=` string. Authz accepts the full `did:key:…/name` form, but a literal `WHERE repo = $1` against the full DID would match zero rows and return a false empty page. The fix keys the SQL on the `RepoRecord` the authz helper resolved, which is the only place the slug the writer actually stores is in scope.

### Out of scope (recorded, not addressed here)

[...]

PR #146, opened 2026-07-02
fix(gl): sign the CLI's /ipfs/pins reads under the #134 auth gate

beardthelion:
PR #134 gates `/api/v1/ipfs/pins` behind authentication (anonymous callers get 401). Two `gl` CLI call sites still hit it unsigned, so once #134 ships they break: `gl ipfs list` returns a hard 401, and the `gl node status` pins panel silently shows nothing. This signs both reads with the caller's identity through the existing `NodeClient::get_signed`, so the CLI keeps working under the gate.

Scope is pins-only:

- `gl ipfs list` now loads the identity (via `--dir`, defaulting to the standard keystore) and signs the request. With no identity it fails with the existing `gl identity new` guidance instead of a raw 401, and it now surfaces a non-2xx response as an error rather than parsing it into an empty list.
- `gl node status` loads the identity gracefully and signs only the pins panel, via an injectable `fetch_pins` helper with four states (pins, empty, unavailable, sign-in-required). The peers, repos, p2p, and events panels stay anonymous, and a pins failure never aborts the dashboard.

`/api/v1/arweave/anchors` (the other endpoint #134 gates) is never called by the CLI, so it is out of scope.

A note on behavior: the pin listing is visibility-filtered (every repo the caller can read, public plus their own private), not scoped to the caller's own objects. The auth requirement exists to stop anonymous enumeration of the node-wide pin index, so the CLI has to authenticate, but the result is the caller's readable view.

This should land in the same release as #134 so no published build ships the broken CLI.

Tests are mockito-based (no database): signed-header presence, the no-identity and non-2xx error paths, and all four node-status pins states including the transport-error and malformed-body branches. The `gl` suite passes (236 tests).

Closes the CLI regression noted in the #134 review.

## Summary by CodeRabbit

* **New Features**
  * `gl node status` now supports an optional `--dir` selector to use a specific identity when checking node status.
  * The Pins section now shows clearer states, including pinned content, empty results, unavailable, and identity-related errors.

* **Bug Fixes**
  * Improved handling of signed pin requests and fallback behavior when no identity is available.
  * Better reporting for empty, rejected, malformed, and network-error responses in the Pins view.

### DIFF
--- a/crates/gl/src/ipfs_cmd.rs
+++ b/crates/gl/src/ipfs_cmd.rs
@@ -49,9 +49,13 @@ async fn cmd_list(node: String, dir: Option<PathBuf>) -> Result<()> {
     // error (it already names `gl identity new`) rather than a bare 401.
     let keypair = crate::identity::load_keypair_from_dir(dir.as_deref())?;
     let client = NodeClient::new(&node, Some(keypair));
-    let resp: Value = client
-        .get_signed("/api/v1/ipfs/pins")
-        .await?
+    let resp = client.get_signed("/api/v1/ipfs/pins").await?;
+    if !resp.status().is_success() {
+        let status = resp.status();
+        let body = resp.text().await.unwrap_or_default();
+        anyhow::bail!("node returned {status} for pins listing: {body}");
+    }
+    let resp: Value = resp
         .json()
         .await
         .context("failed to parse pins response")?;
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P2] Let `gl node status` use the same identity directory as the signed pins list
  `crates/gl/src/node.rs:22`
  This PR adds `--dir` to `gl ipfs list` and signs that request with the selected identity, but the other pins caller still has no identity-directory option: `gl node status --help` only exposes `--node`, and the implementation always calls `load_keypair_from_dir(None)`. Users who created or selected an identity with `--dir` can make `gl ipfs list --dir ...` work, but the status dashboard will still render the pins panel as “sign in to view” because it cannot load that same key. Please add a `dir` option to `NodeCmd::Status` and pass it into the pins fetch path so both CLI callers can authenticate consistently.

beardthelion:
Addressed in cf588d0. `gl node status` now takes `--dir` and threads it into the signed pins fetch, so it selects the same identity directory as `gl ipfs list` (both resolve through `load_keypair_from_dir(dir.as_deref())`). An explicit `--dir` that fails to load renders `IPFS pins: no usable identity in <dir>` and issues no request, instead of the misleading `sign in to view`; a missing default keystore (no `--dir`) still degrades quietly, and a pins failure never aborts the dashboard.

f882ee9 adds a follow-up test asserting the pins read signs with the `--dir` identity rather than the default keystore, plus a guard that an absent `--dir` never resolves to the unusable-dir state.

jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #148, opened 2026-07-03
fix(node): make creation-route rate limit configurable; fix(gl): surface real errors on pr create failure

carlitos973:
Two small, independent fixes found while building an autonomous agent on gitlawb:

## 1. Configurable creation-route rate limit (gitlawb-node)

The rate limiter was hardcoded to 10 requests/hour per DID, shared across the whole creation-routes group (repo create, issue create, PR create, register, fork):

```rust
let rate_limiter = rate_limit::RateLimiter::new(10, std::time::Duration::from_secs(3600));
```

This is workable for a human clicking around, but a real agent/bot doing modest volume (or retrying after a transient error) hits the wall fast, with no way to raise it short of recompiling. We hit this directly running an ACP-hired worker agent — it also caused a real job to expire mid-SLA while rate-limited.

This PR adds two config fields following the existing clap env-var pattern already used throughout `Config`:

- `GITLAWB_RATE_LIMIT_MAX` (default 10, unchanged)
- `GITLAWB_RATE_LIMIT_WINDOW_SECS` (default 3600, unchanged)

`main.rs` now builds the limiter from config instead of the hardcoded values. No behavior changes unless a node operator explicitly opts in.

## 2. `gl pr create` masks the real error on failure

```rust
let pr: Value = resp.json().await.context("invalid JSON")?;
if !status.is_success() {
    let msg = pr["message"].as_str().unwrap_or("unknown error");
    anyhow::bail!("create PR failed ({status}): {msg}");
}
```

Calling `.json()` before checking `status` means any non-2xx response with an empty or non-JSON body (e.g. a 404 from a malformed URL) crashes with a generic "invalid JSON" error instead of the real one. Fixed by reading the raw response body first and including it in the error message when the request fails.

## Testing

Both changes verified with `cargo check -p gitlawb-node -p gl` — compiles clean, no warnings.

## Summary by CodeRabbit

* **New Features**
  * Added configurable request rate limits with defaults, allowing creation throttling to be adjusted via environment settings.

* **Bug Fixes**
  * Improved create-request error messages so failures now include the HTTP status and full server response.
  * Preserved successful response handling while making error reporting clearer.

### DIFF
--- a/crates/gitlawb-node/src/config.rs
+++ b/crates/gitlawb-node/src/config.rs
@@ -152,6 +152,16 @@ pub struct Config {
     /// in flight and exits. Default: 30s.
     #[arg(long, env = "GITLAWB_SHUTDOWN_GRACE_SECS", default_value_t = 30)]
     pub shutdown_grace_secs: u64,
+
+    /// Maximum number of creation requests (repo create, issue create, PR
+    /// create, register, fork) allowed per DID within the rate-limit window.
+    #[arg(long, env = "GITLAWB_RATE_LIMIT_MAX", default_value_t = 10)]
+    pub rate_limit_max_requests: u32,
+
+    /// Rate-limit window size, in seconds, over which
+    /// `rate_limit_max_requests` applies.
+    #[arg(long, env = "GITLAWB_RATE_LIMIT_WINDOW_SECS", default_value_t = 3600)]
+    pub rate_limit_window_secs: u64,
 }
 
 impl Config {
--- a/crates/gitlawb-node/src/main.rs
+++ b/crates/gitlawb-node/src/main.rs
@@ -190,7 +190,10 @@ async fn main() -> Result<()> {
     let repo_store =
         git::repo_store::RepoStore::new(config.repos_dir.clone(), tigris, db.pool().clone());
 
-    let rate_limiter = rate_limit::RateLimiter::new(10, std::time::Duration::from_secs(3600));
+    let rate_limiter = rate_limit::RateLimiter::new(
+        config.rate_limit_max_requests as usize,
+        std::time::Duration::from_secs(config.rate_limit_window_secs),
+    );
 
     // Initialize the iCaptcha proof gate (inert unless ICAPTCHA_MODE is set).
     icaptcha::init().await;
--- a/crates/gl/src/pr.rs
+++ b/crates/gl/src/pr.rs
@@ -234,12 +234,11 @@ async fn cmd_create(
         .await
         .context("failed to connect to node")?;
     let status = resp.status();
-    let pr: Value = resp.json().await.context("invalid JSON")?;
-
+    let raw_text = resp.text().await.context("failed to read response body")?;
     if !status.is_success() {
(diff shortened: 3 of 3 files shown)


jatmn (changes requested):
Thanks for the update. I found one issue to clean up before this is ready.

## Findings

- [P3] Update the PR description to match the final migration
  `PR description`
  The current description still says migration v10 adds both the `owner_did TEXT` column and the `idx_ref_updates_owner` index, but the final patch intentionally dropped that index and the migration now only adds the nullable column. Since the index was deferred to the follow-up gate work, please update the PR body so reviewers and operators do not come away thinking this branch already ships that index.

beardthelion (changes requested):
The core change is sound and behavior-preserving at the defaults: I confirmed `Config`'s defaults (10/3600) reproduce the old hardcoded limiter (10 allowed, 11th blocked), and that an unsigned creation POST still 401s at `require_signature` before the limiter runs, so the `did == None` branch is unreachable on these routes. The key is the verified DID, not a forgeable header, so there is no throttle bypass here. A few things to tighten before merge, highest first.

## Findings

- **[P2] Reject 0 for both new rate-limit knobs** `crates/gitlawb-node/src/config.rs:155-164`
  Both accept 0 via env/CLI. `GITLAWB_RATE_LIMIT_WINDOW_SECS=0` silently disables the limiter (every timestamp is pruned, so `check` always passes) and `GITLAWB_RATE_LIMIT_MAX=0` blocks all creation (`len() >= 0` is always true). Both are execution-confirmed. The window case is the worse one: it fails open on a DoS brake with no signal to the operator, and `0` reads as "unlimited" to most people. Add `value_parser = clap::value_parser!(u32).range(1..)` (and the `u64` equivalent on the window) so a bad value fails fast at startup instead of at runtime.

- **[P3] Sanitize and bound the raw server body before printing it** `crates/gl/src/pr.rs:239`
  The new error path embeds the entire response body verbatim. A hostile or compromised node can return ANSI/terminal control sequences (confirmed: an ESC byte passes through unsanitized) or a very large body, both of which land directly in the caller's terminal. An empty body also yields a dangling `create PR failed (404): `. Strip control characters, cap the length, and substitute a placeholder when the body is empty.

- **[P3] Derive retry-after from the configured window** `crates/gitlawb-node/src/rate_limit.rs:81`
  The 429 sends a hardcoded `retry-after: 60` while the window is now operator-configurable (confirmed still `60` with a 7200s window). A client that honors Retry-After will retry into another 429 until the real window elapses. Compute the header from the window, or the caller's remaining time.

- **[P3] Add a regression test for a non-2xx non-JSON body in `cmd_create`** `crates/gl/src/pr.rs`
  The point of the change is to stop masking non-JSON error bodies as "invalid JSON", but the only failure test uses a JSON body and passes under both the old and new code, so it does not lock the behavior in. Add a `with_status(502)` case with a non-JSON body asserting the raw text surfaces.

- **[Process] Rebase onto current main before merge**
  The base is 11 commits behind and overlaps `main.rs`; confirm the `RateLimiter::new` wiring survives the rebase.

PR #149, opened 2026-07-03
fix(node): bound list_ref_certificates with LIMIT and add upsert to prevent unbounded growth (#147)

### LINKED ISSUE #147: list_ref_certificates fetches a repo's entire cert set with no LIMIT (permissionless read amplification on public repos)
## Summary

`list_ref_certificates` (`crates/gitlawb-node/src/db/mod.rs:1846`) runs `SELECT ... FROM ref_certificates WHERE repo_id = $1 ORDER BY issued_at DESC` with `fetch_all` and no `LIMIT`, loading every certificate row for a repo into memory. The table grows one row per ref per push (`api/repos.rs:919` loops `issue_ref_certificate` over every advanced ref; `cert.rs:44` mints a fresh UUID per call), with no upsert/dedup and no prune or retention anywhere, so it accumulates permanently. An anonymous caller reading a public repo turns a single cheap GET into an unbounded fetch, allocation, and response body.

This is an availability/cost problem, separate from #120 (which adds the missing visibility gate to these handlers) and #114 (which bounded only the gossip half of the events feed). #120's fix does not help here: for a public repo `authorize_repo_read` allows an anonymous caller straight through, so the load stays fully permissionless after that gate lands.

## Where

Two consumers of the unbounded fetch, both on the anonymous read group (`optional_signature`):

- `crates/gitlawb-node/src/api/certs.rs:20` — `list_certs` (`GET /api/v1/repos/{owner}/{repo}/certs`, routed at `server.rs:317`). No `limit`/cursor param and no truncate: it serializes the entire cert set into the response.
- `crates/gitlawb-node/src/api/events.rs:181` — `list_repo_events` caps the response with `all_events.truncate(limit)`, but only after the full fetch, so the DB read and the intermediate `Vec` allocation stay unbounded regardless of `?limit`.

Growing the table needs an authenticated pusher (git-receive-pack sits behind `require_signature`), but any registered agent can push (`enforce_owner_push` defaults false) and `git_write_routes` carries no rate limit, so a single writer can inflate one repo's cert count without bound. Every read afterward is permissionless and amplified.

## Impact

Measured against a live DB (200k synthetic cert rows for one repo, then reverted): ~60 MB of column data; the no-`LIMIT` query returns all rows in ~120 ms; `list_certs` would serialize ~82 MB per request. The node holds the `Vec<RefCertificate>` from `fetch_all`, the re-mapped `Vec<serde_json::Value>`, and the response body concurrently, roughly 200+ MB transient heap per in-flight `/certs` request from one unauthenticated GET. Reads have no per-caller rate limit, so N concurrent anonymous GETs multiply toward OOM. Same availability/amplification class as #82.

## Fix

Bound the fetch at the DB layer instead of in memory:

- Add `LIMIT` (plus cursor/offset) to `list_ref_certificates` and a paged `/certs` handler, mirroring `list_ref_updates_page` that #114 built for the gossip table.
- Move the `list_repo_events` bound into the query so the cert half is capped before `fetch_all`, not by the post-fetch `truncate`.
- Separately, consider a retention/prune policy or a `(repo_id, ref_name)` upsert so the table cannot grow without limit per ref.

Found as a follow-up to the PR #143 events-feed work. Verified by execution: `list_ref_certificates` returns all 250 seeded rows past the 200-row feed ceiling, and the figures above are from a live-DB measurement.

Gravirei:
## Summary

Bound `list_ref_certificates` with `LIMIT` at the DB layer and add a `(repo_id, ref_name)` upsert so the table cannot grow without bound per ref.

## Motivation & context

Closes #147

`list_ref_certificates` ran `SELECT ... WHERE repo_id = $1` with `fetch_all` and no `LIMIT`, loading every certificate row for a repo into memory. An anonymous caller on a public repo could amplify a single GET into an unbounded DB fetch, allocation, and response body.

## Kind of change

- [x] Security fix

## What changed

- `gitlawb-node` — `db/mod.rs`: added `LIMIT $2` to `list_ref_certificates` query; changed `insert_ref_certificate` to upsert on `(repo_id, ref_name)` conflict (v10 migration creates the unique index)
- `gitlawb-node` — `api/certs.rs`: added `?limit` query param (default 50, max 200) and `count` in response
- `gitlawb-node` — `api/events.rs`: pass the existing `limit` to `list_ref_certificates` so the cert half is bounded before `fetch_all`

## How a reviewer can verify

```sh
cargo test -p gitlawb-node migration_tests
```

## Before you request review

- [x] Migration unit tests pass

## Summary by CodeRabbit

* **New Features**
  * Added `limit` query support for certificate listings (`default 50`, clamped to `1–200`).
  * Certificate responses now include a top-level `count` field.
* **Bug Fixes**
  * Event listings now enforce a minimum `limit` of `1`, and related certificate lookups follow the same bound.
* **Database / Reliability**
  * Removed duplicate reference certificates and added uniqueness to keep only the newest entry per repo+ref.
* **Security / Access**
  * Improved authorization behavior for certificate endpoints.
* **Tests**
  * Expanded coverage for `limit`, ordering, `count`, and private-repo access.

### DIFF
--- a/crates/gitlawb-node/src/api/certs.rs
+++ b/crates/gitlawb-node/src/api/certs.rs
@@ -1,23 +1,33 @@
 //! API handlers for ref certificates.
 
-use axum::extract::{Extension, Path, State};
+use std::collections::HashMap;
+
+use axum::extract::{Extension, Path, Query, State};
 use axum::Json;
 
 use crate::auth::AuthenticatedDid;
 use crate::error::{AppError, Result};
 use crate::state::AppState;
 
-/// GET /api/v1/repos/{owner}/{repo}/certs
+/// GET /api/v1/repos/{owner}/{repo}/certs?limit=50
 pub async fn list_certs(
     State(state): State<AppState>,
     Path((owner, name)): Path<(String, String)>,
+    Query(params): Query<HashMap<String, String>>,
     auth: Option<Extension<AuthenticatedDid>>,
 ) -> Result<Json<serde_json::Value>> {
+    let limit = params
+        .get("limit")
+        .and_then(|v| v.parse::<i64>().ok())
+        .map(|v| v.max(1))
+        .unwrap_or(50)
+        .min(200);
+
     let caller = auth.as_ref().map(|e| e.0 .0.as_str());
     let (record, _rules) =
         crate::api::authorize_repo_read(&state, &owner, &name, caller, "/").await?;
 
-    let certs = state.db.list_ref_certificates(&record.id).await?;
+    let certs = state.db.list_ref_certificates(&record.id, limit).await?;
     let certs_json: Vec<serde_json::Value> = certs
         .iter()
         .map(|c| {
@@ -35,7 +45,10 @@ pub async fn list_certs(
         })
         .collect();
 
-    Ok(Json(serde_json::json!({ "certificates": certs_json })))
+    let count = certs_json.len();
+    Ok(Json(
+        serde_json::json!({ "certificates": certs_json, "count": count }),
+    ))
 }
 
 /// GET /api/v1/repos/{owner}/{repo}/certs/{id}
--- a/crates/gitlawb-node/src/api/certs.rs
+++ b/crates/gitlawb-node/src/api/certs.rs
@@ -27,7 +27,18 @@ pub async fn list_certs(
     let (record, _rules) =
         crate::api::authorize_repo_read(&state, &owner, &name, caller, "/").await?;
 
-    let certs = state.db.list_ref_certificates(&record.id, limit).await?;
+    // When a prefix is given (short-ID resolution from the CLI) use a
+    // generous limit and delegate to the prefix-matched query so the
+    // caller can resolve IDs regardless of how many certs exist.
+    let prefix = params.get("prefix").filter(|p| !p.is_empty());
+    let certs = if let Some(prefix) = prefix {
+        state
+            .db
+            .list_ref_certificates_by_prefix(&record.id, prefix, 200)
+            .await?
+    } else {
+        state.db.list_ref_certificates(&record.id, limit).await?
+    };
     let certs_json: Vec<serde_json::Value> = certs
         .iter()
         .map(|c| {
--- a/crates/gitlawb-node/src/api/events.rs
+++ b/crates/gitlawb-node/src/api/events.rs
@@ -129,6 +129,7 @@ pub async fn list_ref_updates(
     let limit = params
         .get("limit")
         .and_then(|v| v.parse::<i64>().ok())
+        .map(|v| v.max(1))
         .unwrap_or(50)
         .clamp(0, MAX_VISIBLE_REF_UPDATES);
 
@@ -177,6 +178,7 @@ pub async fn list_repo_events(
     let limit = params
         .get("limit")
         .and_then(|v| v.parse::<i64>().ok())
+        .map(|v| v.max(1))
         .unwrap_or(50)
         .clamp(0, MAX_VISIBLE_REF_UPDATES);
 
@@ -209,7 +211,7 @@ pub async fn list_repo_events(
     // into an empty 200, matching the gossip half below.
     let cert_events: Vec<serde_json::Value> = state
         .db
-        .list_ref_certificates(&record.id)
+        .list_ref_certificates(&record.id, limit)
         .await?
         .iter()
         .map(|c| {
--- a/crates/gitlawb-node/src/api/events.rs
+++ b/crates/gitlawb-node/src/api/events.rs
@@ -129,7 +129,7 @@ pub async fn list_ref_updates(
     let limit = params
         .get("limit")
         .and_then(|v| v.parse::<i64>().ok())
-        .map(|v| v.max(1))
+        .map(|v| v.max(0))
         .unwrap_or(50)
         .clamp(0, MAX_VISIBLE_REF_UPDATES);
 
@@ -178,7 +178,7 @@ pub async fn list_repo_events(
     let limit = params
         .get("limit")
         .and_then(|v| v.parse::<i64>().ok())
-        .map(|v| v.max(1))
+        .map(|v| v.max(0))
         .unwrap_or(50)
         .clamp(0, MAX_VISIBLE_REF_UPDATES);
 
--- a/crates/gitlawb-node/src/cert.rs
+++ b/crates/gitlawb-node/src/cert.rs
@@ -40,7 +40,7 @@ pub async fn issue_ref_certificate(
 
     let signature = state.node_keypair.sign_b64(&payload_bytes);
 
-    let mut cert = RefCertificate {
+    let cert = RefCertificate {
         id: Uuid::new_v4().to_string(),
         repo_id: repo_id.to_string(),
         ref_name: ref_name.to_string(),
@@ -52,7 +52,7 @@ pub async fn issue_ref_certificate(
         issued_at,
     };
 
-    let persisted_id = state.db.insert_ref_certificate(&cert).await?;
-    cert.id = persisted_id;
-    Ok(cert)
+    // Persist and return the row as it exists in the database (on a
+    // conflict the existing row survives when it is newer).
+    state.db.insert_ref_certificate(&cert).await
 }
--- a/crates/gitlawb-node/src/cert.rs
+++ b/crates/gitlawb-node/src/cert.rs
@@ -40,7 +40,7 @@ pub async fn issue_ref_certificate(
 
     let signature = state.node_keypair.sign_b64(&payload_bytes);
 
-    let cert = RefCertificate {
+    let mut cert = RefCertificate {
         id: Uuid::new_v4().to_string(),
         repo_id: repo_id.to_string(),
         ref_name: ref_name.to_string(),
@@ -52,6 +52,7 @@ pub async fn issue_ref_certificate(
         issued_at,
     };
 
-    state.db.insert_ref_certificate(&cert).await?;
+    let persisted_id = state.db.insert_ref_certificate(&cert).await?;
+    cert.id = persisted_id;
     Ok(cert)
 }
--- a/crates/gitlawb-node/src/db/mod.rs
+++ b/crates/gitlawb-node/src/db/mod.rs
@@ -1971,20 +1971,28 @@ impl Db {
 
 impl Db {
     /// Insert a ref certificate, or update it if a row for `(repo_id, ref_name)`
-    /// already exists. Returns the id of the row that is now in the database
-    /// (the original id on upsert; the passed id on insert).
+    /// already exists.  The update only applies when the incoming row is newer
+    /// (compared by `issued_at`), so a late-landing older cert cannot regress a
+    /// ref's persisted state.  Returns the id of the row that is now in the
+    /// database (the original id on upsert; the passed id on insert).
     pub async fn insert_ref_certificate(&self, cert: &RefCertificate) -> Result<String> {
         let row: (String,) = sqlx::query_as(
             "INSERT INTO ref_certificates
              (id, repo_id, ref_name, old_sha, new_sha, pusher_did, node_did, signature, issued_at)
              VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)
              ON CONFLICT (repo_id, ref_name) DO UPDATE SET
-                old_sha   = EXCLUDED.old_sha,
-                new_sha   = EXCLUDED.new_sha,
-                pusher_did = EXCLUDED.pusher_did,
-                node_did  = EXCLUDED.node_did,
-                signature = EXCLUDED.signature,
-                issued_at = EXCLUDED.issued_at
+                old_sha   = CASE WHEN EXCLUDED.issued_at > ref_certificates.issued_at
+                                 THEN EXCLUDED.old_sha   ELSE ref_certificates.old_sha   END,
+                new_sha   = CASE WHEN EXCLUDED.issued_at > ref_certificates.issued_at
+                                 THEN EXCLUDED.new_sha   ELSE ref_certificates.new_sha   END,
+                pusher_did = CASE WHEN EXCLUDED.issued_at > ref_certificates.issued_at
+                                  THEN EXCLUDED.pusher_did ELSE ref_certificates.pusher_did END,
+                node_did  = CASE WHEN EXCLUDED.issued_at > ref_certificates.issued_at
+                                 THEN EXCLUDED.node_did  ELSE ref_certificates.node_did  END,
+                signature = CASE WHEN EXCLUDED.issued_at > ref_certificates.issued_at
+                                 THEN EXCLUDED.signature ELSE ref_certificates.signature END,
+                issued_at = CASE WHEN EXCLUDED.issued_at > ref_certificates.issued_at
+                                 THEN EXCLUDED.issued_at ELSE ref_certificates.issued_at END
              RETURNING id",
         )
         .bind(&cert.id)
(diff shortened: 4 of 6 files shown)


jatmn (changes requested):
I found a couple of issues that need to be addressed before this is ready.

## Findings

- [P2] Complete CodeRabbit's request to gate the cert endpoints
  `crates/gitlawb-node/src/api/certs.rs:24`
  CodeRabbit's current-head review item is still valid: both `list_certs` and `get_cert` resolve the repo with `get_repo(...)` and never call the shared `authorize_repo_read(...)` path, while the routes sit on the anonymous read router. This means an unauthenticated caller who can name a private repo can still read its ref certificate metadata (`ref_name`, old/new SHAs, pusher DID, node DID, signature), and `get_cert` can return any certificate ID after only proving that the requested repo path exists. Please thread the optional authenticated caller through these handlers, use `authorize_repo_read(&state, &owner, &name, caller, "/")`, and add the private-repo denial test CodeRabbit requested.

- [P2] Complete Copilot's limit-clamp request for the events path
  `crates/gitlawb-node/src/api/events.rs:54`
  The PR fixed negative `limit` handling in `/certs`, but `list_repo_events` still parses the query with `unwrap_or(50).min(200)`, so `?limit=-1` remains negative. This PR now passes that value into `list_ref_certificates(&record.id, limit)`, and the DB helper binds it directly into `LIMIT $2`; malformed input therefore trips the DB query and the handler silently falls back to an empty local-cert/event result instead of applying the same bounded limit as normal requests. Please clamp the lower bound in the events path and add the DB-boundary clamp/test requested by the earlier reviewer comment so every caller of `list_ref_certificates` stays bounded.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found one item that still needs to be addressed.

## Findings

- [P2] Complete Copilot's DB-boundary limit clamp request
  `crates/gitlawb-node/src/db/mod.rs:1965`
  The API handlers now clamp their query parameters before calling `list_ref_certificates`, but the DB helper itself still accepts any `i64` and binds it directly into `LIMIT $2`. Copilot's current thread on this helper is still valid, and the previous review also asked for the DB-boundary clamp/test so every caller of this helper remains bounded. Please clamp or reject invalid limits inside `list_ref_certificates` itself, and add a focused test for a negative/raw helper limit so a future caller cannot reintroduce this path.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found one item that still needs to be addressed.

## Findings

- [P2] Gate the repo events certificate path as well
  `crates/gitlawb-node/src/api/events.rs:63`
  This update correctly moves `/certs` through `authorize_repo_read`, but the same local certificate data is still reachable through `GET /api/v1/repos/{owner}/{repo}/events`. That route sits on the optional-signature read router, `list_repo_events` resolves the repo with `get_repo(...)` instead of a caller-aware visibility check, and then the local-cert branch calls `list_ref_certificates(&record.id, limit)` and returns `ref_name`, `old_sha`, `new_sha`, `pusher_did`, and `node_did`. For a private repo, an unauthenticated caller who can name the repo therefore still gets the ref-certificate metadata this PR just gated on `/certs`. Please thread the optional authenticated caller into `list_repo_events` and gate the local repo branch with `authorize_repo_read(..., "/")` before fetching or returning local cert/gossip events.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found one item that still needs to be addressed.

## Findings

- [P2] Finish gating the repo-events gossip path after auth denial
  `crates/gitlawb-node/src/api/events.rs:80`
  The current update gates the local certificate query, but when `authorize_repo_read(...)` returns an error for this repo the handler falls back to `format!("{owner}/{repo_name}")` and still calls `list_repo_ref_updates(&repo_id_str, limit)`. If that URL slug matches rows in `received_ref_updates` for a local repo the caller cannot read, `GET /api/v1/repos/{owner}/{repo}/events` still returns the repo's ref metadata (`ref_name`, `old_sha`, `new_sha`, `pusher_did`, `node_did`, `cert_id`, timestamps) even though the repo-read gate denied access. Please complete the previous repo-events gating request by distinguishing "repo not local" from "local repo exists but caller is unauthorized", and avoid fetching either local certs or local/gossip event rows after an authorization denial for a local repo.

beardthelion (dismissed):
Verified against the current head (f3e4d885). The certificate gating, the DB-boundary limit clamp, and the v10 migration structure all hold up. The blocking item is the one jatmn already flagged, still present on this head; a few smaller items follow.

## Findings

- [P2] Withhold the gossip branch of `list_repo_events` after an authorization denial
  `crates/gitlawb-node/src/api/events.rs:80`
  Seconding jatmn's open finding, still live on f3e4d885. The local certificate query is gated, but on the `authorize_repo_read` error branch the handler falls back to `format!("{owner}/{repo_name}")` and still calls `list_repo_ref_updates`, a raw `WHERE repo = $1` read with no visibility filter. Driving the endpoint as an anonymous caller against a private local repo whose slug has a `received_ref_updates` row returns that row's `ref_name`, `old_sha`, `new_sha`, `pusher_did`, and `node_did`. The row has to be present first (peer-injected via gossip or `notify_sync`, since a private repo does not announce its own pushes), which keeps this at P2, but the endpoint should distinguish "repo not local" from "local repo exists but caller denied" and skip both the cert and gossip reads once a local repo denies the caller.

- [P3] Keep the certificate id stable across re-pushes
  `crates/gitlawb-node/src/db/mod.rs:1932`
  `insert_ref_certificate` now upserts with `ON CONFLICT (repo_id, ref_name) DO UPDATE SET id = EXCLUDED.id`, so every re-push to a ref rotates the row's primary key. `get_ref_certificate` looks up by id alone, so a cert id issued earlier (returned by `issue_ref_certificate`, logged, and carried as `received_ref_updates.cert_id`) stops resolving through `GET /certs/{id}` after the next push. Note that just dropping `id = EXCLUDED.id` is not enough on its own: `issue_ref_certificate` returns the freshly minted UUID to its caller, so the returned id would then diverge from the persisted one. Return the persisted row (`RETURNING id`) or derive the id deterministically from `(repo_id, ref_name)`.

- [P3] Add a deny-path test for `list_repo_events` and an upgrade-path test for the v10 dedup
  `crates/gitlawb-node/src/api/events.rs:51`, `crates/gitlawb-node/src/db/mod.rs:826`
  The new tests are all happy-path. Nothing drives an unauthorized caller against a private local repo and asserts the gossip rows are withheld, which is exactly where the leak above sits. Separately, the v10 dedup `DELETE` only matters on already-deployed nodes that accumulated duplicates, but every new test runs on a fresh `#[sqlx::test]` database, so that branch never executes; a test that seeds duplicates at the prior schema version and then migrates would cover it. While there, give the dedup `ORDER BY issued_at DESC` an `id` tiebreaker so the surviving row is deterministic when two share a timestamp.

The `CREATE UNIQUE INDEX` is non-concurrent and runs inside the migration transaction, so it briefly blocks `ref_certificates` writes during deploy. That is fine while the table stays small per ref, but worth a line in the deploy notes.

jatmn (approved):


@kevincodex1 LGTM

beardthelion (dismissed):
Re-reviewed the current head (2460b1b8). Every item from my previous review and jatmn's blocking finding is resolved, and I verified each fix by execution: the repo-events gossip path now returns 404 to a non-owner before any cert or gossip read (I drove the full, bare, and truncated owner forms, so the deny arm's `get_repo` recheck can't be sidestepped by an alternate owner slug); the negative and oversized `?limit=` cases return a clamped 200 rather than a 500; the upsert preserves the cert id across re-pushes; and the v10 dedup keeps a deterministic survivor. The core is sound. Two things before it can merge.

## Findings

- [P1] Rebase onto current main and re-review the resolved ref-update diff before merge
  `crates/gitlawb-node/src/api/events.rs`, `crates/gitlawb-node/src/db/mod.rs`
  This branch's merge-base predates #143, which gated the ref-update feeds (the GraphQL `ref_updates` query, the global `/api/v1/events/ref-updates` feed, and the subscription broadcast). #149 also edits `events.rs` and `db/mod.rs`, so the rebase will conflict in exactly those files, and a wrong resolution can silently drop #143's `collect_visible_ref_updates` gating and reopen the private-repo ref-metadata leak. Please rebase onto current main and re-review the resolved diff across `events.rs`, `graphql/query.rs`, `repos.rs`, and every `received_ref_updates` reader before merging.

- [P2] Add an upgrade-path migration test for the v10 dedup
  `crates/gitlawb-node/src/db/mod.rs`
  The dedup `DELETE` runs before the `CREATE UNIQUE INDEX` in one migration transaction and `v10_dedup_removes_old_duplicates` passes, so the migration itself is sound. But that test runs `run_migrations()` on a fresh database and then hand-copies the dedup SQL as string literals, so the real migration's dedup branch never runs on a seeded prior-version node and the test can drift from `MIGRATIONS[v10]` without failing. Add a test that seeds `schema_migrations` at v9 with duplicate `ref_certificates` (including a pair sharing `issued_at`), runs `Db::run_migrations()`, and asserts one deterministic survivor per ref plus the index present.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found one item that still needs to be addressed.

## Findings

- [P3] Return the persisted certificate id after an upsert
  `crates/gitlawb-node/src/cert.rs:55`
  The upsert now preserves the original row id for `(repo_id, ref_name)` conflicts, but `issue_ref_certificate` still returns the freshly generated `cert` after `insert_ref_certificate(&cert)` succeeds. On a re-push to an existing ref, that generated UUID is discarded by the `ON CONFLICT` update, so the caller logs an “issued ref certificate” id that `get_ref_certificate` cannot resolve and future callers of this helper would receive a certificate object that was never persisted. Please have the insert/upsert return the stored row, or reload by `(repo_id, ref_name)`, so `issue_ref_certificate` returns the id that actually exists after conflicts.

beardthelion (dismissed):
Re-reviewed the current head (5ea26d1). Both items from my previous review are resolved and I verified each by execution: the rebase preserved #143's `collect_visible_ref_updates` gating (private-repo ref-update and gossip reads still 404 to a non-owner), and `v10_upgrade_dedup_via_migration` drives the real `run_migrations()` and is load-bearing (mutating the migration's `ORDER BY` turns it red). The cert-endpoint gating, the `get_cert` cross-repo guard, and the negative/oversized `?limit=` clamps hold up under the deny and must-not cases. One new issue the upsert introduces needs addressing before merge.

## Findings

- [P2] Gate the cert upsert on `issued_at` so a late-landing older cert can't overwrite a newer one
  `crates/gitlawb-node/src/db/mod.rs:1951`, `crates/gitlawb-node/src/api/repos.rs:905`
  `git_receive_pack` releases the repo write guard before the cert-issuance loop, so `insert_ref_certificate` runs unguarded and its `ON CONFLICT (repo_id, ref_name) DO UPDATE` overwrites the row unconditionally. Two sequential same-ref pushes (`OLD→A` then `A→B`) whose cert inserts land in reverse order leave the stale `OLD→A` cert as the survivor, with an `issued_at` behind the ref's real state. That is a regression from the prior append plus `ORDER BY issued_at DESC`, which always surfaced the newest cert. I reproduced it by persisting the newer cert first, then the older one, for the same ref: the survivor is the older `new_sha`. Only adopt the incoming row when it is newer, e.g. a per-column `CASE WHEN EXCLUDED.issued_at > ref_certificates.issued_at THEN EXCLUDED.x ELSE ref_certificates.x END`, or issue the certificate while the write guard is still held. Note a bare `DO UPDATE ... WHERE ...` no longer works now that the statement uses `RETURNING id`: a skipped update returns no row and `fetch_one` would error. I confirmed the `CASE` form closes the race and keeps id-preservation and the existing upsert tests green.

beardthelion (dismissed):
The lost-update race I flagged last round is fixed. The per-column `issued_at` guard on the upsert holds up: I ran an older cert landing after a newer one for the same `(repo_id, ref_name)` both sequentially and as two concurrent transactions with the older one committing last, and the newer content survives in both cases; reverting the guard reproduces the regression. `RETURNING id` still returns the persisted row id on the no-op path, and `issue_ref_certificate` now propagates it, so the id-resolution issue from the prior round is addressed on this head too. Full suite green here.

Two non-blocking notes, both latent rather than reachable today.

## Findings

- [P3] Note that the tie rule differs from the v10 dedup by design
  `crates/gitlawb-node/src/db/mod.rs:1954`
  On an equal `issued_at` the runtime `CASE` (strict `>`) keeps the incumbent row, while the v10 dedup migration at `db/mod.rs:837` (`ORDER BY issued_at DESC, id DESC`) keeps the highest id, so the two select opposite rows on a tie. It is not reachable in practice, since issuance is serial per push at nanosecond resolution, but a one-line comment would make the divergence read as intentional instead of an oversight.

- [P3] Compare `issued_at::timestamptz` before any peer-cert relay path lands
  `crates/gitlawb-node/src/db/mod.rs:1954`
  `issued_at` is TEXT and the guard compares it with `>`, so the ordering is lexicographic. That is correct only because the one writer is `cert.rs:27` `Utc::now().to_rfc3339()`, which always emits a `+00:00` offset. A value stored as `...Z` or with a non-UTC offset would sort wrong: `'…Z' > '…+00:00'` and `'…+02:00' > '…+00:00'` are both true as strings but false as instants, which lets an older cert win. No such writer exists today, so this is latent. If ingesting or relaying certificates from peers is on the roadmap, `EXCLUDED.issued_at::timestamptz > ref_certificates.issued_at::timestamptz` closes it and is worth landing with this change; otherwise it is just a robustness note.

Core change is sound. @kevincodex1 good to merge once you have looked over the two notes.

jatmn (approved):


@kevincodex1 LGTM

kevincodex1:
@Gravirei please rebase to main and fix conflicts

beardthelion (approved):
Re-reviewed on 89607ab. Approving. I verified the security-critical properties by execution on this head, not by inspection.

The freshness upsert is load-bearing: the cert suite is green on this head, and flipping the guard to `<` turns insert_ref_certificate_upserts_on_repo_ref and v10_upgrade_dedup_via_migration red, so the older-cannot-regress-newer property is genuinely tested. Read gating holds on all three cert surfaces: list_certs, get_cert with the repo_id cross-check, and the events local_cert path all gate through authorize_repo_read before any fetch. I drove the deny matrix through the real handlers: a signed non-owner and an anonymous caller both get 404 on a private repo's certs while the owner gets 200. The limit is clamped at both callers and backstopped in the DB, so no negative or oversized value reaches Postgres. The v10 dedup is a new highest-version migration, runs in one transaction, keeps the newest row per (repo_id, ref_name), and its upgrade-path test fires the real migration.

Three non-blocking notes for a follow-up, none gate this:

- issued_at is a wall-clock string, so the guard assumes a monotonic clock. It is correct for normal operation since issued_at is always server-side Utc::now in +00:00, but a backward clock step between two pushes to the same ref would keep the earlier-real-time cert. A timestamptz cast would not change that, since the values are already UTC, so a one-line comment noting the assumption is enough.
- On a losing upsert, issue_ref_certificate returns cert.id from the existing row while the returned struct still carries the rejected cert's data. Only c.id is read today, so nothing breaks, but returning the persisted row or re-fetching would remove the latent trap.
- v10_dedup_removes_old_duplicates hand-copies the migration SQL and will not catch drift in the v10 block; v10_upgrade_dedup_via_migration is the load-bearing one.

@kevincodex1 good to merge.

jatmn (changes requested):
Thanks for the update. I rechecked the changed paths and found an issue that still needs to be addressed.

## Findings

- [P1] Preserve `gl cert show` short-ID resolution after bounding the list endpoint
  `crates/gitlawb-node/src/api/certs.rs:19`
  The new default now returns only the newest 50 certificates, but `crates/gl/src/cert.rs:184-210` still resolves a short certificate ID by fetching that unparameterized list and searching its entire response. Once a repository has more than 50 distinct current refs, a valid older short ID is omitted; the resolver then sends that prefix to the exact-ID endpoint, which returns 404. Before this change the CLI searched the complete list. Please add pagination or a server-side prefix lookup (and a regression test with more than 50 certificates) so previously valid short IDs remain usable.

jatmn (approved):

beardthelion (changes requested):
The core here is solid — I verified the v10 dedup migration, the negative-limit clamp, the cert read-gate, and the upsert freshness guard all hold under a break (each goes red if I neuter it). One thing to fix on the P1 remedy before this merges:

## Findings

- [P3] Make the >50 prefix regression test actually guard the regression
  `crates/gitlawb-node/src/test_support.rs:3891`
  `list_certs_prefix_resolves_deep_cert` asserts `any(id.starts_with("deep-0"))`, which the default newest-50 list already satisfies via the in-window certs `deep-0005..deep-0009`. I confirmed it's vacuous by forcing the prefix branch off (`prefix -> None`): the test still passes, because the fallback list contains `deep-000x`. jatmn's P1 was specifically about resolving a short-ID for a cert *beyond* the default 50, and this test doesn't exercise that path — a reintroduction of the bug would ship green.

  The fix itself is correct: against the real code, `?prefix=deep-0000` resolves `deep-0000` (the oldest of the 55, beyond the window). Assert a specific beyond-window cert (e.g. that the result contains `deep-0000`, or query `?prefix=deep-0000` and assert exactly that id) so the test actually protects the fix.

Everything else checks out — nice work on the dedup-before-unique-index and the DB-boundary clamp.

Issue #62, opened 2026-06-19
Harden served-git process handling: timeout, cross-env pid cap, wiring test (follow-up to #61)


beardthelion:
Follow-ups deliberately left out of #61 (the #53 zombie-reaping fix) to keep that change focused. Each is independent.

## 1. Bound a hung (not disconnected) served git with a timeout

#61 tears down git on client disconnect, but a wedged `upload-pack`/`pack-objects` that neither finishes nor disconnects can still block the request indefinitely and hold a PID. Wrap the `wait_with_output()` in `run_git_service` (`crates/gitlawb-node/src/git/smart_http.rs`) in a `tokio::time::timeout`. On expiry the existing `KillGroupOnDrop` guard already fires when the future is dropped, so the process group gets cleaned up; the work is choosing a sane bound (large repos need headroom) and surfacing a clear error.

## 2. Cross-environment process cap, not just compose

`pids_limit: 1024` lives in `docker-compose.yml` only, and the cgroup pids controller counts threads, so per-request fan-out (git + pack-objects + helper threads) eats into it faster than one PID per request suggests. It also does not apply on Fly. Two parts:

- An application-level concurrency cap (a semaphore in front of `command.spawn()` in the serve path) so the node sheds load with a clean 503 instead of every concurrent git op failing with a 500 mid-stream once the table is exhausted.
- A matching limit wherever the node runs outside compose (e.g. Fly), so the safety net is not compose-only.

## 3. Wiring test for the disconnect/reap path

The new tests in #61 exercise the `KillGroupOnDrop` guard in isolation. There is no test asserting that `run_git_service` actually applies `process_group(0)` and arms the guard, so a refactor that dropped either would pass CI. A fake `git` on `PATH` (a script that sleeps, or exits immediately to force the stdin-write-error path) makes both the disconnect-teardown and the reap-on-error paths deterministically testable end to end.

Refs #53, #61.

PR #150, opened 2026-07-03
test(node,git): wire run_git_service teardown end-to-end (#62)

beardthelion:
Part 3 of #62 (served-git process hardening), and the piece to land first: it locks the teardown invariant the timeout work (part 1) will lean on.

## Why
The #61 tests exercise `KillGroupOnDrop` in isolation. They hand-build the guard and never call `run_git_service`, so the production wiring is unprotected: deleting `command.process_group(0)`, the guard-arming, or the post-reap `disarm()` leaves the whole suite green.

## What
- A minimal injectable `git_bin` parameter on the private `run_git_service` (production callers pass `"git"`). It exists purely so a fake `git` can drive the real function end-to-end without mutating the process-global PATH, which keeps the tests parallel-safe.
- Three end-to-end tests that drive `run_git_service` through that seam and assert the production spawn path actually wires the teardown:
  - dropping the request future mid-flight SIGTERMs the whole group, reaping git and its `pack-objects` grandchild;
  - a request that runs to completion (both non-zero and zero exit) disarms the guard, leaving a still-live grandchild untouched.

## Verification
Every assertion is load-bearing, shown RED against its mutation and green when restored:
- remove `process_group(0)` or the guard-arming, and the future-drop test fails;
- remove `disarm()`, and both completion tests fail;
- gate `disarm()` on failure only, and the success-path test fails while the error-path test passes.

No production behavior change: the wrapper callers pass `"git"`, identical to the previous hardcoded value, and `run_git_service` has no other callers. `cargo fmt` and `clippy -D warnings` are clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of Git service requests, reducing the chance of stray background processes after interrupted or completed operations.
  * Made Git upload and receive actions handle failures more cleanly and consistently.

### DIFF
--- a/crates/gitlawb-node/src/git/smart_http.rs
+++ b/crates/gitlawb-node/src/git/smart_http.rs
@@ -52,7 +52,7 @@ pub async fn info_refs(repo_path: &Path, service: &str) -> Result<Response> {
 /// Serves pack data for a clone or fetch. This is stateless — the entire
 /// negotiation happens in a single request/response.
 pub async fn upload_pack(repo_path: &Path, request_body: Bytes) -> Result<Response> {
-    let output = run_git_service("git-upload-pack", repo_path, request_body).await?;
+    let output = run_git_service("git", "git-upload-pack", repo_path, request_body).await?;
 
     Ok(Response::builder()
         .status(StatusCode::OK)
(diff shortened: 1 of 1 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):
LGTM

PR #151, opened 2026-07-04
ci: pin cargo-audit on the blocking gate, float the weekly canary

beardthelion:
The PR-blocking `audit` job installs cargo-audit with an unpinned `cargo install --locked cargo-audit`, so the scanner floats to the latest release on every run. A regressed release then reds the gate repo-wide with no code change, the same spontaneous-breakage class the quinn-proto advisory exposed but on the scanner itself. #57 added the unpinned install; nothing has pinned it since.

**pr-checks.yml (the gate):** pinned to `--version 0.22.2` (what the unpinned install currently resolves to). `cargo install --version X` is an exact pin, not a caret, so it freezes the binary; the RustSec advisory DB is still fetched fresh at scan time, so this does not weaken what the gate catches. The step also asserts the resolved version after install, so an accidental future unpin fails loud instead of silently running a different scanner, and the cargo-audit version is now part of the rust-cache key so hits stay deterministic.

**audit-schedule.yml (weekly canary):** left deliberately unpinned. It is cron-only and never gates a merge, so floating it cannot block a PR; if a newer cargo-audit regresses, the weekly run reds or its report changes, which is the drift signal that latest has moved and the gate's pin should not be bumped yet.

If 0.22.2 is ever yanked the gate's install fails loud until bumped; that bump path is documented in the step comment.

## Summary by CodeRabbit

* **Chores**
  * Updated CI audit checks to use a pinned `cargo-audit` version for PR validation, with an added runtime version check to prevent unexpected tool changes.
  * Adjusted the scheduled audit job to use a floating scanner as a drift canary, with clearer workflow documentation.

### DIFF
--- a/.github/workflows/audit-schedule.yml
+++ b/.github/workflows/audit-schedule.yml
@@ -37,7 +37,18 @@ jobs:
         with:
           key: audit-schedule
 
-      - name: Install cargo-audit
+      - name: Install cargo-audit (floating; drift canary)
+        # Deliberately UNPINNED, unlike the blocking gate in pr-checks.yml. This
+        # job is cron-only (weekly); it is never pull_request- or push-triggered,
+        # so it NEVER gates a merge no matter what it does. Floating the scanner
+        # here is the drift canary: if a newer cargo-audit regresses, at worst
+        # this weekly run reds (the install step below is unguarded) or its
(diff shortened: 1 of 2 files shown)


jatmn (approved):

@kevincodex1 LGTM

kevincodex1 (approved):

PR #152, opened 2026-07-05
fix(node): close two spam-vector root causes (trust upsert + ungated push)

kevincodex1:
The June 2026 spam wave kept re-materializing after cleanup because two paths let an unauthenticated-by-captcha actor keep operating:

1. `update_trust_score` was an upsert. Any authenticated push/issue/PR from a DID with no agent row silently INSERTed one with a fresh `registered_at`, bypassing the iCaptcha gate on /api/register — so deregistered spam DIDs came back the moment they pushed. Make it an UPDATE-only no-op for unregistered DIDs; registration is the sole way into the agents table. Regression test added.

2. `git-receive-pack` had no rate limit. The per-DID limiter on the creation routes can't brake a push flood from a DID farm (one throwaway identity per repo), so the node absorbed several pushes/sec per IP, each a Tigris round-trip + peer-notify fan-out. Add a per- client-IP limiter (Fly-Client-IP / X-Forwarded-For) on the push path, default 600/h, `GITLAWB_PUSH_RATE_LIMIT` override, 0 disables. Fails open without a proxy header so self-hosted nodes aren't broken.

## Summary by CodeRabbit

* **New Features**
  * Added a per-client-IP “push-path flood brake” for `git-receive-pack`, enforced during both `info/refs` (advertisement) and receive phases.
  * Over-limit requests now return **429 Too Many Requests** with `retry-after`.
  * Added configurable push throttling via `GITLAWB_PUSH_RATE_LIMIT` (default 600; `0` disables) and trusted-proxy IP selection via `GITLAWB_TRUSTED_PROXY` (including rightmost `X-Forwarded-For` in trusted modes).

* **Bug Fixes**
  * Trust score updates no longer create agent records for unknown DIDs.
  * Rate limiting reliably disables when set to `0` and doesn’t break on malformed/missing forwarded-IP headers.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -90,6 +90,21 @@ GITLAWB_PUBLIC_READ=true
 # Maximum git smart-HTTP pack request size, in bytes.
 GITLAWB_MAX_PACK_BYTES=2147483648
 
+# ── Push rate limiting (git-receive-pack flood brake) ─────────────────────
+# Max receive-pack requests (info/refs advertisement + push POST) per client
+# IP per hour. 0 disables. Default 600.
+GITLAWB_PUSH_RATE_LIMIT=600
+
+# Which forwarded header the edge is trusted to set, used to resolve the real
+# client IP for the push limiter. One of:
+#   (unset) — no trusted proxy: key on the socket peer address, ignore headers.
+#   fly     — behind Fly's edge (trust Fly-Client-IP).
+#   x-forwarded-for — behind a single reverse proxy like Caddy/NGINX (trust the
+#             rightmost X-Forwarded-For hop).
+# Only set this when a proxy you control actually fronts the node; trusting a
+# forwarded header on a directly-exposed node lets clients spoof the key.
+GITLAWB_TRUSTED_PROXY=
+
 # ── Sync ─────────────────────────────────────────────────────────────────
 # Enable automatic background sync from known peers
 GITLAWB_AUTO_SYNC=false
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -487,6 +487,8 @@ pub async fn git_info_refs(
     State(state): State<AppState>,
     Path((owner, repo)): Path<(String, String)>,
     Query(query): Query<InfoRefsQuery>,
+    crate::rate_limit::PeerAddr(peer): crate::rate_limit::PeerAddr,
+    headers: axum::http::HeaderMap,
     auth: Option<Extension<AuthenticatedDid>>,
 ) -> Result<Response> {
     let name = repo.trim_end_matches(".git");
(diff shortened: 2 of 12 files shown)


jatmn (changes requested):

## Findings

- [P1] Rate-limit the receive-pack advertisement before it refreshes from Tigris
  `crates/gitlawb-node/src/api/repos.rs:529`
  The new limiter only wraps `POST /{owner}/{repo}/git-receive-pack`, but every push first performs `GET /{owner}/{repo}/info/refs?service=git-receive-pack`. That GET is still in `git_read_routes` with only `optional_signature`, and `git_info_refs` calls `repo_store.acquire_fresh()` for `service=git-receive-pack`, so an attacker can repeatedly hit the unauthenticated advertisement phase and keep forcing the fresh Tigris check/download path without ever sending the limited POST. Please apply the same push flood brake before the receive-pack `info/refs` refresh, or otherwise avoid the expensive fresh-acquire work until after the request is authenticated and throttled.

- [P1] Do not trust client-supplied forwarded headers for the push limiter
  `crates/gitlawb-node/src/rate_limit.rs:109`
  The new push limiter keys requests directly from `Fly-Client-IP` or the first `X-Forwarded-For` value, but clients can send those headers themselves on direct/self-hosted deployments, and any proxy that does not strip or overwrite incoming values leaves the same bypass. A push flooder can rotate either header on each `git-receive-pack` request and never fill a bucket, so the PR does not actually stop the per-IP flood it is meant to stop outside tightly controlled edge setups. Please derive the key from the real peer address unless the request came through an explicitly trusted proxy, or otherwise ignore/reject forwarded headers that were not injected by trusted infrastructure.

beardthelion (changes requested):
Concur with jatmn's two P1s (advertisement refresh, forwarded-header trust); I won't repeat them. The findings below are separate issues on the new IP limiter, plus a note that the trust-score half is sound.

## Findings

- [P2] Bound the per-IP limiter's key set
  `crates/gitlawb-node/src/rate_limit.rs:43`
  `check()` inserts a HashMap entry for every distinct key before it can reject, and `cleanup()` (line 62, on the 300s loop) only evicts keys whose timestamps are all older than the 3600s window. Since the key is the client-controlled IP header from the P1 above, varying the header adds a resident entry per request that lives up to an hour, and the header value is attacker-sized. The limiter's own state becomes an unauthenticated memory-exhaustion vector, and the insert-before-reject ordering means even rejected requests still allocate. Cap the number of tracked keys (or evict oldest past a bound).

- [P2] Reject the empty leading X-Forwarded-For hop instead of failing open
  `crates/gitlawb-node/src/rate_limit.rs:112`
  `X-Forwarded-For: ,1.2.3.4` splits to an empty first hop, which `.filter(|ip| !ip.is_empty())` turns into `None`, and `rate_limit_by_ip` skips the limiter entirely when `client_ip` is `None`. A single static malformed header disables the brake, no rotation needed. Separately the `Fly-Client-IP` branch lacks the same `is_empty` filter, so an empty value returns `Some("")` and collapses that traffic onto one shared bucket rather than skipping; align the two branches. This is a distinct parsing defect that compounds the header-trust P1.

- [P2] Test the 429 path and the advertisement's throttle status
  `crates/gitlawb-node/src/rate_limit.rs:197`
  The suite covers `RateLimiter::check()` and `client_ip()` parsing, but nothing drives `rate_limit_by_ip` to an actual 429, and nothing asserts whether `GET /{owner}/{repo}/info/refs?service=git-receive-pack` (`server.rs:382`) is throttled. The load-bearing new behavior (the reject branch) and the exact gap in the first P1 both ship untested, so neither the current hole nor a fix has a guardrail. A middleware test sending over-limit requests from a fixed IP asserting 429, plus one asserting the advertisement's throttle state, locks both down.

Not a finding: the `update_trust_score` upsert-to-UPDATE-only change is sound. `register_agent` inserts before the update, and the other four callers are fire-and-forget bumps that now correctly no-op for unregistered DIDs. No reader gates on the row existing (`get_trust_score` returns `0.0` for a missing row), so the change is strictly more restrictive, and the added regression test is load-bearing.

kevincodex1:
do a full review @coderabbitai  approve if all good

PR #156, opened 2026-07-06
refactor(visibility): route is_owner through the shared did_matches matcher (#153)

### LINKED ISSUE #153: is_owner: bare-key mirror owner denied read to own repo; stale did_matches comment
## What

`visibility::is_owner` (`crates/gitlawb-node/src/visibility.rs`) matches the caller against `normalize_owner_key(owner_did)`, which only strips `did:key:` off the *owner* side. When a row stores `owner_did` in bare short-key form and the caller authenticates with their canonical full `did:key:`, the owner fails to match and is denied read to their own repo.

Executed against current main (throwaway unit test, reverted):

```
visibility_check(&[], false, "z6MkOwner",         Some("did:key:z6MkOwner"),        "/") => Deny    // bare owner, full caller: LOCKED OUT
visibility_check(&[], false, "did:key:z6MkOwner", Some("z6MkOwner"),                "/") => Allow   // reverse direction is fine
visibility_check(&[], false, "did:key:z6MkOwner", Some("did:gitlawb:z6MkOwner"),    "/") => Deny    // cross-method correctly denied
```

This is fail-*closed* (an owner is denied), so it is a correctness/usability bug, not a leak.

## Reachability

Mirror rows store the bare form: `upsert_mirror_repo` runs `let owner_key = normalize_owner_key(owner_did)` before the INSERT. `get_repo` prefers the canonical (full-DID) row over the mirror when both exist (#124), so the lockout is scoped to a **mirror-only** repo (no canonical row) whose owner reads it while authenticated with their full `did:key:`. Worth confirming how often an owner-authenticated read resolves to a bare-owner mirror row in practice.

## Remedy (verified)

Route `is_owner` through `crate::api::did_matches`, which strips `did:key:` on both sides and refuses cross-method matches. Same throwaway test:

```
did_matches("z6MkOwner", "did:key:z6MkOwner")             => true    // lockout fixed
did_matches("did:key:z6MkOwner", "z6MkOwner")             => true
did_matches("did:key:z6MkOwner", "did:gitlawb:z6MkOwner") => false   // cross-method still denied
```

## Stale comment

The `is_owner` doc comment says using `did_matches` "would let a non-key canonical row bypass the #124 visibility gate." The current `did_matches` denies cross-method (third case above), so that reasoning is stale; the comment should be dropped or corrected whichever way this lands.

## Context

Surfaced while bringing #113 current with main. #113 had independently switched `is_owner` to `did_matches`; I kept main's `normalize_owner_key` in that merge rather than reverse #124/#141 through a conflict resolution, and am filing this so the change gets its own decision. If adopted, add a bare-mirror-owner + cross-method regression pair to keep owner matching in parity with `did_matches`.

Gravirei:
## Summary

Switch `is_owner` from `normalize_owner_key` (single-side normalization) to `did_matches` (two-side normalization with cross-method rejection) so a mirror-only repo owner authenticated with their full `did:key:` is not denied read to their own repo.

## Motivation & context

Closes #153

`visibility::is_owner` used `normalize_owner_key(owner_did)` which only strips `did:key:` off the owner side. When a mirror row stores `owner_did` as bare short key and the caller authenticates with full `did:key:`, the comparison fails and the owner is locked out.

The old doc comment warned that `did_matches` "would let a non-key canonical row bypass the #124 visibility gate" — but `did_matches` already rejects cross-method collisions (`did:key:X` != `did:gitlawb:X`), so that reasoning is stale.

## Kind of change

- [x] Bug fix

## What changed

- **`crates/gitlawb-node/src/visibility.rs`** — Replaced `is_owner` body with `crate::api::did_matches(owner_did, caller)` and updated the doc comment to reflect current semantics.
- **Tests** — Added `bare_owner_full_caller_allows_owner` (regression for the lockout) and `cross_method_did_denied_with_bare_owner` (cross-method bypass guard).

## How a reviewer can verify

```sh
cargo test -p gitlawb-node -- visibility::tests
```

## Protocol & signing impact

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved owner matching for visibility checks so equivalent `did:key` representations are recognized consistently.
  * Preserved access restrictions across different DID methods, preventing incorrect matches when identifiers only share the same trailing value.
  * Added unit test coverage for the DID matching behavior, including empty-input and regression cases.


jatmn (approved):

@kevincodex1 LGTM

beardthelion (dismissed):
Verified the security-critical property by execution, not inspection: routing `is_owner` through `did_matches` widens who counts as owner by exactly one case, the bare-key mirror owner matched by the same key's full `did:key:` caller (the #153 fix). Reverting `is_owner` makes `bare_owner_full_caller_allows_owner` fail, so it's load-bearing. Cross-method stays denied because only `did:key:` is stripped, so `did:gitlawb:zX` keeps its method and can't equal a bare id. Ran the full owner/DID truth table and the node suite (424 green). The short-circuit isn't exploitable by the widening: the newly-matched caller is the Ed25519-verified owner, and this converges the read gate onto the same `did:key`-aware normalization the mutation gate and SQL owner filter already use, so it reduces gate/query skew.

## Findings

- [P3] Pin the empty-DID invariant on the owner gate
  `crates/gitlawb-node/src/api/mod.rs:70`
  `did_matches("", "did:key:")` and `did_matches("", "")` return true. Unreachable today (the read-gate caller is always a verified `did:key` or None, and an empty-owner mirror slug is quarantined before the gate) and it predates this PR since `require_repo_owner` already uses `did_matches`. But this change makes `did_matches` load-bearing on the read path too, so a one-line test pinning "the gate caller is a validated `did:key` or None" is cheap insurance. Optional.

Concur with the approval; the note is defense-in-depth, not a blocker.

jatmn (approved):
LGTM

beardthelion (dismissed):
Re-approving on `73bd7a4`. My earlier approval was on `dce4330`; the only change since is the added `empty_did_matches` test, so I re-ran the owner-gate matrix by execution against the current head.

Routing `is_owner` through `did_matches` fixes the #153 mirror-owner lockout: a bare owner key with a full `did:key` caller now resolves to Allow, and reverting `is_owner` reproduces the lockout. The security-critical negatives hold under mutation. Cross-method (`did:gitlawb:` / `did:web:` against a bare key) denies, and a trailing-segment over-widen makes `cross_method_did_denied_with_bare_owner` go red, so that guard is load-bearing. A distinct full key denies. The empty-DID match is unreachable: an empty or bare `did:key:` id never resolves through `to_verifying_key`, so the caller reaching the owner arm is always a verified `did:key` or `None`. Quarantine still drops before the owner short-circuit (`authorize_repo_read` and the git-serve paths), so a newly bare-key-matched mirror cannot surface a hidden repo.

## Findings

- [P3] Pin the fail-safe negative in `empty_did_matches`. `crates/gitlawb-node/src/api/mod.rs:118`. The test pins the degenerate positives (`did_matches("","")`, `did_matches("","did:key:")`) but not the security-relevant negative: an empty `owner_did` must never match a real key. It is unreachable today, so this is not a live issue, but a future over-widening refactor would keep the test green. I confirmed a `strip + starts_with` matcher leaves this test passing, while adding `assert!(!did_matches("", "did:key:z6MkReal"))` turns it red. Consider adding that negative, and optionally an empty-id guard inside `did_matches` so the invariant lives at the matcher (the #153 tests still pass with the guard).

Optional and non-blocking. @kevincodex1 good to merge.

kevincodex1:
please fix conflicts

beardthelion (dismissed):
Verified by execution that this is behaviorally a no-op on the current base, not a live lockout fix. Routing is_owner through did_matches is identical to the two-side normalize_owner_key already on main after #157: I ran the owner gate across the clone/fetch path, list_repos, and subtree withholding, as the owner, as a cross-method did:gitlawb: caller, and anonymous, against both the did_matches body and the base body, with identical Allow/Deny outcomes. The code is safe and consolidating onto the shared matcher is a genuine improvement. The changes I'm asking for are to the framing and one test comment so the record stays accurate.

## Findings

- [P2] Fix the false "previously denied (lockout)" comment; the case already passes on the current base.
  `crates/gitlawb-node/src/visibility.rs:434`
  The lockout lived on the pre-#157 single-side is_owner. On this PR's base (26bc3f6, which already normalizes both sides) is_owner("z6MkOwner", "did:key:z6MkOwner") returns Allow, so reverting is_owner to the base body leaves both new tests green. Keep them as forward guards against a revert to single-side, but correct the comment (and the "fixes #153 lockout" assertion message on line 438) to say that, rather than claiming this diff lifts a lockout.

- [P3] Retitle from fix(...lockout) to refactor(...) so the changelog is accurate.
  `crates/gitlawb-node/src/visibility.rs:32`
  Since #157 already resolved the read lockout, a fix title records a changelog Bug Fix for a behavioral no-op. Suggest: refactor(visibility): route is_owner through the shared did_matches matcher (#153). It still legitimately closes #153, which also asked for the stale did_matches doc comment to be corrected, and this does that.

beardthelion (approved):
Re-reviewed on 96cd9e1d. The is_owner→did_matches reroute is a verified no-op, and my earlier request-changes was framing-only — its substantive ask is resolved on this head, so I've dismissed it and I'm approving.

Verified by execution on this head: did_matches is byte-identical to the old double-normalize_owner_key body across an adversarial truth table (cross-method, did:key:z:extra, empty, bare-vs-full) — zero divergence, so no visibility decision changes on any reader. Both new guards are load-bearing — reverting is_owner to single-side normalize reds bare_owner_full_caller_allows_owner, and to a method-blind segment match reds cross_method_did_denied_with_bare_owner. Cross-method stays denied, parity with the SQL owner key holds (normalize_owner_key_matches_sql_case), the owner short-circuit still serves the owner through the real feed reader, and the quarantine drop still precedes it (feed_quarantined_* stays withheld from the owner even with is_owner neutralized). The stale "lockout" comment is corrected.

## Findings

- [P3] Retitle the PR so the changelog stays accurate. `crates/gitlawb-node/src/visibility.rs` (#153). The commit is `refactor(visibility): …` but the PR title still reads `fix(visibility): … to fix bare-key mirror lockout (#153)`. Since #157 already lifted the read lockout, a squash-merge on the current title records a Bug Fix for a behavioral no-op. Retitle to `refactor(visibility): route is_owner through the shared did_matches matcher (#153)`, or squash using the commit subject.

@kevincodex1 good to merge once the title's sorted.

jatmn (approved):
LGTM

kevincodex1 (approved):

PR #157, opened 2026-07-06
fix(visibility): gate repo-scoped read surfaces on visibility (#120)

### LINKED ISSUE #120: Repo-scoped read surfaces gated by existence only, not visibility (certs, issues, labels, bounties, stars)

Gravirei:
## Summary

Add `authorize_repo_read` visibility gates to 8 repo-scoped read handlers that previously only checked repo existence, leaking private-repo metadata to anonymous callers.

## Motivation & context

Closes #120

Several repo-scoped read handlers verified the repo exists (`get_repo`) but never ran the visibility gate, so a private repo's ref certs, issues, labels, bounties, and star count were accessible without authentication.

## Kind of change

- [x] Security fix

## What changed

- `certs.rs`: `list_certs`, `get_cert` — gate ref certificate visibility
- `issues.rs`: `list_issues`, `get_issue`, `list_issue_comments` — gate issue visibility
- `labels.rs`: `list_labels` — gate label visibility
- `bounties.rs`: `list_repo_bounties` — gate bounty visibility (had no repo lookup at all)
- `stars.rs`: `get_star_status` — gate star count visibility
- `api/mod.rs`: removed gated handlers from `known_ungated` allowlist
- Test coverage for all handlers (anon 404 / owner admitted)

## How a reviewer can verify

```sh
cargo check -p gitlawb-node
cargo test -p gitlawb-node -- api::authz_guard
```

## Before you request review

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened private-repo visibility enforcement across repo-scoped read endpoints: certs, issues, issue comments, labels, repo bounties, and star count.
  * Anonymous callers are now denied with **404 Not Found**, while authorized owners can access listing and fetch routes.
  * Cert access is more consistent: requesting a cert that doesn’t belong to the authorized repo results in **404**.

* **Tests**
  * Added an integration test suite validating private-repo gating behavior (including **404** for anonymous access).


jatmn (approved):

@kevincodex1 LGTM

beardthelion (dismissed):
The gate additions on the 8 handlers are correct and the drift guard enforces them (I confirmed by removing `get_cert`'s gate: `every_repo_scoped_handler_is_gated` fails). But the PR does not close #120 for the bounty class, and it locks the owner out of the one bounty endpoint it does gate.

## Findings

- [P1] Gate `get_bounty` and `list_all_bounties` — the #120 bounty leak is still open
  `crates/gitlawb-node/src/api/bounties.rs:150` and `:137`
  `list_repo_bounties` is gated, but the same private-repo bounty metadata is served unauthenticated by `GET /api/v1/bounties/{id}` and `GET /api/v1/bounties`. Neither takes a caller or calls `authorize_repo_read`, and their route group `bounty_read_routes` (server.rs:240) has no `optional_signature` layer. `create_bounty` read-gates, so bounties legitimately sit on private repos carrying `repo_owner`/`repo_name`/`issue_id`/`title`/`creator_did`. I seeded a private repo with a bounty; an anonymous GET to both endpoints returned 200 with `"repo_name":"secret-repo"`, the issue id, title, and creator DID. Resolve each bounty's repo and run `authorize_repo_read` at `"/"` before returning; drop unreadable rows in the list. The drift guard cannot see these because they key on a global id, not `Path<(owner, repo)>`.

- [P2] `list_repo_bounties` denies the owner in production; the admit test masks it
  `crates/gitlawb-node/src/server.rs:240`
  `bounty_read_routes` is merged without `optional_signature` (and there is no global layer), so `list_repo_bounties` always sees `auth = None` and denies the owner on a private repo, not just anonymous callers. I confirmed with a real RFC-9421 signed owner request: 404 against the route with no layer (the production wiring), 200 with the layer added. The owner-admit test misses this because `signed_request_as` injects the `AuthenticatedDid` extension directly (test_support.rs:104) instead of routing through the middleware. Attach `optional_signature` to `bounty_read_routes` (or move the route into `read_routes`), and route the gate tests through the real middleware.

- [P2] Add cross-repo IDOR regression tests for the two fixes that lack them
  `crates/gitlawb-node/src/api/issues.rs:204` and `crates/gitlawb-node/src/api/certs.rs:57`
  The `list_issue_comments` issue-belongs-to-repo check and `get_cert`'s `cert.repo_id != record.id` guard each close a real cross-repo IDOR but neither is covered. I removed the comments check and the whole issue test set still passed; `get_cert`'s admit test uses a matching `repo_id`, so the `!=` branch never runs. Add a repo-A-authorized / repo-B-resource-id test asserting 404 for each.

- [P3] Cover the vacuous deny tests and skipped degenerate states
  `crates/gitlawb-node/src/test_support.rs`
  The `get_cert`/`get_issue`/`list_issue_comments` deny tests target a nonexistent id, so they 404 with or without the gate (I removed `get_cert`'s gate and its deny test still passed); seed a matching-`repo_id` resource so the 404 comes only from the gate. Also untested: a quarantined repo, a public repo anon read (the gate must not over-deny), and owner bare-key vs full did:key (#153/#156).

The gating direction is right and the drift guard is a solid backstop; the bounty leak is the blocker. @kevincodex1 holding this until the bounty readers are gated.

jatmn (approved):
@kevincodex1 LGTM

beardthelion (dismissed):
The gating direction is right and my prior blocker is resolved: all ten repo-scoped readers route through `authorize_repo_read`, the drift guard enforces the eight repo-scoped ones (I stripped `get_cert`'s gate and both its deny test and the drift guard go red), `is_owner` normalizes without cross-method collision, and the deny tests seed real resources so they're load-bearing. Two things block, and the two new global-id gates ship untested.

## Findings

- **[P1] Clamp `limit` and authorize each repo once in `list_all_bounties` — the diff makes an anonymous GET an unbounded DB scan**
  `crates/gitlawb-node/src/api/bounties.rs:137`
  The handler fetches `q.limit` rows (unclamped: `list_bounties` binds it straight into `LIMIT`, db/mod.rs:2964) then calls `authorize_repo_read` per row, each running three sequential queries (`get_repo` + `is_repo_quarantined` + `list_visibility_rules`, mod.rs:45-57). The route carries only `optional_signature` and the outer layer is `TraceLayer`, no rate limit, so `GET /api/v1/bounties?limit=<huge>` costs `1 + 3N` serial queries for an unauthenticated caller, N growable by anyone who can `create_bounty`. Pre-PR this was one query. A negative `limit` also reaches Postgres and 500s (confirmed). Clamp with the `raw_limit.clamp(1, 200)` pattern (repos.rs:313) and memoize the per-repo decision. Same clamp applies to `list_repo_bounties`.

- **[P2] Return the identical 404 body in `get_bounty` on an authz denial — it currently leaks the private repo's owner/name**
  `crates/gitlawb-node/src/api/bounties.rs:150`
  `get_bounty` resolves the bounty first (`NotFound("bounty {id} not found")` when absent) then authorizes (`RepoNotFound("{owner}/{name}")` when private/quarantined). Both are 404 but the bodies differ — I drove it: a missing id returns `{"error":"not_found",…}` while an id on a private repo returns `{"error":"repo_not_found","message":"repository 'did:key:…/secret-repo' not found"}`, so an anonymous holder of a bounty id learns it maps to a private repo and reads that repo's owner/name. `get_cert` in this same PR already returns a byte-identical `NotFound("certificate {id}")` on both branches (certs.rs:55,58) — mirror that.

- **[P2] Add gate tests for `get_bounty` and `list_all_bounties` — the two new confidentiality branches have none**
  `crates/gitlawb-node/src/api/bounties.rs:137`
  These are the global-id handlers `authz_guard` cannot see (it keys on `Path<(owner, repo)>`), so no automated backstop; a crate-wide grep finds zero tests for either. `list_all_bounties`'s `if authorize_repo_read(...).is_ok()` filter is the only thing keeping private-repo bounties out of the anonymous global feed. Add an anon-deny test for `get_bounty` on a private repo and a `list_all_bounties` filter test (one private + one public bounty, assert anon sees only the public one), driven through `build_router` + a real signature like `list_repo_bounties_gate_admits_owner_on_private` (test_support.rs:2963).

- **[P3] `list_issue_comments` resolves an issue-id prefix for the existence check but fetches comments by the exact string**
  `crates/gitlawb-node/src/api/issues.rs:204`
  The belongs-to guard calls `git_issues::get_issue`, which resolves an 8-char prefix to the full id, but the next line does `list_issue_comments(&issue_id)` with the raw param. I drove it: a prefix passes the gate then returns `{"comments":[]}` for an issue that has a comment, and an ambiguous prefix 500s instead of 404s. Bind the DB fetch to the id `get_issue` returns.

- **[P3] `list_all_bounties` filters after the limit, so a non-owner can get zero of many readable bounties**
  `crates/gitlawb-node/src/api/bounties.rs:137`
  It fetches the newest `limit` rows then drops unreadable ones with no cursor. I drove it: with three newest private bounties and one older public bounty, an anon `?limit=3` returns `{"bounties":[]}` even though the public one is readable. Paginate by a stable cursor and filter before the page limit.

Separate follow-up issue (out of this PR's scope): the bounty mutation endpoints (`claim`/`submit`/`approve`/`cancel`/`dispute`) fetch by global id and return the full `BountyRecord` with no `authorize_repo_read`, so `claim_bounty` discloses a private repo's owner/name/title to any authenticated agent — the same leak class this PR closes on the read side.

@kevincodex1 holding for the amplification and the get_bounty error-channel.

jatmn (changes requested):
Thanks for the updates. I rechecked the current bounty/read-gate paths and found issues that still need to be addressed.

## Findings

- [P1] Do not treat a denied repo as allowed after the first bounty
  `crates/gitlawb-node/src/api/bounties.rs:154`
  The global bounty feed records only that a repo key has been seen, not whether `authorize_repo_read` admitted that repo. When the first bounty from a private repo is denied, `seen.insert(key)` still runs; the next bounty from that same private repo then hits `seen.contains(&key)` and is pushed into `allowed` without any authz check. An anonymous `GET /api/v1/bounties` can therefore leak every later bounty from a private repo that has more than one bounty in the fetched window. Please memoize the authorization result per repo, not just the repo key, and add a regression with two private bounties from the same repo.

- [P2] Keep paging past hidden bounties instead of stopping at the first 1000 rows
  `crates/gitlawb-node/src/api/bounties.rs:145`
  The over-fetch fix still applies the SQL `LIMIT` before visibility filtering: with `limit=200`, the handler fetches at most the newest 1000 rows, filters private rows in memory, and never looks beyond that window. If those 1000 rows are private and an older public bounty exists, anonymous callers get an empty/short page even though readable bounties are available. Please page/scan with a bounded loop until the requested number of visible rows is collected, or otherwise move the visibility decision ahead of the page limit, and cover the hidden-window case in tests.

beardthelion (dismissed):
jatmn already has the two I'd otherwise lead with, and I re-verified both by execution: the `list_all_bounties` memoize leak (P1, `bounties.rs:154`) still reproduces on this head (two bounties on one private repo, an anonymous `GET /api/v1/bounties` returns the second), and it clears with a per-repo decision cache; the paging window (P2, `bounties.rs:145`) is real too. The prior round's other fixes hold: `get_bounty` no longer leaks the repo identity on denial and the `limit` clamp kills the negative-limit 500, both confirmed by reverting them and watching the tests go red.

One blocker those reviews don't cover, plus a note on the paging fix.

## Findings

- **[P2] Gate `star_repo` and `unstar_repo` on repo read, the same as `get_star_status`**
  `crates/gitlawb-node/src/api/stars.rs:17` and `:52`
  This PR gated the star-count read (`get_star_status`) and the module doc now says "Star count is gated on repo read access," but both write handlers still do a plain `get_repo` and then return `star_count`, confirm the repo exists, and insert or delete a star on a repo the caller cannot read. Driving a signed non-reader at a private repo: `PUT /api/v1/repos/{owner}/{repo}/star` returns `201 {"star_count":1,"status":"starred"}` and `DELETE` returns `200`, so any self-registered keypair gets an existence oracle, the star count, and an unauthorized write. Swap the `get_repo` lookup in both handlers for `authorize_repo_read(&state, &owner, &repo, Some(caller), "/")`; the non-reader then gets the same 404 as an absent repo, and the owner is still admitted (I checked both directions). The doc comment overclaims until this lands.

## Note on the paging fix (jatmn's P2, `bounties.rs:145`)

A direction, not a measured claim: prefer moving the visibility filter into the query over a bounded scan-until-N-visible loop. `list_all_bounties` already over-fetches up to 1000 rows and runs one `authorize_repo_read` (three sequential queries) per distinct repo, so a single anonymous GET can cost on the order of 1000 fetched rows plus ~3000 serial authz queries. A loop that keeps fetching until it has collected the requested number of visible rows can amplify that further, whereas filtering to the caller's readable repos in SQL fixes the window and the cost together.

The gating direction is right and the drift guard is a good backstop. Out of scope for #120 but worth its own issue: the bounty mutation endpoints (`claim`, `submit`, `approve`, `cancel`, `dispute`) fetch by global id and return the full `BountyRecord` with no read gate, the same leak class on untouched write handlers.

jatmn (changes requested):
Thanks for the updates. I rechecked the current visibility-gated bounty paths and found issues that still need to be addressed.

## Findings

- [P1] Sign the CLI and MCP reads that now require repo visibility
  `crates/gl/src/issue.rs:218`, `crates/gl/src/cert.rs:71`, `crates/gl/src/bounty.rs:196`, `crates/gl/src/repo.rs:661`, `crates/gl/src/star.rs:118`, `crates/gl/src/mcp.rs:973`
  The server handlers now correctly use `optional_signature` plus `authorize_repo_read`, but the real clients for those reads were not updated to send a signature. `gl issue list/show/comments`, `gl cert list/show`, `gl bounty list/show`, `gl bounty show`, `gl repo label-list`, `gl star count`, and the issue section in `gl status` all build unsigned clients and call `.get(...)`; the MCP tools build a client with the loaded keypair but still use unsigned `.get(...)` for the newly gated issue/bounty reads. As a result, the owner/authorized-reader cases covered by the server tests are not reachable through the shipped CLI/MCP for private repos: they arrive as anonymous and get the same 404 as a stranger. Please load the identity for these gated reads and call `get_signed` when a keypair is available, while keeping anonymous reads for public repos working.

- [P2] Move global bounty visibility filtering ahead of the scan loop
  `crates/gitlawb-node/src/api/bounties.rs:147`
  The current fix follows the bounded scan-until-visible direction, but the loop still stops after scanning 10,000 rows and then returns whatever visible rows it found with no cursor, `has_more`, or incomplete marker. With 10,000 newer hidden bounties and an older public bounty, an anonymous `GET /api/v1/bounties?limit=1` still returns a complete-looking empty page even though a readable bounty exists; the new regression covers six hidden rows, not the capped path. Please push the readable-repo predicate into the query, or otherwise filter before applying the page limit, so the endpoint cannot present partial results as complete. If the implementation keeps a bounded scan loop instead, return an explicit incomplete/failure state when the cap is reached before filling the requested page.

beardthelion (dismissed):
jatmn's two findings hold and I verified both: the unsigned CLI/MCP reads (`NodeClient::new(node, None)` + `.get()`, so an authorized owner arrives anonymous and 404s on their own private repo; `get_signed` already exists at `crates/gl/src/http.rs:50`), and the capped-scan completeness gap. The read-surface core is solid and I confirmed it by execution: the memoize decision-cache, the multi-page loop, and the star/unstar gate each go RED when neutralized and GREEN on this head, and every `BountyRecord` reader is gated (`get_bounty` 404, `list_repo`/`list_all` per-row) or aggregate-only.

Two things jatmn's findings don't cover:

## Findings

- [P2] Gate `claim_bounty` on repo read
  `crates/gitlawb-node/src/api/bounties.rs:211`
  A permissionless signed non-reader POSTing `/api/v1/bounties/{id}/claim` on a private repo's open bounty gets a 200 with the full record (`repo_owner`/`repo_name`/`title`/`creator_did`) and becomes its claimant. It is the same data class this PR closes on the read side, reached through a mutation; pre-existing rather than introduced here, but it belongs with the rest of #120. Mirror `get_bounty`: on `authorize_repo_read(...).is_err()` return `NotFound("bounty {id} not found")`. A bare `authorize_repo_read(...)?` is not enough on its own; its `RepoNotFound` body still echoes `owner/repo`.

- [P2] Fix the OFFSET paging instability while reworking the scan
  `crates/gitlawb-node/src/api/bounties.rs:135`, `crates/gitlawb-node/src/db/mod.rs:2959`
  Separate from jatmn's cap point: `list_bounties` orders by `created_at DESC` with no `id` tie-break, and the new loop pages by growing OFFSET, so an ordinary concurrent insert shifts every later offset and a readable bounty is duplicated or skipped across the page boundary (reproduced: page 1 `[b00..b09]`, page 2 `[b09, ...]`). The same loop is also an anonymous cost lever: no rate limiter, up to `max_rows=10_000` rows scanned via deep OFFSET plus up to ~10k `authorize_repo_read` calls per request. The repo already has the pattern that fixes cap-completeness, instability, and cost together: `list_ref_updates_keyset` (`(timestamp, id) < (after_ts, after_id)`, `ORDER BY timestamp DESC, id DESC`). Reusing it resolves this and jatmn's P2 in one change.

Once claim is gated and the feed moves to the keyset cursor, this is close.

jatmn (approved):
Thanks for the updates. I rechecked the previously discussed paths and do not see any remaining actionable issues from my side.

@beardthelion LGTM

beardthelion (approved):
Re-reviewed the new head 0be5e42d against my earlier blockers and jatmn's. All five are resolved, and I verified each by execution against a local Postgres rather than by reading.

- `claim_bounty` now gates on `authorize_repo_read` before the status check and the mutation, and maps an authz failure to the same `bounty {id} not found` so the body carries no repo path (`bounties.rs:245`). A signed non-reader gets 404 with no owner/name/title and no status flip; reverting the gate turns that into 200 plus a real claim, so the gate is load-bearing.
- `list_all_bounties` moved from OFFSET to a `(created_at, id)` keyset and the scan cap now returns 422 `Incomplete` instead of a complete-looking partial page (`bounties.rs:158`, `db/mod.rs`). Confirmed the keyset is stable across a page boundary under a concurrent insert where OFFSET duplicated a row, and that a feed with 10k+ hidden rows returns 422, not a partial 200.
- `?limit` is clamped on both bounty read handlers; a negative limit no longer reaches Postgres as `LIMIT -1` (500).
- The gated reads now sign via `get_authed`, so an owner reaches their own private-repo bounty (signed 200) where an anon gets 404. The net-new gates in this PR (bounties/certs/issues/labels/stars) all have their CLI/MCP clients converted, and the drift guard was tightened correctly (the eight `known_ungated` entries removed now that their gates landed).

Non-blocking follow-ups, none of which should hold this PR:

- The list-style clients (`bounty/issue/cert list`, `label-list`, `issue comments`) parse the body without checking `resp.status()`, so a transient node 5xx renders as "No X" and exits 0. The single-item `show` paths already check status; worth aligning the list paths.
- `signed_client` uses `load_keypair_from_dir(dir).ok()`, which collapses "no identity" (fine, unsigned public read) and "identity present but unreadable" (a real error) into an unsigned request. On a caller-gated read the second case surfaces as a false "not found" rather than an identity error.
- The scan cap can return 422 for an honest query on a very large feed when the caller's visible rows sit past the window. Exposing the internal keyset cursor as a client page param would let those queries page instead.
- Two behaviors here ship without a test: the `claim_bounty` gate and the 422 cap path. Both are quick to add.

@kevincodex1 good to merge once you're happy with the above.

kevincodex1 (approved):
nice catch

PR #161, opened 2026-07-07
fix(node): gate POST /api/v1/sync/trigger and rate-limit the peer-sync routes (#82)

### LINKED ISSUE #82: Unauthenticated POST /api/v1/sync/trigger drives outbound fan-out and queue flooding
`POST /api/v1/sync/trigger` is reachable unauthenticated in the default configuration and drives the node to fan out HTTP requests to every known peer and enqueue per-repo sync work, with no caller identity check or effective rate limit.

### Where
- `require_signed_peer_writes` defaults to false (`crates/gitlawb-node/src/config.rs:48`).
- With the flag false, the route is gated only by `auth::optional_signature` (`server.rs:248`), which passes unsigned requests through (`auth/mod.rs:239`).
- `trigger_sync` (`crates/gitlawb-node/src/api/peers.rs:184`) takes no auth extractor and no body, loops over all peers issuing outbound GETs to `{peer}/api/v1/repos`, and enqueues a sync row per returned repo.
- The route is publicly merged (`server.rs:248`). The per-DID rate limiter (`rate_limit.rs`) keys on the authenticated DID and early-returns when none is present, so it does nothing for an unauthenticated caller.

### Impact
An anonymous caller can repeatedly force outbound fan-out and flood the sync queue (amplification/abuse). Combined with the outbound-redirect behavior and any poisoned peer row, it also becomes an unauthenticated trigger for the SSRF path.

### Suggested fix
Require a valid signature on `sync/trigger` even when `require_signed_peer_writes` is false. (A "known-peer" check doesn't fit here: the request carries no peer identity and pulls from all peers.) Rate-limiting only helps once auth is enforced, since the limiter keys on the authenticated DID, so the two changes are coupled rather than independent. The sole caller, `gl sync trigger`, already signs when an identity is present, so requiring a signature is low-collateral. `notify_sync` is acceptable as-is since it gates on a known-peer DID.

beardthelion:
`POST /api/v1/sync/trigger` was reachable unauthenticated in the default config and drove an O(peers) outbound fan-out plus one `sync_queue` enqueue per repo, an anonymous amplification/DoS (#82). The same `enqueue_sync` sink is also reachable through `/sync/notify`, which accepts unsigned known-peer requests and had no rate limit, so gating only trigger would have left an equivalent hole.

What changed:

- `/sync/trigger` now requires a valid RFC 9421 signature in both config modes (its own always-signed route group), independent of `require_signed_peer_writes`.
- Two dedicated per-client-IP rate limiters: a tight one (60/hr) on trigger, a generous one (600/hr) on the `announce`+`notify` group. Per-IP rather than per-DID, because a did:key farm self-registers and never trips a per-DID bucket; the brake is keyed before auth so DID rotation can't bypass it. `notify`'s signature behavior is unchanged for rolling-upgrade compat, it only gains the brake. Knobs: `GITLAWB_SYNC_TRIGGER_RATE_LIMIT`, `GITLAWB_PEER_WRITE_RATE_LIMIT` (0 disables).
- `gl sync trigger` now checks HTTP status before parsing (a 401/429 can no longer render as a fake "sync triggered" success), bounds the error-body read, and strips control and Unicode bidi characters from the node's error before printing.

Known limitation: the per-IP brake caps a single-source flood. A distributed multi-IP notify flood stays open until `/peers/announce` is gated (the free peer self-registration that enables it), which is tracked separately.

Tests are an adversarial RED to GREEN matrix: unsigned to 401 in both modes, IP flood to 429, DID-farm and forged-header resistance, cross-bucket isolation, disabled-when-0, the notify and announce brakes, the real-signature positive through the full router, and the client 401/429/sanitize/oversized-body paths.

Closes #82.

## Summary by CodeRabbit

* **New Features**
  * Added separate, configurable per-client rate limits for `POST /api/v1/sync/trigger` and peer-write routes (`/api/v1/peers/announce`, `/api/v1/sync/notify`), controlled via new CLI/env settings (including optional disable via `0`).
  * Split `sync/trigger` into its own authenticated route with dedicated throttling.
* **Bug Fixes**
  * Improved `gl sync trigger` to fail locally when no identity is configured and to surface server errors correctly.
  * Rate-limit responses now use a single generic 429 message for relevant sync/peer routes.
* **Documentation**
  * Updated API surface and CLI usability notes to reflect `sync/trigger` signature requirements and staged-rollout behavior.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -95,6 +95,16 @@ GITLAWB_MAX_PACK_BYTES=2147483648
 # IP per hour. 0 disables. Default 600.
 GITLAWB_PUSH_RATE_LIMIT=600
 
+# ── Peer-sync rate limiting (per client IP, uses GITLAWB_TRUSTED_PROXY below) ─
+# /api/v1/sync/trigger requires a signature and fans out to every peer per call,
+# so it gets a tight bucket. 0 disables. Default 60.
+GITLAWB_SYNC_TRIGGER_RATE_LIMIT=60
+# /api/v1/peers/announce and /api/v1/sync/notify accept unsigned requests from
+# known peers and run at higher frequency, so a generous bucket. Separate from
+# the trigger bucket so an unsigned notify flood can't drain trigger's quota.
+# 0 disables. Default 600.
+GITLAWB_PEER_WRITE_RATE_LIMIT=600
+
 # Which forwarded header the edge is trusted to set, used to resolve the real
 # client IP for the push limiter. One of:
 #   (unset) — no trusted proxy: key on the socket peer address, ignore headers.
--- a/.env.example
+++ b/.env.example
@@ -96,14 +96,14 @@ GITLAWB_MAX_PACK_BYTES=2147483648
 GITLAWB_PUSH_RATE_LIMIT=600
 
 # ── Peer-sync rate limiting (per client IP, uses GITLAWB_TRUSTED_PROXY below) ─
-# /api/v1/sync/trigger requires a signature and fans out to every peer per call,
-# so it gets a tight bucket. 0 disables. Default 60.
-GITLAWB_SYNC_TRIGGER_RATE_LIMIT=60
 # /api/v1/peers/announce and /api/v1/sync/notify accept unsigned requests from
 # known peers and run at higher frequency, so a generous bucket. Separate from
 # the trigger bucket so an unsigned notify flood can't drain trigger's quota.
 # 0 disables. Default 600.
 GITLAWB_PEER_WRITE_RATE_LIMIT=600
+# /api/v1/sync/trigger requires a signature and fans out to every peer per call,
+# so it gets a tight bucket. 0 disables. Default 60.
+GITLAWB_SYNC_TRIGGER_RATE_LIMIT=60
 
 # Which forwarded header the edge is trusted to set, used to resolve the real
 # client IP for the push limiter. One of:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2293,9 +2293,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+checksum = "2d6914041f254d6e9176c01941b21115dcfb7089e55135a35411081bd106ef3f"
 dependencies = [
  "crossbeam-utils",
 ]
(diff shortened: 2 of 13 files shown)


jatmn (changes requested):

## Findings

- [P2] Update the peer-write rollout docs for the always-signed trigger route
  `README.md:305`
  The README still lists `POST /api/v1/sync/trigger` under the staged peer-write rollout and says unsigned legacy peers are accepted when `GITLAWB_REQUIRE_SIGNED_PEER_WRITES=false`; `docs/OSS-READINESS-AUDIT.md` also still says `gl sync trigger` falls back to an unsigned request for live compatibility. This PR intentionally moves `/sync/trigger` into its own always-signed route group, so operators following the docs will expect the flag to preserve unsigned trigger compatibility even though the route now returns 401 without a signature. Please split `/sync/trigger` out of the staged peer-write list and document that only `/peers/announce` and `/sync/notify` keep the rolling-upgrade unsigned behavior.

- [P2] Stop silently sending unsigned `gl sync trigger` requests
  `crates/gl/src/sync.rs:33`
  `gl sync trigger` still loads the keypair with `.ok()` and then builds `NodeClient::new(&args.node, keypair)`, so a missing or unreadable identity is silently converted into an unsigned request. Since this PR makes `/api/v1/sync/trigger` always require a signature, that path can only fail remotely with a 401 and keeps the legacy unsigned fallback alive in the client instead of telling the operator to create or fix their identity before calling the route. Please make `Trigger` require a keypair locally, as the other signed CLI writes do, and cover the missing-identity path so the command fails before issuing an unsigned request.

beardthelion:
@jatmn both addressed in c84c703:

- Docs: split `/sync/trigger` out of the staged unsigned peer-write list in the README and dropped the unsigned-fallback line in `docs/OSS-READINESS-AUDIT.md`. Only `/peers/announce` and `/sync/notify` keep the rolling-upgrade unsigned behavior now.
- Client: `gl sync trigger` loads the identity with `?` like the other signed writes and fails locally with "identity not found" before issuing a request. Added `trigger_requires_identity_fails_before_request`, which points at an unreachable node so it can only pass if the command bails before dialing.

Also bumped crossbeam-epoch to 0.9.20 (RUSTSEC-2026-0204) so cargo audit is green again.

beardthelion:
@coderabbitai the `.env.example` ordering is fixed (`GITLAWB_PEER_WRITE_RATE_LIMIT` now precedes `GITLAWB_SYNC_TRIGGER_RATE_LIMIT`). Leaving the limiter-setup dedup helper and the server.rs route-grouping note for a separate cleanup rather than widening this security PR's diff near approval.

jatmn (approved):

kevincodex1 (approved):
LGTM

PR #168, opened 2026-07-09
test(gl): add tests for iCaptcha transparent retry flow (#167)

### LINKED ISSUE #167: Registration hard-blocked for new users: node v0.5.0 requires iCaptcha proof but released gl v0.3.8 predates icaptcha-client
### What happened?

Fresh install, new user, following https://gitlawb.com/start exactly.

Repro:
1. Install gl v0.3.8 (latest release) on Linux (WSL2 Ubuntu)
2. gl identity new → OK
3. gl register → 403: "iCaptcha proof required (missing proof header). Solve a challenge at https://icaptcha.gitlawb.com for level >= 3 and resend with the x-icaptcha-proof header."

Cause: node v0.5.0 gates register/repo-create behind iCaptcha, but v0.3.8 (current latest release) contains no icaptcha-client — verified via strings on the release binary: no x-icaptcha-proof handling. The transparent solve-and-retry flow exists on main (crates/icaptcha-client, crates/gl/src/http.rs) but hasn't shipped.

Impact: every new identity is blocked from registering with the public node. Likely affects everyone starting via the Hy3 builder bounty today.

Fix: cut a release from main. Workaround in the meantime: build gl + git-remote-gitlawb from source.

DID that hit this: did:key:z6MksApMMMq14ycCqrviT7o3byEymutQNBEM6BMRgwTnLEo1

### Expected behavior

Following https://gitlawb.com/start on a fresh install: gl register should register the new identity with the public node (solving the iCaptcha challenge automatically, as the icaptcha-client code on main does).

### Steps to reproduce

1. Install gl v0.3.8 (latest release) on Linux (WSL2 Ubuntu, x86_64)
2. gl identity new → OK
3. gl register → 403 icaptcha_proof_required

### Affected crate

gitlawb-node

### Version or commit

gl 0.3.8 (latest release binary from Gitlawb/releases, gitlawb-v0.3.8-x86_64-unknown-linux-musl.tar.gz)

### Environment

_No response_

### Relevant logs

```shell

```

Gravirei:
## Summary

Add comprehensive tests for the iCaptcha transparent solve-and-retry flow in the `gl` CLI HTTP client, covering the critical registration path that was blocking new users.

## Motivation & context

Closes #167

Node v0.5.0 requires iCaptcha proof for registration, but the released gl v0.3.8 predates the icaptcha-client code. The fix existed on `main` (transparent solve-and-retry in `crates/gl/src/http.rs`, `crates/icaptcha-client`) but lacked test coverage for the retry flow.

## Kind of change

- [ ] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [x] Tests / CI

## What changed

- **`crates/icaptcha-client/src/lib.rs`**: Added `#[cfg(test)]` support in `is_https` to allow localhost/127.0.0.1 URLs in test builds, enabling mockito-based integration tests for the full iCaptcha retry path.
- **`crates/gl/src/http.rs`**: Added 16 tests:
  - 7 unit tests for `icaptcha_cfg` (header parsing, defaults, keypair required)
  - 4 tests for `send_once` (proof header attachment, signing)
  - 3 tests for `send_signed` (non-iCaptcha 403 passthrough, success, 405 handling)
  - 2 full integration tests using dual mock servers (node + iCaptcha): happy path retry to 201, and retry exhaustion returning 403

## How a reviewer can verify

```sh
cargo test -p gl -p icaptcha-client
cargo clippy -p gl -p icaptcha-client
cargo fmt -p gl -p icaptcha-client --check
```

All 266 tests in `gl` and 12 tests in `icaptcha-client` pass with no warnings.

## Before you request review

- [x] `cargo test --workspace` passes locally (node DB-dependent tests are pre-existing PG auth failures)
- [x] Commit titles use Conventional Commits (`test(gl): ...`)

## Notes for reviewers

The two `send_signed` integration tests (`send_signed_solves_icaptcha_and_retries_to_success`, `send_signed_returns_403_after_icaptcha_retries_exhausted`) use `std::env::set_var` for `GITLAWB_ICAPTCHA_URL`. An RAII guard (`IcaptchaEnv`) ensures cleanup. Tests use `#[tokio::test]` default single-threaded runtime, so env var races are not a concern.

## Summary by CodeRabbit

* **Bug Fixes**
  * Profile updates now save/persist content under the correct identifier, improving reliability.
  * Improved iCaptcha and signed request handling: captcha proofs and signing headers are applied consistently, and retry behavior correctly follows iCaptcha challenge outcomes.
  * In insecure mode, HTTP loopback endpoints (`localhost`/`127.0.0.1`) are supported, with stricter origin matching (including port) to avoid trusting an unintended solver.


beardthelion (dismissed):
Thanks for adding coverage here. The blocker is that the two "full integration" tests don't exercise the iCaptcha retry path they're named for — they fall through to the real production service — so the transparent-retry flow has no real coverage.

I verified by execution: the mockito iCaptcha server is never contacted, and the solve resolves to DEFAULT_URL (https://icaptcha.gitlawb.com) and makes a live network call. Root cause: is_https gates the localhost allowance behind #[cfg(test)], but cfg(test) is not set for icaptcha-client when it's compiled as gl's dependency, so that branch is compiled out during `cargo test -p gl`. With is_https(http://127.0.0.1:port) false, resolve_solver_url drops both the operator env URL and the advertised header URL and falls back to DEFAULT_URL. Pointing DEFAULT_URL at a bogus host makes both tests fail with a DNS error on {DEFAULT_URL}/v1/challenge; removing the relaxation entirely leaves the tests passing, confirming it's inert.

## Findings

- [P1] Make the integration tests exercise the mock, not the live service. `crates/gl/src/http.rs` (send_signed_solves_icaptcha_and_retries_to_success, send_signed_returns_403_after_icaptcha_retries_exhausted). Both resolve the solve to https://icaptcha.gitlawb.com and make a real network call — the mock's /v1/challenge and /v1/answer are never hit, and with DEFAULT_URL pointed at a bogus host both fail with a DNS error. So they're network-dependent (fail in a no-egress CI, hit production on every run) and the retry flow is uncovered; the exhaustion test even panics on the live-call error before its assert_eq!(403), so retry-exhaustion is untested. The #[cfg(test)] approach can't work because cfg(test) doesn't propagate to a dependency. Test the localhost solve inside icaptcha-client's own #[cfg(test)] module instead, where the relaxation is live — I confirmed the localhost path is reachable there; alternatively inject the base URL / reqwest client into gl's solve path.

- [P2] Assert the mocks and the retry count. `crates/gl/src/http.rs`. Both tests bind their mocks with `_` and never call .assert(); the exhaustion test comments "3 node calls" but sets no .expect(3). That silence is exactly why the live-service fallthrough passes unnoticed — add .assert()/.expect(N) so an unreached mock fails loudly.

- [P2] Drop or rework the is_https #[cfg(test)] relaxation. `crates/icaptcha-client/src/lib.rs`. It's inert where it's meant to help (removing it changes no test) and nothing depends on it. It does not reach release builds — I confirmed a non-test release build rejects a localhost URL and falls back to DEFAULT_URL — so it's not a security hole, just dead code that implies coverage which isn't there.

- [P3] The clippy borrow fix in `crates/gitlawb-node/src/api/profiles.rs` is unrelated to a test(gl) PR and duplicates the identical one-liner already in #156 and #134; drop it here to avoid a conflict/no-op once one of those merges.

Happy to re-review once the tests drive the mock.

beardthelion (dismissed):
Approving. The blocker from the last round is genuinely fixed: the two integration tests now exercise the iCaptcha retry flow against their mock instead of falling through to the live service. I confirmed it by poisoning `DEFAULT_URL` to a bogus host and re-running — both tests stay green, so the mock is actually consumed and no live call happens; removing the new `GITLAWB_ICAPTCHA_INSECURE` branch makes them fail with a DNS error on that poisoned host, so the flag is load-bearing. The `.expect(N)` + `.assert()` on every mock closes the earlier no-assertions gap. The runtime flag is contained: with it set, a hostile `x-icaptcha-url: http://127.0.0.1` (or `169.254.169.254` / `[::1]` / a userinfo trick) is still rejected and falls back, and the API key stays with the operator origin — I drove those cases through `resolve_solver_url` and `IcaptchaCfg::new`.

Three small non-blocking follow-ups:

## Findings
- [P3] Tighten the flag's presence check — `crates/icaptcha-client/src/lib.rs:97`. `std::env::var_os("GITLAWB_ICAPTCHA_INSECURE").is_some()` treats any value as enabled, so `GITLAWB_ICAPTCHA_INSECURE=0` or `=false` still turns the insecure loopback path on. A truthy check (`== "1"` / parse a bool) would honor disable-intent.
- [P3] The relaxation is now reachable in a non-test build, not just tests — `crates/icaptcha-client/src/lib.rs:90-104`. I verified it's inert without a localhost operator URL (a hostile node can't reach it, no key leaks), so this is fine as a default-off escape hatch. If you'd rather keep insecure code out of the shipped binary entirely, injecting the base URL/client at a test seam in `gl` would avoid the runtime flag.
- [P3] `crates/gitlawb-node/src/api/profiles.rs` — the clippy borrow tweak is unrelated to this PR and is the same one-liner as #169; it'll no-op once either merges.

Core is solid. @kevincodex1 good to merge once you're happy with it.

jatmn (changes requested):
I found two issues that need to be addressed before this is ready.

## Findings

- [P2] Keep a loopback test origin from trusting a different loopback port
  `crates/icaptcha-client/src/lib.rs:97`
  With `GITLAWB_ICAPTCHA_INSECURE=1`, an operator URL of `http://localhost:3000`, and `GITLAWB_ICAPTCHA_API_KEY` configured, a hostile node can advertise `x-icaptcha-url: http://localhost:9000`. The new branch accepts both URLs, while `resolve_solver_url` allowlists and marks the selected endpoint key-trusted using only `host_str`; it therefore chooses `:9000` and sends the operator's bearer key there. Compare normalized origins (including the effective port) before honoring an insecure advertised URL, or keep advertised HTTP URLs untrusted, and add the distinct-port regression case.

- [P2] Assert that the retry uses the proof produced by the solver
  `crates/gl/src/http.rs:503`
  The success-path node mock accepts any `x-icaptcha-proof`, so this test remains green if `send_signed` substitutes a stale or unrelated nonempty token instead of the `mock.proof` returned by the iCaptcha answer mock. Match `x-icaptcha-proof: mock.proof` here so the new full-flow test makes the solve-to-retry value transfer load-bearing.

kevincodex1:
hello @Gravirei thanks for the contribution please address @jatmn comments

jatmn (changes requested):

## Findings

- [P3] Restore the inherited iCaptcha test environment after each mock test
  `crates/gl/src/http.rs:428`
  `crates/icaptcha-client/src/lib.rs:437`
  Both new guards replace process-wide iCaptcha variables and then unconditionally remove them on drop. A `cargo test` run launched with `GITLAWB_ICAPTCHA_URL` or `GITLAWB_ICAPTCHA_INSECURE` set therefore loses its original configuration after these tests, which can make later in-process tests (including consumers of the configured URL) order-dependent. Save the prior `OsString` values before installing the mock settings, then restore each value on drop (removing it only when it was originally absent).

jatmn (changes requested):

## Findings

- [P2] Make the negative `send_once` checks reject the headers they prohibit
  `crates/gl/src/http.rs:305`
  `send_once_omits_proof_header_when_not_provided` and `send_once_does_not_sign_when_no_keypair` create unconstrained Mockito handlers, so both tests still receive 200 if `send_once` starts attaching `x-icaptcha-proof` with `None` or emits `Signature`, `Signature-Input`, and `Content-Digest` without a keypair. Match those headers with `mockito::Matcher::Missing` so the newly added negative cases actually protect the unsigned/no-proof contract.

beardthelion (approved):
Approving at 58769e7d. jatmn's two open findings are both addressed here, and I confirmed each holds under a break rather than just that the suite is green.

- The negative `send_once` tests now protect the contract: attaching `x-icaptcha-proof` when none was provided, or emitting the signature headers with no keypair, makes each test fail (the `Matcher::Missing` matchers reject the request). Before this they passed regardless — that was the P2.
- The retry integration tests exercise the real path: disabling the retry drops `send_signed_solves_icaptcha_and_retries_to_success` to a 403, and raising the retry cap breaks the exhaustion test's call count. A retry-exhausted 403 is returned to the caller (register reads `!status.is_success()` and bails), and an unsolvable challenge surfaces as an `Err`, not a success.
- The env guards now restore an inherited `GITLAWB_ICAPTCHA_URL`/`INSECURE` on drop instead of deleting it (the P3 from the prior round).

Every denial path is surfaced to the caller rather than rendered as empty/success: retry-exhausted → `Ok(403)` surfaced, unsolvable → `Err`, no iCaptcha headers → `Ok(403)` surfaced. The loop is capped and origin trust compares scheme+host+port, so a hostile advert or a different loopback port can't redirect the proof or key.

Three optional follow-ups, none blocking:
- [P3] The status half of the "only a 403 retries" gate isn't pinned — the header gate short-circuits it, so a non-403 carrying `x-icaptcha-*` isn't exercised.
- [P3] No end-to-end test for the level-only advert (`x-icaptcha-level` present, `x-icaptcha-url` absent).
- [P3] A panic while holding the env lock surfaces a sibling test as a poison error rather than its real assertion; `lock().unwrap_or_else(|e| e.into_inner())` fixes the ergonomics.

One housekeeping item: the base is behind main and the one-line `profiles.rs` change is already identical in main, so please rebase before merge — it resolves to a no-op, no gate dropped.

@kevincodex1 good to merge once rebased.

kevincodex1 (approved):
lgtm

PR #170, opened 2026-07-10
Fix/db startup resilience

### LINKED ISSUE #171: Node survives DB outages but lies about it: unreachable 503 mapping, wedgeable retry loop, constant /ready, dark metrics
## Problem

The database-outage resilience added in "Keep node alive during database outages"
keeps the process running, but the surrounding signals are wrong or fragile. An
intensive review of the branch found the outage behavior mostly doesn't do what
it claims:

1. **Outages return 500, not 503.** The `AppError::Db(PoolTimedOut | PoolClosed)
   → 503` mapping is unreachable dead code: every `Db` method returns
   `anyhow::Result`, so sqlx errors reach handlers as `AppError::Internal` and
   clients get `500 internal_error` (with raw error text) during an outage.
   Nothing in the crate ever constructs `AppError::Db`.

2. **The startup retry loop can wedge forever.** The connect timeout wraps only
   the pool connect; `migrate()` runs unbounded afterward and blocks on
   `SELECT pg_advisory_lock(...)`. If another instance (or a wedged session)
   holds the lock, `Db::connect` hangs, retries stop, and the node sits in
   degraded 503 mode indefinitely with no logs and nothing to restart it.

3. **One slow client can block startup / cause downtime windows.** After the DB
   becomes ready, main awaits the degraded server's graceful drain with no
   deadline (a slowloris request stalls the full server forever), then drops the
   listener and re-binds — so there's also a closed-port window (peer pruning,
   p2p bootstrap, S3 client init) where every connection is refused.

4. **`/ready` only tells the truth at boot.** The full server's `/ready` is a
   constant 200 that never probes the pool: a mid-life Postgres outage leaves
   readiness green while every real endpoint fails. Degraded `/health` returns
   HTTP 200 (`status: degraded`), so peer pings (`is_success` on GET /health)
   and uptime monitors count a DB-less node as alive.

5. **The degraded window is invisible.** The `:9091` metrics listener only
   starts after the DB connects, so the exact alive-but-unusable window this
   feature creates can't be scraped; the degraded router also 503s `/metrics`
   on the main port.

6. **A bad `DATABASE_URL` now looks like a successful deploy.** The retry loop
   has no attempt cap and no error classification, and none of the fly configs
   define health checks — a typo'd secret used to fail the deploy via
   crash-loop; now the machine goes green and serves 503s forever with only
   warn-level logs.

Smaller items found in the same review: the backoff's `.min(5)` exponent cap
silently defeats `GITLAWB_DB_RETRY_MAX_SECS`; five new `GITLAWB_DB_*` env vars
bypass the clap `Config` (no `--help`, silent fallback on malformed values); the
outer connect timeout is redundant with sqlx's `acquire_timeout`; the
`max_connections` default of 50 needs sizing against small Postgres instances;
smart-HTTP repo names use `trim_end_matches(".git")`, which strips repeatedly
and can alias a repo literally named `foo.git` (creatable via peer mirror
ingestion) onto `foo`.

## Expected behavior

- DB-unreachable errors surface as `503 db_unavailable` end to end, from one
  shared classification.
- The startup retry loop always makes progress: every attempt (including
  migrations) is bounded; misconfiguration-class errors (Postgres 28xxx/3D000)
  are loudly distinguishable from transient outages.
- The degraded→full handoff never closes the port and can't be stalled by a
  client.
- `/ready` reflects actual DB health for the whole process lifetime, degraded
  `/health` is a 503, and Fly routing/deploys gate on `/ready` so a broken node
  sheds traffic and a bad secret fails the deploy visibly.
- Metrics are scrapeable during the degraded window.

## Environment

All four prod nodes (node1/2/3 on Fly, manila on EC2), `gitlawb-node` v0.5.0,
branch `fix/db-startup-resilience`.

kevincodex1:
## Summary

Keep the node alive and honest through PostgreSQL outages: a degraded HTTP server
answers 503s (with real status) while the node retries the database, `/ready` now
reflects actual DB health for the whole process lifetime, and the degraded→full
handoff shares one socket so the port never closes.

## Motivation & context

A database outage or a bad `DATABASE_URL` used to crash-loop the node and hammer
the DB provider. The first cut of this branch kept the process alive but had gaps
an intensive review surfaced: the 503 mapping was unreachable dead code (outages
returned 500s), a held migration advisory lock could wedge the retry loop forever,
one slow client could block the full server from ever starting, readiness only
existed at boot, the metrics port was dark during the exact window that needed
observing, and a permanently bad secret produced a "green" deploy serving 503s
with no operator signal. This PR closes all of them.

Closes #

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

All in `gitlawb-node` (plus `infra/fly/fly.toml`, `.env.example`):

- **error.rs** — `From<anyhow::Error>` now downcasts sqlx errors back out of
  anyhow chains into `AppError::Db` (the 503 arm was unreachable: every db method
  returns `anyhow::Result`, so outages surfaced as 500 `internal_error`).
  Widened the "unavailable" class to `PoolTimedOut | PoolClosed | Io | Tls` and
  exported a shared `db_unavailable` code/message used by all three surfaces.
- **main.rs** — one `TcpListener` bound at startup and shared (dup'd fd) between
  the degraded and full servers: no closed-port window during the handoff, and
  the degraded drain is no longer awaited, so a slowloris client can't block the
  full server from starting. Each connect+migrate attempt is bounded
  (`GITLAWB_DB_CONNECT_TIMEOUT_SECS`, default 60s) so a held migration advisory
  lock can't silently wedge the retry loop. Auth/catalog errors (Postgres class
  28/3D) log at error level and jump to max backoff. Degraded `/health` returns
  503 so peer pings (`is_success` on GET /health) and uptime monitors stop
  counting a DB-less node as alive. Metrics listener starts before the DB
  connects. Backoff exponent cap no longer defeats `GITLAWB_DB_RETRY_MAX_SECS`.
  Degraded handlers consolidated to one JSON body; startup counters are atomics.
- **server.rs** — `/ready` probes the pool (`SELECT 1`, 2s bound) instead of
  returning a constant 200, so mid-life outages flip it to 503 and back.
- **config.rs / db/mod.rs** — the five `GITLAWB_DB_*` knobs moved from ad-hoc
  `std::env::var` reads onto clap-derived `Config` (validated, in `--help`;
  malformed values now fail startup instead of silently defaulting). Dropped the
  outer connect timeout (dead code: sqlx bounds the initial connect via
  `acquire_timeout`). Pool default is 20 connections (lazy cap, tunable).
- **api/repos.rs** — smart-HTTP repo names strip at most one `.git` suffix
  (`trim_end_matches` strips repeatedly, aliasing a repo literally named
  `foo.git` — reachable via peer mirror ingestion — onto `foo`).
- **infra/fly/fly.toml** — `[[http_service.checks]]` on `/ready`: fly-proxy
  won't route to a machine until it's actually ready, and a deploy with a bad
  `DATABASE_URL` fails visibly instead of going green. (Per-node tomls are
  gitignored; the same block was added to the local node1/2/3 files on disk.)

## How a reviewer can verify

```sh
cargo test --workspace   # sqlx::test cases need a Postgres DATABASE_URL (CI provisions one)

# Outage drill (verified end-to-end on this branch with postgres@16):
# 1. Start the node with Postgres DOWN:
#    - GET /, /health, /ready, any API route → 503 {"error":"db_unavailable",...}
#      with db_attempts / db_next_retry_secs visible; :9091/metrics → 200
# 2. Start Postgres: node connects and swaps to the full router with ZERO
#    refused connections (hammer `/` through the transition); /ready → 200
# 3. Stop Postgres mid-life: /ready → 503 and e.g. GET /api/v1/repos → 503
#    db_unavailable (was 200-ready + 500 internal_error before this PR)
# 4. Restart Postgres: /ready recovers on the next poll
# 5. Point DATABASE_URL at a nonexistent db: single ERROR-level
#    "database rejected our configuration" log, retry pinned at max backoff
# 6. gl init → git push → git clone round trip passes
```

Notes for reviewers

- The full server's /health stays a liveness-style 200 during mid-life outages
by design — /ready is the readiness signal, and peers pinging /health on
older nodes would 404 on /ready, so the ping target wasn't changed.
- Deploy semantics change: with the /ready check in fly.toml, fly deploy
now waits for readiness — a deploy during a real DB outage fails instead of
silently shipping a 503-serving machine. That's intentional.
- Pool default dropped 50 → 20: lazy cap, and 50 risked crowding a
max_connections=100 Postgres next to admin tooling; override per node with
GITLAWB_DB_MAX_CONNECTIONS.
- The degraded/readiness behavior was verified by a live outage drill (above)
rather than automated tests — the transition needs a real Postgres to
start/stop, which #[sqlx::test] can't model. Happy to add a harness in a
follow-up if wanted.
- Unit tests cover the backoff curve fix and the single-.git-suffix rule.

## Issue
Fixes #171 

## Summary (updated)

* **New Features**
  * Added a **GET `/ready`** endpoint that checks database responsiveness.
  * Added a deployment readiness probe (Fly) that gates rollout on **`/ready`**.
  * Introduced **degraded mode** that serves 503 responses while database connection/retries run.

* **Improvements**
  * Added PostgreSQL pool sizing and startup/retry timing configuration.
  * Database-unavailable errors now return consistent **`503 Service Unavailable`** JSON status messaging.

* **Bug Fixes**
  * Improved smart-HTTP repo name normalization to remove only one `.git` suffix and added coverage for edge cases.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,6 +21,23 @@ GITLAWB_REPOS_DIR=/data/repos
 # When using the bundled docker-compose, this is wired automatically.
 DATABASE_URL=postgresql://gitlawb:changeme@localhost:5432/gitlawb
 
+# ── Database pool & startup resilience ────────────────────────────────────
+# Maximum connections in the PostgreSQL pool. A cap, not a floor —
+# connections open lazily. Size against the DB server's max_connections,
+# remembering admin tooling opens its own pool.
+GITLAWB_DB_MAX_CONNECTIONS=20
+# Seconds a request waits for a pool connection before failing with 503.
+GITLAWB_DB_ACQUIRE_TIMEOUT_SECS=5
+# Upper bound on each startup connect+migrate attempt, in seconds. Keep it
+# generous enough for a peer instance to finish migrating (the migration
+# advisory lock is held across a deploy); on expiry the attempt is retried.
+GITLAWB_DB_CONNECT_TIMEOUT_SECS=60
+# Startup retry backoff while the database is unavailable: first delay and
+# ceiling, in seconds. The node stays up serving 503s (with a truthful
+# /ready) instead of crash-looping.
+GITLAWB_DB_RETRY_INITIAL_SECS=5
+GITLAWB_DB_RETRY_MAX_SECS=60
+
 # ── IPFS pinning (Pinata) ─────────────────────────────────────────────────
 # Get a JWT at https://app.pinata.cloud/developers/api-keys
 GITLAWB_PINATA_JWT=
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -488,7 +488,10 @@ pub async fn get_tree(
 // ── Git smart HTTP endpoints ──────────────────────────────────────────────
 
 fn smart_http_repo_name(repo: &str) -> Result<&str> {
-    let name = repo.trim_end_matches(".git");
+    // Strip at most one ".git" suffix: trim_end_matches strips repeatedly,
+    // which would misdirect a repo literally named "foo.git" (creatable via
+    // the peer mirror path, which skips API name validation) to repo "foo".
+    let name = repo.strip_suffix(".git").unwrap_or(repo);
     if name.is_empty() {
         return Err(AppError::BadRequest("missing repository name".into()));
     }
@@ -1914,6 +1917,10 @@ mod tests {
     #[test]
     fn smart_http_repo_name_rejects_empty_after_git_suffix() {
         assert_eq!(smart_http_repo_name("demo.git").unwrap(), "demo");
+        assert_eq!(smart_http_repo_name("demo").unwrap(), "demo");
+        // Only one suffix is stripped: a repo literally named "demo.git"
+        // stays addressable and never aliases to "demo".
+        assert_eq!(smart_http_repo_name("demo.git.git").unwrap(), "demo.git");
         assert!(matches!(
             smart_http_repo_name(".git"),
             Err(AppError::BadRequest(_))
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -487,6 +487,14 @@ pub async fn get_tree(
 
 // ── Git smart HTTP endpoints ──────────────────────────────────────────────
 
+fn smart_http_repo_name(repo: &str) -> Result<&str> {
+    let name = repo.trim_end_matches(".git");
+    if name.is_empty() {
+        return Err(AppError::BadRequest("missing repository name".into()));
+    }
+    Ok(name)
+}
+
 /// GET /:owner/:repo.git/info/refs?service=git-upload-pack|git-receive-pack
 pub async fn git_info_refs(
     State(state): State<AppState>,
@@ -496,7 +504,7 @@ pub async fn git_info_refs(
     headers: axum::http::HeaderMap,
     auth: Option<Extension<AuthenticatedDid>>,
 ) -> Result<Response> {
-    let name = repo.trim_end_matches(".git");
+    let name = smart_http_repo_name(&repo)?;
     tracing::info!(owner = %owner, repo = %name, "info/refs request");
     let record = state
         .db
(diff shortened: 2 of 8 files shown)


kevincodex1:
@coderabbitai please do a full review

PR #184, opened 2026-07-11
test(icaptcha): assert obtain_proof consumes the mocked service

beardthelion:
Adds an integration test that runs `obtain_proof` against a `mockito` server and asserts both `/v1/challenge` and `/v1/answer` are consumed. This proves the iCaptcha client actually calls the service rather than short-circuiting or making a live call: a regression that skipped the network would fail the test, since the mock-consumed assertions go red when the endpoints are not hit.

Adds `mockito` as a dev-dependency.

Verified: the test is load-bearing (red when the mock is under-consumed or the client points elsewhere, green when the service is consumed), and `cargo build --workspace`, `fmt --check`, and `clippy -D warnings` are clean.

## Summary by CodeRabbit

* **Tests**
  * Added an integration test that verifies the CAPTCHA flow performs the challenge and answer HTTP POST requests against the configured mocked service.
  * Ensures each mocked endpoint is consumed exactly once and that the returned proof matches the mocked response.
  * Hardened the test environment to prevent accidental live network calls by blackholing non-local traffic.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3300,7 +3300,7 @@ dependencies = [
 
 [[package]]
 name = "git-remote-gitlawb"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "anyhow",
  "gitlawb-core",
@@ -3312,7 +3312,7 @@ dependencies = [
 
 [[package]]
 name = "gitlawb-attest"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "base64",
  "ed25519-dalek",
@@ -3329,7 +3329,7 @@ dependencies = [
 
 [[package]]
 name = "gitlawb-core"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "anyhow",
  "base64",
@@ -3356,7 +3356,7 @@ dependencies = [
 
 [[package]]
 name = "gitlawb-node"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "alloy",
  "anyhow",
@@ -3412,7 +3412,7 @@ dependencies = [
 
 [[package]]
 name = "gl"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "alloy",
  "anyhow",
@@ -3821,9 +3821,11 @@ version = "0.4.0"
 dependencies = [
  "anyhow",
  "gitlawb-core",
+ "mockito",
  "reqwest",
  "serde",
  "serde_json",
+ "sha2",
  "thiserror 2.0.18",
  "tracing",
 ]
--- a/crates/icaptcha-client/Cargo.toml
+++ b/crates/icaptcha-client/Cargo.toml
@@ -16,3 +16,6 @@ anyhow = { workspace = true }
 thiserror = { workspace = true }
 tracing = { workspace = true }
 sha2 = { workspace = true }
+
+[dev-dependencies]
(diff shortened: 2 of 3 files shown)


jatmn (changes requested):
These assertions only verify that the local mock received one request for each route. A regression can make those two requests and also send the challenge, answer, or configured API key to `DEFAULT_URL` (or another origin), while this test still passes. That leaves the stated guard against unintended live iCaptcha traffic and token exposure untested. Please either constrain the test's HTTP transport so every non-mock destination fails, or narrow the test name/comments and PR claim to endpoint consumption only.

beardthelion:
Addressed via your option (a) on `8c25af6`. The test now blackholes every non-loopback origin: `NO_PROXY` lets the loopback mock through, `ALL_PROXY` routes any other host to a closed port, so the client contacting anything but the injected mock fails the request. Verified load-bearing: pointing `cfg.url` at `DEFAULT_URL` makes `obtain_proof` fail with a proxy tunnel error, and the loopback-mock path still passes.

jatmn (changes requested):

## Findings

- [P2] Make the non-mock egress block override scheme-specific proxies
  `crates/icaptcha-client/tests/icaptcha_mock_consumed.rs:26`
  This still does not guarantee the requested no-live-call behavior on a proxied runner. `obtain_proof` uses reqwest's default system-proxy configuration, and its `HTTPS_PROXY`/`https_proxy` (or HTTP equivalents) take precedence over the `ALL_PROXY` set here. Therefore an environment with a working HTTPS proxy can let a regression additionally send the DID, answer, or API key to `DEFAULT_URL` while both loopback mock requests are consumed and this test remains green. Override or clear the scheme-specific proxy variables too (restoring the prior environment safely), or enforce the transport policy directly, so the test actually detects that egress.

beardthelion:
Confirmed and fixed at `593b83f`.

The finding is real. I reproduced the precedence directly: with `ALL_PROXY` pointed at a closed port and `HTTPS_PROXY` at a live listener, reqwest routed a `https://` request to `HTTPS_PROXY`, past the `ALL_PROXY` blackhole. Since `obtain_proof` builds a default `reqwest::blocking::Client` (`lib.rs:252`) with no transport control and `DEFAULT_URL` is https, an `ALL_PROXY`-only blackhole leaves an https leak open on any runner whose environment already has a working `HTTPS_PROXY`.

Fix: the test now blackholes the scheme-specific variables too (`HTTPS_PROXY`/`https_proxy`/`HTTP_PROXY`/`http_proxy`, both cases) to the closed port, with `NO_PROXY` still letting the loopback mock through. No env restore: this file has a single `#[test]`, so it runs in its own process and the override never races a sibling.

Vetted both ways:
- The override defeats a pre-set working `HTTPS_PROXY` — with one set to a reachable listener, applying the blackhole sends the https request to the closed port and the listener is never contacted.
- No regression — the guard still passes with `cfg.url` at the loopback mock.
- Still load-bearing — pointing `cfg.url` off-loopback turns it RED (obtain_proof blackholed), confirmed by a revert-check.

`fmt` and `clippy -D warnings` clean. @jatmn ready for another look.

jatmn (approved):
LGTM

beardthelion:
@kevincodex1 #184 is ready for your review/merge: jatmn approved the current head and CI is green (up to date with main).

kevincodex1 (approved):
LGTM

PR #188, opened 2026-07-11
fix: strip Cf bidi/format controls in terminal-bound error sanitizers (#183)

### LINKED ISSUE #183: Strip Cf bidi/format controls in the terminal-bound error sanitizers (Trojan Source residual)
## Summary

Two terminal-bound error sanitizers strip only Unicode `Cc` (control) characters and let `Cf` (bidi/format) characters through, so a right-to-left override in attacker-influenceable bytes survives to the terminal and can visually reorder the displayed line (Trojan Source, CWE-451). The `gl` client already handles this class; these two do not.

## Affected

- **`git-remote-gitlawb::safe_error_body_excerpt`** (`crates/git-remote-gitlawb/src/main.rs:556`, filter at `:566`) — runs on the untrusted HTTP error body from the node (`main.rs:543`) and reaches the terminal via the error path.
- **`icaptcha-client::sanitize_excerpt`** (`crates/icaptcha-client/src/lib.rs:53`, filter at `:61`) — runs on the service-supplied `reason`/body from a node-advertised iCaptcha origin and is `bail!`'d to the terminal (`lib.rs:295`).

Both filter `!c.is_control()` only. `char::is_control()` matches category `Cc` (C0/C1 + DEL); it does **not** match `Cf` (bidi overrides/isolates: `U+202A–202E`, `U+2066–2069`, `U+200E/200F/061C`).

## Reference impl (correct)

`gl::sanitize_node_msg` (`crates/gl/src/sync.rs:135`) filters `!c.is_control() && !is_bidi_format(c)`, where `is_bidi_format` (`sync.rs:144`) covers the `Cf` ranges. That is the shape to port.

## Reproduction (executed)

Each function was called with `"error: \u{202e}drowssap\u{202c} ok"`:

- `safe_error_body_excerpt` → output still contains `U+202E` (**FAIL**)
- `sanitize_excerpt` → output still contains `U+202E` (**FAIL**)
- `sanitize_node_msg` → `U+202E` stripped (**PASS**)

The `Cc` class (ESC/BEL/NUL) is stripped correctly in all three; only the `Cf` residual remains in the first two.

## Reachability

The bytes are attacker-influenceable (a hostile or MITM'd node's error body; an attacker-selectable or compromised iCaptcha service response) and render to the operator's TTY during a normal git operation. The dangerous escape-introducer (`Cc`) class is already neutralized, so this is the residual bidi/format vector, not the full escape-injection one.

## Fix

Add an `is_bidi_format` predicate (port `gl`'s) and extend both filters to `!c.is_control() && !is_bidi_format(c)`. Add a per-crate test asserting `U+202E` is stripped. Note that `icaptcha-client`'s existing `sanitize_strips_control_chars_and_escapes` test asserts only that `is_control()` chars are gone, which is why the gap survived — the new test must assert the `Cf` case specifically.

## Severity

Hardening completeness of an already-partially-mitigated terminal-injection vector: bounded blast radius (visual spoofing/phishing, not leak or RCE), known one-function fix. Suggested `sev:medium`.

Related: the untrusted-bytes-to-terminal sanitization class; the prior `gl` fix in #82 that added the bidi predicate there.

beardthelion:
Two terminal-bound error sanitizers filtered only `Cc` control bytes and let `Cf` bidi/format controls through to the operator's terminal, so a right-to-left override in an attacker-influenceable error line could visually reorder it (Trojan Source, CWE-451). `gl` already stripped these; `git-remote-gitlawb::safe_error_body_excerpt` and `icaptcha-client::sanitize_excerpt` did not.

## Change

- Centralize the bidi predicate as `gitlawb_core::sanitize::is_bidi_format`: the 12 reordering `Cf` code points (LRM/RLM/ALM, the LRE..RLO embedding/override range, the LRI..PDI isolate range). All three client crates already depend on `gitlawb-core`, so this adds no dependency edges and leaves one audited definition instead of three copies that can drift.
- Migrate `gl`'s local copy to the shared predicate.
- Add a drop-and-continue branch in the two affected sanitizers, dropping bidi chars without setting `pending_space` (they are not whitespace, so stripping leaves no phantom gap).

This deliberately diverges from the issue's "port into each crate" wording. Centralizing removes the exact drift that produced the gap: two divergent copies, one incomplete.

## Scope

Strips only the bidi-reordering subset, not all of `Cf`. Legitimate characters like ZWJ (`U+200D`) and soft hyphen pass through unchanged; homoglyph and other Trojan-Source vectors are out of scope.

Not addressed here (separate class, follow-up): several `gl` commands print node-advertised strings (PR/issue titles, agent model, registration errors) straight to the terminal with no sanitizer at all. Same vector, different shape, tracked separately.

## Tests

Per-class strip coverage (each `Cf` class asserted individually), a must-not-over-strip guard, a transport-path proof through `handle_connect`, and a cap-boundary test. Every strip and preserve assertion was verified load-bearing by mutation: predicate returning false, predicate over-broad, and each of the 12 arms dropped individually all go red; the real predicate is green.

Closes #183.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error-message rendering by removing hidden Unicode formatting characters that could reorder or obscure displayed text.
  * Prevented invisible characters from introducing unexpected spaces or consuming available error-message length.
  * Preserved legitimate right-to-left text and other meaningful Unicode characters.
  * Applied the improved sanitization consistently across connection, synchronization, and challenge-related error messages.

### DIFF
--- a/crates/git-remote-gitlawb/src/main.rs
+++ b/crates/git-remote-gitlawb/src/main.rs
@@ -1185,7 +1185,11 @@ mod tests {
         for c in [
             '\u{202e}', '\u{202a}', '\u{202c}', '\u{2066}', '\u{2069}', '\u{200e}', '\u{061c}',
         ] {
-            assert!(!message.contains(c), "bidi U+{:04X} leaked: {message:?}", c as u32);
+            assert!(
+                !message.contains(c),
+                "bidi U+{:04X} leaked: {message:?}",
+                c as u32
+            );
         }
         assert!(
             message.contains("errok"),
--- a/crates/git-remote-gitlawb/src/main.rs
+++ b/crates/git-remote-gitlawb/src/main.rs
@@ -570,6 +570,13 @@ fn safe_error_body_excerpt(body: &str) -> String {
             continue;
         }
 
+        if gitlawb_core::sanitize::is_bidi_format(ch) {
+            // Drop Cf bidi/format controls (INV-6, #183) without marking
+            // pending_space: unlike a newline/tab they are not whitespace, so
+            // "a\u{202e}c" collapses to "ac" with no phantom gap.
+            continue;
+        }
+
         if ch.is_whitespace() {
             pending_space = true;
             continue;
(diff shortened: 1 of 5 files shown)


jatmn (approved):

@kevincodex1 LGTM

beardthelion:
@kevincodex1 following up on #188, still ready and green with jatmn's approval on the current head (up to date with main). It's been about a week since the last nudge, whenever you get a chance.

kevincodex1 (approved):
LGTM

PR #194, opened 2026-07-12
test(node): real-node deny harness for trust-boundary regressions (owner-gated mutations, path-scoped reads, client-surfaced denials)

beardthelion:
## What

A real-socket, end-to-end security regression harness for gitlawb-node. It boots a real node on `127.0.0.1:0`, drives trust-boundary DENY paths through a real reqwest client with real RFC-9421 signing, and asserts both the refusal status and that no withheld data leaks. It turns the per-PR real-node-verify step into executed tests, and covers ground `tower::oneshot` can't: the full production middleware stack over an actual socket, plus a systematic no-empty-200 (denial-as-success) assertion.

## Why the crate split

gitlawb-node was binary-only, so an out-of-crate integration test could not reach `build_router` / `AppState` / `Config`. The first commit splits it into lib+bin: the module tree and boot logic move to `src/lib.rs` (exposing a minimal `run()` plus the boot surface), and `src/main.rs` becomes a thin `#[tokio::main]` shim. No behavior change, the 488-test node suite stays green and the production binary is unaffected. The harness itself lives behind a `test-harness` feature so its spawn surface never compiles into the release binary.

## Coverage

Fourteen executed cases, one strong case per invariant plus a high-value owner-gate fan-out:

- **Denials are never rendered as empty/success:** unsigned git-receive-pack rejected 401, and an anonymous `/ipfs/{cid}` read of a withheld blob denied 404 with no leak.
- **Mutations are owner-gated, not merely authenticated:** a validly-signed non-owner is rejected 403 on set_visibility, plus protect/unprotect branch, webhook create/delete, and visibility removal. Each carries an owner-reachability check so a 403 from an earlier layer cannot masquerade as a pass.
- **Reads and replication gate on the requested path:** a withheld blob read denied 404 with no leak, the same withhold over the content-addressed `/ipfs` surface, and the git-upload-pack replication path where the served pack must omit the withheld blob while keeping the sibling public one.

Every case is mutation-verified load-bearing: the specific gate was broken, the test observed to go red (the secret leaking, or the withheld object appearing in the pack), then reverted.

## Notes for review

- The upload-pack case drives the `git-upload-pack` POST directly (v0 stateless-RPC) instead of `git clone`, because a default `git clone` negotiates protocol v2 and hangs against the node's v0 server. That hang (rather than a clean error) may be worth a separate look if standard-client interop matters. The assertion is packfile-aware (`git index-pack` + `verify-pack`), not a raw byte scan, since a leaked OID would otherwise hide inside the zlib stream.
- CI runs the harness explicitly with `--features test-harness`, since `cargo test --workspace` skips it by design.
- Deliberately deferred: the lower-impact owner-gates and read surfaces (labels, PR comments, list_visibility, and similar) keep their existing source-level authz-table guard and gate-helper unit tests, where a full-stack case per endpoint adds near-zero marginal safety. Replica register/unregister were excluded because they are signer-self, not owner-gated.

## Summary by CodeRabbit

* **Tests**
  * Added a real-node deny-harness suite with end-to-end coverage for authorization denials, request signature validation, protected content access, and owner vs. non-owner mutation safeguards.
  * Added denial assertion helpers to verify correct 4xx responses and prevent leakage of withheld tokens/identifiers.
  * Added replication checks to ensure protected objects are excluded from returned packfiles.
* **CI / Release Engineering**
  * Extended PR CI to run the deny-harness regression suite on every pull request using the existing Postgres test service.


jatmn (changes requested):

## Findings

- [P1] Create identity keys atomically with owner-only permissions
  `crates/gitlawb-node/src/lib.rs:1007`
  The new library retains the existing `exists()` → `fs::write()` flow. On Unix, `fs::write` creates the PEM using umask-derived permissions and only then changes it to `0600`; a local process can read the node private key in that window. The separate existence check also lets concurrent node starts overwrite each other's generated identity. Create the file with `create_new` and mode `0600`, then handle `AlreadyExists` by loading the winning key.

- [P2] Enforce the configured HTTP shutdown grace period
  `crates/gitlawb-node/src/lib.rs:539`
  `with_graceful_shutdown` begins draining when the signal fires but has no deadline; the computed `grace` is explicitly discarded at line 571. A long-lived request can therefore prevent termination until the orchestrator hard-kills the process, defeating `GITLAWB_SHUTDOWN_GRACE_SECS` and risking interrupted cleanup. Bound the drain with that duration and force completion once it expires.

beardthelion:
Addressed the review feedback in 30672cf.

@jatmn:

**P1 (identity key).** Created atomically with `OpenOptions::create_new(true).mode(0o600)` and written through that handle, so there is no umask-derived 0644 window. A lost race hits `AlreadyExists` and loads the winner's key instead of overwriting, with a short bounded retry so it can't read a half-written PEM; loading an existing key also tightens loose perms to 0600.

**P2 (shutdown grace).** The drain is now bounded by the configured grace, measured from the signal rather than server start (extracted as `drive_serve_with_grace`) and forced to complete once it expires. The discarded `grace` is gone and the misleading comment corrected.

Unit tests cover 0600-on-create, tighten-on-load, 8-thread concurrent-start convergence, and the grace abandon / normal-drain / signal-gated paths.

Also took CodeRabbit's nitpick: the deny-harness clients now go through a shared 30s `bounded_client` so a wedged route fails fast instead of hanging to the CI timeout.

jatmn (changes requested):

## Findings

- [P1] Remove a failed first-write identity file
  `crates/gitlawb-node/src/lib.rs:1123`
  The new `create_new(true)` path fixes the original permission window, but if the PEM write itself fails after the file has been created, the just-created key path is left behind as an empty or partial PEM. Every later start then takes the `key_path.exists()` branch, retries parsing that same bad file in `load_racing`, and exits with `invalid PEM key` instead of generating a fresh identity. A transient ENOSPC/EIO/quota failure during first boot can therefore permanently wedge the node until an operator manually deletes the file. Please remove the newly-created file on `write_all` failure in both the Unix and non-Unix branches before returning the error.

- [P2] Do not ignore failed key permission tightening
  `crates/gitlawb-node/src/lib.rs:1059`
  The load path now advertises that loose existing identity keys are tightened to `0600`, but the `set_permissions` result is discarded. If the file is readable but chmod fails, for example on a read-only mount or an ownership/ACL mismatch, the node still reads and uses a world/group-readable private key while logging a normal "loaded existing identity" path. That leaves the exact key exposure this follow-up is trying to close. Please surface the chmod failure or otherwise verify the final mode before continuing with the key.

beardthelion:
Both addressed on `fb8685d`.

**Remove a failed first-write** (F1). Extracted `write_key_or_cleanup`, which removes the just-created file when `write_all` fails, and wired it into both the Unix and non-Unix create branches. A transient ENOSPC/EIO on first boot now leaves no file, so the next start regenerates instead of re-parsing an empty/partial PEM forever. Covered by `failed_write_removes_the_partial_key_file` (an injected write error removes the file; RED with the `remove_file` disabled), and I confirmed the non-Unix branch type-checks by compiling it via a temporary cfg swap.

**Do not ignore a failed permission tighten** (F2). The load path now propagates the `set_permissions` error and then calls `ensure_key_mode_0600`, which fails closed if the key is not `0600` after the attempt — so a chmod that failed or silently no-opd (read-only mount, ACL mismatch) is refused rather than read and used exposed. Covered by `loose_key_mode_is_rejected_not_used` (a 0644 key is rejected, a 0600 key accepted; RED with the check disabled). `existing_key_is_loaded_and_tightened` still passes (a loose key on a writable mount is tightened and loaded).

One behavior note on F2: a loose key that genuinely cannot be tightened is now rejected rather than used, which narrows the original "never reject a loose key" leniency — but only in the exposed-and-unfixable case, which is the exposure this follow-up closes. The real ENOSPC/chmod-fail I/O triggers are not driven end to end (no portable fault injection), but the error handling is proven at the helper level and the wiring is a one-line pass-through of the write/chmod result into it.

RED->GREEN for each; no other production behavior changes.

jatmn (changes requested):

## Findings

- [P2] Do not fail concurrent startup after an arbitrary 100 ms key-write window
  `crates/gitlawb-node/src/lib.rs:1123`
  `create_new` exposes the final key path before the winner has completed `write_all`, and every other process that sees that inode gives up after 50 2-ms retries. On a slow or temporarily stalled filesystem, a winner can legitimately take longer than that interval, so all losing node starts return `invalid PEM key` even though the winning write later succeeds. This reintroduces an availability failure for the concurrent-start case the new code is intended to make safe. Keep retrying until a meaningful startup deadline, or publish a fully written temporary key atomically so readers never observe a partial final file.

beardthelion:
Confirmed and fixed at `e3eac37`.

The finding is right. The create path did `create_new(final)+write_all`, so the final path appears as an empty inode before the PEM is flushed, and `load_racing` only retried `50 × 2ms ≈ 100ms` — a slow/stalled-FS winner can exceed that, and every other start then fails boot with `invalid PEM key`.

I took option (b), keeping the single-winner guarantee: write the full PEM to a sibling temp, then `hard_link` it into place. `hard_link` is atomic and fails if the target exists, which gives three things at once — the final path only ever appears **complete** (no partial-read window), a lost race never clobbers the winner, and a crashed writer leaves only the temp rather than a partial final that would wedge later boots. `load_racing_key` now polls on a wall-clock deadline (5s) instead of a fixed 100ms count. I hoisted `load_existing_key`/`load_racing_key` to module level so the deadline is directly testable.

Vetted both ways:
- A 250ms-slow winner is now waited out — RED when I reverted the deadline to 100ms.
- A reader watching the final path never observes an empty/partial file — RED when I reverted the publish to `create_new(final)+write`.
- A losing publish does not clobber the winner (and temps are cleaned up); the existing convergence test stays green.

Full suite 500 pass; `fmt` and `clippy -D warnings` clean.

Two deliberate tradeoffs worth flagging:
1. The atomic publish uses `hard_link`, so it now requires a filesystem that supports hard links. Every normal node key location (ext4/xfs/overlay/tmpfs) does; a hardlink-less mount would fail boot loudly with a clear error rather than silently. Happy to add a `create_new`+write fallback for that case if you'd prefer universal-FS support over failing loud.
2. A genuinely corrupt existing key now blocks boot up to the 5s deadline before erroring (a valid key still parses on the first attempt, no delay).

@jatmn ready for another look.

jatmn (changes requested):

## Findings

- [P1] Recover from a stale pre-publication key file
  `crates/gitlawb-node/src/lib.rs:1174`
  The temporary path is deterministically derived from the PID and a process-local counter. If the process dies after writing `.identity.pem.tmp.<pid>.0` but before the cleanup at line 1195, that file remains. A restarted container commonly runs as PID 1 again and starts its counter at zero, so `create_new` fails on the same stale name before the final key exists; the node cannot start until an operator deletes the temp file. Retry a new temporary name (or safely handle stale temps) so an interrupted first provision remains recoverable.

- [P1] Make identity-key publication durable before exposing it
  `crates/gitlawb-node/src/lib.rs:1185`
  `write_all` followed by `hard_link` only atomically changes the namespace; neither the completed temporary file nor the containing directory is synced. A host/power crash after the link can therefore leave `identity.pem` durable while its PEM bytes are missing or truncated. On restart the existing-path branch retries that invalid key and then exits, recreating the permanent startup wedge this change is intended to eliminate. Sync the temporary file before linking and the parent directory after publication, with appropriate error/cleanup handling.

- [P2] Wait for the spawned node before SQLx drops its database
  `crates/gitlawb-node/src/test_harness.rs:48`
  Dropping `TestNode` only sends the shutdown watch signal; the detached `axum::serve` task, which owns the router's `PgPool`, is never joined. `#[sqlx::test]` starts `cleanup_test` immediately after the async test future returns, so it can issue `DROP DATABASE` while that server still retains connections. SQLx prints and ignores that cleanup failure, making the new CI job leak test databases and race/flap under parallel execution. Retain the task handle and add an async shutdown/teardown path that signals, awaits the server, and releases the pool before the test returns.

beardthelion:
All three addressed on 688b506, with a follow-up in de21c64 for a test-timing consequence of the durability fix.

**Stale pre-publication temp file (P1).**

> A restarted container commonly runs as PID 1 again and starts its counter at zero, so `create_new` fails on the same stale name before the final key exists

The process-global counter is gone. The temp name is now picked by a bounded per-call retry over `.identity.pem.tmp.<pid>.<attempt>` (64 names), and `AlreadyExists` skips to the next name, so a crashed start's leftover no longer wedges a PID-stable restart. The same loop also covers two same-PID containers publishing over a shared volume, which the old counter never did. Deliberate limit: a temp this call did not create is never deleted, because a stale temp is indistinguishable from a live concurrent publisher's temp, and unlinking a live one between its write and its hard_link would fail that healthy start. A crashed start therefore leaks at most one 0600 temp file, which later starts skip; the function comment documents this. Tests: `publish_recovers_from_a_stale_crashed_temp` (pre-seeded attempt-0 stale with partial bytes, publish must win with the complete PEM; fails under a single deterministic name, verified by mutation) and `publish_errors_when_all_temp_names_are_taken` (all 64 names taken: bounded error naming the directory and prefix, final path untouched).

**Durability (P1).**

> A host/power crash after the link can therefore leave `identity.pem` durable while its PEM bytes are missing or truncated.

The temp write now chains `sync_all` through `write_key_or_cleanup`, so the PEM bytes are durable before the name can appear and a failed fsync removes the temp exactly like a failed write. After a successful link plus temp unlink, the parent directory is fsynced before `Won` is reported (unix only; opening a directory fails on Windows). On dir-fsync failure the final file stays in place: it is complete and data-synced, a concurrent loser may already have loaded it, and the next start loads it via the exists() path. Verified by syscall trace: one publish issues fsync on the temp file fd and then on the directory fd, both returning 0.

The added fsyncs consumed the timing margin of `publish_never_exposes_a_partial_final` (400ms reader window minus the 200ms hold-open), which started flaking about 1 in 5 local runs under parallel-suite disk load. de21c64 lets the reader poll a 5s deadline and stop at the first complete observation; the never-see-partial assertion still runs before the break, and an exposure mutation (empty final visible during the window) still fails the test.

**TestNode teardown (P2).**

> Retain the task handle and add an async shutdown/teardown path that signals, awaits the server, and releases the pool before the test returns.

`TestNode` keeps the serve JoinHandle and a pool clone, and `shutdown()` signals, joins the serve task under a 10s timeout (a wedged graceful shutdown fails loudly instead of hanging CI), closes the pool, and removes the repos dir. Every deny test now ends with it (8 sites, not 10; grep-verified), with Drop kept as the best-effort fallback for tests that panic first. Regression test `shutdown_joins_server_and_closes_pool` asserts through an observer clone that `is_closed()` holds before the test future returns, which is exactly the state sqlx's DROP DATABASE cleanup sees, and that a fresh request to the old address fails to connect. With drop-only teardown the assert fails (verified by mutation). Observed while implementing: with an idle keep-alive connection open, the serve task joins in about a second, so `with_graceful_shutdown` does close idle connections and the 10s bound is slack, not load-bearing.

Full run: 502 unit plus 15 integration tests green, clippy and fmt clean at both heads. One incidental change: the pre-push gate flagged the pre-existing bare git spawn in `seed_bare_repo` once this diff touched the file; it now carries an `allow-unbounded-git` annotation (test-harness-only fixture seeding, feature-gated out of the production binary, local git against a tempdir).

jatmn (changes requested):

## Findings

- [P1] Verify the mode before using a newly published identity key
  `crates/gitlawb-node/src/lib.rs:1277`
  The new-key path returns `kp` immediately after `publish_key_atomically`, while the existing-key path is the only place that calls `ensure_key_mode_0600`. `OpenOptionsExt::mode(0o600)` is only a creation request, and mounts with mode/ACL remapping can leave the freshly linked file group- or world-readable even though the node then starts with that private key. Apply the same post-publication verification (and fail closed) before logging and returning the winner.

- [P1] Do not make first startup depend on hard-link support
  `crates/gitlawb-node/src/lib.rs:1231`
  `publish_key_atomically` unconditionally publishes with `hard_link`, so a new node whose configured key directory is on a filesystem without hard links (for example a supported removable, FUSE, or network-backed volume) cannot start at all. The previous create/write path worked on those paths; preserve the no-clobber/complete-publish guarantee with a platform-appropriate fallback instead of turning `--key-path` into a hard-link-only storage contract.

- [P2] Surface an unexpected real-server failure to the deny harness
  `crates/gitlawb-node/src/test_harness.rs:142`
  The spawned task discards the result of `axum::serve`, so `shutdown()` observes only a successful `JoinHandle` even if the real socket server exited with an I/O error before teardown. That lets the test harness represent an unexpected server lifecycle failure as a clean shutdown, weakening the new full-stack regression tests. Propagate the serve result through the task (and make `shutdown()` fail on it).

- [P2] Abort the server on panic-path test teardown
  `crates/gitlawb-node/src/test_harness.rs:59`
  When a deny test panics before its explicit `node.shutdown().await`, `Drop` only signals the watch channel and immediately returns while the detached Axum task can still hold `PgPool` clones. `#[sqlx::test]` then starts `DROP DATABASE`, racing the graceful server shutdown and leaking per-test databases or cascading cleanup failures exactly when a regression makes a test fail. The fallback needs to abort/release the server-owned resources rather than relying on an unawaited graceful signal.

- [P2] Align the enforced shutdown behavior with the configured contract
  `crates/gitlawb-node/src/lib.rs:610`
  When the grace timer wins, `drive_serve_with_grace` drops the Axum `serve` future and reports `Ok(())`; in-flight clients receive a closed connection/reset, not the 503 promised by `GITLAWB_SHUTDOWN_GRACE_SECS` in `config.rs`. This matters especially for mutations, where a client cannot distinguish a completed request from an interrupted one. Either implement an explicit bounded 503/error response path or update the operator-facing contract and test the actual real-socket behavior.

beardthelion:
All five addressed on `09bfe36`. Full `cargo test -p gitlawb-node` is 517 pass / 1 ignored; `deny_harness` 17 pass; fmt and clippy clean.

**F2 (hard-link portability).** `publish_key_atomically` now falls back to a direct `create_new(true).mode(0o600)` on the final path when `hard_link` returns any non-`AlreadyExists` error, so a key directory on a filesystem without hard links starts a node instead of failing outright (probed: `hard_link` returns `EPERM` on a vfat mount). No-clobber holds through the exclusive create (`AlreadyExists` routes to the existing lost-race load), and the cleanup is phase-split: a `write_all` failure removes the partial final, but a file- or dir-fsync failure keeps it, because it is complete and a concurrent loser may already have loaded it (the Won-arm rule). Six new tests cover both fallback arms, the lost race, concurrent convergence, and both failure phases. The fallback reintroduces the partial-visibility window on that path only; the rustdoc now scopes the complete-only-visibility and durability guarantees to the hard-link path and documents the fallback's weaker, loud-failure contract.

**F1 (mode before use).** Both creation sites (the temp on the hard-link path and the final on the fallback path) now run the loader's tighten-then-verify pair before any PEM bytes are written: `set_permissions(0o600)`, then the verify. The pair, not the verify alone, is load-bearing — `create_new`'s mode is narrowed by the umask, so a `0277` umask lands `0400` and the tighten repairs it (previously the publish succeeded and shipped the `0400` key); on a mode-ignoring mount the tighten no-ops and the verify fails closed with the empty file removed. Each call site is mutation-vetted (removing it turns a test RED), and a `0277` case executes the repair.

**F4 (teardown vs. sqlx cleanup).** Reframed after tracing it: sqlx 0.8.6 has no `catch_unwind`, so on a panicking test its `DROP DATABASE` cleanup never runs — the panic-path race isn't reachable. The reachable case is a test that returns successfully without `shutdown()`, so `Drop` now aborts any remaining server handle. `abort()` is a synchronous request; the task's pool clones drop on a later scheduler tick that sqlx cleanup's own awaits provide before the drop statement. Because axum 0.8.8 runs each connection on its own task holding its own state clone, an abort-only regression isn't observable at the database surface through the graceful path, so I added a test-only flag that suppresses the graceful signal and fences the abort in isolation: remove `handle.abort()` and that test goes RED (leaked sessions) while the graceful-path test stays green.

**F5 (shutdown contract).** The `config.rs` prose promised a 503 to in-flight clients at grace expiry; `drive_serve_with_grace` abandons them (connection close), which is the standard grace-timeout behavior, and a retroactive 503 on an in-flight response isn't constructible under hyper. Fixed the prose to state the abandon contract (`config.rs` is the only operator-facing surface — README and `.env.example` don't mention the knob) and added a real-socket test: a handler held past a 250ms grace, asserting the client never sees a `503` status line and the driver returns bounded. A 503-shed path for not-yet-dispatched requests during the drain window is a separate feature, not this fix.

**F3 (serve result).** `shutdown()` now checks `axum::serve`'s result instead of discarding it. On axum 0.8.8 the `Err` arm is unreachable (graceful shutdown resolves `Ok`, accept errors retry), so this is forward-compat insurance and is documented inline as such; the `Ok` path runs in every harness test.

One residual worth naming, not fixed here: on a no-hardlink filesystem, a concurrent first-start whose peer takes the fallback path and then fails mid-write will poll the removed final for the full deadline and fail its start rather than regenerating. It needs the three conditions together (no hardlink support, concurrent starts, a mid-fallback I/O failure), both starts fail loudly, and a restart recovers — no clobber or key exposure in any interleaving. I left it for a follow-up rather than widen this round.

jatmn (changes requested):

## Findings

- [P1] Do not retain an unsuccessfully synced fallback identity key
  `crates/gitlawb-node/src/lib.rs:1431`
  The fallback is specifically used when the key directory cannot hard-link, but after `sync_all` fails it returns the error while retaining the newly created final pathname. An fsync failure (for example deferred ENOSPC/EIO) means the PEM is not known durable; a subsequent crash can leave that visible final file empty or truncated. On the next start `load_or_create_keypair` takes the existing-file path and retries parsing it rather than regenerating it, permanently wedging startup. The hard-link path removes its still-unpublished temporary file on the same failure; the fallback needs an equally recoverable publish strategy or cleanup/quarantine of an unsuccessfully synced final.

- [P2] Abort the serve task when graceful teardown times out
  `crates/gitlawb-node/src/test_harness.rs:216`
  `shutdown()` moves the `JoinHandle` into `timeout`; if the ten-second guard expires, dropping that future detaches the still-running task and `self.server` is already `None`, so `Drop` cannot abort it. The detached router retains its `PgPool` clones while `#[sqlx::test]` begins `DROP DATABASE`, leaking the test database or causing the cleanup race that this harness is intended to prevent. Time out a mutable handle and abort/join it on expiry before panicking or returning.

beardthelion:
Both findings fixed in `33ea63b`; I verified each by execution at that head.

**F1 (fallback retains an unsynced identity key).** The hard-link fallback now removes the final on a file-fsync failure instead of keeping it, so a bytes-accepted-but-not-durable final can't survive to wedge the next start. I split the write and the fsync so each removes-on-failure but keeps a distinct error context (write-rejected vs durability-failed), which preserves the phase signal for an operator debugging ENOSPC/EIO on a network mount. Verified: I reversed the fallback-fsync test to assert the final is gone. It was RED on the old keep behavior (file still present) and GREEN after; neutralizing the remove line turns it RED again. All identity-key tests stay green.

This is the cleanup path you asked for over keep. The tradeoff is the concurrent-start case: a loser that already loaded the key via `load_racing_key` keeps running while the winner regenerates on its next start. I'm accepting that here, and it's the residual already noted in the prior round rather than something new. The reason keep isn't viable: `load_existing_key` is deliberately fail-loud on a truncated final (`fallback_crash_residue_fails_loudly_not_silently` never regenerates it), so keeping the un-synced final would reintroduce exactly the startup wedge this closes, and this publish path is first-start-only, so there's no established identity to strand in the common single-node case.

**F2 (serve task detached on teardown timeout).** `shutdown` now retains the `JoinHandle` by `&mut` and aborts and reaps it on timeout before returning, so the task and its pool clones are gone before `#[sqlx::test]` runs `DROP DATABASE`. Verified load-bearing: the wedged-task test asserts the handle is finished (aborted and reaped) after the timeout; a detach variant leaves it running and the test fails.

Reworking that path, I also closed three teardown gaps in the same change: a task that finishes in the timeout race is now honored as a graceful exit rather than panicking as wedged; a genuine serve-task panic is propagated instead of being flattened into a timeout; and the pool is closed on every path, including the wedged arm, before the panic. The previous order panicked first and skipped `pool.close()` on the one path that most needs it.

Full suite is green at `33ea63b`: node lib 519, `deny_harness` 17, fmt and clippy clean. Re-requesting your review.

jatmn (changes requested):

## Findings

- [P2] Make the hard-link fallback recoverable after interruption
  `crates/gitlawb-node/src/lib.rs:1401`
  This fallback is explicitly used for filesystems that reject `hard_link`, but it creates `identity.pem` before writing and syncing the PEM. A SIGKILL or power loss in that window leaves an empty or partial final key; the next boot takes the `exists()` path at `load_or_create_keypair` and only retries parsing it, so it fails permanently instead of provisioning a key. The base direct-write path had this interruption risk too, but the new atomic primary path fixes it and this newly added compatibility fallback retains it for the network/overlay filesystems it is meant to support. Use an atomic no-clobber publish primitive there as well, or leave enough recoverable state to distinguish and remove an interrupted first write.

beardthelion:
Fixed at `32903c4` (rebased onto the #195 test landing), 15 commits, each verified by execution at its head. Final-head validation: 581 lib tests plus 39 deny-harness integration tests pass with the test-harness feature, fmt and clippy clean. 62 new identity-key tests; every behavior-bearing one was observed failing before its fix, and the load-bearing guards are mutation-vetted (each goes red when its check is removed).

The finding is right, and it overruled round five's stated trade-off for the right reason. That trade-off rested on a partial first write being indistinguishable from a corrupted real key. The fix removes the indistinguishability instead of accepting the wedge: the fallback brackets its non-atomic window with a durable publish marker, created and fsynced before `create_new(final)`. A boot that finds a marker beside a content-failed final provably has an interrupted first write: it claims the round, quarantines the final with bytes preserved, and regenerates. `recovery_regenerates_after_crash_leaves_empty_final` and its truncated-PEM sibling fail on the old code with the exact permanent wedge and pass now, in well under two seconds.

The must-not cases are locked by execution, not inspection:

- A bad final with no marker keeps today's loud error, file untouched (`bad_final_without_marker_still_fails_loudly`).
- A non-content load failure (transient EIO, EACCES, failed mode tighten) never recovers, marker or not; the class is a typed root error, not message matching (`transient_error_with_marker_never_recovers`).
- A loadable final with a stale marker loads unchanged, and stale state is swept only after the sweeper itself fsyncs the final it loaded, because a page-cache-loadable final is not necessarily durable.

The concurrency half went through repeated adversarial review beyond the finding itself, and the protocol that survived is stricter than the first cut: recovery must atomically claim a round before touching anything, with per-process nonces in claim names so PID-1 containers on a shared volume cannot alias each other; live rounds heartbeat their claims under an RAII stop-and-join guard, which makes claim age a sound orphan signal for sweeping and claimability; the winner's own-marker removal is its commit check, and a demoted publisher waits for claim absence and fails closed if any claim survives the bound; post-write publish failures remove nothing and hand the marker-plus-partial state to boot recovery; a quarantined key that turns out complete is restored through the atomic no-clobber publisher; a stranded quarantine younger than the age floor is adopted for identity continuity, while anything that provably lost its round is superseded, so deleting `identity.pem` for a fresh DID can never resurrect an outcompeted key; forensic files are retention-swept after 24 hours.

Honest residuals, all stated in the code docs: on dir-fsync-hostile mounts the marker fsync is warn-and-continue, so recovery there narrows to SIGKILL-class crashes rather than never provisioning; a live publish whose pre-write window outlasts the 250ms crash-signature persistence can be hijacked by a concurrent starter, which converges through demotion and restore at the cost of churn; non-Unix keeps this file's existing reasoned-not-run posture. The adversarial passes had converged to P2-and-below for several rounds when I stopped them, but I would not claim the multi-actor lifecycle space is exhausted; extra scrutiny on the shared-volume interleavings is welcome.

One settled design call: the identity machinery is now large enough to deserve its own module, but I kept it in place this round so the diff stays reviewable against the finding it fixes; the split is queued as a follow-up.

jatmn (changes requested):

## Findings

- [P2] Fail when the denial body cannot be read
  `crates/gitlawb-node/tests/support/assert.rs:55`
  `assert_denied` turns a `resp.text()` error into an empty body. A denial can send its expected 4xx headers and then reset or truncate the response body; this helper then accepts the status and observes no withheld token, so the new no-leak regression passes without inspecting the body that might contain the leak. Propagate the body-read failure before applying `check_denied`.

- [P2] Close every ordinary spawned node before SQLx cleanup
  `crates/gitlawb-node/tests/deny_harness.rs:154`
  The protected-branch test returns without `node.shutdown().await`; the same is true for the PR-diff and registry-sweep tests. They therefore use `TestNode::Drop`, which cannot await/reap the Axum task or close its pool, while `#[sqlx::test]` immediately drops the test database. Server-held pool connections can race that cleanup and leak databases or make the new CI job flaky. End each normal test with the awaited shutdown (leaving only the tests intentionally exercising `Drop` as exceptions).

- [P2] Bound the fixture setup requests as well as the probes
  `crates/gitlawb-node/tests/support/probe.rs:319`
  The registry test builds its fixture before it reaches its bounded probe client, but the fixture seeders still use unbounded `reqwest::Client::new()` calls. A hung create/claim/push route during setup can therefore wait until the 45-minute CI job timeout, despite the harness's stated fast-fail timeout contract. Reuse `support::bounded_client()` (or pass a bounded client through the seed helpers) for these setup requests.

beardthelion:
All three are addressed in 8d81cbe, with a follow-up in 996e1b3. Full `deny_harness` suite is green (43 tests).

**Fail when the denial body cannot be read** (`support/assert.rs`): `assert_denied` now propagates the body-read failure instead of folding it into an empty string, so an unreadable denial can no longer pass the withheld-token check vacuously. A legitimately empty body is still `Ok("")`, so the empty-body 401 denials pass unchanged.

**Close every ordinary spawned node before SQLx cleanup** (`deny_harness.rs`): added `node.shutdown().await` to the three tests you named (protected-branch push, PR-diff, registry sweep). To keep the contract from drifting again, I added a completeness guard that scrapes this file and fails if any `spawn_node` test omits `shutdown()`, with a closed allowlist for the two Drop-regression tests that deliberately exercise `Drop`. It is load-bearing: removing any one `shutdown()` turns it red (verified), and the guard is itself unit-tested against a commented-out call and a brace-in-string body so it cannot pass vacuously.

**Bound the fixture setup requests** (`support/probe.rs`): the eight seeder clients now use the bounded `support::bounded_client()` (30s) instead of `reqwest::Client::new()`, so a wedged setup route fails fast instead of running to the CI job timeout.

@jatmn ready for another look.

jatmn (changes requested):

## Findings

- [P1] Reintegrate the harness branch with current `main` without restoring removed behavior
  `.github/workflows/pr-checks.yml:3`, `.github/workflows/release.yml:49`, `crates/gl/src/cert.rs:151`, `crates/gl/src/init.rs:41`, `crates/gl/src/status.rs:147`
  The PR head does not descend from the current PR base (its merge-base is an older commit), and the resulting base-to-head diff rolls back several safeguards that are already on `main`: merge-queue validation, certificate signature verification, OIDC npm publication, native/smoke-tested arm64 releases, and the branch/remote-aware `gl` behavior. These changes are outside the deny-harness scope and would reintroduce the failures those merged fixes addressed. Please rebase or otherwise reintegrate with current `main`, resolving the integration so these existing protections remain, then request a fresh review of the resulting full diff.

beardthelion:
Reintegrated with current main at `8f7af96` (a merge of `b484a24`, keeping the branch's reviewed history rather than rewriting it, which I read as within your "rebase or otherwise reintegrate").

On the rollback concern: I believe that reading came from a two-dot comparison against the new main. The branch's merge-base diff never touched four of the five files you named. Post-merge, `git diff origin/main -- .github/workflows/pr-checks.yml .github/workflows/release.yml crates/gl/src/cert.rs crates/gl/src/init.rs crates/gl/src/status.rs` shows exactly one hunk, the added deny-harness CI step in pr-checks.yml, sitting alongside the merge-queue guards from main. Merge-queue validation, cert signature verification, OIDC npm publish, native arm64 releases, and the gl status/init/doctor fixes are all intact at main's versions.

Verified on the merged state: full workspace suites green (601 node tests among them), and the deny harness runs 43/43, matching the pre-merge count, with the shutdown-guard vacuity tests present by name in the output. One deliberate non-change: Cargo.lock still carries the 0.5.1 stamps because the manifest/lock version drift exists on main itself and #243 owns that sync; absorbing it here would have added an unreviewed hunk outside this PR's scope.

Ready for a fresh look at the full diff.

jatmn (changes requested):

## Findings

- [P1] Preserve the released 0.7.0 version baseline
  `.release-please-manifest.json:2`
  The current base is already tagged `v0.7.0`, but this head changes the release manifest and every shipped crate back to `0.6.0` and deletes the 0.7.0 changelog section. As a result, a release from this branch reports an already superseded version and release-please calculates from a stale baseline, which can collide with existing tags/packages. Rebase or regenerate the release metadata so it retains the base's 0.7.0 state.

- [P1] Do not remove the gitlawb-core dependency-purity gate
  `.github/workflows/pr-checks.yml:257`
  This PR deletes `core-deps-purity` together with its checker and exhaustive allowlist, with no replacement. That job is the only check that rejects new normal dependencies in `gitlawb-core`, which is embedded by the node, `gl`, and the remote helper; normal test, clippy, and audit runs do not enforce this contract. A subsequent dependency addition will therefore pass PR CI and silently propagate to every shipped consumer.

- [P2] Refuse symlinked identity-key paths before tightening permissions
  `crates/gitlawb-node/src/lib.rs:1257`
  `metadata` and `set_permissions` both follow `GITLAWB_KEY` symlinks. If an attacker can replace the configured key path in a writable directory, a privileged node start chmods the symlink target to `0600` before validating or parsing it. That permits filesystem-integrity denial of service against any service-readable target and is newly introduced by the key-hardening path. Open the key without following symlinks and verify it is a regular file before changing permissions or reading it.

- [P2] Close the temp key before trying to remove it after a failed write
  `crates/gitlawb-node/src/lib.rs:2031`
  The error path calls `write_key_or_cleanup` while `f` is still open; that helper immediately calls `remove_file`. On Windows an open file cannot be removed, so a write or fsync failure leaves `.{stem}.tmp.<pid>.<n}` behind. Repeated restarts with a reused service/container PID can exhaust all 64 deterministic names and prevent recovery after storage becomes healthy. Drop the handle before cleanup (and keep the cleanup result observable) so the documented failed-write recovery works cross-platform.

beardthelion:
All four findings are addressed on `83da867`, though two of them were already satisfied before this round and I want to be precise about which.

## The two P1s were fixed by the merge from main, not by me

`.release-please-manifest.json` reads `{".": "0.7.0"}` at this head, byte-identical to main, and `core-deps-purity` is present in `pr-checks.yml` with the same job and allowlist as main. Every commit between your review and this head arrived from `origin/main` through the 08-04 merge, so the branch had drifted rather than deliberately reverted either one, and reintegrating current main resolved both. Nothing further to do there, but I would rather say the merge fixed them than let it read as work I did.

## Refuse symlinked identity-key paths

Fixed, and the finding was live: nothing in the file matched `symlink_metadata`, `is_symlink`, `O_NOFOLLOW`, or `custom_flags`, and the load path made four separate symlink-following lookups (`metadata`, `set_permissions`, the mode verify, `read_to_string`).

On Unix it now opens once with `O_NOFOLLOW` and does the mode check, the tightening, and the read through that single descriptor. That refuses a symlink outright, and it also closes the check-to-use window the path-based sequence had even without one. A non-regular file is refused for the same reason. A regular key with loose bits is still narrowed rather than rejected, since rejecting it would break existing deployments; the pre-existing `existing_key_is_loaded_and_tightened` runs unmodified as the proof.

Two notes from doing it. `ErrorKind::FilesystemLoop` is not usable at our toolchain (still behind `io_error_more`), so the symlink case matches `raw_os_error() == Some(libc::ELOOP)`. And the recovery arm downcasts an `io::Error` for its NotFound case, so I checked `vanished_final_at_reparse_still_regenerates` still passes with NotFound now coming from the open rather than the read.

## A gap in my own fix, found afterwards

I ran the head past a second model before pushing, and it caught something my tests could not: the non-regular-file refusal sits after the open, and opening a FIFO read-only blocks until a writer connects. So a named pipe planted at the key path hung startup indefinitely instead of being refused, reachable by exactly the actor your finding assumes. My test planted a directory, which fails the open immediately with EISDIR and does reach the refusal, so it passed while the FIFO case hung.

Confirmed directly before believing it: an open with a 3-second alarm was killed rather than returning, and with `O_NONBLOCK` the open returns at once while `fstat` still reports a FIFO. The open now passes `O_NOFOLLOW | O_NONBLOCK`. It has no effect on a regular file, and a non-regular descriptor never survives the type check, so nothing needs to clear it afterwards. The regression runs the call on its own thread with a bounded wait, so if this ever regresses the test fails instead of hanging the suite.

## Close the temp key before removing it

Fixed. The write and fsync result is computed, the handle is dropped, then the cleanup helper runs; `sync_all` needs the live descriptor so it stays ahead of the drop. A removal that fails is now logged, since that is the precursor to exhausting the 64 deterministic per-pid names, while the write error remains the one surfaced.

I also checked the fallback publish path, since the same helper shape appears there. It deliberately removes nothing on a failed write, because a removal would race a recovery that already claimed the round, and `failed_fallback_write_leaves_the_partial_final_for_recovery` pins that. So it does not have this bug.

## The Docker failure was this branch

The dependency-cache layer died on `couldn't read crates/gitlawb-node/src/lib.rs`. This branch gives `gitlawb-node` a lib target alongside its binary, and the stub layer only wrote a `main.rs` stub for it, so cargo aborted while resolving targets. Added the matching stub, keeping the block's deliberate lack of `|| true`.

That conversion has a consequence beyond CI worth flagging: `gitlawb-node` is now lib+bin, so unit tests in `lib.rs` run under `--lib` rather than `--bin gitlawb-node`. That changes how the crate is tested for everyone, not just here.

## Evidence and what is not proven

Every guard above is proven by injection rather than inspection: four mutations, each reverting the specific defect, each observed going red on its named message, isolated so dropping `O_NOFOLLOW` reddens the symlink test and dropping `O_NONBLOCK` reddens the FIFO test. Locally the lib suite, `fmt`, and `clippy -D warnings` all pass; CI on this head will say whether the Docker layer is actually fixed, since only a full release build proves that end to end.

Two things I am not claiming. The non-Unix arm is reasoned, not run: `cargo check` against the Windows target fails inside `aws-lc-sys`'s C build for reasons unrelated to this change, so the Windows behavior of the temp-close fix is unverified locally, though that arm is byte-identical to before. And `tighten_and_verify_created` still chmods by path, called on the temp and on the final, which is the same class as the symlink finding at sites this review did not cite. I left it alone rather than widen a PR whose size you have already objected to, but it is a real remaining instance and I would rather name it than have you find it.

jatmn (changes requested):

## Findings

- [P1] Rebase onto current `main` again — `0.7.1` landed after the August merge
  `.release-please-manifest.json:2`, `crates/gitlawb-node/Cargo.toml:4`, `CHANGELOG.md:1-3`
  Your August 5 comment is accepted: the August 4 merge from `origin/main` was the right fix for the earlier drift, and at `83da867` the manifest matched `main` as it then stood (`0.7.0`), with `core-deps-purity` intact. That is no longer true against **current** `main`: three commits have landed since (`cd810b3` release `0.7.1`, `5f44515` lockfile sync, `fdf716dd` release merge), and `main` now reads `0.7.1` while this head still reads `0.7.0` and omits the `0.7.1` changelog section. This is fresh stale-base drift, not a harness logic bug, but merging without another rebase would again regress the released version baseline. Please rebase onto current `main` and let the `0.7.1` manifest, crate versions, lockfile, and changelog come forward.

- [P3] Add `--locked` to the deny-harness CI step
  `.github/workflows/pr-checks.yml:114`
  No response yet on this point. The workspace test step uses `cargo test --locked --workspace`, but the deny-harness step runs `cargo test -p gitlawb-node --features test-harness --test deny_harness` without `--locked`. The shared `Cargo.lock` makes practical drift unlikely, but the harness step should match the locked-resolution contract used everywhere else in PR CI.

## Acknowledged from your comments

- **Symlink refusal (`O_NOFOLLOW`) and FIFO hang (`O_NONBLOCK`)** — verified on head; accepted as fixed on `83da867`.
- **Temp handle closed before removal** — verified (`drop(f)` before `write_key_or_cleanup`); accepted as fixed. Your note that the fallback post-write path intentionally keeps partial state for recovery is correct and matches the tests.
- **`tighten_and_verify_created` still chmods by path** — your disclosure that this is a deliberate, out-of-scope deferral is noted; not blocking this round.

euxaristia:
@coderabbitai full review

beardthelion:
Reintegrated onto current main (merge at bfc44f92, 0.7.1). The deny-harness CI step now runs `cargo test --locked -p gitlawb-node --features test-harness --test deny_harness`.

/ipfs withhold uses production pin CIDs (`from_git_object_bytes`) with anon deny, signed stranger 404, and reader 200 legs. Identity key hardening: parent dir tightened to 0700 before publish only; fallback markers land 0600. Follow-up harness work: `assert_denied` scans response headers; inline `is_claimant` multi-principal orphan guard; gossip ping unit tests mock `/ready` 404 before `/health` fallback.

At head `26014754`: deny_harness 50 passed; `cargo test -p gitlawb-node --lib --locked` 1175 passed; clippy `-D warnings` green on push; `cargo check --release --locked -p gitlawb-node` green.

gl/git-remote client denials stay outside this PR's HTTP harness scope.

jatmn (changes requested):

## Merge readiness

- [P1] Resolve the failing `CodeQL` status before merge
  GitHub reports two high-severity CodeQL annotations in the new harness: the test sends its deliberately withheld CID over local HTTP, and the denial assertion includes an OID in a panic witness. These may be test-fixture false positives, but the aggregate CodeQL check is failing. Classify and suppress only confirmed fixture-only alerts (with narrow, documented suppression), or change the harness if either flow can reach non-test behavior; do not merge while the required security check remains unresolved.

## Findings

- [P1] Restore the dedicated advisory-lock pool
  `crates/gitlawb-node/src/lib.rs:300`
  The lib/bin extraction now passes `db.pool().clone()` to `RepoStore`, instead of the cancellation-safe `build_lock_pool(...)` used by the prior boot path. `acquire_write` holds a PostgreSQL session advisory lock while receive-pack/Tigris work is in flight. If the request is cancelled in that interval, returning the ordinary pooled connection does not run the lock-pool release hook, so a later checkout of that session can retain the lock and other sessions block on it. Long pushes also consume the query pool itself, starving authorization, visibility, and post-receive database work.

  Restore `lock_pool_size` plus `build_lock_pool(db.pool(), ..., db_acquire_timeout_secs)` at the library boot boundary and pass that pool to `RepoStore`. Keep the existing `after_release(pg_advisory_unlock_all())` cleanup and the separate capacity budget; the root cause is that the extraction silently substituted a plain cloned `PgPool` for this behavior-bearing constructor.

- [P1] Keep the cross-field DB/push configuration check on the new boot path
  `crates/gitlawb-node/src/lib.rs:83`
  `run()` no longer calls `Config::validate()`. Clap validates each setting independently, so a deployment can start with (for example) fewer DB connections than `max_concurrent_git_pushes + DB_POOL_APP_HEADROOM`. Slow receive-packs can then occupy the available connections and turn unrelated DB-backed requests into acquire-timeout failures.

  Call the existing `config.validate().map_err(...)` immediately after `bootstrap::merge_seeds(&mut config)` and before key generation, database connection, or listener setup. Preserve the existing validation and its error wording rather than duplicating only today’s pool check: the root cause is that the moved `run()` bypasses the single cross-field validation boundary.

- [P2] Preserve peer reachability hysteresis in the extracted gossip loop
  `crates/gitlawb-node/src/lib.rs:1084`
  The previous `gossip_ping_round` kept a `failed_once` set and persisted `last_ping_ok = false` only after two consecutive failed probes. The new inline loop writes every probe result immediately. A short network timeout, peer restart, or transient database error therefore hides the peer’s federated data for the full five-minute gossip interval, despite the prior contract intentionally treating one failure as non-authoritative.

  Move the existing `gossip_ping_round`/`peer_ping_db_update` state machine with the extraction, retain the missed-tick behavior, and restore the two-failure unit and DB-round regression tests. The root cause is not the probe implementation: it is the lost in-memory failure state when the loop was inlined.

beardthelion:
Addressed the boot-path regressions and the follow-on gaps on head 2a7ed47c.

Boot path (your Aug 30 review):
- `Config::validate()` runs again after `bootstrap::merge_seeds` before key load.
- `RepoStore` uses `build_lock_pool(db.pool(), lock_pool_size(...), acquire_timeout)` instead of `db.pool().clone()`.
- Gossip reachability is back on the two-failure state machine (`gossip_ping_round` / `peer_ping_db_update`); `lock_pool_sizing` and both hysteresis unit tests pass.

CodeQL:
- Alert #39: inline `codeql[rust/cleartext-transmission]` on the loopback GET in `deny_harness.rs` (fixture-only; must request the withheld CID over HTTP).
- Alert #37: inline `codeql[rust/cleartext-logging]` on the intentional leak-witness panic in `assert.rs`.

Also on this head:
- CI deny-harness step now runs `--lib --test deny_harness` so the `join_or_abort_*` guards execute in PR checks.
- Existing-key load repairs a group/world-writable parent directory (`mode & 0o022`) before `boot_load_key`; read-only parents (e.g. `0555` in `sweep_failure_tolerated`) are left alone.
- Gossip aborts on HTTP shutdown and polls shutdown between peer probes; `list_peers` failures log a warning instead of being swallowed.

Checks run locally:
- `cargo test --locked -p gitlawb-node --features test-harness --lib --test deny_harness` (1182 lib + 50 deny_harness)
- `cargo test --locked --workspace`
- `cargo clippy --locked --workspace --all-targets -- -D warnings`

Ready for another look.

jatmn (changes requested):
I found issues that need to be addressed before this is ready. This review is written to **stop drip feedback**, not add another open-ended round. Read the overall guidance first — it explains why this PR has produced twelve review cycles and what to do differently so the next push is the last one before merge.

---

## Overall guidance — why feedback keeps dripping, and how to end it

This branch has **54 commits** and **twelve** prior `CHANGES_REQUESTED` reviews since July. That is not because reviewers are moving goalposts at random. It is the predictable outcome of how the PR has been structured. The remaining items below are small on head `2a7ed47`; the **pattern** is what keeps generating new rounds.

### 1. The PR is several projects in one branch

What started as “real-node deny harness” has accumulated, in the same series:

- lib+bin split and public module surface
- identity-key atomic publish, quarantine, recovery claims, delayed resweep, parent-dir repair
- shutdown grace enforcement (`drive_serve_with_grace`)
- gossip hysteresis restoration and shutdown polling
- deny registry / completeness guards (absorbed from #195)
- CodeQL config and CI wiring

Each layer is legitimate work, but **each production boot change re-opens security review** even when the stated deliverable is test infrastructure. Reviewers cannot sign off on “tests only” while `lib.rs` still changes identity recovery semantics every round.

**What to do:** For the merge push, **freeze production behavior**. No new boot-path logic, no new recovery edge cases, no “small follow-up” commits in `lib.rs`. If something still needs hardening, open a **follow-up PR** with its own review. The deny harness can merge without solving every identity-key corner case in the same branch.

Update the PR body to match reality. It still says “No behavior change, the 488-test node suite stays green.” Head materially changes boot (identity recovery, shutdown grace, gossip). Stale claims invite reviewers to re-audit production on every round because the stated contract does not match the diff.

### 2. Fix-one-review-item, break-another — the lib/bin extraction pattern

A recurring drip pattern on this branch:

1. Reviewer finds boot regression from lib/bin split (missing `validate()`, lock pool, gossip hysteresis).
2. Author fixes in `692b894`.
3. Reviewer finds new gaps in the fix (`2a7ed47`: parent-dir repair threshold, gossip shutdown, CodeQL still red).

This is classic **regression churn from moving code without a frozen contract**. The extraction moved `main.rs` → `lib.rs` but did not carry a single checklist of boot invariants on the first try.

**What to do:** Before requesting another review, run this **boot invariant checklist** yourself against `run()` vs pre-split `main.rs` (or vs `main` today) and paste the result in the PR comment so reviewers do not rediscover items:

- `Config::validate()` after `merge_seeds`
- `build_lock_pool` + `lock_pool_size`, not `db.pool().clone()` for `RepoStore`
- `build_http_client()` redirect policy, not `Client::new()`
- gossip `failed_once` hysteresis + `shutdown_rx` in ping round + `gossip_handle.abort()` after serve
- `drive_serve_with_grace` wired to `shutdown_grace_secs`
- identity key path: publish vs reload parent-dir policy **consistent**

If all six are true on your branch (they are on head today), say so explicitly and ask reviewers to treat further boot tweaks as **out of scope for this PR**.

### 3. The harness over-claims “comprehensive” while documenting exclusions

The registry, completeness guards, and PR prose claim runtime discharge of INV-1/INV-2/INV-8 across deny-bearing routes. The same codebase **documents** intentional exclusions:

- GraphQL mutations (#219)
- `get_encrypted_blob` (no IPFS stub)
- global pin/anchor listings (#121)
- `git_receive_pack` owner-push wiring (only signature row + protected-branch ad hoc test)
- replica register/unregister

That design is defensible, but **every exclusion is a hole reviewers will probe** unless the PR draws a hard boundary. “Comprehensive” invites “what about X?” forever.

**What to do:** Add a short **“Harness scope contract”** section to the PR body (not scattered comments). For each excluded surface, one line: *excluded because …, tracked in #N, covered by [unit test / other PR / accepted risk]*. Then stop expanding the registry in this PR unless a production gate is actually broken.

For `enforce_owner_push` specifically (see P3 finding below): **make a one-time decision** — either add one real-socket test and close the topic, or add one sentence to the scope contract (“wiring covered by `repos.rs` unit tests; socket sweep does not own push policy”) and **do not entertain further rounds on it**.

### 4. CI fixes are being attempted without verifying on push

CodeQL is still **FAIL** on head after `paths-ignore`, inline `codeql[...]` comments, and multiple “clear false positive” commits. Each attempt that does not green the check creates another review round.

**What to do:** Treat CodeQL as a **verify-on-push** gate, not a code-change guess:

1. Push the fix to the PR branch.
2. Wait for the CodeQL run to finish.
3. Only then request re-review, with a link to the green run **or** a link to Security-tab dismissals with rationale.

Do not stack another “maybe this suppression works” commit without that evidence. If dismissals are the team’s chosen path for fixture-only alerts, **do that once**, document it in the PR, and stop iterating on comment placement.

### 5. What actually blocks merge vs what does not

To end the back-and-forth, separate **merge blockers** from **follow-ups**:

| Item | Blocks merge? | Notes |
|---|---|---|
| CodeQL check FAIL | **Yes** | Required status; fixture false positives still fail the aggregate check |
| Parent-dir `0o022` vs `0o077` on reload | **No** (P3) | Real inconsistency in new code; not worse than `main` for `0755`; fix in this PR or #follow-up |
| `enforce_owner_push` socket coverage | **No** (P3) | Harness scope / claims alignment; production gate + unit tests exist |
| GraphQL / encrypted blob / global listings | **No** | Already documented out of scope |
| Further registry rows | **No** | Unless you choose to expand scope; otherwise freeze |

**Recommended close-out for the author:**

1. Green CodeQL (verify on push).
2. Update PR body: honest scope, boot invariant checklist signed off, harness scope contract with exclusions.
3. Freeze `lib.rs` production changes; follow-ups in separate PRs.
4. Optionally fix the two P3 items in this PR **or** open tracked follow-ups and stop debating them here.

If you do (1)–(3), the remaining review surface is intentionally narrow and this should be the last substantive round.

---

## Merge readiness

- [P1] Resolve the failing CodeQL check before merge
  `crates/gitlawb-node/tests/deny_harness.rs:307`
  `crates/gitlawb-node/tests/support/assert.rs:105`
  `.github/codeql/codeql-config.yml`

  **What is failing:** GitHub Advanced Security on head `2a7ed47` reports **2 high-severity** alerts and the aggregate `CodeQL` status is **FAIL** (`mergeStateStatus: BLOCKED`).

  1. **`anon_ipfs_read_of_withheld_blob_is_denied`** — `client.get(format!("{}{}", node.base_url, secret_path))` where `secret_path` is `/ipfs/{secret_cid}`. Query: `rust/cleartext-transmission`.
  2. **`assert_denied`** — `panic!("{reason}")` may include withheld OID/secret tokens. Query: `rust/cleartext-logging`.

  Both are intentional harness fixtures (loopback deny probe; leak witness on test failure). Not production exposure.

  **Root cause:** Mitigations on this branch did not clear the head run:

  - **`paths-ignore` for `**/tests/**`** — alerts still fire on changed test files in the PR diff. Do not assume this config alone greens PR scanning without a verified run.
  - **Inline `codeql[rust/...]` comments** — on line 305 (two lines above `.get`) and on the `panic!` line; GitHub still reports both on head.

  **How to fix (pick one path and verify on push):**

  - **Restructure** so CodeQL does not trace sensitive data into the flagged expressions (e.g. redact panic message; build URL without embedding `secret_cid` in the expression the query sees).
  - **Working suppression** on the **exact** flagged lines, confirmed against a finished CodeQL run.
  - **Security-tab dismissal** of both as false positives (“loopback deny-harness fixture only”), documented in the PR comment with dismissal links.

  Do not disable CodeQL or ignore production `src/`.

---

## Findings (non-blocking — fix here or in a follow-up, but do not extend this PR’s production scope to chase them)

- [P3] Align identity-key parent repair on reload with publish-time checks
  `crates/gitlawb-node/src/lib.rs:3434`
  `crates/gitlawb-node/src/lib.rs:1379`

  Reload repairs parent only when `mode & 0o022 != 0`; publish uses `ensure_key_parent_dir_private` when `mode & 0o077 != 0`. Example: parent `0755` is left loose on restart but tightened on first publish. **Not a regression from `main`** (base had no reload repair). Root cause: reload fast-path used a narrower mask than publish. Fix: call `ensure_key_parent_dir_private(parent)` unconditionally on existing-key load (or gate on `0o077`), add a `0755` reload test. Acceptable as post-merge follow-up if you freeze boot changes now.

- [P3] Decide once on `enforce_owner_push` real-socket coverage
  `crates/gitlawb-node/tests/support/routes.rs:458`
  `crates/gitlawb-node/tests/deny_harness.rs:145`
  `crates/gitlawb-node/src/api/repos.rs:1860`

  `git-receive-pack` has three signed push gates: signature (registry), `owner_push_rejection` (default on, unit-tested only), branch protection (ad hoc test). Registry sweep does not catch wiring regressions for `owner_push_rejection`. Root cause: registry design + documented exclusion of `git_receive_pack` from owner orphan scan. **Production gate is correct on head.** Fix: add one socket test on unprotected branch **or** add one line to harness scope contract and close the topic. Do not leave implicit — that is what causes drip.

---

## Summary for the author

**Merge when:** CodeQL is green (verified on push) and PR body reflects actual scope.

**Stop dripping when:** Production boot is frozen in this PR, harness exclusions are written down once, and P3 items are either fixed or explicitly deferred with issue links — not left implicit for the next reviewer to reopen.

beardthelion:
Head is now `988e8ea4`. All four findings from the last review are closed.

**CodeQL cleartext-transmission** (`deny_harness.rs:307`): the inline anonymous GET traced a withheld CID directly into a URL. Routed through `support::anon_get` so URL construction crosses a function boundary, matching the unflagged `signed_request` shape.

**CodeQL cleartext-logging** (`assert.rs:105`): `panic!("{reason}")` emitted withheld tokens. Added `redact_withheld` at the panic boundary; `check_denied` error messages are unchanged (their unit tests intentionally name the token).

**Parent-dir mask inconsistency** (`lib.rs:3434`): the reload fast-path checked only write bits (`0o022`) while the publish helper checked all group/world bits (`0o077`). A `0755` parent survived reload but was tightened on publish. Aligned the reload mask to `0o077`. Regression test `reload_tightens_0755_parent_dir_to_0700` proves a 0755 parent is tightened on reload (RED with the old mask, GREEN after).

**enforce_owner_push socket coverage**: the existing protected-branch test could not distinguish which gate fired the 403. Added `signed_stranger_push_to_unprotected_branch_is_forbidden` on an unprotected repo, isolating `enforce_owner_push` from the protected-branch gate, with an owner control (asserts NOT 403).

Verification:
- `cargo test --features test-harness --test deny_harness`: 53 passed, 0 failed
- `cargo test --lib -- identity_key_tests`: 101 passed, 0 failed, 1 ignored
- `cargo fmt --check`: clean
- `cargo clippy --workspace --all-targets -D warnings`: clean (pre-push gate)
- RED-then-GREEN verified for the mask fix

The production behavior change is limited to enforcing the already-documented owner-only `0700` parent-directory invariant during reload, not just first publish.

jatmn (changes requested):

This review is scoped to what head `988e8ea` actually changes and what blocks merge today. The deny harness itself looks structurally sound on head (parent-dir reload mask and `enforce_owner_push` socket coverage landed in the latest commit). The items below are the remaining gaps — not a request to reopen boot-path design or expand harness registry scope.

---

## Merge readiness

- [P1] Green the required CodeQL check (fixture false positive, not a production leak)
  `crates/gitlawb-node/tests/support/assert.rs:115-118`
  `.github/codeql/codeql-config.yml`

  **What is failing:** On head `988e8ea`, the aggregate **CodeQL** required check is **FAIL** (`mergeStateStatus: BLOCKED`). GitHub Advanced Security reports one open alert on this PR:

  | Rule | Location | Severity |
  |------|----------|----------|
  | `rust/cleartext-logging` | `assert.rs:118` (`panic!("{redacted}")`) | warning |

  The earlier `rust/cleartext-transmission` alert on `deny_harness.rs` is **cleared** on head (URL built via `secret_path` variable instead of embedding the CID in the `.get()` expression CodeQL traced).

  **This is not a production security defect.** `assert_denied` runs only in the deny harness. It panics when a deny probe fails — i.e., when the test has already detected a problem. At runtime, `redact_withheld` strips withheld OIDs/CIDs from the message before `panic!`. No production path logs secrets; no loopback client receives them from this helper on the success path.

  **Root cause:** CodeQL's taint analysis follows withheld tokens from `check_denied_with_headers` into the `reason` string (the error path embeds the token in `withheld token {token:?} leaked…`), then through `redact_withheld`, and still flags the `panic!` sink. Static analysis does not treat `redact_withheld` as a sanitizer. Branch mitigations did not clear the head run:

  - `paths-ignore: **/tests/**` — does exclude `tests/support/assert.rs`, but **PR diff scanning still raised the alert** on the changed file; do not assume `paths-ignore` alone greens PR checks without a verified push.
  - `paths-ignore: **/test_support/**` — does **not** match anything in this repo (`tests/support/` is the actual layout; in-crate `src/test_support.rs` is a file, not a directory).
  - Inline `codeql[...]` comments — did not clear the head alert on the panic line.

  **How to fix (pick one path, verify on push before re-requesting review):**

  1. **Restructure the failure witness so taint never reaches the sink** (preferred if you want zero Security-tab noise):
     - Build the panic message from fixed strings only, e.g. `panic!("deny assertion failed: status/body/header leak check")`.
     - Log redacted detail via a separate channel CodeQL does not trace to a log sink, or assert with `assert!(false, "…")` using only static text and pass redacted detail as separate arguments that do not embed raw tokens in the format string.
     - Keep `redact_withheld` unit tests — they still validate the redaction helper even if `assert_denied` no longer panics with interpolated secrets.

  2. **Security-tab dismissal** (acceptable for fixture-only alerts if the team uses dismissals for test harnesses):
     - Dismiss `rust/cleartext-logging` on `assert.rs:118` as false positive ("loopback deny-harness test failure witness; tokens redacted before panic; no production path").
     - Link the dismissal URLs in a PR comment. Dismissal greens the aggregate check only after GitHub processes it — confirm on the finished run.

  3. **Working suppression on the exact flagged line** — only if you confirm on a finished CodeQL run that the suppression is honored (prior inline attempts on this branch did not).

  **Do not:** disable CodeQL, add broad `paths-ignore` for `src/`, or weaken `check_denied` leak detection on the success path.

---

## Findings

- [P2] Restore gossip readiness tests lost in the lib+bin move (coverage regression, not harness logic)
  `crates/gitlawb-node/src/lib.rs` — `gossip_ssrf_tests` (~3640–3815)
  merge-base reference: `crates/gitlawb-node/src/main.rs` (pre-split)

  **What broke:** The lib+bin extraction moved `ping_peer_readiness_with_timeout` into `lib.rs` **unchanged** (function body is byte-identical to merge-base). The **unit tests** beside it were not fully carried over: merge-base had **7** `ping_peer_*` tests; head has **3** renamed `ping_peer_health_*` tests. Production gossip behavior is the same; **CI guardrails shrank**.

  **Root cause:** Mechanical `main.rs` → `lib.rs` move copied production code but dropped part of the `gossip_ssrf_tests` module. Commit `692b894` ("restore lib/bin boot regressions dropped in the deny-harness extraction") fixed several boot-path gaps but did not restore this gossip suite. This is collateral damage from the crate split, not a defect in the deny harness routes, registry, or `spawn_node` wiring.

  **What head still covers:**

  | Retained test | What it guards |
  |---------------|----------------|
  | `ping_peer_health_does_not_follow_redirect` | Legacy `/health` 302 must not be followed (SSRF) when `/ready` is 404 |
  | `ping_peer_health_reports_success_on_200` | Legacy `/health` 200 counts as healthy when `/ready` is 404 |
  | `ping_peer_health_reports_unhealthy_on_connection_error` | Connection errors map to unhealthy |

  **What is missing and why it matters** (distinct from retained tests):

  | Dropped test (merge-base name) | Gap if not restored |
  |------------------------------|---------------------|
  | `ping_peer_readiness_reports_success_on_200` | `/ready` 200 happy path never exercised — retained tests only hit `/ready` 404 then `/health` |
  | `ping_peer_readiness_ignores_liveness_only_health` | **Most important:** `/ready` 503 + `/health` 200 must **not** mark peer ready. Current code handles this at ```1260:1266:crates/gitlawb-node/src/lib.rs``` (non-404, non-success → `false` without fallback), but nothing in CI would catch a refactor that reintroduced `/health` fallback on 503 |
  | `ping_peer_readiness_legacy_fallback_shares_deadline` | Serial `/ready` + `/health` probes must share one timeout envelope — subtle regression class |
  | `ping_peer_readiness_falls_back_for_legacy_peer` | Explicit expect-count contract for 404→200 sequence (mostly overlapping with retained health test, lowest priority) |

  `ping_peer_readiness_legacy_fallback_does_not_follow_redirect` overlaps heavily with the retained redirect test; restoring it is optional.

  **Smallest fix:** Copy the four merge-base tests (or the top three rows excluding the optional overlap) into `gossip_ssrf_tests` in `lib.rs`, using `ping_peer_readiness_with_timeout` where the shared-deadline test needs it. No production code change required unless a test fails.

  **Scope boundary:** This is lib-move test parity, not deny-harness expansion. Do not fold this into registry rows, GraphQL coverage, or further boot-path changes. If you prefer to defer, open a tracked follow-up and note it in the PR comment — but merging as-is **reduces** `main`'s federation-readiness coverage for unchanged code.

---

## Author guidance (close-out without more review drift)

These are not additional findings — they are how to land this PR without another drip round:

1. **Treat CodeQL as verify-on-push.** Push the fix or dismissal, wait for the CodeQL run to finish, link the green run (or dismissal URLs) in a PR comment, then request re-review. Do not stack another untested suppression commit.

2. **Freeze production boot changes for the merge push.** Head already includes identity recovery, shutdown grace, and gossip hysteresis restoration. Further `lib.rs` boot edits in this branch re-open security review unrelated to the harness deliverable. Follow-ups belong in separate PRs.

3. **Update the PR body once, honestly.** Replace "No behavior change" with a short scope contract: what the harness covers (REST deny paths, INV-1/2/8 cases listed), what it excludes (GraphQL #219, iCaptcha enforce mode, global listings), and that the lib+bin split also carries boot hardening already on head. A signed boot-invariant checklist (validate, lock pool, `build_http_client`, gossip hysteresis+abort, `drive_serve_with_grace`, parent-dir 0o077) lets reviewers treat further boot tweaks as out of scope.

4. **Do not expand harness scope in this PR** to chase lower-impact owner gates, GraphQL, or encrypted-blob surfaces — those are documented deferrals with existing unit-test guards. The registry/completeness machinery is already large; more rows invite "what about X?" rounds without proportional safety gain.

beardthelion:
Both findings from the 2026-09-08 review are closed on 4c42d78b.

CodeQL cleartext-logging (P1): replaced the interpolated `panic!("{redacted}")`
with a fixed string so no tainted data reaches the sink. CodeQL's taint analysis
traced withheld tokens through `redact_withheld` without recognizing it as a
sanitizer; a fixed message sidesteps that entirely. `redact_withheld` is still
called (keeping it live) and its unit tests still pass. This is a test-only
fixture: no production path calls `assert_denied`. Will verify the CodeQL run
greens on push.

Gossip readiness tests (P2): restored the 4 tests dropped in the lib+bin split:
`ping_peer_readiness_reports_success_on_200`, `ping_peer_readiness_ignores_liveness_only_health`,
`ping_peer_readiness_falls_back_for_legacy_peer`, and
`ping_peer_readiness_legacy_fallback_shares_deadline`. No production code
change; function signatures are unchanged. 9/9 gossip tests pass, 53/53 deny
harness tests pass, fmt and clippy clean.

Per your close-out guidance: no boot-path changes, no harness scope expansion,
no further production edits in this branch.

jatmn (dismissed):
LGTM

jatmn (approved):
LGTM (for real this time lol)

PR #195, opened 2026-07-13
test(node): invariant deny-prober (driven registry + completeness cross-check)

beardthelion:
Stacks on #194 (base `feat/real-node-deny-harness`), reusing its real-node spawn rig and RFC-9421 signing client. Review after #194.

This turns the hand-written deny cases into a driven registry: one declarative table of every route that carries a runtime deny, and a sweep that boots a real node and drives each one hostile-then-authorized.

## What's here

- **Registry** (`tests/support/routes.rs`): every deny-bearing route as a row (owner-gate 403, read-gate 404, signature 401), each classified by reading its handler, not the route name. 12 owner-gate + 1 signature + 14 repo-scoped read rows.
- **Probe generator** (`tests/support/probe.rs`): a two-repo fixture (public owner-gate substrate, private read-gate substrate, plus a seeded PR and issue so the author-or-owner close gates are reached rather than 404'd on an absent entity) and `probes_for`, which expands each row into the hostile request plus a positive twin so a deny for the wrong reason cannot false-pass.
- **Runtime sweep** (`deny_bearing_registry_...`): walks the registry against a real node. Hostile probe must return the exact deny and leak nothing; owner-gate twins must reach the handler (not 403); read twins must return a non-empty 2xx (an empty 2xx is a denial rendered as success). A terminal invariant asserts one hostile probe per row was actually driven, so the sweep can't pass by testing nothing.
- **Completeness cross-check** (`completeness::...`, no DB): a source scrape of `server.rs` (multiline `.route` parser, floor 90) that keeps the registry honest: no row points at a route that no longer mounts, and no handler carrying an owner-gate marker escapes the registry. Complements the in-crate `authz_guard` egress guard (which proves handlers are gated) by proving the deny-bearing ones are actually driven, and reaches the non-API git mounts it never sees.

## Verification

Every deny class was mutation-verified load-bearing, RED-then-GREEN by execution: breaking `did_matches` drove the owner-gate hostile to 500 and, inverted, drove the owner twin to 403; neutering `authorize_repo_read` drove the read hostile to 200 and, as always-deny, the read twin to 404; the signature assertion was isolated at `git-receive-pack`; both completeness floors and both drift guards were tripped and named their target. All reverted. 22 harness tests green.

CI already runs this file on every PR (the `--features test-harness --test deny_harness` step added in #194), so U3 and U4 run there with no config change.

### DIFF
--- a/crates/gitlawb-node/src/test_harness.rs
+++ b/crates/gitlawb-node/src/test_harness.rs
@@ -169,6 +169,16 @@ impl TestNode {
         record.id
     }
 
+    /// Mark `branch` as protected on `repo_id`, so a non-owner push to it is
+    /// rejected by the branch-protection gate (used by the deny-prober's
+    /// protected-push probe — #195, F3).
+    pub async fn seed_protected_branch(&self, repo_id: &str, branch: &str, created_by: &str) {
+        self.db
+            .protect_branch(repo_id, branch, created_by)
+            .await
+            .expect("seed protected branch");
+    }
+
     /// Seed an open PR `number` in `repo_id`, authored by `author_did`. The
     /// author-or-owner close gate (`close_pr`) loads the PR *before* it runs, so a
     /// deny-prober row for `.../pulls/{number}/close` needs the PR to exist or a
--- a/crates/gitlawb-node/src/test_harness.rs
+++ b/crates/gitlawb-node/src/test_harness.rs
@@ -26,7 +26,7 @@ use uuid::Uuid;
 use gitlawb_core::identity::Keypair;
 
 use crate::config::Config;
-use crate::db::{Db, RepoRecord, VisibilityMode};
+use crate::db::{Db, PullRequest, RepoRecord, VisibilityMode};
 use crate::rate_limit::{RateLimiter, TrustedProxy};
 use crate::state::AppState;
 
@@ -167,6 +167,41 @@ impl TestNode {
         };
         self.db.create_repo(&record).await.expect("seed repo");
         record.id
+    }
+
+    /// Seed an open PR `number` in `repo_id`, authored by `author_did`. The
+    /// author-or-owner close gate (`close_pr`) loads the PR *before* it runs, so a
+    /// deny-prober row for `.../pulls/{number}/close` needs the PR to exist or a
+    /// stranger gets a 404 (absent entity) instead of the 403 the gate emits.
+    pub async fn seed_pr(&self, repo_id: &str, number: i64, author_did: &str) {
+        let now = Utc::now().to_rfc3339();
+        let pr = PullRequest {
+            id: Uuid::new_v4().to_string(),
+            repo_id: repo_id.to_string(),
+            number,
+            title: "seed pr".to_string(),
+            body: None,
+            author_did: author_did.to_string(),
+            source_branch: "feature".to_string(),
+            target_branch: "main".to_string(),
+            status: "open".to_string(),
+            merged_by_did: None,
+            merged_at: None,
+            created_at: now.clone(),
+            updated_at: now,
+        };
+        self.db.create_pr(&pr).await.expect("seed pr");
+    }
+
+    /// Seed issue `issue_id` (a git-ref JSON blob authored by `author_did`) in the
+    /// repo's bare on-disk store. Like [`Self::seed_pr`], the `close_issue`
+    /// owner-or-author gate loads the issue before it runs, so the deny-prober row
+    /// for `.../issues/{id}/close` needs the issue present to reach the 403.
+    pub fn seed_issue(&self, owner_did: &str, name: &str, issue_id: &str, author_did: &str) {
+        let slug = owner_did.replace([':', '/'], "_");
+        let bare = self.repos_dir.join(&slug).join(format!("{name}.git"));
+        let json = format!(r#"{{"author":"{author_did}"}}"#);
+        crate::git::issues::create_issue(&bare, issue_id, &json).expect("seed issue");
     }
 
     /// Add a path-scoped visibility rule restricting `path_glob` to
--- a/crates/gitlawb-node/tests/deny_harness.rs
+++ b/crates/gitlawb-node/tests/deny_harness.rs
@@ -11,6 +11,8 @@ mod support;
 use std::process::Command;
 
 use support::assert::assert_denied;
+use support::probe::{probes_for, Expect, Fixture, Probe, Signer};
+use support::routes::deny_bearing_routes;
 use support::signing::signed_request;
 
 use gitlawb_core::cid::Cid;
@@ -403,6 +405,142 @@ async fn anon_clone_excludes_withheld_subtree_blobs(pool: sqlx::PgPool) {
     assert!(
         listing.contains(&public_oid),
         "public blob {public_oid} must be present (withhold is subtree-scoped); listing:\n{listing}"
+    );
+}
+
+// ── U3: drive the whole deny-bearing route registry over the real stack ───────
+
+/// Send one [`Probe`] against the running node, signing as the probe's identity.
+/// Anon requests are unsigned; Owner/Stranger requests carry a real RFC-9421
+/// signature for the fixture's owner / stranger keypair.
+async fn send_probe(
+    client: &reqwest::Client,
+    base_url: &str,
+    fixture: &Fixture,
+    probe: &Probe,
+) -> reqwest::Response {
+    let rb = match probe.signer {
+        Signer::Anon => client
+            .request(probe.method.clone(), format!("{base_url}{}", probe.path))
+            .body(probe.body.clone()),
+        Signer::Owner => signed_request(
+            client,
+            probe.method.clone(),
+            base_url,
+            &probe.path,
+            probe.body.clone(),
+            &fixture.owner,
+        ),
+        Signer::Stranger => signed_request(
+            client,
+            probe.method.clone(),
+            base_url,
+            &probe.path,
+            probe.body.clone(),
+            &fixture.stranger,
+        ),
+    };
+    let rb = if probe.json {
+        rb.header("content-type", "application/json")
+    } else {
+        rb
+    };
+    rb.send().await.unwrap_or_else(|e| {
+        panic!(
+            "probe failed to send: {} [{} {}]: {e}",
+            probe.label, probe.method, probe.path
+        )
+    })
+}
+
+/// Walk every deny-bearing route (U1 registry), expand each into its hostile
+/// probe plus positive twin (U2), and drive them against a real node: the
+/// hostile request must return the exact deny status and leak nothing, and the
+/// twin must reach the handler (owner-gate: not 403; read-gate: 2xx). This is
+/// the runtime discharge of INV-1/INV-2/INV-8 across the whole registry, not one
+/// hand-written case at a time.
+///
+/// Terminal anti-vacuous-green invariant: exactly one hostile probe per row is
+/// driven and the count equals the registry size, so a row that silently
+/// produced no probe fails here instead of the sweep passing by testing nothing.
+#[sqlx::test]
+async fn deny_bearing_registry_denies_hostile_and_admits_authorized(pool: sqlx::PgPool) {
+    let node = spawn_node(pool).await;
+    // A bounded timeout so a wedged route fails the suite rather than hanging it.
+    let client = reqwest::Client::builder()
+        .timeout(std::time::Duration::from_secs(30))
+        .build()
+        .unwrap();
+    let fixture = Fixture::seed(&node).await;
+
+    let rows = deny_bearing_routes();
+    assert!(
+        !rows.is_empty(),
+        "deny-bearing route registry is empty — nothing to sweep"
+    );
+
+    let mut hostile_driven = 0usize;
+    let mut twins_driven = 0usize;
+
+    for row in rows {
+        let probes = probes_for(row, &fixture);
+        assert!(
+            !probes.is_empty(),
+            "row {} {} produced no probes",
+            row.method,
+            row.path
+        );
+        for probe in &probes {
+            let ctx = format!("{} [{} {}]", probe.label, probe.method, probe.path);
+            let resp = send_probe(&client, &node.base_url, &fixture, probe).await;
+            let status = resp.status();
+            match &probe.expect {
+                Expect::Deny(code) => {
+                    hostile_driven += 1;
+                    assert_eq!(
+                        status.as_u16(),
+                        *code,
+                        "hostile probe must deny with {code}, got {status}: {ctx}"
+                    );
+                    // INV-8 shape guard: rechecks the status and that the body is
+                    // neither an empty-200-as-success nor carrying withheld data.
+                    assert_denied(resp, *code, &[]).await;
+                }
+                Expect::Not403 => {
+                    twins_driven += 1;
+                    assert_ne!(
+                        status.as_u16(),
+                        403,
+                        "owner-reachability twin must reach the handler (not 403): {ctx}"
+                    );
+                }
+                Expect::Ok2xx(token) => {
+                    twins_driven += 1;
+                    assert!(
+                        status.is_success(),
+                        "read-reachability twin must be 2xx, got {status}: {ctx}"
+                    );
+                    if let Some(tok) = token {
+                        let body = resp.text().await.unwrap_or_default();
+                        assert!(
+                            body.contains(tok),
+                            "read twin 2xx body missing expected token {tok:?}: {ctx}"
+                        );
+                    }
+                }
+            }
+        }
+    }
+
+    assert_eq!(
+        hostile_driven,
+        rows.len(),
+        "expected exactly one hostile probe per deny-bearing row ({} rows), drove {hostile_driven}",
+        rows.len()
+    );
+    assert!(
+        twins_driven >= 1,
+        "no positive twins were driven — the reachability proof is missing"
     );
 }
 
--- a/crates/gitlawb-node/tests/deny_harness.rs
+++ b/crates/gitlawb-node/tests/deny_harness.rs
@@ -374,6 +374,191 @@ async fn withheld_path_blob_read_is_denied(pool: sqlx::PgPool) {
     );
 }
 
+// ── U3: get_pr_diff SECOND (per-path) gate — a diff touching a withheld path ──
+
+/// Build a git v0 receive-pack body creating `refs/heads/<branch>` at a fresh
+/// commit that adds `files` on top of the existing `main`, force-updating main to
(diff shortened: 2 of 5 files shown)


jatmn (changes requested):

## Findings

- [P2] Cover the deferred read-gate routes in the completeness invariant
  `crates/gitlawb-node/tests/support/routes.rs:351`
  The registry explicitly leaves out `get_issue`, `get_pr`, PR diff/review/comment reads, `get_cert`, `get_bounty`, `get_encrypted_blob`, and path-scoped `get_tree`, although these handlers return the same 404 authorization denial before looking up their entities.  U3 only drives registry rows, and U4 checks mounted rows plus owner-gate markers only, so removing or bypassing `authorize_repo_read` on any of these endpoints leaves the new real-node sweep green.  Seed the required entities and include the routes (with positive twins), or make the completeness check derive and require every mounted read-denying handler.

- [P2] Assert that hostile registry responses do not expose the private fixture data
  `crates/gitlawb-node/tests/deny_harness.rs:507`
  Every hostile probe calls `assert_denied` with an empty withheld-token list.  That verifies only the status: a 403/404 body containing the private blob content, object ID, or repository metadata would pass even though this sweep claims the denial "leak[s] nothing."  Carry the relevant private fixture tokens on the probes/rows and pass them to `assert_denied` for the affected read routes.

- [P2] Do not let inline owner gates bypass the registry drift check
  `crates/gitlawb-node/tests/deny_harness.rs:788`
  The orphan scan deliberately looks only for `require_repo_owner(`/`require_owner(` and excludes `did_matches`, but existing owner-only handlers such as `protect_branch` and `unprotect_branch` use the latter idiom.  A future mounted owner-only handler using that established pattern can be omitted from `deny_bearing_routes()` while U4 still passes, so the advertised completeness guard does not keep the owner-gate registry complete.  Extend the classifier to recognize the owner-specific inline pattern (or otherwise tie mounted owner routes to the registry) rather than excluding that gate class.

beardthelion:
All three fixed; each guard now goes red when the protection it checks is removed.

**Read-gate orphans** — added a read-gate orphan guard symmetric to the owner one: it scans `src/api` for every `authorize_repo_read` / `visibility_check` handler and requires each to be either a driven `ReadGate` row or an enumerated `READ_GATE_NOT_DRIVEN` entry with a reason, replacing the free-text DEFERRED/EXCLUDED prose as the single source of truth. Deleting a driven row (`get_repo`) or an allowlist entry (`get_cert`) both fail it, and a staleness check rejects a dead exemption. The scan surfaced three repo-root reads the prose had missed (`get_star_status`, `get_icaptcha_proof`, `replicate_encrypted_blobs`), now allowlisted with a reason rather than silently dropped.

**Sweep leak assertion** — the private fixture now seeds a distinctive secret and captures its blob OID; the read-gate hostile probes carry those tokens into `assert_denied` (was `&[]`), so a 404/403 body echoing withheld content or the OID fails. Owner-gate (403) and signature (401) probes fire on NO_ENTITY repos before any entity lookup, so they keep an honest empty list.

**Owner did_matches idiom** — the owner-orphan scan now folds in `has_owner_did_matches`, matching the inline `did_matches(caller, &record.owner_did)` form `protect_branch` / `unprotect_branch` use, without matching the signer-self (`register_replica`) or author (`close_pr`) forms — pinned by unit tests on the exact bodies. Removing the `protect_branch` row now fails where it passed before.

jatmn (changes requested):

## Findings

- [P2] Exercise read denial for a signed non-reader
  `crates/gitlawb-node/tests/support/probe.rs:185`
  Every `ReadGate` hostile probe is anonymous, although the fixture also creates a signed `stranger`. This leaves the signed-non-reader branch of every private read untested: a regression that keeps anonymous callers at 404 but treats any valid HTTP signature as authorized exposes private repositories while the registry sweep remains green. Generate a `Signer::Stranger` 404/no-leak hostile probe for each read row (and adjust the hostile-count invariant accordingly).

- [P2] Do not exempt probeable private root reads from the runtime sweep
  `crates/gitlawb-node/tests/deny_harness.rs:1082`
  `get_star_status`, `get_icaptcha_proof`, and `replicate_encrypted_blobs` are allowlisted despite needing no additional fixture entity. The new guard only proves that their functions retain an `authorize_repo_read(`/`visibility_check(` marker; it never drives a request. Consequently an early return or conditional bypass that leaves the marker in the body can leak private-repo metadata while every new test passes. Add these as `ReadGate` rows with hostile and authorized probes rather than treating them as a source-only exemption.

- [P2] Keep the protected-push owner gate under a real probe
  `crates/gitlawb-node/tests/deny_harness.rs:929`
  The owner-marker scan recognizes the `did_matches(&auth.0, &record.owner_did)` check in `git_receive_pack`, then removes that handler because its registry row only tests the unsigned 401 path. A signed stranger pushing to a protected branch therefore never reaches a test; deleting or inverting the 403 at `src/api/repos.rs:897-923` leaves both the sweep and completeness checks green. Add a signed-stranger protected-branch push probe with an owner control, or keep this gate in the completeness invariant until it is driven.

- [P2] Withhold private repository metadata in the denial-body assertion
  `crates/gitlawb-node/tests/support/probe.rs:199`
  The leak assertion carries only the private blob content and its object ID. `Fixture::seed` discards the private repository record/ID, so a 404 that serializes the private record's internal UUID or visibility metadata still passes `assert_denied`. That leaves the repository-metadata portion of the earlier no-leak request unverified. Preserve distinctive private repository metadata in the fixture, include it in the hostile probes' withheld tokens, and add the corresponding synthetic leaking-body check.

beardthelion:
All four addressed on `a5fa334`, each guard proven load-bearing by reverting the exact production auth-line it protects.

**Signed non-reader** (F1). Every ReadGate row now drives a `Signer::Stranger` 404/no-leak probe alongside the anon one, and a dedicated `readgate_stranger_driven` counter asserts one per read row so a dropped probe fails loudly rather than being absorbed into the total. Reverting `authorize_repo_read` to authorize any signed caller makes the stranger probe leak (200) while the anon probe stays 404; dropping the stranger probe fails the counter.

**Driven, not source-only** (F2). `get_star_status`, `get_icaptcha_proof`, and `replicate_encrypted_blobs` are now real ReadGate rows and out of `READ_GATE_NOT_DRIVEN`. For each, a runtime bypass that keeps the `authorize_repo_read`/`visibility_check` marker but ignores the denial makes the driven probe leak (200) while the source scan stays green — the exact bypass a marker check cant see.

**Protected-push 403 under a real probe** (F3). Added `signed_stranger_protected_branch_push_is_forbidden`: a signed non-owner push to a protected branch expects 403, with an owner control that must not be 403 (plus a `seed_protected_branch` harness helper). Inverting the gate drops the stranger 403; denying all on a protected branch makes the owner control fire. git_receive_pack stays out of the owner-orphan scan because that gate is now driven here, not because it is unchecked.

**Repo metadata withheld** (F4). The fixture preserves the private repo id and the read-gate no-leak assertion withholds it; a synthetic 404 body carrying the UUID fails `check_denied` (it passed before the token was added). No live denial path leaks it today, so this is regression-hardening.

No production behavior ships — the auth-line reverts were RED-proof scaffolding, all restored.

jatmn (changes requested):

## Findings

- [P1] Restore the configured shutdown-grace deadline
  `crates/gitlawb-node/src/lib.rs:539`
  This replaces the base branch's `drive_serve_with_grace` race with a plain `with_graceful_shutdown(...).await`, then explicitly discards `grace` at line 571. Axum will now wait indefinitely for a stalled in-flight request after SIGTERM, so teardown and a deployment restart can hang forever despite `GITLAWB_SHUTDOWN_GRACE_SECS` documenting that the server exits after that duration. Rebase onto the target (or restore the deadline race and its tests) rather than rolling this behavior back.

- [P1] Do not revert the atomic identity-key creation path
  `crates/gitlawb-node/src/lib.rs:1007`
  The target branch's `create_new(...).mode(0o600)`/AlreadyExists handling is replaced by `exists()` followed by `fs::write()` and a later chmod. Concurrent first starts can both generate and overwrite the same PEM, leaving running nodes with divergent DIDs; on Unix the PEM is also created with umask-derived permissions before line 1027 narrows it, exposing the private signing key during that window. Keep the atomic create, winner-load retry, and existing-key permission tightening from the base branch.

- [P2] Exercise both sides of the owner-or-author close rule
  `crates/gitlawb-node/tests/support/probe.rs:65`
  The fixture seeds each close target with `author_did == owner_did`, and the generated positive probe only signs as that same key. Consequently the new sweep passes if either close handler loses its `is_author` branch: a real contributor can no longer close their own PR/issue, but the owner-as-author control still reaches the handler and the stranger still receives 403. Seed these resources with a distinct author and add an author-signed positive probe (with the owner control retained) so the declared owner-or-author contract is actually load-bearing.

- [P2] Drive the remaining deferred private read endpoints
  `crates/gitlawb-node/tests/deny_harness.rs:1073`
  `get_issue`, `get_pr`, the PR/comment/review reads, `get_cert`, `get_bounty`, `get_encrypted_blob`, and path-scoped `get_tree` are still source-only allowlist entries. The new guard only checks that their bodies contain an authorization-marker call, so an early return or ignored authorization result that leaves that call present keeps the guard green while exposing private data. This leaves the earlier request to cover the deferred 404-denying reads unresolved; seed the necessary entities and run their anonymous/signed-stranger denial and authorized positive probes.

beardthelion:
Both P2s are fixed, and the same pass turned up a third instance of the class, so I generalized the invariant rather than patch the two.

**Owner-or-author (probe.rs).** The close rows were driven with only an owner twin while the fixture seeded `author == owner`, so the author arm was dead. Reworked into a general `MultiPrincipalGate` that carries its authorizing-principal set and drives a `Not403` twin per arm plus the stranger 403. `close_pr`/`close_issue` now drive owner+author; reverting either `is_owner` or `is_author` on either handler turns that arm's twin red. The same rework surfaced `dispute_bounty` (creator-or-claimant) as the third untested-arm gate, so it's driven too. `registry_internal_consistency` now derives the required twin set from each row's declared arms, so a future multi-arm handler can't register with an arm untested.

**Deferred reads (deny_harness).** The ten `READ_GATE_NOT_DRIVEN` entries were source-scanned for a marker, never driven. Nine are now real `ReadGate` rows (anon/signed-stranger 404, owner 2xx), each carrying its own seeded marker in the withheld set, so a 404 that leaks that read's own content fails, not just a status check. `get_encrypted_blob` stays deferred under a closed structural reason (its owner twin needs a live IPFS backend the harness doesn't have), and the free-text reason list is now a closed enum. Added a completeness gate that derives the required read-gate set from the mounted route table, so a read that gates via an unrecognized helper can't slip past.

Every guard proven load-bearing by reverting the exact production line and observing red.

**The two lib.rs findings (shutdown-grace, atomic-key)** aren't in this PR's own diff, which is test-only. They're inherited from the base branch at `30672cf`, a point before #194 landed its own grace and atomic-key fixes (both present on #194's head). They clear when this branch rebases onto #194's current tip.

Separately, I hit a pre-existing gap while generalizing the invariant: `submit_bounty`'s claimant gate is vacuous under the existing `every_in_scope_mutation_has_its_gate` substring check. Out of scope here; filed as #FOLLOWUP.

jatmn (changes requested):

## Findings

- [P2] Pin the object format in the temporary Git repositories
  `crates/gitlawb-node/tests/support/probe.rs:506`
  The new fixture initializes its worktree without an object-format argument, then compares its blob ID with the SHA-1 ID produced by `seed_bare_repo(..., "sha1")`. On a machine with `init.defaultObjectFormat=sha256`, this produces 64-character IDs and the assertion at line 517 fails before the registry sweep runs. The analogous helper in `tests/deny_harness.rs:407` has the same assumption and will construct incompatible ref updates. Initialize both worktrees explicitly with `--object-format=sha1`, matching their served fixture repositories.

- [P2] Include the remaining identity-gated bounty mutations in the invariant sweep
  `crates/gitlawb-node/tests/support/routes.rs:172`
  `submit_bounty`, `approve_bounty`, and `cancel_bounty` each return a runtime 403 for a non-claimant or non-creator, but none is a registry row or has a real-node hostile/control probe. The source guard in `api/mod.rs` only checks that their function text still contains `did_matches(`, so an ignored result or conditional bypass can expose these mutations while every new registry and completeness check stays green. Seed status-appropriate bounty fixtures and drive stranger denials plus claimant/creator controls for these routes, or make an explicit, enforced exclusion that does not claim the registry covers every runtime deny.

beardthelion:
Both addressed on 07abcc6.

**Object format (P2).**

> On a machine with `init.defaultObjectFormat=sha256`, this produces 64-character IDs and the assertion at line 517 fails before the registry sweep runs.

Both worktree inits now pass `--object-format=sha1`, with a comment tying the pin to the served sha1 fixture repos. One precision on the repro: git here is 2.43.0, which predates the `init.defaultObjectFormat` key (git 2.47), so that config alone is silently ignored; the same hazard reproduces via `GIT_DEFAULT_HASH=sha256` (git 2.29+), and the pin closes both spellings. Observed before the fix: the registry sweep panics at probe.rs:518 with a 64-character blob OID against the fixture's sha1 OID, and with only the probe site pinned, `get_pr_diff_withheld_path_is_denied` independently fails from the second site's incompatible ref update (owner push 400 vs 200). After the fix, both tests pass under `GIT_DEFAULT_HASH=sha256` and under the default config.

**Bounty mutation gates (P2).**

> `submit_bounty`, `approve_bounty`, and `cancel_bounty` each return a runtime 403 for a non-claimant or non-creator, but none is a registry row or has a real-node hostile/control probe.

All three are now registry rows driven by execution. Each is a `MultiPrincipalGate` row (claimant arm for submit, creator arm for approve and cancel) with a signed-stranger hostile asserting exactly 403 and a Not403 twin per declared arm; the gate-class doc now covers one-or-more named principals, since the prober only ever required non-empty arms. These handlers check bounty status before the auth gate, so each row seeds its own bounty through the real API in the gate-reaching status (`seed_bounty_at_stage`: create for open, +claim for claimed, +submit for submitted); the twins then consume that status, which is why the rows cannot share a bounty. The hostile-count invariant recomputed from 66 to 69 and the multi-principal structural floor went from 3 to 6.

Load-bearing, by execution rather than inspection: disabling each handler's enforcement while keeping its `did_matches(` text makes the sweep fail with "must deny with 403, got 200 OK" naming exactly that row, for all three handlers. Under the submit mutation the `api/mod.rs` text guard stayed green while the sweep was red, which is precisely the #203 scenario; these gates no longer depend on that guard. Two more executed negatives: pointing the submit row at the wrong-status (open) bounty fails the sweep with "got 400", proving the status wiring is load-bearing and not decorative, and all three arm twins were observed reaching 200 on first execution. This covers the enforcement half of #203 for these three routes; the guard-hardening half of #203 stays open.

Full run: 488 unit plus 33 integration tests green, fmt and clippy clean.

jatmn (changes requested):

## Findings

- [P2] Keep every real-node probe on a bounded client
  `crates/gitlawb-node/tests/deny_harness.rs:145`
  The new protected-branch probe (and the new PR-diff probe at line 488) uses `reqwest::Client::new()` instead of the suite's bounded client, so a wedged router, Git subprocess, or stalled response makes `.send().await` wait indefinitely until the CI job kills it. That hides precisely the route failure this harness is meant to report; use the shared timeout-bearing client for these real-node requests.

- [P2] Actually withhold the pushed PR-diff object's ID
  `crates/gitlawb-node/tests/deny_harness.rs:541`
  `oids` is captured before `secret/x.txt` is created, so this lookup is always empty and the conditional never supplies an object ID to `assert_denied`. The test therefore checks the secret text but not the full or short SHA it promises to protect; a 404 path that exposes the withheld blob ID still passes. Have `build_branch_push_body` retain the new blob ID and assert that neither representation is present in the denial body.

- [P2] Do not describe the registry as covering every runtime denial while excluding signer-self gates
  `crates/gitlawb-node/tests/support/routes.rs:1`
  The PR describes a driven table of every route with a runtime deny, but this file deliberately omits signer-self 403 routes and the completeness checks only cover owner/read gates. For example, `tasks::create_task` rejects a signer who supplies somebody else's `delegator_did`; bypassing that `did_matches` result would let callers impersonate a delegator while the new sweep and completeness checks stay green. The same class exists in the task claim/complete/fail paths. Drive those gates in the real-node invariant, or narrow the registry and PR claims so they do not promise coverage they do not provide.

beardthelion:
All three fixed on `35577d3`, each guard proven load-bearing by reverting the exact production line it protects.

**Bounded client (F1).** Both new real-node probes (`signed_stranger_protected_branch_push_is_forbidden`, `get_pr_diff_withheld_path_is_denied`) now use a shared `bounded_client()` (30s timeout, the pattern already inline at the upload-pack and registry-sweep sites), so a wedged git subprocess fails the suite instead of hanging it until CI kills the job.

**Pushed OID withhold (F2).** You were right that the branch was dead: `oids` came from `seed_bare_repo`, which never seeds the pushed paths, so the lookup was always empty. `build_branch_push_body` now returns each pushed file's blob OID, and the PR-diff denial withholds the secret blob's full sha1 plus its `[..12]` short form (matching the `private_blob_oid_short` convention). Load-bearing: a 404 body carrying either OID form is rejected, and dropping the OID tokens lets the same body pass.

**Registry coverage (F3).** Drove the class rather than narrowing the claim. Every caller-self runtime deny now has a real-node hostile-then-authorized probe:

- `create_task` / `claim_task` / `register` as a new `SignerSelfGate` (a body DID field bound to the authenticated signer),
- `complete_task` / `fail_task` as a `MultiPrincipalGate` assignee arm (the caller must equal the task's stored assignee, driven against a seeded claimed task).

`register` is driven, not excused: the harness never initializes iCaptcha, so `verify_request` is inert and a signed stranger reaches the `did_matches` 403 exactly like `create_task`. A completeness guard derives the required set from a scan of `src/api` and requires every `did_matches(caller, X)` handler (X not the repo owner) to be a driven registry row; its non-vacuous floor is pinned to the exact caller-self count. The module doc is reconciled to the REST surface only, and states the scan's two limits (a check hidden behind a helper, and a gate in a new subdirectory module, same limits the owner/read sibling guards carry). The parallel GraphQL mutation caller-self gates stay fenced separately in #219, not claimed here.

Each of the five gates is mutation-verified: reverting its production `did_matches` line drives the sweep RED at that gate's hostile, and the completeness guard goes RED when a driven row is dropped or the classifier is starved. Harness green, no production change (`git diff src/` is empty).

jatmn (approved):
LGTM

beardthelion:
@kevincodex1 #195 is ready for merge. jatmn approved the current head, CI is green, and it's mergeable (clean, up to date with main). Yours when you have a moment.

kevincodex1 (approved):
LGTM

PR #198, opened 2026-07-14
fix(gl): point doctor's version check at Gitlawb/node, not the frozen Gitlawb/releases repo

### LINKED ISSUE #197: gl doctor checks the wrong/stale release repo for version updates
## Summary

`check_version` in `crates/gl/src/doctor.rs` queries `https://api.github.com/repos/gitlawb/releases/releases/latest` to check whether the installed `gl` CLI is up to date. That repo path is wrong on two counts: the org segment should be `Gitlawb` (capitalized, matching every other reference in this codebase), and more importantly `Gitlawb/releases` is a separate, now-frozen repo whose last tag is v0.3.8 from 2026-03-30.

Releases have been cut directly on `Gitlawb/node` since v0.4.0 (2026-06-30), via release-please, and `install.sh` already treats `Gitlawb/node` as the canonical release source (`REPO="${GITLAWB_RELEASE_REPO:-Gitlawb/node}"`, line 9). So `gl doctor`'s version check is silently comparing against a repo that will never receive another tag.

## Impact

Right now this happens to still report "up to date" by luck, since v0.3.8 sorts below the current v0.5.1. But once `Gitlawb/node` ships a version whose comparison against the frozen v0.3.8 tag stops mattering isn't the point: the check can never again correctly detect that an update is available, because it's asking the wrong repo. `gl doctor`'s version check is effectively permanently broken.

## Fix

Point `check_version` at `Gitlawb/node`'s releases endpoint instead, matching `install.sh`'s own source of truth.

I have a fix ready (with tests) and will open a PR referencing this issue.

euxaristia:
## Summary

`check_version` in `crates/gl/src/doctor.rs` queries `gitlawb/releases/releases/latest`. `Gitlawb/releases` is a separate repo that stopped receiving tags after v0.3.8 (2026-03-30). Releases have shipped directly on `Gitlawb/node` since v0.4.0, via release-please, and `install.sh` already treats `Gitlawb/node` as the canonical release source (`REPO="${GITLAWB_RELEASE_REPO:-Gitlawb/node}"`). The old query target meant the version check could never again detect a real update once past v0.3.8.

Closes #197

## What changed

- `check_version` now queries `Gitlawb/node`'s releases endpoint instead of `gitlawb/releases`.
- The GitHub API base is now a parameter (`github_api_base`), so the fix can be covered by a test against a mocked server instead of hitting the real GitHub API.
- Two new tests: one asserting a warning fires when a newer tag is returned, one asserting "up to date" when the tag matches current.

## How a reviewer can verify

```
cargo test -p gl doctor
cargo fmt --all -- --check
cargo clippy -p gl --all-targets -- -D warnings
```

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `gl doctor` version checks by retrieving release information from the correct project endpoint.
  * Added support for configurable API endpoints, improving reliability in environments that use alternate GitHub API URLs.

### DIFF
--- a/crates/gl/src/doctor.rs
+++ b/crates/gl/src/doctor.rs
@@ -475,4 +475,3 @@ mod tests {
         assert!(check.detail.contains("GitHub API returned HTTP 403"));
     }
 }
-
--- a/crates/gl/src/doctor.rs
+++ b/crates/gl/src/doctor.rs
@@ -343,6 +343,13 @@ async fn check_version(current: &'static str, github_api_base: &str) -> Check {
         Err(_) => return Check::pass("version", format!("v{current} (offline — could not check)")),
     };
 
+    if !resp.status().is_success() {
+        return Check::pass(
+            "version",
+            format!("v{current} (GitHub API returned HTTP {})", resp.status()),
+        );
+    }
+
     let body: serde_json::Value = match resp.json().await {
         Ok(v) => v,
         Err(_) => return Check::pass("version", format!("v{current} (invalid response)")),
@@ -451,5 +458,6 @@ mod tests {
 
         let check = check_version("0.1.0", &server.url()).await;
         assert!(matches!(check.state, CheckState::Ok));
+        assert!(check.detail.contains("up to date"));
     }
 }
--- a/crates/gl/src/doctor.rs
+++ b/crates/gl/src/doctor.rs
@@ -460,4 +460,19 @@ mod tests {
         assert!(matches!(check.state, CheckState::Ok));
         assert!(check.detail.contains("up to date"));
     }
+
+    #[tokio::test]
+    async fn test_check_version_http_error() {
+        let mut server = mockito::Server::new_async().await;
+        let _m = server
+            .mock("GET", "/repos/Gitlawb/node/releases/latest")
+            .with_status(403)
+            .create_async()
+            .await;
+
+        let check = check_version("0.1.0", &server.url()).await;
+        assert!(matches!(check.state, CheckState::Ok));
+        assert!(check.detail.contains("GitHub API returned HTTP 403"));
+    }
 }
+
--- a/crates/gl/src/doctor.rs
+++ b/crates/gl/src/doctor.rs
@@ -16,6 +16,7 @@ use std::path::PathBuf;
 use crate::http::NodeClient;
 
 const PUBLIC_NODE: &str = "https://node.gitlawb.com";
+const GITHUB_API_BASE: &str = "https://api.github.com";
 
 #[derive(Args)]
 pub struct DoctorArgs {
@@ -263,7 +264,7 @@ pub async fn run(args: DoctorArgs) -> Result<()> {
 
     // ── 7. Version up to date ─────────────────────────────────────────────
     let current = env!("CARGO_PKG_VERSION");
-    checks.push(check_version(current).await);
+    checks.push(check_version(current, GITHUB_API_BASE).await);
 
     // ── Render ────────────────────────────────────────────────────────────
     for check in &checks {
@@ -319,8 +320,9 @@ fn which_in_path(name: &str) -> bool {
         .unwrap_or(false)
 }
 
-/// Fetch the latest release tag from gitlawb/releases and compare to current version.
-async fn check_version(current: &'static str) -> Check {
+/// Fetch the latest release tag from Gitlawb/node (the actual release repo — see install.sh's
+/// `GITLAWB_RELEASE_REPO` default) and compare to current version.
+async fn check_version(current: &'static str, github_api_base: &str) -> Check {
     let client = match reqwest::Client::builder()
         .user_agent(format!("gl/{current}"))
         .timeout(std::time::Duration::from_secs(5))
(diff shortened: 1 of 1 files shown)


euxaristia:
Re: the needs-tests note — this PR does add tests, in the same commit and file. \`test_check_version_queries_gitlawb_node_releases\` and \`test_check_version_up_to_date\` are both in \`crates/gl/src/doctor.rs\`'s existing \`mod tests\` block, covering the warn and up-to-date branches of \`check_version\` against a mocked GitHub API.

The triage workflow's test-detection regex (\`.github/workflows/pr-triage.yml:73-75\`) only matches \`#[test]\` or \`#[cfg(test)\` literally. Both new tests use \`#[tokio::test]\`, since \`check_version\` is async, which doesn't match either alternative. That looks like a gap in the regex rather than anything about this PR: it would false-flag any PR whose only new tests are async (\`#[tokio::test]\`, \`#[async_std::test]\`, etc.), which given this codebase already uses \`#[tokio::test]\` throughout is probably worth a follow-up fix on its own.

euxaristia:
Pushed a fixup addressing both CodeRabbit comments:

- \`check_version\` now checks \`resp.status().is_success()\` before parsing the body, so a rate-limit (403) or other API error reports its actual HTTP status instead of falling through to the generic "could not parse latest tag".
- \`test_check_version_up_to_date\` now also asserts \`check.detail.contains("up to date")\`, so the test can't silently pass if the mock is never hit (mockito's unmatched-request fallback is a 501, which \`check_version\` also maps to \`Ok\`, so the state-only assertion wasn't proving the mock was exercised).

\`cargo test -p gl doctor\`, \`cargo fmt --check\`, and \`cargo clippy -D warnings\` all still clean.

beardthelion (dismissed):
Verified by execution: reverting just the production URL (leaving the mocks) turns both new tests red, so they're genuinely pinned to the fix, and the retarget to `Gitlawb/node` matches the `GITLAWB_RELEASE_REPO` default in `install.sh`/`install.ps1`. The core change is correct. One thing to add before merge.

## Findings

- [P2] Add a test for the new non-2xx branch
  `crates/gl/src/doctor.rs:346`
  The `if !resp.status().is_success()` guard is the branch the fixup was written for (the 403 rate-limit case), but it has no test. Disabling the whole block leaves all 9 tests green, so nothing currently drives the error response class, only the 200 path. This is the adversarial gap that matters on a diff whose second half exists specifically to handle non-2xx.

  A `with_status(403)` mock closes it (I ran this against the branch; it passes):

  ```rust
  #[tokio::test]
  async fn test_check_version_http_error() {
      let mut server = mockito::Server::new_async().await;
      let _m = server
          .mock("GET", "/repos/Gitlawb/node/releases/latest")
          .with_status(403)
          .create_async()
          .await;

      let check = check_version("0.1.0", &server.url()).await;
      assert!(matches!(check.state, CheckState::Ok));
      assert!(check.detail.contains("403"));
  }
  ```

  Assert both the `Ok` state and `contains("403")` — a state-only check would still pass if a regression dropped the status into the JSON-parse path instead.

The retarget itself is sound and the two added tests are load-bearing, so this is the only blocker. The `is_newer` numeric-only parse (pre-release / 4-part tags) is pre-existing and unchanged here, out of scope for this PR.

jatmn (changes requested):

## Findings

- [P2] Remove the extra blank line at EOF
  `crates/gl/src/doctor.rs:478`
  The added terminal blank line makes `cargo fmt --all -- --check` fail; the repository's `fmt + clippy` workflow runs that command as a required formatting gate. The current head therefore cannot pass CI until this line is removed.

euxaristia:
Pushed a fixup addressing jatmn's finding: removed the trailing blank line at EOF in doctor.rs. cargo fmt --all -- --check now passes clean, and all 10 doctor tests still pass.

jatmn (approved):
@beardthelion LGTM

beardthelion (approved):
Both blockers are closed. `test_check_version_http_error` is load-bearing: I removed the `if !resp.status().is_success()` block and it goes red (panic at the 403-detail assertion) while the other 9 doctor tests stay green, then restored to 10/10. That pins the exact non-2xx branch the fixup was written for. The EOF blank line is gone and `fmt + clippy` is green.

The rest still holds from my last pass: the two URL tests remain load-bearing (reverting the retarget turns both red), and the retarget itself matches the canonical release source (`install.sh` and `install.ps1` both default `GITLAWB_RELEASE_REPO` to `Gitlawb/node`). I traced the full response-class matrix on the current head, offline / 3xx / 4xx / 5xx / 2xx-non-JSON / 2xx-without-tag / 2xx-newer / 2xx-up-to-date, and every branch returns a truthful detail string with the status guard correctly ahead of the JSON parse. The injected base URL is test-only; the sole production caller passes the const. LGTM.

One non-blocking note for later, not this PR: now that the URL actually resolves, `is_newer`'s numeric-only parse (`splitn(3, '.')`) starts seeing real tags, so a pre-release or 4-part tag (`v1.2.3-rc1`, `v1.2.3.4`) would compare imperfectly. It degrades gracefully (no panic), it's pre-existing, and it's out of scope here, worth a follow-up if release tags ever take that shape.

@kevincodex1 approved and ready to merge.

kevincodex1 (approved):

kevincodex1:
nice catch thanks!

PR #200, opened 2026-07-14
fix(core): honor wildcard capabilities in Ucan::can, matching is_attenuated_by

### LINKED ISSUE #199: Ucan::can does not honor wildcard capabilities that is_attenuated_by grants during delegation
## Summary

`Capability::is_attenuated_by` (`crates/gitlawb-core/src/ucan.rs:52-57`) treats `with: "*"` and `can: "*"` / `can: "repo/admin"` on a parent capability as wildcards that cover any delegated child capability. So a validly-delegated UCAN chain can end with a capability entry like `{"with": "*", "can": "git/push"}`, meaning "push to anything," and `verify_chain` will accept it as a correctly-attenuated delegation.

But the runtime authorization check, `Ucan::can` (`ucan.rs:190-195`), does strict equality:

```rust
pub fn can(&self, resource: &str, action: &str) -> bool {
    self.payload
        .att
        .iter()
        .any(|cap| cap.with == resource && cap.can == action)
}
```

A capability of `{"with": "*", "can": "git/push"}` will never satisfy `cap.with == resource` for any concrete `resource` string, so `can()` returns `false` even though the delegation chain that produced this UCAN is valid and intentionally broad.

## Impact

This is a functional bug, not a security issue: the mismatch only produces false negatives (an agent holding a validly-delegated wildcard capability gets denied), never false positives. There's no path here to authorize something that shouldn't be authorized.

It's dormant today — grepped the tree, `Ucan::can` is only called from its own unit tests and a CLI test in `gl/src/ucan_cmd.rs`; no production authorization path consults it yet. `auth::caller_authorized_to_push` (`gitlawb-node/src/auth/mod.rs:28-30`) is pure `did_matches(...)`, with a comment noting UCAN-capability checking is future work. So this is latent until that Phase 2 wiring lands, at which point a correctly wildcard-delegated agent would be incorrectly denied.

## Fix

Make `can()` mirror `is_attenuated_by`'s wildcard semantics from the matching side (here, the stored capability plays the role `is_attenuated_by` calls "parent"):

```rust
pub fn can(&self, resource: &str, action: &str) -> bool {
    self.payload.att.iter().any(|cap| {
        let resource_ok = cap.with == resource || cap.with == "*";
        let action_ok = cap.can == action || cap.can == "*" || cap.can == caps::REPO_ADMIN;
        resource_ok && action_ok
    })
}
```

I have a fix ready (with tests) and will open a PR referencing this issue.

euxaristia:
## Summary

`Capability::is_attenuated_by` (`crates/gitlawb-core/src/ucan.rs:52-57`) treats `with: "*"` and `can: "*"` / `"repo/admin"` on a parent capability as wildcards during delegation-chain verification, so a valid UCAN can end with a capability like `{"with": "*", "can": "git/push"}`. But `Ucan::can` did strict equality on both fields, so that same capability could never match a concrete `(resource, action)` request at runtime.

Closes #199

## Why this is a bug fix, not a security fix

The mismatch only produces false negatives: a validly-delegated wildcard capability gets denied, never something unauthorized getting granted. It's also dormant today — `Ucan::can` isn't called from any production authorization path yet (`auth::caller_authorized_to_push` is pure `did_matches(...)`, with UCAN-capability checking noted as future Phase 2 work). This closes the gap before that wiring lands.

## What changed

- `Ucan::can` now mirrors `is_attenuated_by`'s wildcard semantics from the matching side: the stored capability plays the "parent" role, so `with == "*"` matches any resource and `can == "*"` or `"repo/admin"` matches any action.
- Two new tests: one for the resource wildcard, one for the `repo/admin` action wildcard.

## How a reviewer can verify

```
cargo test -p gitlawb-core ucan
cargo test -p gl ucan
cargo fmt --all -- --check
cargo clippy -p gitlawb-core --all-targets -- -D warnings
cargo clippy -p gl --all-targets -- -D warnings
```

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## Protocol & signing impact

No wire-format or signing change — `can()` is a local, in-process capability check over an already-verified UCAN's payload. Behavior only changes for capabilities that already contain a literal `"*"` or `"repo/admin"`, which `is_attenuated_by` already treats as wildcards during chain verification.

### DIFF
--- a/crates/gitlawb-core/src/ucan.rs
+++ b/crates/gitlawb-core/src/ucan.rs
@@ -358,6 +358,24 @@ mod tests {
     }
 
     #[test]
+    fn can_honors_action_wildcard() {
+        let issuer = Keypair::generate();
+        let audience = Keypair::generate().did();
+
+        let ucan = Ucan::issue(
+            &issuer,
+            audience,
+            vec![Capability::new("gitlawb://repos/test/repo", "*")],
+            None,
+        )
+        .unwrap();
+
+        assert!(ucan.can("gitlawb://repos/test/repo", caps::GIT_PUSH));
+        assert!(ucan.can("gitlawb://repos/test/repo", caps::PR_MERGE));
+        assert!(!ucan.can("gitlawb://repos/other/repo", caps::GIT_PUSH));
+    }
+
+    #[test]
     fn bootstrap_ucan() {
         let issuer = Keypair::generate();
         let audience = Keypair::generate().did();
--- a/crates/gitlawb-core/src/ucan.rs
+++ b/crates/gitlawb-core/src/ucan.rs
@@ -187,11 +187,17 @@ impl Ucan {
     }
 
     /// Check if this UCAN grants a specific capability on a resource.
+    ///
+    /// Mirrors `Capability::is_attenuated_by`'s wildcard semantics: a stored
+    /// capability of `with: "*"` or `can: "*"` / `"repo/admin"` covers any
+    /// requested resource/action, since a valid delegation chain can produce
+    /// exactly that capability (see `is_attenuated_by`).
     pub fn can(&self, resource: &str, action: &str) -> bool {
-        self.payload
-            .att
-            .iter()
-            .any(|cap| cap.with == resource && cap.can == action)
+        self.payload.att.iter().any(|cap| {
+            let resource_ok = cap.with == resource || cap.with == "*";
+            let action_ok = cap.can == action || cap.can == "*" || cap.can == caps::REPO_ADMIN;
+            resource_ok && action_ok
+        })
     }
 
     /// Encode to a compact JSON string (the wire format).
@@ -310,6 +316,45 @@ mod tests {
         assert_eq!(ucan.payload.aud, audience);
         assert!(ucan.can("gitlawb://repos/test/repo", caps::GIT_PUSH));
         assert!(!ucan.can("gitlawb://repos/test/repo", caps::PR_MERGE));
+    }
+
+    #[test]
+    fn can_honors_resource_wildcard() {
+        let issuer = Keypair::generate();
+        let audience = Keypair::generate().did();
+
+        let ucan = Ucan::issue(
+            &issuer,
+            audience,
+            vec![Capability::new("*", caps::GIT_PUSH)],
+            None,
+        )
+        .unwrap();
+
+        assert!(ucan.can("gitlawb://repos/test/repo", caps::GIT_PUSH));
+        assert!(ucan.can("gitlawb://repos/other/repo", caps::GIT_PUSH));
+        assert!(!ucan.can("gitlawb://repos/test/repo", caps::PR_MERGE));
+    }
+
+    #[test]
+    fn can_honors_repo_admin_action_wildcard() {
+        let issuer = Keypair::generate();
+        let audience = Keypair::generate().did();
+
+        let ucan = Ucan::issue(
+            &issuer,
+            audience,
+            vec![Capability::new(
+                "gitlawb://repos/test/repo",
+                caps::REPO_ADMIN,
+            )],
+            None,
+        )
+        .unwrap();
+
+        assert!(ucan.can("gitlawb://repos/test/repo", caps::GIT_PUSH));
+        assert!(ucan.can("gitlawb://repos/test/repo", caps::PR_MERGE));
+        assert!(!ucan.can("gitlawb://repos/other/repo", caps::GIT_PUSH));
     }
 
     #[test]


beardthelion (dismissed):
Verified the core is correct by execution: `can()` is an exact mirror of `is_attenuated_by` with the stored capability as the parent, so it grants exactly what the delegation chain already accepts. I enumerated the full cap x resource x action matrix (including `"*"`, `"repo/admin"`, and empty strings) and found no case where `can()` grants beyond `is_attenuated_by` — only false negatives removed, no new grant. Reverting `can()` to the old strict equality turns both new tests red, and there's no production caller yet, so this is a correct, dormant fix. One gap before merge.

## Findings

- [P2] Add a test for the new `cap.can == "*"` branch
  `crates/gitlawb-core/src/ucan.rs:198`
  Of the three match alternatives the fix introduces, only `with == "*"` and `repo/admin` are exercised. Dropping just `|| cap.can == "*"` leaves both `can_honors_*` tests green, so the full-action-wildcard branch has no coverage. On an authorization primitive every match branch should be load-bearing. One test closes it:

  ```rust
  let ucan = Ucan::issue(
      &issuer,
      audience,
      vec![Capability::new("gitlawb://repos/test/repo", "*")],
      None,
  )
  .unwrap();
  assert!(ucan.can("gitlawb://repos/test/repo", caps::GIT_PUSH));
  assert!(ucan.can("gitlawb://repos/test/repo", caps::PR_MERGE)); // two actions via can == "*"
  assert!(!ucan.can("gitlawb://repos/other/repo", caps::GIT_PUSH)); // wrong resource still denied
  ```

- [P3, pre-existing] `constraints` (`nb`) is unenforced. Both `can()` and `is_attenuated_by` ignore the field, so a capability like `{with, can: "repo/admin", nb: {...}}` grants broadly regardless of the constraint. Not introduced here, but this PR widens where it matters once `can()` gets wired into the authorization path — worth settling constraint enforcement before that Phase 2 wiring lands. Note, not a blocker.

- [P3, pre-existing] `repo/admin` is treated as a universal action wildcard (equal to `"*"`), so it grants `network/join`, `agent/deploy`, etc. on the resource, not just repo actions. This lives in `is_attenuated_by`; `can()` faithfully inherits it. Worth confirming that breadth is intended. Design note.

The retarget of the matching semantics to mirror `is_attenuated_by` is sound and correctly scoped. The only blocker is the untested `can == "*"` branch.

jatmn (changes requested):

## Findings

- [P2] Add load-bearing coverage for the `can == "*"` wildcard branch
  `crates/gitlawb-core/src/ucan.rs:198`
  The implementation adds three wildcard alternatives, but the new tests exercise only `with == "*"` and `can == "repo/admin"`. Removing just `|| cap.can == "*"` leaves every added test green, so a regression would silently reintroduce false negatives for valid full-action-wildcard UCANs once this method is used for authorization. Add a test with a concrete resource and `can: "*"` that proves multiple actions match while the same action remains denied for a different resource.

euxaristia:
Pushed a fixup addressing beardthelion's and jatmn's shared finding: added can_honors_action_wildcard, a test with a concrete resource and can: "*" that asserts multiple actions match while the same action stays denied for a different resource. All 18 ucan tests pass.

jatmn (approved):
@beardthelion LGTM

beardthelion (approved):
The blocker from my last pass is closed. `can_honors_action_wildcard` is load-bearing: I dropped `|| cap.can == "*"` from `can()` and it goes red (panic at the `pr/merge` assertion) while the other 17 ucan tests stay green, then restored and confirmed 18/18. That pins the exact branch jatmn and I flagged. The core is unchanged from my prior review and still correct: `can()` is a term-for-term mirror of `is_attenuated_by` with the stored capability in the parent slot, so it only removes false negatives and never grants past a real delegation. I re-checked the requested-side wildcard case too (`can("*", ...)` against a concrete stored cap denies, matching `is_attenuated_by`'s "wildcards on self carry no meaning").

Confirmed dormant on this head: the only production authz path is `verify_chain()` at `crates/gitlawb-node/src/auth/mod.rs:301`; every `.can()` caller in the tree is test-only. LGTM.

One forward-looking note, not blocking this PR: `can()` trusts `payload.att` raw, with no signature/expiry/chain check of its own. That's the right split for a capability-matching primitive, but the wildcard change widens the blast radius if a future caller forgets to validate first (one `("*","repo/admin")` cap now answers `can(any, any)` = true). When Phase 2 wires `can()` into the auth path, the caller must gate on `verify_chain` before calling it. Worth a one-line precondition on the doc comment while it's cheap; fine to fold into the Phase 2 change.

@kevincodex1 approved and ready to merge.

kevincodex1 (approved):

PR #202, opened 2026-07-14
fix(ci): recognize namespaced test attributes in pr-triage's needs-tests check

### LINKED ISSUE #201: PR triage's test-detection regex misses namespaced test attributes (#[tokio::test], etc.)
## Summary

The triage workflow's `needs-tests` check (`.github/workflows/pr-triage.yml:73-75`) detects an inline test addition with:

```js
/^\+.*#\[(test\]|cfg\(test\))/m
```

This only matches a literal `#[test]` or `#[cfg(test)` immediately after `#[`. It does not match namespaced test attributes like `#[tokio::test]` or `#[async_std::test]`, since the characters between `#[` and `test]`/`test)` don't fit either alternative.

## Impact

Any PR whose only new tests are async (`#[tokio::test]`, which this codebase already uses throughout, e.g. `crates/gl/src/agent.rs`) gets falsely labeled `needs-tests` and receives the "no tests changed" guidance comment, even when tests were in fact added in the same commit.

Reproduced against PR #198: it added two `#[tokio::test]`-only tests to `crates/gl/src/doctor.rs` in the same commit as the source change, and the triage bot still commented `needs-tests`.

## Fix

Widen the regex to also match any attribute ending in `::test]`, covering namespaced test macros regardless of depth:

```js
/^\+.*(#\[test\]|#\[cfg\(test\)|::test\])/m
```

Verified against representative patches (including PR #198's actual diff) — the new regex correctly detects `#[test]`, `#[cfg(test)]`, `#[tokio::test]`, and `#[async_std::test]`, while still correctly rejecting a non-test Rust change.

I have a fix ready and will open a PR referencing this issue.

euxaristia:
## Summary

The `needs-tests` check in `.github/workflows/pr-triage.yml` uses:

```js
/^\+.*#\[(test\]|cfg\(test\))/m
```

to detect whether a PR added an inline test. This only matches a literal `#[test]` or `#[cfg(test)` immediately after `#[`, so it misses namespaced test attributes like `#[tokio::test]` or `#[async_std::test]`. Any PR whose only new tests are async gets falsely labeled `needs-tests`, even when tests were added in the same commit.

Closes #201

## Reproduction

PR #198 added two `#[tokio::test]`-only tests to `crates/gl/src/doctor.rs` in the same commit as the source fix, and the triage bot still commented `needs-tests`.

## Fix

```js
/^\+.*(#\[test\]|#\[cfg\(test\)|::test\])/m
```

Matches any attribute ending in `::test]`, covering namespaced test macros at any depth, while still matching the original `#[test]` / `#[cfg(test)]` forms.

## How a reviewer can verify

I don't see any existing test harness for the inline `github-script` workflows in this repo, so I verified the regex directly against representative diff patches (a standalone Node script, not committed — just the verification, shown here):

```js
const oldRe = /^\+.*#\[(test\]|cfg\(test\))/m;
const newRe = /^\+.*(#\[test\]|#\[cfg\(test\)|::test\])/m;

// old=true, new=true  — "+    #[test]\n+    fn foo() {}"
// old=true, new=true  — "+#[cfg(test)]\n+mod tests {"
// old=false, new=true — "+    #[tokio::test]\n+    async fn foo() {}"   <- the bug
// old=false, new=true — "+    #[async_std::test]\n+    async fn foo() {}"
// old=false, new=false — "+    let x = 1;\n+    println!(\"{x}\");"    <- correctly still rejected
// old=false, new=true — PR #198's actual patch excerpt
```

All six cases behave as expected under the new regex, including the real PR #198 diff that triggered this.

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [x] Tests / CI

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pull request triage detection for newly added Rust inline test gates.
  * More accurately recognizes `#[cfg(test)]` and async test attribute forms such as `#[tokio::test(...)]` and `#[async_std::test]`.
  * Reduced false positives by analyzing added lines with context-aware matching and ignoring occurrences inside comments and string/char literals.


beardthelion (dismissed):
Verified by running both the old and new regex over the full `addsInlineTest` predicate (filename guard, `f.patch` guard, `^\+` anchor, `.some()`, and the alternation), 20+ inputs each way. The direction is right and the dominant `#[tokio::test]` form now resolves with zero regressions. Two things to fold in before merge.

## Findings

- [P2] Cover the parametrized `#[tokio::test(flavor = ...)]` form
  `.github/workflows/pr-triage.yml:78`
  The `::test]` alternative needs `]` immediately after `test`, but the args form is `::test(...)`, so `test]` never appears and it's missed. This is the PR's own target class and it's live in the tree at `crates/gitlawb-node/src/git/smart_http.rs:753` and `:818`. A PR whose only new tests use the flavor form still gets falsely labeled `needs-tests`. The bare form (248 hits) is fixed, so this is a partial fix rather than a wrong one.

- [P3] Anchor the match to an attribute so `::test]` stops over-matching non-test code
  `.github/workflows/pr-triage.yml:78`
  `::test]` is unanchored, so it also fires on ordinary added lines like `let r = registry[Region::test];` and on the macro appearing in a comment or string literal. Effect is bounded (it only suppresses the advisory label, which isn't a merge gate), but it's a new false-positive class the old regex didn't have.

Both close with a single regex. I ran this against the same matrix: 0 regressions, 0 false-positives, and it catches the flavor form:

```
/^\+\s*#\[\s*(cfg\(\s*test|[\w-]+(::[\w-]+)*::test|test)\b/m
```

Anchoring to a leading `#[` removes the over-match, and `::test\b` (instead of `::test]`) also clears `#[tokio::test(...)]`.

Net: strict improvement as it stands, and the above makes it complete against the title. Non-gating label, so nothing here is severe, but the flavor form defeats the stated purpose on code that already exists in the repo.

jatmn (changes requested):

## Findings

- [P3] Keep the detector on an added diff line
  `.github/workflows/pr-triage.yml:80`
  `\s*` after the diff `+` also consumes newlines. As a result, a Rust PR can add only a blank line immediately before an existing context-line `#[tokio::test]`; the expression then treats that unchanged test as newly added and suppresses `needs-tests`. Use horizontal whitespace (for example, `[ \t]*`) at this position so the attribute itself must be on an added line.

- [P3] Complete the `cfg(test)` predicate before accepting it as a test
  `.github/workflows/pr-triage.yml:80`
  The widened `cfg\(\s*test\b` branch also matches valid non-test predicates such as `#[cfg(test = "triage_only")]`. Adding only that attribute to an otherwise testless Rust PR makes `addsInlineTest` true and bypasses the advisory test signal; the former expression did not do so. Require the predicate to close (for example, `cfg\(\s*test\s*\)`) before classifying it as a test gate.

- [P3] Handle valid whitespace in the namespaced attribute syntax
  `.github/workflows/pr-triage.yml:80`
  Rust accepts token whitespace in attributes, including `#[tokio :: test]` and `#[cfg (test)]`, but the new matcher requires both `::` and `cfg(` to be adjacent. A PR whose only added test uses either valid form is still labeled `needs-tests`, so the claimed namespaced/cfg recognition remains incomplete. Allow horizontal token whitespace around these separators while keeping the match confined to the added line.

- [P3] Do not treat raw-string contents as executable test attributes
  `.github/workflows/pr-triage.yml:80`
  Anchoring at `#[` prevents ordinary comment and string-expression lines from matching, but it does not distinguish a multiline raw string: an added fixture containing a line `#[tokio::test]` makes this expression true despite adding no runnable test. This is a new false positive for namespaced attributes and lets a testless Rust change suppress `needs-tests`; account for Rust string state (or otherwise avoid presenting the leading-attribute check as excluding strings).

euxaristia:
Pushed a fixup addressing jatmn's four findings and this morning's CodeRabbit finding:

- The regex now requires only horizontal whitespace before `#[`, so a blank added line before an unchanged context-line test no longer counts as adding a test.
- `cfg(test...)` now has to close before it's accepted as a test gate, so `#[cfg(test = "triage_only")]` no longer counts.
- Whitespace around `::` and inside `cfg(` is now allowed, matching valid Rust like `#[tokio :: test]` and `#[cfg (test)]`.
- Added a small Rust lexer (line comments, block comments, char literals, normal and raw strings) so attribute-looking text inside string/comment content is no longer mistaken for a real test.

For the lexer's cross-hunk state gap CodeRabbit flagged: instead of scanning the diff patch text directly, the workflow now fetches the head-version file content via the contents API and lexes it in file order, then checks only the line numbers the diff actually added (computed from the hunk headers). This means lexical state (open raw string, block comment, etc.) is always accurate regardless of where a hunk's context window starts.

Verified against 12 cases covering all of the above, including the specific scenario CodeRabbit described (an added line landing inside a raw string opened outside the hunk's visible context).

jatmn (changes requested):

## Findings

- [P2] Preserve nesting while skipping block comments
  `.github/workflows/pr-triage.yml:89`
  Rust block comments nest, but this lexer represents comment state as a boolean and clears it at the first `*/`. A source-only PR can add `#[tokio::test]` after an inner comment closes but before the enclosing comment closes (`/* outer`, `/* inner */`, `#[tokio::test]`, `*/`); Rust still treats the attribute as comment text, while this code treats it as live and suppresses `needs-tests`. Track block-comment depth (and cover nested comments) before accepting an attribute.

- [P2] Bound the per-file Contents API scan and do not silently classify failures as no test
  `.github/workflows/pr-triage.yml:170`
  Every changed Rust file with a patch now causes a serial `repos.getContent` call, while every failure is silently skipped. A large PR can therefore exhaust the workflow's ten-minute/API budget, and rate-limit or transient failures on the only file containing a new test make the triage job incorrectly add `needs-tests` (or prevent label reconciliation altogether). The previous implementation only examined the already-fetched file patches. Put a bounded strategy/fallback around this scan and surface a failure rather than treating an unavailable file as testless.

- [P3] Require `test` to be the terminal attribute segment
  `.github/workflows/pr-triage.yml:80`
  The `test\b` alternatives stop at a word boundary rather than a valid attribute terminator, so non-test attributes such as `#[foo::test::helper]`, `#[test::helper]`, and `#[foo::test = "x"]` match. Adding one of those to a Rust-only PR suppresses the advisory test signal despite adding no test. Require optional horizontal space followed by `]` or `(` after the bare/namespaced `test` segment.

- [P3] Recognize valid multiline test attributes
  `.github/workflows/pr-triage.yml:80`
  Rust permits whitespace, including newlines, between attribute tokens. A valid added test formatted as `#\n[test]` or `#[tokio\n::test]` is scanned one line at a time and neither line matches this horizontal-whitespace-only expression, so the PR still receives `needs-tests`. Carry enough attribute-token state across added lines (or parse the added attribute span) to cover the valid forms the workflow is intended to recognize.

jatmn (changes requested):

## Findings

- [P2] Do not abort triage when the bounded scan is exhausted
  `.github/workflows/pr-triage.yml:199`
  The cap is reached using the broad `test`/`cfg` substring prefilter rather than a real attribute match. A normal Rust PR that edits 16 files with (for example) `#[cfg(feature = ...)]` or even comments containing those words reaches the `throw` on the sixteenth file. That exits before the label reconciliation below, so every managed label (including stale labels that should be removed) and the guidance comment are left untouched. Keep the API work bounded, but use a non-fatal conservative fallback so an otherwise valid multi-file PR is still triaged.

- [P2] Handle Contents API responses that do not carry Base64 text
  `.github/workflows/pr-triage.yml:206`
  `getContent` is decoded unconditionally, but GitHub returns an empty `content` with `encoding: "none"` for the 1–100 MB range unless the request uses the raw representation. An added `#[tokio::test]` in such a Rust file therefore gets lexed as an empty file and is incorrectly labeled `needs-tests` (and an absent `content` response would instead fail the job). Check the response encoding/size and fetch a supported raw/blob representation or take an explicit conservative fallback.

- [P3] Require an added attribute token, not merely added whitespace inside one
  `.github/workflows/pr-triage.yml:226`
  The expression permits newlines between `#` and `[`, then marks a match as new when *any* character in its span is on an added line. Thus a source-only PR can insert just a blank line into an existing multiline attribute (`#`, added blank line, `[test]`) and make `addsInlineTest` true without adding a test. Rust accepts that formatting, so this suppresses `needs-tests` for the rest of an otherwise testless change. Tie the match to an added attribute token/span rather than any whitespace it crosses.

- [P3] Preserve token boundaries while stripping literals
  `.github/workflows/pr-triage.yml:136`
  Removing a literal without a placeholder can join unrelated Rust tokens into a synthetic test attribute. For example, `consume!(# "not an attribute" [test]);` is valid input to a `macro_rules! consume { ($($tt:tt)*) => {}; }` matcher, but the stripper produces `#[test]`; the regex then treats the added invocation as a test. Preserve a separator for skipped comments/literals (or match lexed attribute tokens) so testless macro changes cannot bypass the advisory label.

- [P3] Recognize absolute namespaced test attributes
  `.github/workflows/pr-triage.yml:80`
  Rust accepts an attribute path beginning with `::`, but the namespaced branch requires an identifier before its first `::`. Consequently `#[::tokio::test]` is not detected even though it is a valid async-test attribute, and a PR that adds only that form still receives `needs-tests`. Allow the optional leading path separator when matching a namespaced `test` attribute.

euxaristia:
Pushed a fixup addressing this round of jatmn's findings:

- The bounded Contents API scan no longer aborts the whole job when the fetch cap is hit; it now logs a warning and stops scanning further candidates, so label reconciliation and the guidance comment still run for the rest of the PR.
- File content now comes from the Git blob API (using the blob sha already in the diff) instead of the Contents API, avoiding the `encoding: "none"` gap that API has for files between 1-100MB. A non-base64 response is logged and skipped rather than silently treated as empty.
- The added-line check now requires an actual non-whitespace attribute character to be on an added line, not just any whitespace the match happens to span, so an added blank line inside an otherwise-unchanged multiline attribute no longer counts as adding a test.
- Stripped comments/strings/char literals now leave a placeholder character behind, so a macro argument list containing a stray `#` and a string literal can't have its stripped literal glue the surrounding tokens into a false `#[test]` match.
- The namespaced-test regex now accepts an optional leading `::`, so `#[::tokio::test]` is recognized.

Verified with a 22-case regression harness run directly against the functions extracted from this file (all prior cases plus one per new finding), plus a syntax check of the extracted script.

jatmn (changes requested):

## Findings

- [P2] Do not classify a truncated scan as evidence that no test was added
  `.github/workflows/pr-triage.yml:215`
  After fifteen broad `test`/`cfg` candidates, this breaks and falls through with `addsInlineTest` still false. A Rust PR can have fifteen earlier files containing incidental `test`/`cfg` text and add its only `#[tokio::test]` in the sixteenth; the workflow then adds `needs-tests` and posts misleading guidance despite the test. Preserve the API bound, but make an exhausted scan inconclusive (or scan the remaining actual attribute candidates) rather than asserting the negative signal.

- [P2] Bound the blob payloads before decoding and lexing them
  `.github/workflows/pr-triage.yml:228`
  Each candidate causes the workflow to download, Base64-decode, split, and build a line map for the entire blob, with no size check. A fork PR can touch a small hunk in up to fifteen very large Rust files whose patches contain `test` or `cfg`, consuming the ten-minute job's bandwidth and memory and leaving labels/comments unreconciled when it times out or is killed. Check the blob size before decoding and use a bounded conservative fallback for oversized files.

- [P3] Do not count the no-newline patch marker as a new-file line
  `.github/workflows/pr-triage.yml:186`
  `\\ No newline at end of file` reaches this context-line branch and increments `newLine`. If a no-newline final line is replaced by `#[test]`, the addition is recorded as line 2 while the head file maps it to line 1, so the valid test is missed and `needs-tests` is applied. Ignore the marker rather than advancing the new-file cursor.

- [P3] Recognize raw-identifier spellings of the test attribute
  `.github/workflows/pr-triage.yml:81`
  Rust accepts `#[r#test]` (and raw identifier path segments such as `#[tokio::r#test]`) as the test attribute, but this expression only accepts the literal `test`. A Rust PR whose sole new inline test uses that valid spelling is therefore falsely labeled `needs-tests`; allow the raw-identifier form when matching the terminal test segment.

euxaristia:
Addressed the latest review round in d18d22a:

- An exhausted 15-file scan now sets the result as inconclusive instead of asserting no test was found, so `needs-tests` is no longer applied off unverified evidence.
- Blob content is checked against a 5MB size limit before decoding, and oversized or non-base64 blobs also mark the scan inconclusive rather than counting as no-test.
- `addedLineNumbers` now skips the `\ No newline at end of file` patch marker instead of counting it as an added context line.
- The test-attribute regex now accepts the raw-identifier spelling (`#[r#test]`, `#[tokio::r#test]`).

jatmn (changes requested):

## Findings

- [P2] Do not treat a missing file patch as a negative test scan
  `.github/workflows/pr-triage.yml:208`
  `pulls.listFiles` can omit `patch` for a large or truncated diff, but this guard skips that changed Rust file without setting `scanIncomplete`. If that file contains the PR's only added `#[tokio::test]`, the loop leaves both `addsInlineTest` and `scanIncomplete` false, and the workflow applies the erroneous `needs-tests` label/comment. This is another unavailable-evidence case, so mark it inconclusive (or obtain authoritative added-line coordinates) before asserting that no inline test was added.

- [P2] Enforce the blob limit before downloading blob content
  `.github/workflows/pr-triage.yml:236`
  The size check occurs only after `git.getBlob` has returned its Base64 payload. A fork can make up to fifteen candidate Rust files (the broad prefilter needs only added `test`/`cfg` text) refer to blobs far above 5 MB, so the privileged triage job downloads all of those large payloads before it can take the conservative fallback. That can exhaust the ten-minute job and leave label reconciliation/commenting untouched—the availability failure this bound is meant to prevent. Determine blob size from metadata before retrieving content, or use a bounded content-fetch path.

- [P3] Preserve token-separating comments when matching attributes
  `.github/workflows/pr-triage.yml:82`
  Rust accepts comments where whitespace separates attribute tokens, e.g. `#[cfg /* reason */ (test)]` or `#[tokio /* reason */ :: test]`. The stripper turns each block comment into `\0`, while this expression permits only whitespace, so those valid test gates are missed and a PR whose only tests use that formatting receives `needs-tests`. Keep a token separator that the matcher can accept in these positions (without reintroducing the literal-gluing false positive), or match the lexed tokens directly.

beardthelion (dismissed):
The lexer works. I had it driven through nested block comments, `r#"..."#` and `r##"..."##`, escaped quotes, char literals, the `'a` lifetime trap that catches most hand-rolled Rust lexers, CRLF, and an unterminated string: `stripped` and `lineMap` stay in parity on every case, there is no zero-length match, no `lastIndex` leak across files, and no catastrophic backtracking (I ran `#[` plus 24 chained `a::` segments and whitespace-amplified variants, all sub-millisecond). On the four false positives that motivated it, an attribute inside a line comment, a string, a block comment opened earlier, and a raw string opened earlier, it rejects all four where the old regex accepted them. As an implementation this is good work.

I am asking for it to be removed anyway, and that call is mine rather than something for you to litigate with the reviewers.

## Findings

- [P2] Cut back to an anchored regex and drop the file fetching and lexing

`.github/workflows/pr-triage.yml:82`

`testAttrRe` starts `#[ \s]*\[` with no line-start anchor, so it matches inside comments and strings, and the lexer exists to strip those first. Anchor the attribute to the start of an added line and that whole class stops being reachable, no fetching and no lexing required. I ran three variants over one corpus of 18 cases (10 real attribute forms including the parametrized flavor form, `::tokio::test` and `tokio::r#test`; 3 decoy attributes taken from this repo; 5 false-positive lines): current main is wrong on 8, this PR's regex applied to the added line is wrong on 4, and `/^\+[ \t]*#\[[ \t]*(?:cfg[ \t]*\([ \t]*test[ \t]*\)|(?:::)?(?:[\w-]+[ \t]*::[ \t]*)*(?:r#)?test\b)/m` is wrong on 0.

The one thing anchoring gives up is an attribute sitting inside a block comment opened on an earlier line. That is worth giving up here: `needs-tests` applies a label and posts one guidance comment, with no `setFailed` anywhere in this workflow and no required status checks on the main ruleset, so nothing gates on being right.

- [P2] The anchoring was reversed for a form this codebase does not contain

`.github/workflows/pr-triage.yml:75`

Round 2 had horizontal-only anchoring and round 4 reversed it to support a multiline attribute (`#` and `[` split across lines), which is what made the added-token check and the head-file lex necessary. That form does not appear anywhere: `git grep -nE '^\s*#\s*$' -- '*.rs'` returns nothing across 112 files carrying 1085 test attributes. The comment on line 75 still describes the horizontal-whitespace anchoring that was removed, so the file currently advertises the mechanism I am asking you to restore.

- [P3] Extract it or inline it, but do not keep a lexer in a YAML block scalar with no harness

Nothing in CI executes `stripRustFile` or `addedLineNumbers`, so each round's behavioral properties get re-litigated by hand and the previous round's fix is free to regress. Two of the current findings are exactly that: the NUL placeholder added in round 4 broke comment-separated attributes, and the size bound added in round 5 sits on the wrong side of the fetch. If this does stay, it belongs in `scripts/` with a runner in `pr-checks.yml`, the way `scripts/test-resolve-release-tag.sh` is wired in #249.

### If you keep the current shape instead

I would rather not ask for fixes to code I am asking you to delete, so these are conditional on the above being declined:

`throw new Error(...)` on a non-404 blob fetch (line 252) aborts the whole step, which skips label reconciliation and the guidance comment for every managed label. That directly contradicts the tradeoff stated at lines 219-222 for the fetch cap. The 404 branch (line 249) is the only "could not check this file" path that does not set `scanIncomplete`, and its "file since deleted" rationale is unreachable because a deletion-only patch never passes the pre-filter. `addedLineNumbers` (line 188) treats a content line starting `--` or `++` as context and shifts subsequent line numbers.

Checked and not raised, so nobody re-derives them: no ReDoS, and the `pull_request_target` posture holds because the fetched fork bytes are only ever lexed in process, never evaluated or interpolated, and `f.sha` comes from this PR's own file list scoped to the base repo. I also confirmed base-repo `getBlob` does resolve a fork head blob, so fork PRs are not silently mislabeled.

One process note: jatmn's last round landed on this same commit and nothing has been pushed since, so those three findings are still open independently of this review. The scope growth here was driven by review feedback rather than by you reaching for it, and unwinding it is my call to make, not a mark against the work.

euxaristia:
Addressed beardthelion's review in 85ed112:

- Dropped the head-file fetch, blob size bound, and in-workflow Rust lexer.
- Restored horizontal-whitespace anchoring on added lines only.
- Needs-tests detection is now a single anchored regex covering bare `#[test]`, `#[cfg(test)]`, namespaced forms including `#[tokio::test(...)]`, optional leading `::`, and raw-identifier spellings (`r#test`).

Verified the regex against the common positive/negative cases (flavor form, `//` comments, strings, `Region::test` decoys).

@coderabbitai full review

beardthelion (approved):
Verified on 85ed112. The lexer and head-file fetch are gone, and the detector is the anchored regex I asked for: added lines only, horizontal whitespace, bare / cfg(test) / namespaced / leading :: / r#test, including the flavor form. I ran a 21-case matrix both ways (0 wrong), plus must-nots for mid-line comments, strings, Region::test, #[testing], and deleted/context lines. Gutting the namespaced arm makes #[tokio::test] and #[tokio::test(flavor = ...)] miss while bare #[test] still hits. needs-tests stays advisory (no setFailed). Prior round asks are closed; jatmn/CodeRabbit lexer-era threads are moot with the cut-back.

One process note, not a finding: cargo audit is red on this head for RUSTSEC-2026-0220 (ruint 1.17.2), same lock as main. That is #292, not this workflow change. Also expect a rebase conflict with #277 on the same pr-triage.yml lines if that lands first.

jatmn (approved):
LGTM

beardthelion:
@kevincodex1 this is ready to merge now. The earlier approve still stands from jatmn and me; the only hold was cargo audit on ruint, and #292 cleared that. Rebased onto main, CI green on 13f6948b.

kevincodex1 (approved):

Issue #94, opened 2026-06-24
Unauthenticated GET /repos/:owner/:repo/hooks leaks webhook target URLs for any repo


beardthelion:
`list_webhooks` (`crates/gitlawb-node/src/api/webhooks.rs:73`) is mounted on the `optional_signature` read-routes group (`crates/gitlawb-node/src/server.rs:344-365`), takes no `AuthenticatedDid`, and performs no ownership or visibility check. Its siblings `create_webhook` and `delete_webhook` both call `require_repo_owner`; `list_webhooks` does not. `get_repo` applies no `is_public` filter, so private repos resolve the same as public ones.

Any unauthenticated caller who knows a repo's `owner/name` can therefore enumerate every webhook registered on it. Secrets are redacted to `***`, but the full target `url`, the `created_by_did`, and the event subscriptions are returned. Webhook URLs routinely embed internal-infra hostnames, third-party integration endpoints, and tokens or signed params in the path or query. The deliberate secret redaction shows these records are treated as sensitive; the directly actionable URL field is left open.

### Where
- `crates/gitlawb-node/src/server.rs:344-365` mounts the route under `optional_signature`
- `crates/gitlawb-node/src/api/webhooks.rs:73-97` handler has no auth param and no `require_repo_owner`
- `crates/gitlawb-node/src/api/webhooks.rs:88-90` redacts `secret` only
- `crates/gitlawb-node/src/db/mod.rs` `get_repo` has no `is_public` predicate

### Fix
Move `list_webhooks` into an authenticated group and add `require_repo_owner` (mirroring `create_webhook` / `delete_webhook`), or gate it on the repo's authorization model if collaborator read access is intended. Keep the existing secret redaction.

### Related
The same `optional_signature` read group also carries `list_protected_branches`, `list_replicas`, and `list_repo_events`. If any are meant to be visibility-gated for private repos, they share this missing-gate pattern and are worth tracing together, since `get_repo` applies no visibility filter.

PR #209, opened 2026-07-16
docs(agents): add AGENTS.md carrying the authorization rules PRs most often break

beardthelion:
Adds a root `AGENTS.md` written for coding agents and human contributors alike, plus a pointer from CONTRIBUTING's PR guidelines. It carries only what a patch author cannot recover from the code alone: the owner-gate forms in use and the write routes that are non-owner by design, the path argument contract on `authorize_repo_read`, denial semantics, the deny-test expectations for new and removed routes, migration versioning, client denial surfacing, and what CI actually blocks on vs what is convention.

Two ground rules held while writing it: every named identifier, test, and CI claim was checked against current main, so the enforcement labels describe only what is merged today; and nothing already covered by README, CONTRIBUTING, SECURITY, or the PR template is repeated. The deny suite in review (#194, #195) is referenced with wording that stays true whether or not it has landed in a given checkout.

Docs-only: no runtime impact, no code or CI changes.

## Summary by CodeRabbit

* **Documentation**
  * Added contributor guidance covering authorization rules, denial-behavior testing, signature and certificate verification, database migration practices, and client handling of denied access.
  * Updated contribution guidelines to require reviewing authorization guidance before modifying endpoints, visibility, or migration workflows.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,38 @@
+# Contributor guide for coding agents
+
+This file carries the rules a patch author cannot recover from the code alone, for humans and coding agents alike. For build, test, and PR basics, read `CONTRIBUTING.md` first.
+
+## Authorization invariants (the rules PRs most often break)
+
+Every repo-scoped endpoint must bind the caller to an authorization decision before serving or mutating anything.
+
+- Owner-only mutations (visibility, webhooks, protected branches, merges) are gated against the repo owner. Three gate forms are in use and all are recognized: webhooks and merges call `require_repo_owner` (`crates/gitlawb-node/src/api/mod.rs`), visibility uses its module-local `require_owner`, and protected branches compare inline with `did_matches`. Use the form the surrounding module already uses.
+- Not every write is owner-only, and that is by design: `star_repo`/`unstar_repo` bind to the signer (they still require repo read access), `register_replica`/`unregister_replica` bind to the replica's own DID, the bounty actions (claim, submit, approve, cancel, dispute) have their own multi-party rules, and closing a PR or issue is owner-or-author. Do not add owner gates to these.
+- Content-serving reads (blobs, trees, raw files) gate through `authorize_repo_read` with the specific path being served, so a withheld subtree is denied even on an otherwise-public repo; repo-level reads (listings) pass `"/"` by the helper's own contract. The source scans below cannot check this argument's value, so review will.
+- A route that reaches repo data through a global id, with no repo in its path, or through an extractor shape the scans do not recognize still needs the same gate; the scans cannot see those.
+- Read-surface denials must not reveal existence: a caller who may not read a repo (or a withheld subtree) gets the same 404 as a missing repo, never a 403. (Owner-gated mutations currently return 403 after the repo lookup; do not widen that shape to read paths.)
+- Two tests in module `authz_guard` (`crates/gitlawb-node/src/api/mod.rs`) scan handler sources and fail on a repo-scoped handler without a recognized gate: `every_in_scope_mutation_has_its_gate` and `every_repo_scoped_handler_is_gated`. If they fail on your handler, add the gate. The second test carries a `known_ungated` allowlist of tracked gaps being closed; never add an entry to it.
+
+## Adding or changing routes
+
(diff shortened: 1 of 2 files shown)


jatmn (commented):

## Findings

- [P1] Make the denial-test instruction match each route's actual authorization policy
  `AGENTS.md:18`
  This tells every new gated handler to reject a signed non-owner, but several routes documented immediately above intentionally authorize non-owners: any reader may star a readable repo, replica registration is bound to the replica DID, and bounty/close actions have their own role policies. A contributor following this mandatory rule will either assert denial for a valid request or add an incorrect owner gate to make the test pass, while potentially missing the identity that is actually unauthorized. Require denial cases based on the handler's specific authorization rule (retaining non-owner denial for owner-only routes), alongside the applicable anonymous/no-leak assertions.

- [P2] Do not describe the repo-scoped read guard as covering handlers it explicitly exempts
  `AGENTS.md:7`
  The new invariant says every repo-scoped endpoint binds a caller before serving data, but `authz_guard::every_repo_scoped_handler_is_gated` currently exempts `list_webhooks`, `list_replicas`, and `list_protected_branches` in `known_ungated`; each handler only looks up the repo and returns its metadata. The following description of the guard does not name those live exceptions, so this security guide presents the existing private-repo metadata exposure as protected. Either land the gates, or call out the exact tracked exceptions and keep contributors from treating those surfaces as safe.

- [P2] Correct the statement that the listed CI jobs block merges
  `AGENTS.md:37`
  The live `main` ruleset requires one approving review but has no required-status-check rule; the branch-protection API likewise has no required checks. The PR workflow does run the listed jobs, but a failing job does not itself block a merge under the current repository rules. Since this section is explicitly presented as CI reality, describe these as workflow checks (or make them required) rather than telling contributors they are merge-blocking.

beardthelion:
Fixed in 8397a4a:

- Line 18: reworded to scope the denial-test requirement per route instead of a blanket non-owner assertion, so it no longer contradicts star_repo/replica/bounty/close-action routes (used CodeRabbit's proposed wording).
- Line 7: named the three known_ungated exceptions (list_webhooks, list_replicas, list_protected_branches) instead of implying full coverage; verified against the current authz_guard::every_repo_scoped_handler_is_gated allowlist.
- Line 37: reworded to describe the listed jobs as workflow checks that run on every PR rather than merge-blocking gates.

jatmn (changes requested):

## Findings

- [P1] Make the forged-certificate case a rejection test
  `AGENTS.md:25`
  This requires a handler to return `valid` for a well-formed but forged artifact, even though the preceding rule correctly says that an artifact self-signed by the DID it names must fail the independent issuer-anchor check. It also reverses the test rationale: invalid-only cases miss an implementation that rejects everything, not one that accepts everything. Following this guidance can make a verification regression test bless the exact untrusted certificate it needs to catch. Require a trusted, anchored artifact for the positive case and assert rejection for the fully populated forged artifact.

- [P2] Rebase the guide and remove the retired ungated-handler exception
  `AGENTS.md:7`
  This is stale-base documentation drift: the branch predates the #94/#113 fixes, but it will add this guide cleanly to the current base. `authz_guard::every_repo_scoped_handler_is_gated` now has an empty `known_ungated` list, and its required-gate rows cover `list_webhooks`, `list_replicas`, and `list_protected_branches` (`list_webhooks` also has the owner gate). Leaving the exception here tells future contributors that those private-repository metadata surfaces are deliberately unprotected, contradicting the protections now on `main` and inviting their removal. Rebase and update the guide to describe the current gates.

beardthelion:
Addressed on 8ccf2b1 (rebased onto current main):

- [P1] forged-certificate case: restored the signatures/certificates section. Positive case is now a trusted, independently anchored artifact; the forged well-formed artifact is a rejection case, not a drive-to-valid.
- [P2] ungated exceptions: `known_ungated` is `&[]` on main, and `list_webhooks` / `list_replicas` / `list_protected_branches` are in the required-gate rows. The guide no longer lists them as exceptions.

jatmn (approved):
LGTM

kevincodex1 (approved):

PR #213, opened 2026-07-17
test(icaptcha-client): add test that new() honors key_trusted gate (#212)

### LINKED ISSUE #212: icaptcha-client: no test binds to IcaptchaCfg::new's API-key gate, so a key-attach regression would ship green
`IcaptchaCfg::new` (`crates/icaptcha-client/src/lib.rs`) decides whether the operator's
`GITLAWB_ICAPTCHA_API_KEY` bearer token is attached to the solve, and attaches it only
when `resolve_solver_url` returns `key_trusted` (the resolved origin is the operator's own
`GITLAWB_ICAPTCHA_URL`, never a node-advertised one). That gate is what stops a hostile node
from redirecting the solve and capturing the bearer token. The pure `key_trusted` decision is
well-tested and correct; what has no test is `new()`'s wiring that consumes it.

The seven key-trust tests in the crate all call `resolve_solver_url` directly and assert on its
returned `key_trusted` bool; none construct through `IcaptchaCfg::new`, so nothing asserts that
`new()` actually honors `key_trusted` when it populates `api_key`. The other construction paths
do not cover it either: the one non-test caller, `crates/gl/src/http.rs`, reaches `new()` but
never sets `GITLAWB_ICAPTCHA_API_KEY` and never asserts on `api_key`; and PR #184 adds
`crates/icaptcha-client/tests/icaptcha_mock_consumed.rs`, which builds `IcaptchaCfg` by struct
literal with `api_key: None`, bypassing `new()` entirely. No test in the workspace sets
`GITLAWB_ICAPTCHA_API_KEY`.

Verified by mutation. The crate's source (`lib.rs` / `pow.rs` / `solvers.rs`) is byte-identical
across `origin/main`, the `fix/issue-135-ipfs-cid-tree-gate` branch, and PR #184's head; PR #184
additionally adds the `tests/` file mentioned above. Running at `origin/main`:

- Baseline: the 18 `icaptcha-client` lib tests are green.
- Defect injected into `new()` (attach the key unconditionally, ignoring `key_trusted`): all 18
  still green. That is the gap. The mutation is exactly the regression that drops the gate, and
  the suite does not notice.

The exposure if such a regression landed: a wrongly-attached key is not inert. `obtain_proof`
sends it via `req.bearer_auth(key)` on both the challenge and answer requests, so a `new()` that
attached the key for a node-advertised origin would hand the operator's bearer token to that
origin. This is the credential-exfil half of the same SSRF boundary the `resolve_solver_url`
tests already guard on the pure-function side.

Suggested fix: add one guard test in `lib.rs`'s `#[cfg(test)] mod tests` (so it shares the
existing `ICAPTCHA_ENV_LOCK`; a `tests/` integration file would not see that lock and its env
mutations would race the sibling env-touching tests). It constructs through `new()` and asserts
both arms:

```rust
#[test]
fn new_withholds_key_from_untrusted_origin_but_attaches_for_operator() {
    let guard = ICAPTCHA_ENV_LOCK.lock().unwrap();
    let prev_key = std::env::var_os("GITLAWB_ICAPTCHA_API_KEY");
    let prev_op = std::env::var_os("GITLAWB_ICAPTCHA_URL");
    std::env::set_var("GITLAWB_ICAPTCHA_API_KEY", "secret-bearer");

    // No operator configured, node advertises an attacker origin: the key has
    // no trusted destination, so new() must NOT attach it.
    std::env::remove_var("GITLAWB_ICAPTCHA_URL");
    let from_untrusted =
        IcaptchaCfg::new("did:key:zTEST", Some("https://evil.example".into()), None).api_key;

    // Operator origin configured and advertised: the key is attached.
    std::env::set_var("GITLAWB_ICAPTCHA_URL", "https://icap.mynode.example");
    let from_operator =
        IcaptchaCfg::new("did:key:zTEST", Some("https://icap.mynode.example/v1".into()), None)
            .api_key;

    // Restore env and release the lock BEFORE asserting, so a failing assertion
    // can't poison the shared lock or cascade into a sibling test.
    match prev_key {
        Some(v) => std::env::set_var("GITLAWB_ICAPTCHA_API_KEY", v),
        None => std::env::remove_var("GITLAWB_ICAPTCHA_API_KEY"),
    }
    match prev_op {
        Some(v) => std::env::set_var("GITLAWB_ICAPTCHA_URL", v),
        None => std::env::remove_var("GITLAWB_ICAPTCHA_URL"),
    }
    drop(guard);

    assert_eq!(from_untrusted, None, "key must not ride to an untrusted origin");
    assert_eq!(from_operator.as_deref(), Some("secret-bearer"));
}
```

This test is load-bearing, verified both ways at `origin/main`: RED against the unconditional-key
mutation (the whole suite shows exactly one failure, on the `from_untrusted` assertion where
`api_key` is `Some` but must be `None`), green against the current code (19 lib tests pass). It
captures both results and restores env before asserting so a RED never poisons the shared lock.
The exact env-restore shape is a starting point, not a mandate; a `temp-env`-style helper would
read cleaner if the crate takes that on.

This is a coverage gap on a gate whose logic is correct today, not a live vulnerability, so it
does not block #184. Filing separately for the same reason #211 was.

Gravirei:
## Summary

Adds a guard test that constructs `IcaptchaCfg` through `new()` and verifies the API key gate is honored — key withheld for untrusted origins, attached for the operator's own origin.

## Motivation & context

Closes #212

All existing tests called `resolve_solver_url` directly; none exercised `new()`'s wiring that consumes `key_trusted`. A regression dropping the gate would ship green.

## Kind of change

- [ ] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [x] Tests / CI

## What changed

- `crates/icaptcha-client/src/lib.rs`: Added `new_withholds_key_from_untrusted_origin_but_attaches_for_operator` test

## How a reviewer can verify

```sh
cargo test -p icaptcha-client
```

## Before you request review

## Protocol & signing impact

No protocol impact.

## Notes for reviewers

This test saves/restores env vars and releases the lock before asserting, so a failure won't poison the shared lock or cascade into sibling tests.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for `IcaptchaCfg::new` to ensure the iCaptcha API key is attached only when the node-advertised iCaptcha origin matches the configured operator URL.
  * Verified URL/origin resolution across scenarios, including default fallback, operator-based resolution, and rejection of untrusted origins (including mismatched ports).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3300,7 +3300,7 @@ dependencies = [
 
 [[package]]
 name = "git-remote-gitlawb"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "anyhow",
  "gitlawb-core",
@@ -3312,7 +3312,7 @@ dependencies = [
 
 [[package]]
 name = "gitlawb-attest"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "base64",
  "ed25519-dalek",
@@ -3329,7 +3329,7 @@ dependencies = [
 
 [[package]]
 name = "gitlawb-core"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "anyhow",
  "base64",
@@ -3356,7 +3356,7 @@ dependencies = [
 
 [[package]]
 name = "gitlawb-node"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "alloy",
  "anyhow",
@@ -3412,7 +3412,7 @@ dependencies = [
 
 [[package]]
 name = "gl"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "alloy",
  "anyhow",
@@ -3824,6 +3824,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
+ "sha2",
  "thiserror 2.0.18",
  "tracing",
 ]
--- a/crates/icaptcha-client/src/lib.rs
+++ b/crates/icaptcha-client/src/lib.rs
@@ -535,6 +535,15 @@ mod tests {
             None,
         );
 
+        // Arm 4: Operator configured, no URL advertised (the node sent only
+        // x-icaptcha-level, not x-icaptcha-url). resolve_solver_url falls back
+        // to the operator origin and marks the key trusted. This is the common
+        // API-key deployment path — a regression that additionally required
+        // url.is_some() for key attachment would leave all current tests green
+        // but make these fallback challenge/answer requests omit Authorization
+        // and fail against the protected operator service.
+        let operator_fallback = IcaptchaCfg::new("did:key:zTEST", None, Some(3));
+
         // Restore env and release the lock BEFORE asserting, so a failing
         // assertion can't poison the shared lock or cascade into a sibling.
         match prev_key {
@@ -564,5 +573,15 @@ mod tests {
             "key must not ride to default origin when operator is configured"
         );
         assert_eq!(default_advert.url, "https://icaptcha.gitlawb.com/v2");
+
+        assert_eq!(
+            operator_fallback.api_key.as_deref(),
+            Some("secret-bearer"),
+            "fallback to operator origin with no advert must still attach key"
+        );
+        assert_eq!(
+            operator_fallback.url, "https://icap.mynode.example",
+            "no-advert path must resolve to operator URL"
+        );
     }
 }
--- a/crates/icaptcha-client/src/lib.rs
+++ b/crates/icaptcha-client/src/lib.rs
@@ -557,6 +557,16 @@ mod tests {
         let operator_hostile_advert =
             IcaptchaCfg::new("did:key:zTEST", Some("https://evil.example".into()), None);
 
+        // Arm 6: Operator configured, same host but different port.
+        // origin_key includes the port, so :8443 must not match the
+        // operator origin at the default :443 — the advert is rejected
+        // and the solver falls back to the operator origin.
+        let operator_diff_port = IcaptchaCfg::new(
+            "did:key:zTEST",
+            Some("https://icap.mynode.example:8443/evil".into()),
+            None,
+        );
+
         // Restore env and release the lock BEFORE asserting, so a failing
         // assertion can't poison the shared lock or cascade into a sibling.
         match prev_key {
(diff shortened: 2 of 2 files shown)


beardthelion (changes requested):
Verified this load-bearing by mutation at 7c24cd7 before anything else: baseline 19/19; making new() attach the key unconditionally goes RED on exactly the untrusted-arm assertion (left: Some("secret-bearer"), right: None); making it never attach goes RED on the operator arm; restore returns 19/19. Both arms bind the gate, the restore path uses remove_var rather than an empty set_var, and both asserts run after env-restore plus lock-drop, so a failure can't poison ICAPTCHA_ENV_LOCK. Ambient GITLAWB_ICAPTCHA_INSECURE can't flip either arm (every URL here is https, and with no operator configured lib.rs:166 pins key_trusted false). CI is green including MSRV. Solid work, and exactly what #212 asked for.

The same mutation pass turned up one state the test doesn't drive, worth closing while you're in this block.

## Findings

- [P2] Drive the third reachable key_trusted state: operator configured, default host advertised
  `crates/icaptcha-client/src/lib.rs:517`
  The untrusted arm removes GITLAWB_ICAPTCHA_URL, so no arm ever observes the key withheld while an operator IS configured. That state is reachable: with the operator var set and a node advertising https://icaptcha.gitlawb.com/v2, the advert is allowlisted and chosen while its origin differs from the operator's, so key_trusted must be false. I ran both regressions that live in this gap: `if key_trusted` -> `if operator.is_some()` at lib.rs:204, and dropping the origin comparison at lib.rs:166. Each passes the full crate suite 19/19 with the test as submitted, and either one hands the operator's bearer key to an origin the operator did not choose, steerable by any node via x-icaptcha-url. Fix: before the restore block, while the operator var is still set, construct once more with Some("https://icaptcha.gitlawb.com/v2".into()) and assert api_key is None. I ran that arm both ways: RED under both mutations, green on the current code.

- [P3] Pin cfg.url in both arms
  `crates/icaptcha-client/src/lib.rs:519`
  The test asserts only api_key, so new()'s URL wiring is unpinned: url: url.unwrap_or(resolved_url) at lib.rs:212 passes the whole crate suite yet routes the solve to the hostile advertised origin. Capturing the cfg and asserting url == DEFAULT_URL in the untrusted arm (and "https://icap.mynode.example/v1" in the operator arm) closes it; I ran those asserts both ways (RED under that mutation with left "https://evil.example", green on current code).

- [P3] Retitle to test(icaptcha-client): the change is test-only
  CONTRIBUTING.md line 55: commit prefixes drive the release-please version bump, so a fix: title on squash-merge would cut a patch release for a tests-only change. Recent test-only merges use test(...) (ad7c2b2, c296e23).

Out of scope here: the empty-string GITLAWB_ICAPTCHA_API_KEY path through new()'s filter (filing that one separately), and wire-level observation that the bearer actually rides both challenge and answer requests, which belongs with the proxy-observer machinery #214 is building rather than a new issue.

jatmn (changes requested):

## Findings

- [P2] Exercise the configured-operator/default-host key gate
  `crates/icaptcha-client/src/lib.rs:522`
  With `GITLAWB_ICAPTCHA_URL=https://icap.mynode.example`, a node may advertise the separately allowlisted `https://icaptcha.gitlawb.com/v2`; `resolve_solver_url` selects that URL, but `key_trusted` must be false because it is not the operator origin. The two current arms both remain green if `new()` regresses from `if key_trusted` to merely testing whether an operator is configured (or if the origin comparison is removed), which sends the operator bearer token to the node-selectable default origin. Add that third construction while the operator environment variable is set and assert that `api_key` is `None`.

- [P2] Assert the resolved URL from the constructor
  `crates/icaptcha-client/src/lib.rs:518`
  This test keeps only `api_key`, so it does not bind `new()` to use `resolved_url`. A regression such as assigning the raw advertised URL to `IcaptchaCfg::url` still passes both assertions and the direct resolver tests, yet makes the untrusted `https://evil.example` advertisement the challenge/answer destination. Retain each constructed config and assert that the rejected advertisement resolves to `DEFAULT_URL` and the matching operator advertisement retains its allowed `/v1` URL.

- [P3] Use the test Conventional Commit type for this test-only change
  `7c24cd72943b0d02289ca1bdbbccc887a7d750d7`
  The PR changes only a unit test, but its `fix(icaptcha-client):` title/commit will cause release-please to select a patch release: `CONTRIBUTING.md` states that Conventional Commit prefixes drive version bumps. Please retitle/recommit it as `test(icaptcha-client): ...` so merging this coverage-only change does not publish an unnecessary release.

jatmn (changes requested):

## Findings

- [P2] Exercise the configured-operator fallback construction
  `crates/icaptcha-client/src/lib.rs:522`
  All key-attached constructions pass an advertised operator URL. A normal API-key deployment also reaches `IcaptchaCfg::new` with `GITLAWB_ICAPTCHA_URL` configured but no `x-icaptcha-url` header (for example, the client receives only `x-icaptcha-level`); `resolve_solver_url(None, Some(operator))` selects that operator and marks the key trusted. Consequently, a regression such as additionally requiring `url.is_some()` when attaching the key leaves all 19 current crate tests green but makes these fallback challenge/answer requests omit `Authorization` and fail against the protected operator service. Construct this `url = None` case while the operator env var is set and assert both the operator URL and attached key.

beardthelion (changes requested):
Re-reviewed `6b2ec93d`. My two prior findings are resolved, both confirmed by mutation: the operator-configured / default-host-advertised state now withholds the key, and each arm pins the resolved URL.

Concurring with jatmn's remaining finding: the operator-configured, no-advertised-URL fallback (`resolve_solver_url(None, Some(operator))`) attaches the key with no arm covering it, so a `url.is_some()`-style regression ships green. I ran it: with `if key_trusted && url.is_some()` the 19 submitted tests all stay green while the real `url = None` path drops the `Authorization` header. One construction with the operator var set and `url = None`, asserting the key attaches, closes it. Not approving over that.

beardthelion (changes requested):
Re-reviewed `9be5eac3`. All three prior P2 asks are in, and I certified each by mutation at this head: `if key_trusted` -> `if operator.is_some()` and the dropped origin comparison both go RED on exactly the arm-3 assertion (left: Some("secret-bearer"), right: None); `key_trusted && url.is_some()` goes RED on arm 4 (left: None, right: Some("secret-bearer")); `url.unwrap_or(resolved_url)` goes RED on arm 1's URL pin. Restore returns 19/19. The Cargo.lock delta is a clean regeneration against main's 0.5.1 manifests (sha2 v0.10.9 was already declared and consumed by pow.rs; no new packages, no checksum churn).

One state is still unpinned, and it is the scenario in the test's own name.

## Findings

- [P2] Pin the operator-configured hostile-advert state (the exfiltration direction)
  `crates/icaptcha-client/src/lib.rs:517`
  Arm 1 is the only hostile-advert arm and it removes GITLAWB_ICAPTCHA_URL first, so its `api_key == None` passes because no operator exists, not because a live key was withheld from an attacker origin. I ran the regression that lives in this gap: `url: if operator.is_some() && url.is_some() { url.clone().unwrap() } else { resolved_url }` at lib.rs:212 keeps all 19 tests green while (operator set, key set, advert `https://evil.example`) yields `url=https://evil.example`, `api_key=Some("secret-bearer")`: the operator bearer rides to a node-steered origin, and arm 1's URL pin cannot see it because the override only fires when an operator is configured. Fix: a fifth construction with the operator var and key still set and advert `Some("https://evil.example".into())`, asserting `url == "https://icap.mynode.example"` and `api_key.as_deref() == Some("secret-bearer")`. I ran that arm both ways: RED under the mutation (it observes the exfil state directly), GREEN on the current code.

- [P3] Retitle the PR itself to test(icaptcha-client)
  Commit 6b2ec93d carries the requested test( type, but this repo squash-merges with COMMIT_OR_PR_TITLE and the PR now has two commits, so the merge commit takes the PR title, which still says fix(icaptcha-client). Per CONTRIBUTING.md that cuts a patch release for a tests-only change. With the title fixed the squash result is right regardless of the second commit's fix( message.

- [P3] Optional: pin the (no operator, no advert, key set) state while you are in this block
  `crates/icaptcha-client/src/lib.rs:204`
  `if key_trusted || (url.is_none() && operator.is_none())` passes all 19 tests and would attach the bearer to the public default host whenever the node omits x-icaptcha-url. One construction in arm 1's env state with `url = None`, asserting `api_key == None` and `url == DEFAULT_URL`, closes it; I ran it both ways (RED under that mutation, GREEN on current code). Take it or leave it in this PR.

Out of scope for this round, deliberately: wire-level observation that the bearer actually rides (or is absent from) the challenge and answer requests stays with the proxy-observer machinery #214 is building, and the empty-string GITLAWB_ICAPTCHA_API_KEY filter is still mine to file separately. Everything else from the prior rounds is resolved and verified.

beardthelion (dismissed):
Re-certified the gate by mutation at `82fec97a`: baseline 19/19, all six prior gate-regressions each go RED on their arm, restore returns 19/19. Arm 5 (operator + hostile advert) and the URL pins are in and correct. One security-relevant state is still unpinned, and it is the same shape as the arms you already added.

## Findings

- **[P2] Pin https port-isolation through `new()`**
  `crates/icaptcha-client/src/lib.rs:116`
  Every port test reaches `origin_key` through the `GITLAWB_ICAPTCHA_INSECURE` http-localhost escape hatch (`rejects_insecure_advertised_url_on_different_port`), so the production https path never pins that a differing-port advert is a distinct origin. I ran the regression: stripping the port from `origin_key` for `scheme == "https"` keeps all 19 tests green, and under it an advert of `https://icap.mynode.example:8443/evil` resolves as the operator origin, so the solve (and the trusted bearer) rides to the node-steered `:8443` URL on the operator host. Production code is correct today; this is the missing regression fence. Add one arm with the operator var set: `IcaptchaCfg::new("did:key:zTEST", Some("https://icap.mynode.example:8443/evil".into()), None)`, assert `url == "https://icap.mynode.example"` and `api_key == Some("secret-bearer")`. I ran that arm both ways: GREEN on current code, RED under the mutant (the `url` pin catches it).

- **[P3] Collect arm failures instead of asserting sequentially**
  `crates/icaptcha-client/src/lib.rs:572`
  The six arms assert in sequence, so the first failing arm masks every later one in a single run. Either split per-arm or push mismatches into a `Vec` and `assert!(failures.is_empty(), ...)` once at the end, so a multi-arm regression shows every affected arm.

- **[P3] Add assert messages to the three bare `assert_eq!`s**
  `crates/icaptcha-client/src/lib.rs:590`
  `operator_match.api_key`, `operator_match.url`, and `default_advert.url` assert without the descriptive message every sibling arm carries. Match the convention so a failure names the arm.

jatmn (changes requested):

## Findings

- [P2] Exercise HTTPS port isolation through `IcaptchaCfg::new`
  `crates/icaptcha-client/src/lib.rs:527`
  The current `origin_key` implementation correctly preserves ports, but the added constructor arms never combine the configured HTTPS operator with a same-host advertised URL on a different port; the only port test enables the insecure localhost escape hatch and calls `resolve_solver_url` directly. If a later change regresses HTTPS port handling, an advert such as `https://icap.mynode.example:8443/evil` matches configured `https://icap.mynode.example`, so `new()` preserves that node-selected URL and attaches the operator bearer. All current tests stay green, but both challenge and answer requests then send the bearer to `:8443`. Add a configured-operator constructor arm for that advert and assert fallback to the operator URL with the key attached.

jatmn (approved):
@beardthelion LGTM

beardthelion (approved):
Certified the current head by mutation (baseline 19/19 green):

- Drop the port from `origin_key` and arm 6 goes red (`got https://icap.mynode.example:8443/evil`), so the same-host/different-port advert really is rejected and falls back to the operator origin.
- Force `key_trusted = true` and arms 1/1b/3 go red, each surfacing the `secret-bearer` token where it must be `None`. The move from fail-fast `assert_eq!` to the accumulate-and-`assert!(failures.is_empty())` harness stays load-bearing.

The gate under test (`resolve_solver_url`/`origin_key`/`new`) is byte-identical to `main`; the only production drift is `sanitize_excerpt`, where `main` is ahead by the bidi-control strip. Please rebase onto `main` before merge to pick that up. It is a clean disjoint-region rebase, no gate is dropped.

LGTM. @kevincodex1 this is ready to merge.

beardthelion:
@kevincodex1 ready to merge. jatmn and I both approved the current head (`1ba9ee5`) and CI is green. Test-only, covers the `key_trusted` gate in `new()` for #212.

kevincodex1 (approved):

PR #214, opened 2026-07-17
test(icaptcha): harden mock-consumed guard with proxy observer and blackhole tripwire

### LINKED ISSUE #211: icaptcha mock-consumed guard: swallowed dual-contact leak is invisible, and the blackhole has no tripwire test
`crates/icaptcha-client/tests/icaptcha_mock_consumed.rs` (#184) guards against a swallowed dual-contact leak with two
halves: mock-consumed asserts, plus a proxy blackhole (`NO_PROXY` for the mock's
host, `127.0.0.1,localhost`, every proxy var pointed at a closed port) so a request
to any other origin fails instead of silently succeeding.

The blackhole half works as documented for leaks whose error propagates. Verified by
mutation at the #184 head:

- `cfg.url` pointed at `DEFAULT_URL` with the blackhole armed: RED, failing in the
  blackholed tunnel (`tunnel error ... Connection refused`), exactly as the test's
  doc comment claims.
- A dual-contact leak injected into `obtain_proof` (a stray call alongside the
  sanctioned flow: `client.get(DEFAULT_URL).send().context("leak")?` before the
  challenge request): RED on the injected call, aborting before the mock is
  reached. The green direction is inferred, not re-run: without the blackhole the
  request bypasses the proxy, and on a runner that can reach `DEFAULT_URL` it
  would succeed (`send()` does not error on non-2xx statuses), leaving the test
  green with the mock consumed. So the env setup is what makes this class fail.

Two gaps remain, both hardening rather than a current vulnerability:

1. **A swallowed leak stays green.** Change the injected leak to
   `let _ = client.get(DEFAULT_URL).send();` and the test passes: the mock is
   consumed and the leak's transport error is discarded. The closed-port design can
   only surface leaks whose failure propagates into the flow's result. Catching
   fire-and-forget exfiltration needs an observer rather than a dead end, e.g. a
   local listener standing in as the proxy that counts connection attempts, with the
   test asserting zero. Scope caveat: an observer only witnesses paths the test
   drives. The current test exercises a single passing round (200 on both endpoints,
   no PoW, no continue/failed branch), so a leak living on an error or retry branch
   makes no connection during the run and a zero-count assertion stays green. An
   observer design should either drive those branches too or state that coverage is
   limited to the exercised path, and its guarantee covers only proxy-honoring
   transports. Candidate direction, not a vetted design.

2. **The blackhole mechanism has no committed tripwire.** `obtain_proof` builds a
   stock `reqwest::blocking::Client`, which is why the proxy env binds its transport.
   If the builder later gains `.no_proxy()` or a custom connector, the guard disarms
   silently and every existing run stays green. A committed negative case should lock
   the mechanism (a test asserting that a non-loopback `cfg.url` fails while the
   blackhole is armed). Design note: the naive versions are wrong in one of two ways.
   Pointing at an unresolvable host keeps the request failing even when the guard is
   disarmed, so a tripwire asserting failure stays green vacuously (a DNS failure
   masquerades as the blackhole working), and pointing at a real host makes a live
   network call the moment the guard disarms, degrading into the same masquerade on
   any runner that cannot reach that host (the connect error passes for the
   blackhole working). A robust tripwire needs a destination that would succeed if
   and only if the blackhole is not applied, e.g. a local listener on a loopback
   alias that `NO_PROXY` does not cover; note this depends on reqwest's `NO_PROXY`
   matching by exact IP (holds in the pinned reqwest/hyper-util, recheck on
   upgrade) and on the OS routing 127/8 aliases (default on Linux, macOS needs an
   explicit alias). Also a candidate direction, not a vetted design.

Neither gap blocks #184, which is why this is an issue and not a review comment.

Gravirei:
## Summary

Hardens the iCaptcha mock-consumed guard with a proxy observer (catches fire-and-forget leaks) and a blackhole tripwire (detects if the proxy guard is silently disarmed).

## Motivation & context

Closes #211

## Kind of change

- [ ] Security fix
- [ ] Tests / CI

## What changed

- **icaptcha-client**: Replaced the closed-port proxy with a `ProxyObserver` (TCP listener counting connections) so fire-and-forget leaks whose errors are discarded are caught
- **icaptcha-client**: Added a blackhole tripwire test asserting `obtain_proof` fails against `[::1]` (IPv6 loopback) while the proxy is armed; goes RED if `.no_proxy()` or a custom connector is added
- **icaptcha-client**: No dependency changes — `mockito` was already a dev-dependency

## How a reviewer can verify

```sh
cargo test -p icaptcha-client
cargo clippy -p icaptcha-client --all-targets -- -D warnings
cargo fmt --check -p icaptcha-client
```

## Before you request review

## Protocol & signing impact

N/A - test-only crate change

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage for network protection during proof requests.
  * Verified proof requests succeed when protection is disabled and fail with a connection error when enabled.
  * Added monitoring to detect unexpected network connections and prevent unintended network access.
  * Improved test isolation by restoring temporary network settings automatically after each test.
  * Documented test limitations and synchronization behavior for more reliable validation.

### DIFF
--- a/crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs
+++ b/crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs
@@ -13,6 +13,9 @@
 //! intercepts the connection and blocks it; without the blackhole the request
 //! would reach the server directly and succeed, turning this test RED.
 //!
+//! A positive control (disarmed blackhole) runs first, proving the fixture is
+//! valid and `127.0.0.2` is reachable.
+//!
 //! Design constraints (see issue #211):
 //!   - An unresolvable host would keep the request failing even when the guard
 //!     is disarmed (DNS failure masquerading as the blackhole), so we must use
@@ -45,10 +48,11 @@ fn serve_icaptcha(listener: TcpListener) {
         for stream in listener.incoming() {
             let Ok(mut stream) = stream else { continue };
             let mut buf = [0; 4096];
-            if stream.read(&mut buf).is_err() {
-                continue;
-            }
-            let request = String::from_utf8_lossy(&buf);
+            let n = match stream.read(&mut buf) {
+                Ok(0) | Err(_) => continue,
+                Ok(n) => n,
+            };
+            let request = String::from_utf8_lossy(&buf[..n]);
             let body = if request.contains("/v1/answer") {
                 r#"{"status":"passed","proof":"PROOF-TRIP"}"#
             } else {
@@ -63,32 +67,48 @@ fn serve_icaptcha(listener: TcpListener) {
     });
 }
 
-#[test]
-fn obtain_proof_fails_when_blackhole_blocks_non_loopback_destination() {
-    // Start a listener on 127.0.0.2, a loopback alias NOT listed in NO_PROXY.
-    let listener = TcpListener::bind("127.0.0.2:0")
-        .expect("bind on 127.0.0.2 (requires OS support for 127/8 loopback)");
-    let port = listener.local_addr().unwrap().port();
-    serve_icaptcha(listener);
-
-    support::arm_blackhole("http://127.0.0.1:1");
-
-    let cfg = IcaptchaCfg {
-        url: format!("http://127.0.0.2:{port}"),
+fn make_cfg(url: &str) -> IcaptchaCfg {
+    IcaptchaCfg {
+        url: url.to_string(),
         did: "did:key:zTEST".to_string(),
         level: 1,
         api_key: None,
+    }
+}
+
+#[test]
+fn obtain_proof_blackhole_tripwire() {
+    // Try to bind to 127.0.0.2 — skip if the OS doesn't route 127/8 as
+    // loopback (macOS needs an explicit alias).
+    let listener = match TcpListener::bind("127.0.0.2:0") {
+        Ok(l) => l,
+        Err(e) => {
+            eprintln!("SKIP: 127.0.0.2 not available ({e}); skipping tripwire test");
+            return;
+        }
     };
+    let port = listener.local_addr().unwrap().port();
+    let url = format!("http://127.0.0.2:{port}");
+    serve_icaptcha(listener);
 
-    let solve = |_c: &Challenge| Some("silent".to_string());
-    let solver: &dyn Fn(&Challenge) -> Option<String> = &solve;
+    // ── Positive control: disarmed blackhole → obtain_proof succeeds ──────
+    let _disarmed_guard = support::disarm_proxy_env();
 
-    let result = obtain_proof(&cfg, Some(solver));
+    let solve: &dyn Fn(&Challenge) -> Option<String> = &|_c| Some("silent".to_string());
+    let cfg = make_cfg(&url);
+    let proof = obtain_proof(&cfg, Some(solve))
+        .expect("positive control: obtain_proof should succeed when the proxy is disarmed");
+    assert_eq!(
+        proof, "PROOF-TRIP",
+        "positive control: unexpected proof value"
+    );
+    drop(_disarmed_guard);
 
-    // The blackhole should block the connection via the proxy, producing a
-    // connect-level failure.  Narrowing the assertion to a reqwest connect
-    // error defends against fixture bugs that would otherwise make any error
-    // (e.g. deserialization of a truncated body) vacuously pass.
+    // ── Negative control: armed blackhole → obtain_proof fails with connect error ──
+    let _armed_guard = support::arm_blackhole("http://127.0.0.1:1");
+
+    let result = obtain_proof(&cfg, Some(solve));
+
     let err = result.expect_err(
         "blackhole tripwire: obtain_proof succeeded against 127.0.0.2, \
          meaning the proxy blackhole did not intercept the request; \
--- a/crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs
+++ b/crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs
@@ -1,0 +1,103 @@
+//! Tripwire: a committed negative case that locks the blackhole mechanism.
+//!
+//! The guard against non-loopback egress works because `obtain_proof` builds a
+//! stock `reqwest::blocking::Client` that inherits the process's proxy env vars.
+//! If the builder later gains `.no_proxy()` or a custom connector, the guard
+//! disarms silently and every existing test stays green (no live call is made
+//! during the normal mock-consumed run because the mock is on loopback and
+//! `NO_PROXY` covers it — the blackhole isn't even exercised).
+//!
+//! This test asserts that a non-loopback destination fails *while the blackhole
+//! is armed*. The destination is a local HTTP server on `127.0.0.2`, a loopback
+//! alias that `NO_PROXY` does NOT cover.  With the blackhole active the proxy
+//! intercepts the connection and blocks it; without the blackhole the request
+//! would reach the server directly and succeed, turning this test RED.
+//!
+//! Design constraints (see issue #211):
+//!   - An unresolvable host would keep the request failing even when the guard
+//!     is disarmed (DNS failure masquerading as the blackhole), so we must use
+//!     a reachable address.
+//!   - A real external host would make a live network call the moment the guard
+//!     disarms, and on any runner that cannot reach that host the connect error
+//!     again masquerades as the blackhole.  A local loopback alias avoids both
+//!     problems — it is always reachable when unblocked and never reaches the
+//!     real network.
+//!   - This depends on reqwest's `NO_PROXY` matching by exact IP (holds in the
+//!     pinned reqwest/hyper-util; recheck on upgrade) and on the OS routing
+//!     `127/8` as loopback (default on Linux; macOS may need an explicit alias).
+
+use std::io::{Read, Write};
+use std::net::TcpListener;
+use std::thread;
+
+use icaptcha_client::{obtain_proof, Challenge, IcaptchaCfg};
+
+mod support;
+
+/// A minimal HTTP server that responds to iCaptcha challenge and answer
+/// requests on a loopback alias NOT covered by NO_PROXY.
+///
+/// When the blackhole is working, the proxy should intercept these requests
+/// and `obtain_proof` should fail.  When the blackhole is disarmed, the
+/// requests reach this server and the flow succeeds.
+fn serve_icaptcha(listener: TcpListener) {
+    thread::spawn(move || {
+        for stream in listener.incoming() {
+            let Ok(mut stream) = stream else { continue };
+            let mut buf = [0; 4096];
+            if stream.read(&mut buf).is_err() {
+                continue;
+            }
+            let request = String::from_utf8_lossy(&buf);
+            let body = if request.contains("/v1/answer") {
+                r#"{"status":"passed","proof":"PROOF-TRIP"}"#
+            } else {
+                r#"{"challengeId":"c1","type":"anagram","difficulty":1,"prompt":"listen","token":"tok-1"}"#
+            };
+            let response = format!(
+                "HTTP/1.1 200 OK\r\nContent-Type: application/json\r\nContent-Length: {}\r\nConnection: close\r\n\r\n{body}",
+                body.len(),
+            );
+            let _ = stream.write_all(response.as_bytes());
+        }
+    });
+}
+
+#[test]
+fn obtain_proof_fails_when_blackhole_blocks_non_loopback_destination() {
+    // Start a listener on 127.0.0.2, a loopback alias NOT listed in NO_PROXY.
+    let listener = TcpListener::bind("127.0.0.2:0")
+        .expect("bind on 127.0.0.2 (requires OS support for 127/8 loopback)");
+    let port = listener.local_addr().unwrap().port();
+    serve_icaptcha(listener);
+
+    support::arm_blackhole("http://127.0.0.1:1");
+
+    let cfg = IcaptchaCfg {
+        url: format!("http://127.0.0.2:{port}"),
+        did: "did:key:zTEST".to_string(),
+        level: 1,
+        api_key: None,
+    };
+
+    let solve = |_c: &Challenge| Some("silent".to_string());
+    let solver: &dyn Fn(&Challenge) -> Option<String> = &solve;
+
+    let result = obtain_proof(&cfg, Some(solver));
+
+    // The blackhole should block the connection via the proxy, producing a
+    // connect-level failure.  Narrowing the assertion to a reqwest connect
+    // error defends against fixture bugs that would otherwise make any error
+    // (e.g. deserialization of a truncated body) vacuously pass.
+    let err = result.expect_err(
+        "blackhole tripwire: obtain_proof succeeded against 127.0.0.2, \
+         meaning the proxy blackhole did not intercept the request; \
+         the no-live-call guard is disarmed",
+    );
+    assert!(
+        err.chain().any(|c| c
+            .downcast_ref::<reqwest::Error>()
+            .is_some_and(|e| e.is_connect())),
+        "expected a connect/proxy error, got: {err:#}",
+    );
+}
--- a/crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs
+++ b/crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs
@@ -27,7 +27,8 @@
 //!     real network.
 //!   - This depends on reqwest's `NO_PROXY` matching by exact IP (holds in the
 //!     pinned reqwest/hyper-util; recheck on upgrade) and on the OS routing
-//!     `127/8` as loopback (default on Linux; macOS may need an explicit alias).
+//!     `127/8` as loopback (default on Linux; macOS needs an explicit alias —
+//!     the test fails loudly, never skips, if `127.0.0.2` cannot be bound).
 
 use std::io::{Read, Write};
 use std::net::TcpListener;
@@ -91,15 +92,19 @@ fn make_cfg(url: &str) -> IcaptchaCfg {
 
 #[test]
 fn obtain_proof_blackhole_tripwire() {
-    // Try to bind to 127.0.0.2 — skip if the OS doesn't route 127/8 as
-    // loopback (macOS needs an explicit alias).
-    let listener = match TcpListener::bind("127.0.0.2:0") {
-        Ok(l) => l,
-        Err(e) => {
-            eprintln!("SKIP: 127.0.0.2 not available ({e}); skipping tripwire test");
-            return;
-        }
-    };
+    // Bind to 127.0.0.2, a loopback alias NOT covered by NO_PROXY.  Linux
+    // routes the whole 127/8 block as loopback; macOS does not alias it by
+    // default (add one with `sudo ifconfig lo0 alias 127.0.0.2`).  Fail
+    // loudly on platforms that cannot bind it: a silent skip would report
+    // this tripwire as green while exercising none of the negative control,
+    // recreating the very gap it exists to close.
+    let listener = TcpListener::bind("127.0.0.2:0").expect(
+        "cannot bind 127.0.0.2 for the blackhole tripwire. Linux routes \
+         127/8 as loopback; on macOS add the alias first \
+         (`sudo ifconfig lo0 alias 127.0.0.2`). This test fails rather than \
+         silently skipping so a platform without the alias cannot count the \
+         negative control as passed.",
+    );
     let port = listener.local_addr().unwrap().port();
     let url = format!("http://127.0.0.2:{port}");
     serve_icaptcha(listener);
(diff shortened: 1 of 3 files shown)


beardthelion (dismissed):
The observer half does what #211 asked. I verified it by mutation at 9c731c8: injecting a swallowed fire-and-forget leak (`let _ = client.get("http://leak.example.com/exfil").send();`, discarded error) into obtain_proof makes the mock-consumed test go RED with the observer counting one connection, exactly the case the old closed-port blackhole missed. That's the real win here.

The tripwire, though, is vacuous as committed: it passes whether or not the blackhole is armed. Details below, each checked by running it.

## Findings

- [P1] Fix the tripwire fixture's Content-Length so the test actually locks the blackhole
  `crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs:54`
  `serve_icaptcha` declares `Content-Length: 78` for an 86-byte challenge body and `42` for a 40-byte answer body. Both are wrong, and it makes the tripwire test nothing. I ran it three ways: with the blackhole disarmed (proxy env cleared) and the committed lengths, obtain_proof returns `Err("parse iCaptcha challenge")` because the truncated body fails to deserialize, so `assert!(result.is_err())` passes with the blackhole off. Correct the lengths to 40/86 and disarmed now returns `Ok("PROOF-TRIP")`; armed returns a connect error. So the fix is load-bearing and the current test is not. Derive the length from the body instead of hardcoding: bind each body to a variable and use `body.len()` in the format string.

- [P2] Assert the tripwire error is a connect/proxy failure, not bare `is_err()`
  `crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs:109`
  Even with the lengths fixed, `result.is_err()` accepts any failure, so a future fixture bug can re-vacuate the test the same way this one did. I confirmed the armed-blackhole error downcasts to `reqwest::Error` with `is_connect() == true` ("tcp connect error: Connection refused"), so narrowing the assertion to that shape is both accurate and self-defending: `err.chain().any(|c| c.downcast_ref::<reqwest::Error>().is_some_and(|e| e.is_connect()))`. As a bonus it fails today's parse-error vacuity directly.

- [P2] State the observer's coverage scope, or drive the branches it can't see
  `crates/icaptcha-client/tests/icaptcha_mock_consumed.rs:14`
  The observer only witnesses leaks issued on the one path the test drives: a single passing round, no PoW, no Continue/Failed, no non-2xx. A leak on any of those branches makes no connection during the run and the zero-count assertion stays green, yet the test is named `..._makes_no_live_call` and the header claims "no leak reached the network" without qualification. It also can't see a transport that ignores proxy env (`.no_proxy()` or a raw TcpStream). #211 asked for exactly this: either drive those branches or state that the guarantee covers only the exercised path and proxy-honoring transports. Add that caveat to the header (and ideally an observer-armed test driving a Continue round and a non-2xx response).

- [P3] Retitle to `test(icaptcha): ...`
  The diff is test-only (two `tests/` files plus a `mockito` dev-dependency; no `src/` change), so `fix(` would have release-please cut a patch release for it. Test-only changes here use `test(...)` (ad7c2b2, c296e23); CONTRIBUTING.md line 55 notes prefixes drive the bump.

- [P3] `arm_blackhole` is near-duplicated across the two test files
  They differ only in the proxy target (the observer port vs the closed `:1`). A `tests/support/mod.rs` exporting one `arm_blackhole(target: &str)` both files pull in via `mod support;` removes the duplication. Minor; fine to leave for two small binaries if you'd rather.

One note, not a change request: the Cargo.lock version bumps (0.5.0 -> 0.5.1 across five crates) are not churn from this PR. Those crates' manifests are already 0.5.1 on main; only the lockfile was stale, and regenerating it to add mockito correctly synced them. Leave them as they are. A one-line mention in the PR description that the lockfile picked up an unrelated sync would be tidy, nothing more.

The observer work is a genuine improvement; once the tripwire actually exercises the blackhole it'll be a solid pair.

jatmn (changes requested):

## Findings

- [P1] Make the blackhole tripwire’s successful fixture valid
  `crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs:50`
  The challenge body is 86 bytes but advertises `Content-Length: 78` (and the 40-byte answer advertises 42).  If a future client bypasses the proxy, it reaches the local listener but receives truncated JSON, so `obtain_proof` returns an error and the `is_err()` assertion still passes.  That leaves the test green with the blackhole disarmed—the exact regression this PR is intended to catch.  Derive the header from `body.len()` (or otherwise make the direct fixture valid) so an unblocked flow succeeds.

- [P2] Assert that the tripwire failed at the proxy connection
  `crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs:109`
  Even with valid response lengths, `result.is_err()` accepts any unrelated failure: a malformed fixture, changed request protocol, or solver error would again leave the test green after the proxy guard was bypassed.  The test should require the expected connect/proxy failure (for example by finding a `reqwest::Error` with `is_connect()` in the error chain), or include an explicit unarmed positive control.

jatmn (approved):
@beardthelion LGTM

beardthelion (dismissed):
Re-reviewed `87c716e3`. The tripwire rework is load-bearing: disarmed, the flow returns `Ok("PROOF-TRIP")` (a real positive control); armed, it fails with a reqwest connect error. Injecting a swallowed fire-and-forget leak into `obtain_proof` makes the observer count 1 and the mock-consumed test go red. Content-Length now derives from `body.len()`, the assertion narrows to a connect error, the observer-scope caveat is stated, and `arm_blackhole` is deduped into `support/mod.rs`. Prior findings resolved.

@kevincodex1 ready to merge.

kevincodex1 (approved):

kevincodex1:
hi @Gravirei thaks for this contribution please rebase to main and fix conflicts

jatmn (changes requested):

## Findings

- [P2] Rebase this branch onto the current `main` before merge
  `e45bf312033f9a51cba239dcd9cdc383838266d7`
  The PR head is based on `e14c2d3f3f4269c61ac8f8ae228223f46d22c078`, while `main` is now `b484a24955548257bb4f3fa92f460ba8fd00eafe`; GitHub currently marks the PR blocked, and a maintainer has requested a rebase. The current merge tree is clean, so this is branch-state drift rather than a code conflict, but please rebase before merging so the final change can be reviewed against the current base.

- [P2] Synchronize the proxy observer before asserting that no egress occurred
  `crates/icaptcha-client/tests/icaptcha_mock_consumed.rs:63`
  The counter is incremented only after the detached accept thread is scheduled, but the test reads it immediately after `obtain_proof` returns. A future fire-and-forget/background request can connect to the listener backlog after that read (or before the accept thread runs), so the test exits green even though the leak this guard is meant to catch occurred. Add a bounded drain/acknowledgement or another lifecycle synchronization before asserting a zero count.

jatmn (changes requested):
## Findings

### [P1] `arm_blackhole` does not clear `REQUEST_METHOD`, which disables all proxies

**Severity:** High  
**Files:** `crates/icaptcha-client/tests/support/mod.rs:3-15`  
**Also affects:** `crates/icaptcha-client/tests/icaptcha_mock_consumed.rs:115`, `crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs:74`

In the pinned `hyper-util 0.1.20` (`src/client/proxy/matcher.rs:228-236`), `Builder::from_env()` sets:

```rust
is_cgi: std::env::var_os("REQUEST_METHOD").is_some(),
```

and `build()` (`matcher.rs:304-311`) returns an empty matcher when `is_cgi` is true, ignoring every proxy variable. `arm_blackhole` sets `ALL_PROXY`, `HTTPS_PROXY`, `HTTP_PROXY`, and `NO_PROXY` in both cases, but it does not remove `REQUEST_METHOD`.

**Impact:** If the test runner, CI wrapper, or developer environment happens to set `REQUEST_METHOD`, the blackhole is silently disarmed. `icaptcha_mock_consumed.rs` still passes because its mock lives on `127.0.0.1` and direct loopback access satisfies the mock-consumed assertions; the proxy observer merely sees zero connections, which is indistinguishable from success. Only the tripwire test fails, and then with a confusing "obtain_proof succeeded" message rather than a clear "blackhole disabled" signal.

**Fix:** Clear `REQUEST_METHOD` in `arm_blackhole`:

```rust
std::env::remove_var("REQUEST_METHOD");
```

or save/restore it the same way `InsecureEnv` in `src/lib.rs` handles `GITLAWB_ICAPTCHA_INSECURE`.

---

### [P2] `ProxyObserver::flush()` can under-count leaked connections

**Severity:** Medium–High  
**File:** `crates/icaptcha-client/tests/icaptcha_mock_consumed.rs:96-107`

The flush is supposed to guarantee that every leaked connection initiated during `obtain_proof` has been counted before the zero-leak assertion. It does not provide that guarantee:

1. **Sentinel-ordering race.** `flush()` opens a sentinel self-connection and assumes that once the sentinel is accepted, all prior connections have been counted. TCP backlogs are not FIFO; a leaked connection whose handshake completes after the sentinel's can remain in the backlog while the sentinel is accepted first. `flush()` then reads `count == 1`, subtracts one, and reports `0`.
2. **Unconditional `saturating_sub(1)`.** If the self-connect fails (backlog saturated, accept thread stalled), no sentinel is counted, yet line 106 still subtracts one, turning a single real leak into a reported `0`.
3. **Ignored 100 ms timeout.** `recv_timeout` errors are discarded; if the accept thread has not processed the sentinel within 100 ms on a loaded runner, the subtraction again under-counts.

This is particularly problematic because the test is explicitly designed to catch "fire-and-forget leaks whose transport error is discarded" (`icaptcha_mock_consumed.rs:16-17`). Those leaks are exactly the kind that can still be handshaking when `obtain_proof` returns, so the flush race undermines the guard's purpose.

**Fix direction:** Make `flush()` wait until the count is stable (e.g., repeatedly flush and verify the non-flush count does not change across a bounded interval), or close the listener and drain the backlog before reading the final count. Also subtract the sentinel only when its ack is actually received.

---

### [P3] Blackhole tripwire is not portable to macOS

**Severity:** Medium  
**File:** `crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs:69-70`

The tripwire binds a listener on `127.0.0.2`. The file's own comment notes that macOS may need an explicit loopback alias, but the test uses `.expect(...)` rather than skipping or failing fast with a helpful message. On a stock macOS runner the test panics before it exercises the blackhole.

Worse, if the bind succeeds but the address is unreachable (or if a future runner configuration changes), the assertion only checks `reqwest::Error::is_connect()` (`icaptcha_blackhole_tripwire.rs:97-102`). A direct connect failure to an unreachable `127.0.0.2` is indistinguishable from a proxy-blocked connect failure, so the test can pass for the wrong reason.

**Fix direction:** Gate the test to Linux, mark it `#[ignore = "requires 127.0.0.2 loopback alias"]`, or add a reachability/positive-control check that verifies `127.0.0.2` is reachable before asserting the blackhole behavior.

---

### [P4] Tripwire lacks a disarmed-blackhole positive control

**Severity:** Medium  
**File:** `crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs:66-103`

The tripwire asserts that `obtain_proof` fails when the blackhole is armed, but it never asserts the opposite: that the same request succeeds when the blackhole is disarmed. Without that positive control, a platform where `127.0.0.2` is unreachable (see P3) or a future change that makes `obtain_proof` fail for unrelated reasons will keep the test green while the guard is not actually being exercised.

**Fix direction:** Add a companion test (or an initial phase in the same test) that clears proxy variables, verifies `127.0.0.2` is reachable, and asserts `obtain_proof` returns `Ok("PROOF-TRIP")`.

---

### [P5] `arm_blackhole` mutates process-global env vars without restoration

**Severity:** Low–Medium  
**File:** `crates/icaptcha-client/tests/support/mod.rs:3-15`

The helper overwrites eight proxy-related environment variables and never restores them. This is safe today because each integration-test file contains a single `#[test]` and Cargo compiles each file as a separate binary. It is fragile, however: adding a second test to either file would leak the mutated env into that test, and any child process or library code that inspects proxy vars inside the test will see the blackhole values.

**Fix direction:** Convert `arm_blackhole` into a drop-guard type that captures prior values (or `None` if absent) and restores them on drop, mirroring `InsecureEnv` in `crates/icaptcha-client/src/lib.rs:392-413`.

---

### [P6] Tripwire fixture reads full buffer instead of bytes returned

**Severity:** Low  
**File:** `crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs:47-51`

```rust
let mut buf = [0; 4096];
if stream.read(&mut buf).is_err() { continue; }
let request = String::from_utf8_lossy(&buf);
```

The code discards the byte count returned by `read` and converts the entire 4096-byte buffer, including trailing NULs. `contains("/v1/answer")` still works because NUL is valid UTF-8, but the server is technically parsing uninitialized padding. This is brittle if the fixture ever needs to inspect headers or body length.

**Fix direction:** Use the count returned by `read`:

```rust
let n = stream.read(&mut buf).unwrap_or(0);
let request = String::from_utf8_lossy(&buf[..n]);
```

---

### [P7] Branch is one commit behind `main`

**Severity:** Process / Low  
**Evidence:** `git log --oneline HEAD..main` shows `111cff7 infra: add production fly configs for gitlawb-node-2 (sjc) and gitlawb-node-3 (nrt)`.

GitHub reports the PR as mergeable, so this is branch-state drift rather than a code conflict. Rebasing onto the current `main` before merge keeps the final change reviewed against the latest base.

---

jatmn (changes requested):

## Findings

- [P2] Parse the complete HTTP request before selecting the fixture response
  `crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs:51`
  A TCP read is not message-framed, so it can return before the complete request line (including `/v1/answer`) arrives. The fixture then treats an answer request as a challenge request and returns challenge JSON; the direct positive control fails to deserialize `AnswerResult` even though the proxy behavior is correct. Read through the HTTP headers/request line before dispatching, or use an HTTP test server, so normal TCP segmentation cannot make this tripwire intermittently fail.

jatmn (changes requested):
I found one low-priority test-coverage issue.

## Findings

- [P3] Narrow the observer's fire-and-forget guarantee
  `crates/icaptcha-client/tests/icaptcha_mock_consumed.rs:82`
  `stabilise` declares success after the first 100 ms interval in which the detached accept loop has not incremented `count`. A regression can spawn a fire-and-forget proxy request on the exercised path just before `obtain_proof` returns, but schedule its connection after that interval; both samples are zero, the test passes, and the leaked DID/answer/API-key request is never observed. Conversely, a repeating leak that keeps changing the counter never lets this unbounded loop return. That makes the new test non-load-bearing for the discarded-error/background egress it says it detects. Use a bounded acknowledgement/lifecycle barrier for the request work being tested (or narrow the assertion to connections already accepted), rather than treating one quiet polling interval as proof that the flow has no outstanding egress.

jatmn (changes requested):

## Findings

- [P2] Do not let the blackhole tripwire report success when it never ran
  `crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs:96`
  When `TcpListener::bind("127.0.0.2:0")` fails (the documented macOS / missing-loopback-alias case), the test prints `SKIP` and returns. Cargo still reports the case as **passed**, so a platform that cannot bind `127.0.0.2` gets zero coverage of the negative control while CI stays green. That recreates the original gap this tripwire was added to close: a future `.no_proxy()` or connector bypass can ship undetected anywhere the test silently no-ops. Please fail fast with a clear message, gate the test with `#[cfg(...)]` / `#[ignore = "..."]`, or otherwise make skipped platforms visible instead of counting them as success.

- [P2] Synchronize the proxy observer before asserting zero egress
  `crates/icaptcha-client/tests/icaptcha_mock_consumed.rs:124`
  The latest head removes `stabilise()` and reads the counter immediately after `obtain_proof` returns, while the accept loop still runs on a detached thread. A fire-and-forget leak whose TCP handshake is still in the listener backlog—or completes just after the synchronous call returns—can therefore pass with `leak_count == 0`, which is exactly the class #211 asked this guard to catch. The module comment now documents a post-return window, but that narrows the guarantee rather than replacing the lifecycle barrier previously requested. Please add a bounded drain/acknowledgement (or join/barrier with the accept thread) before the zero-count assertion, or narrow the test name/docs to match what is actually proved.

- [P3] Document the single-test-per-binary constraint on env helpers
  `crates/icaptcha-client/tests/support/mod.rs:48`
  `arm_blackhole` and `disarm_proxy_env` mutate process-wide proxy variables and restore them on drop. Each consuming integration test binary currently has exactly one `#[test]`, so this is safe today, but `cargo test` runs multiple tests in the same binary concurrently by default. A future second test in either file would race on these vars without `serial_test` or an explicit comment on the helpers. Please document that constraint on the public helpers (or add the project's serial-test guard if one exists).

- [P3] Retitle to `test(icaptcha): ...`
  PR title / lead commit
  The diff is test-only (`tests/` plus lockfile sync; no `src/` change). This repo uses Conventional Commits prefixes for release-please versioning, and prior test-only icaptcha work here uses `test(...)`. `fix(...)` can trigger an unnecessary product patch release for harness-only changes.

beardthelion (dismissed):
Re-reviewed b9fff1a. The happy path is sound on this Linux runner: tripwire positive returns PROOF-TRIP, armed negative fails with a connect error, and injecting a sync discarded-error leak through the armed proxy flips leak_count to 1. Gutting arm_blackhole on the negative control leaves Ok("PROOF-TRIP") and the tripwire fails. jatmn's round on this same head is still open; I verified the same holes.

## Findings

- [P2] Fail or ignore when 127.0.0.2 cannot bind
  `crates/icaptcha-client/tests/icaptcha_blackhole_tripwire.rs:99`
  The Err arm prints SKIP and returns. Cargo reports the test passed. I forced that path and got 1 passed; 0 ignored. This tripwire is the only check that the blackhole still intercepts a non-NO_PROXY destination; mock_consumed stays green under .no_proxy() because the mock is loopback. Prefer .expect(...) with the macOS alias hint, or #[ignore = "..."] / cfg so skipped platforms show as ignored, never passed.

- [P2] Match the observer claim to the assertion window
  `crates/icaptcha-client/tests/icaptcha_mock_consumed.rs:133`
  Sync discarded-error leaks are caught (injected get(...).send() with discarded Result went RED). The accept loop is still detached and the zero-count read is immediate, so a handshake that completes after return is outside the window. The module doc already says that; the test name still reads like full fire-and-forget coverage. Add a bounded drain/barrier before asserting zero, or narrow the name/docs to "no proxy accepts by the time obtain_proof returned."

- [P3] Document the single-test-per-binary rule on the env helpers
  `crates/icaptcha-client/tests/support/mod.rs:48`
  arm_blackhole / disarm_proxy_env mutate process-wide proxy vars. Each consuming binary has one #[test] today, so it is safe, but a second test would race under default cargo concurrency. The crate already serializes GITLAWB_ICAPTCHA_INSECURE via ICAPTCHA_ENV_LOCK in src/lib.rs. Doc the constraint on the helpers, or take the same Mutex / serial_test pattern.

- [P3] Retitle to test(icaptcha): ...
  PR title
  Diff is tests plus lock sync, no src/ change. fix: can cut an unnecessary release-please patch. Prefer test(icaptcha): ....

One process note, not a finding: base is 31 commits behind main. Rebase before the next push; the lockfile 0.5.1 to 0.7.0 churn should collapse onto current main, and the PR body still claims mockito was added when it is already there.

beardthelion (dismissed):
Re-reviewed at `59378ea4`. The four asks from b9fff1a landed: soft-skip is gone (`[::1]` + `.expect`), the observer drains before the zero assert, env-helper concurrency is documented, and the title is already `test(icaptcha):`. Gutting the negative arm still returns `Ok("PROOF-TRIP")`, and a sync discarded leak still flips `leak_count` to 1. The drain accounting still has a hole those checks do not catch.

## Findings

- [P2] Acknowledge the flush connection itself, not the next counter tick
  `crates/icaptcha-client/tests/icaptcha_mock_consumed.rs:106`
  `drain` samples `target = count + 1`, then waits until `count >= target`. A leak accepted after that sample and before the flush is counted can satisfy the wait; the test then does `count - before - 1` and treats the leak as the flush, so `leak_count` can be 0 with a real leak present. I reproduced that false green with a one-permit accept loop (leak counted, flush held, accounting returns 0). Wait until the flush connection is the one that was counted (or otherwise tag it) before subtracting.

Not an ask, recorded only: `[::1]` + `.expect` stays the right fail-loud shape for missing IPv6 loopback. Do not bring back a soft SKIP. Port `127.0.0.1:1` as the armed proxy is fine on current CI; an owned reset-proxy is optional hardening, not required for this round.

beardthelion (approved):
Re-reviewed at `dc6b19dd`. The flush-identity drain from the last round is load-bearing: under a one-permit accept loop, a leak accepted before the flush leaves drain waiting, and after the flush is accepted `count()` is 1. Gutting `arm_blackhole` on the tripwire still returns `Ok("PROOF-TRIP")`. A sync discarded proxy leak still flips `leak_count` to 1. Neutering the flush-port register makes the happy-path zero assert fail on the flush itself. Soft-skip is still gone (`[::1]` + `.expect`). Prior ask resolved.

Not an ask, recorded only: a leak whose handshake starts after drain is outside the window (module docs already say so). A delayed post-return spawn stays green by that design; ephemeral source-port collision under loopback TIME_WAIT reuse is the same class and did not reproduce in 64 cycles here. Do not chase another apparatus hop for that residual.

@kevincodex1 ready to merge.

jatmn (approved):
LGTM

kevincodex1 (approved):
lgtm

Issue #185, opened 2026-07-11
release-please leaves Cargo.lock stale: committed lockfile drifts from the manifests each release


beardthelion:
## Summary

release-please bumps the crate versions in every `Cargo.toml` but never updates `Cargo.lock`, so the committed lockfile drifts further from the manifests with each release. It is harmless today because nothing in the build/release pipeline uses `--locked`, but it turns into a hard build failure the moment anything does.

## Current state (main @ ad7c2b2)

- `.release-please-manifest.json` and every crate `Cargo.toml`: `0.5.1`
- `Cargo.lock` workspace entries: `0.5.0`

The pending 0.6.0 release (#182) widens the gap to `0.6.0` in the manifests vs `0.5.0` in the lock.

## Why it is harmless right now

Every cargo build in CI and the release pipeline runs unlocked, so cargo regenerates the lock in place and the build succeeds:

- `pr-checks.yml`: `cargo test --workspace`, `cargo build --release --workspace`
- `release.yml` binaries job: `cargo build --release --target <t> -p ...`
- `Dockerfile` (release image) and `Dockerfile.bins`: `cargo build --release ...`

The only `--locked` in the repo is `cargo install --locked cargo-audit`, which installs the audit tool and has nothing to do with the workspace lock.

## Why it is worth fixing

It flips from cosmetic to a hard failure (exit 101) the instant any build adds `--locked` (a common reproducible-build hardening step), or on a `cargo install --locked` / downstream vendor step. Reproduced against the current main state in a throwaway worktree:

```
$ cargo metadata --locked
error: cannot update the lock file ... because --locked was passed to prevent this
$ echo $?
101
```

Unlocked, the same command exits 0 and rewrites the lock on disk, which is why the drift never gets committed back.

## Cause

`release-please-config.json` uses `release-type: "simple"` with an `extra-files` list that covers the five `Cargo.toml` files but not `Cargo.lock`, so the lock is never part of the version bump.

## Direction (not yet verified against this repo's config)

Either switch the package to release-type `rust` (its strategy maintains the workspace lockfile), or keep `simple` and add a post-generation step that regenerates the lock (`cargo update -p <crates>` or equivalent) and commits it in the release PR. Both need confirming against our setup before wiring in.

beardthelion:
Superseded by #242, which covers the same drift with current evidence and has the fix open in #243 (lock-sync job on the release branch plus --locked builds in PR CI). Closing this one in its favor.

PR #215, opened 2026-07-17
fix(repo_store): use SHA-256 for stable advisory-lock key (#210)

### LINKED ISSUE #210: Advisory-lock key uses DefaultHasher — cross-machine write exclusion can silently break across node builds
## Summary

The per-repo Postgres advisory-lock key is computed with `std::collections::hash_map::DefaultHasher`, whose output is not guaranteed stable across Rust releases. Two nodes built with different toolchains can therefore compute different `i64` keys for the same `(owner_slug, repo_name)`, and `pg_advisory_lock` only mutually excludes callers using the *same* key. On a shared-Postgres deployment this silently voids the cross-machine write exclusion the lock is meant to provide.

## Where

`crates/gitlawb-node/src/git/repo_store.rs`, `advisory_lock_key`:

```rust
fn advisory_lock_key(owner_slug: &str, repo_name: &str) -> i64 {
    let mut hasher = std::collections::hash_map::DefaultHasher::new();
    owner_slug.hash(&mut hasher);
    repo_name.hash(&mut hasher);
    hasher.finish() as i64
}
```

The `acquire_write` doc comment claims this lock "prevents concurrent writes to the same repo across machines," and it is the only cross-machine write-exclusion primitive.

## Why it's a hazard

- **Cross-version key mismatch (primary).** `DefaultHasher::new()` is `SipHasher13` with fixed keys — deterministic within one build, but the std docs explicitly reserve the algorithm to change across releases ("its hashes should not be relied upon over releases"), and the `derive(Hash)`/`str::hash` byte feed is likewise not guaranteed stable. On a decentralized network where operators build and upgrade nodes independently, two nodes on different rustc versions can hash the same repo to different keys, so `pg_advisory_lock` stops excluding them — a silent no-op, no error surfaced.
- **Collision (secondary, low probability).** `finish() as i64` folds 2^64 hash outputs into 2^63 keys; the birthday bound (~N²/2^63) is negligible per deployment but non-zero, and a collision produces phantom contention (one repo's lock blocks an unrelated repo) or wrong-repo exclusion.

## Reachability

Real in the **shared-Postgres** topology, which the project supports: `infra/aws/compose.yaml.tftpl` lets multiple nodes point at one external DB (`db_host`), and `purge-spam` is a separate admin process (`main.rs` `run_admin_command`) that connects its own pool to the same `config.database_url` and competes with the live node's `acquire_write` on the same advisory-lock namespace via `try_lock_repo`. That cross-process exclusion is the lock's designed purpose (`admin.rs`: "a concurrent receive-pack cannot land a ref in the window"). A version mismatch or collision silently breaks it.

## Impact

Loss of cross-machine write exclusion means two nodes can apply `git-receive-pack` (or merge/issue writes, which share `acquire_write`) to the same logical repo concurrently with no serialization — a corruption / lost-push path — and purge-spam's lock-vs-live-push protection can skip or block the wrong repo.

## Related / broader (should be addressed in the same fix)

Under the **federated per-node-DB** topology (each node its own Postgres, e.g. Fly with per-node `DATABASE_URL`), advisory locks never cross Postgres instances at all, so the "across machines" comment is architecturally false there *regardless of the hash*. The fix should both (a) use a stable hash and (b) scope the doc comment to the deployment model where cross-machine exclusion actually holds.

## Suggested fix direction

- Derive the `i64` key with a fixed, version-stable algorithm instead of `DefaultHasher`. `sha2` is already a workspace dependency (used in `crates/gitlawb-core/src/cid.rs` and `crates/gitlawb-node/src/{encrypted_pin,webhooks}.rs`), so the simplest drop-in is SHA-256 over `owner_slug` + a separator byte + `repo_name`, taking the first 8 bytes of the digest as the key (`i64::from_le_bytes(digest[..8].try_into().unwrap())`). That is identical across builds, machines, and rustc versions. (Avoid `std::hash::SipHasher` — it is deprecated; a fixed-key SipHash would pull in the external `siphasher` crate for no benefit over the `sha2` already vendored.)
- Add a golden-value unit test pinning the key for a known `(owner_slug, repo_name)` so an accidental algorithm change fails CI.
- Correct/scope the `acquire_write` doc comment to state the cross-machine guarantee holds only on a shared Postgres.
- **Rollout:** changing the algorithm re-keys every node. During a mixed-version upgrade window, an in-flight lock held under the old key will not exclude a new-key caller, so this needs a staged rollout (or an accepted brief window), not a hot swap.

## Context

Surfaced during review of #196 (which hardened the per-repo advisory-lock protocol: Drop-safe guards, upload serialization, purge lock coverage). Fixing the key algorithm was deferred out of that PR because it changes every node's lock key and needs the rollout thought above. Tracking here; not a code change in #196.

Gravirei:
## Summary

Replace `DefaultHasher` with SHA-256 for the Postgres advisory-lock key so the same `(owner_slug, repo_name)` produces the same `i64` across all Rust toolchain versions, OSes, and machines — fixing silent cross-machine write exclusion breakage.

## Motivation & context

Closes #210

## Kind of change

- [x] Bug fix

## What changed

- **gitlawb-node:** `advisory_lock_key()` in `repo_store.rs` now uses `sha2::Sha256` instead of `std::collections::hash_map::DefaultHasher` (not guaranteed stable across Rust releases).
- **gitlawb-node:** `acquire_write` doc comment scoped to clarify cross-machine guarantee holds only on shared Postgres.
- **gitlawb-node:** Golden-value unit test pins the key for a known input so CI catches accidental algorithm changes.

## How a reviewer can verify

```sh
cargo test -p gitlawb-node -- git::repo_store::tests::advisory_lock_key_is_stable
```

## Before you request review

- [x] `cargo test --workspace` passes locally (relevant tests pass)

## Protocol & signing impact

N/A — internal implementation detail, no protocol change.

## Notes for reviewers

Changing the algorithm re-keys every node. During a mixed-version upgrade window, an in-flight lock held under the old key will not exclude a new-key caller, so this needs a staged rollout (or an accepted brief window), not a hot swap.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-machine write-lock consistency by using deterministic lock keys across toolchains and machines.
  * Added validation to ensure lock keys vary correctly by owner and repository.

* **Documentation**
  * Added rolling-upgrade guidance for version 0.7.0, including recommendations to drain writes or perform a single-node cutover.

### DIFF
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -5691,13 +5691,15 @@ mod tests {
 
         // Reproduce acquire_write's session-level advisory-lock key exactly so the
         // second-connection lock collides with the handler's pg_try_advisory_lock
-        // (repo_store.rs: advisory_lock_key over owner_slug then repo_name).
+        // (repo_store.rs: SHA-256 advisory_lock_key over owner_slug, ':', repo_name).
         fn advisory_lock_key(owner_slug: &str, repo_name: &str) -> i64 {
-            use std::hash::{Hash, Hasher};
-            let mut hasher = std::collections::hash_map::DefaultHasher::new();
-            owner_slug.hash(&mut hasher);
-            repo_name.hash(&mut hasher);
-            hasher.finish() as i64
+            use sha2::Digest;
+            let mut hasher = sha2::Sha256::new();
+            hasher.update(owner_slug.as_bytes());
+            hasher.update(b":");
+            hasher.update(repo_name.as_bytes());
+            let digest = hasher.finalize();
+            i64::from_le_bytes(digest[..8].try_into().expect("sha256 output is >= 8 bytes"))
         }
 
         let owner = "z6acqdead";
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -5683,24 +5683,12 @@ mod tests {
     /// stays pinned past the deadline. Restore to return GREEN.
     #[sqlx::test]
     async fn receive_pack_acquire_deadline_sheds_and_releases_permit(pool: sqlx::PgPool) {
+        use crate::git::repo_store::advisory_lock_key;
         use axum::extract::{Path, State};
         use axum::Extension;
         use std::net::SocketAddr;
         use std::sync::Arc;
         use tokio::sync::Semaphore;
-
-        // Reproduce acquire_write's session-level advisory-lock key exactly so the
-        // second-connection lock collides with the handler's pg_try_advisory_lock
-        // (repo_store.rs: SHA-256 advisory_lock_key over owner_slug, ':', repo_name).
-        fn advisory_lock_key(owner_slug: &str, repo_name: &str) -> i64 {
-            use sha2::Digest;
-            let mut hasher = sha2::Sha256::new();
-            hasher.update(owner_slug.as_bytes());
-            hasher.update(b":");
-            hasher.update(repo_name.as_bytes());
-            let digest = hasher.finalize();
-            i64::from_le_bytes(digest[..8].try_into().expect("sha256 output is >= 8 bytes"))
-        }
 
         let owner = "z6acqdead";
         let name = "acq1";
--- a/crates/gitlawb-node/src/git/repo_store.rs
+++ b/crates/gitlawb-node/src/git/repo_store.rs
@@ -152,7 +152,34 @@ impl RepoStore {
     }
 
     /// Take a write lock (Postgres advisory lock), ensure repo is local, return guard.
-    /// The lock prevents concurrent writes to the same repo across machines.
+    ///
+    /// # Cross-machine guarantee
+    ///
+    /// When multiple nodes share a single Postgres database (the shared-Postgres
+    /// deployment model), this lock prevents concurrent writes to the same repo
+    /// across machines. The lock has no effect across separate Postgres instances
+    /// (federated per-node-DB topology).
+    ///
+    /// # Rolling upgrade caveat (SHA-256 re-keying)
+    ///
+    /// This function hashes `(owner_slug, repo_name)` with SHA-256 to produce a
+    /// stable `i64` key. Earlier builds used `std::collections::DefaultHasher`,
+    /// whose algorithm is not frozen by the Rust standard and has already
+    /// shifted across toolchain versions. The SHA-256 swap re-keys *every*
+    /// repo: an old-binary node holding the legacy `DefaultHasher` key and a
+    /// new-binary node holding the SHA-256 key for the same repo compute
+    /// *different* i64 keys, so PostgreSQL treats them as independent locks and
+    /// the cross-machine write-exclusion this lock provides is lost for the
+    /// duration of a rolling upgrade.
+    ///
+    /// The accepted remediation (see issue #210) is operational: during a
+    /// shared-Postgres rolling upgrade, **drain in-flight writes or cut over
+    /// through a single node** (e.g. stop receive-pack / issue / pull / archive
+    /// writers on the old version) before bringing new-binary nodes online. The
+    /// window is bounded by the operator's rollout cadence. A future
+    /// transition release could acquire both legacy and new keys for one cycle
+    /// and drop the legacy one a release later; that's optional given the
+    /// accepted-window path.
     pub async fn acquire_write(&self, owner_did: &str, repo_name: &str) -> Result<RepoWriteGuard> {
         let (owner_slug, local_path) = self.local_path(owner_did, repo_name)?;
         let lock_key = advisory_lock_key(&owner_slug, repo_name);
@@ -518,12 +545,19 @@ impl RepoWriteGuard {
 }
 
 /// Compute a stable i64 hash for a Postgres advisory lock key.
+///
+/// Uses SHA-256 (not `DefaultHasher`) so the same `(owner_slug, repo_name)`
+/// produces the same `i64` key across every Rust toolchain version, operating
+/// system, and machine — the algorithm is frozen by the SHA-2 standard rather
+/// than by a std-internal implementation detail.
 fn advisory_lock_key(owner_slug: &str, repo_name: &str) -> i64 {
-    use std::hash::{Hash, Hasher};
-    let mut hasher = std::collections::hash_map::DefaultHasher::new();
-    owner_slug.hash(&mut hasher);
-    repo_name.hash(&mut hasher);
-    hasher.finish() as i64
+    use sha2::Digest;
+    let mut hasher = sha2::Sha256::new();
+    hasher.update(owner_slug.as_bytes());
+    hasher.update(b":");
+    hasher.update(repo_name.as_bytes());
+    let digest = hasher.finalize();
+    i64::from_le_bytes(digest[..8].try_into().expect("sha256 output is >= 8 bytes"))
 }
 
 #[cfg(test)]
@@ -931,5 +965,47 @@ mod tests {
                 "owner_did={bad:?} must be rejected"
             );
         }
+    }
+
+    // ── advisory_lock_key stability ─────────────────────────────────────────
+
+    #[test]
+    fn advisory_lock_key_is_stable() {
+        // Golden value: SHA-256("did_key_...:<repo_name>")[..8] as i64 little-endian.
+        // If this test fails, the hashing algorithm has changed — the new key
+        // must be backward-compatible or the rollout planned accordingly.
+        let key = advisory_lock_key(
+            "did_key_z6MkqDnb7Siv3Cwj7pGJq4T5EsUisECqR8KpnDLwcaZq5TPr",
+            "hello",
+        );
+        assert_eq!(key, -6680856138670956537_i64);
+    }
+
+    #[test]
+    fn advisory_lock_key_differs_for_different_inputs() {
+        // Vary one axis at a time so a regression that drops either parameter
+        // from the hash is caught, not just one that drops both. The golden
+        // test above backstops a total algorithm swap.
+        let base = advisory_lock_key("owner_a", "repo_a");
+
+        // Same owner, different repo: a regression that hashes only owner_slug
+        // would make these collide.
+        let same_owner_diff_repo = advisory_lock_key("owner_a", "repo_b");
+        assert_ne!(
+            base, same_owner_diff_repo,
+            "key must depend on repo_name, not just owner_slug"
+        );
+
+        // Same repo, different owner: a regression that hashes only repo_name
+        // would make these collide.
+        let diff_owner_same_repo = advisory_lock_key("owner_b", "repo_a");
+        assert_ne!(
+            base, diff_owner_same_repo,
+            "key must depend on owner_slug, not just repo_name"
+        );
+
+        // Sanity: both axes varying at once still differs (the original shape).
+        let both_differ = advisory_lock_key("owner_b", "repo_b");
+        assert_ne!(base, both_differ);
     }
 }
--- a/crates/gitlawb-node/src/git/repo_store.rs
+++ b/crates/gitlawb-node/src/git/repo_store.rs
@@ -762,7 +762,16 @@ impl Drop for RepoWriteGuard {
 /// produces the same `i64` key across every Rust toolchain version, operating
 /// system, and machine — the algorithm is frozen by the SHA-2 standard rather
 /// than by a std-internal implementation detail.
+///
+/// Domain separation is `owner_slug + ":" + repo_name` with no length prefix,
+/// so the mapping is injective only while `owner_slug` contains no `:` (the
+/// `did:key:`→`did_key_` slug form `local_path` produces). A raw DID would
+/// collide: `("did:key:abc", "x")` and `("did", "key:abc:x")` hash the same.
 pub(crate) fn advisory_lock_key(owner_slug: &str, repo_name: &str) -> i64 {
+    debug_assert!(
+        !owner_slug.contains(':'),
+        "advisory_lock_key owner_slug must not contain ':' (domain-separation guarantee)"
+    );
     use sha2::Digest;
     let mut hasher = sha2::Sha256::new();
     hasher.update(owner_slug.as_bytes());
--- a/crates/gitlawb-node/src/git/repo_store.rs
+++ b/crates/gitlawb-node/src/git/repo_store.rs
@@ -762,7 +762,7 @@ impl Drop for RepoWriteGuard {
 /// produces the same `i64` key across every Rust toolchain version, operating
 /// system, and machine — the algorithm is frozen by the SHA-2 standard rather
 /// than by a std-internal implementation detail.
-fn advisory_lock_key(owner_slug: &str, repo_name: &str) -> i64 {
+pub(crate) fn advisory_lock_key(owner_slug: &str, repo_name: &str) -> i64 {
     use sha2::Digest;
     let mut hasher = sha2::Sha256::new();
     hasher.update(owner_slug.as_bytes());
@@ -771,7 +771,6 @@ fn advisory_lock_key(owner_slug: &str, repo_name: &str) -> i64 {
     let digest = hasher.finalize();
     i64::from_le_bytes(digest[..8].try_into().expect("sha256 output is >= 8 bytes"))
 }
-
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/docs/RUN-A-NODE.md
+++ b/docs/RUN-A-NODE.md
@@ -178,6 +178,7 @@ GITLAWB_ENFORCE_OWNER_PUSH=true
 | Operator key | Dedicated wallet, small ETH balance, not your main treasury |
 | Monitoring | Watch `lastHeartbeat` on-chain; alert if > 22h since last beat |
 | Public URL | Must resolve and serve `/health` — peers will ping it |
+| Rolling upgrade (advisory-lock algorithm) | The SHA-256 advisory-lock key swap (replacing `DefaultHasher`) re-keys every repo in the first release that carries it. During a shared-Postgres rolling upgrade, old nodes compute a different `i64` key than new nodes for the same repo, so PostgreSQL treats them as independent locks and cross-machine write-exclusion is lost. **Drain in-flight writes** or cut over through a **single node** before bringing new-binary nodes online. |
 
 ---
 


beardthelion (dismissed):
The fix is correct and the golden test is load-bearing. I verified the pinned value is the real SHA-256 (recomputed `SHA-256("did_key_...:hello")[..8]` as LE i64 == -6680856138670956537, and big-endian would differ, so byte order is pinned), then ran a mutation matrix at the head: reverting to DefaultHasher, flipping the separator, flipping to from_be_bytes, and dropping owner_slug from the hash each turn the golden test RED, and it's GREEN restored. The separator is injective because owner_slug is colon-sanitized upstream (`owner_did.replace([':','/'], "_")`), so there's no wrong-repo aliasing, and the byte-slice path can't panic. This closes #210 the way it asked.

One thing to resolve before it lands, plus a small test note.

## Findings

- [P2] Persist the re-keying rollout caveat where operators will see it after merge
  `crates/gitlawb-node/src/git/repo_store.rs:407`
  Switching the algorithm re-keys every node. On a shared-Postgres, multi-node deployment doing a rolling upgrade, an old-binary node (DefaultHasher key) and a new-binary node (SHA-256 key) compute different keys for the same repo and stop mutually excluding for the deploy window, a transient loss of the cross-machine write-exclusion this lock exists to provide (concurrent same-repo pushes could race). That topology is real (the shared-DB compose target, plus purge-spam as a separate process on the same lock namespace). Right now the caveat lives only in the PR body, which disappears on merge. I filed #210 with "a staged rollout OR an accepted brief window" as acceptable, so I'm not asking you to change the swap itself, only to make the guidance durable: a short note on `acquire_write` (drain writes or cut over single-node during the upgrade) plus a CHANGELOG line. If you'd rather remove the window entirely, the mechanism is a transition release where `acquire_write` takes both the legacy and the new key, dropping the legacy one a release later, but that's optional given the accepted-window path.

- [P3] Strengthen or drop `advisory_lock_key_differs_for_different_inputs`
  `crates/gitlawb-node/src/git/repo_store.rs:594`
  It varies both inputs at once (`owner_a`/`repo_a` vs `owner_b`/`repo_b`), so it can't isolate a regression that drops one parameter, and I confirmed that by mutation: hashing only repo_name leaves this test green while the golden test goes red. So the golden test already backstops that case and this one adds little. If you want it to pull weight, assert a same-owner/different-repo pair and a different-owner/same-repo pair; otherwise it's fine to drop it and lean on the golden test.

Not a change request: the Cargo.lock version bumps (0.5.0 -> 0.5.1 across five crates, plus a sha2 line under icaptcha-client) aren't from this PR. The manifests are already 0.5.1 on main and gitlawb-node already declares sha2; the lockfile was just stale (that's issue #185), and regenerating it here corrected the drift. Leave them as they are; a one-line mention in the PR description is enough.

Solid, well-scoped fix. Once the rollout note is durable I'm happy to approve.

jatmn (changes requested):

## Findings

- [P1] Retain the lock-owning database session for the write guard
  `crates/gitlawb-node/src/git/repo_store.rs:164`
  This change makes the key deterministic, but `pg_try_advisory_lock` is still executed through `&PgPool` and the guard later unlocks through that pool again. Those are session-scoped PostgreSQL locks: SQLx returns the acquiring connection after `fetch_one`, so `release()` can be assigned another session and silently fail to unlock. Worse, a later writer can be given the idle lock-owning session and reentrantly acquire the same lock. The advertised shared-Postgres exclusion therefore remains unreliable even with the SHA-256 key. Hold a `PoolConnection<Postgres>` in `RepoWriteGuard` from successful acquisition through unlock (including error cleanup), and add a competing-session test. PR #196 already has this exact dedicated-connection repair, so either make this PR depend on/rebase onto it or bring that repair here before merging.

- [P2] Preserve exclusion while old and new lock keys coexist
  `crates/gitlawb-node/src/git/repo_store.rs:407`
  Replacing `DefaultHasher` re-keys every repository. In a rolling shared-Postgres upgrade, an old node can hold the legacy key while a new node acquires the SHA-256 key for the same repository; PostgreSQL treats them as independent locks, allowing concurrent receive-pack/issue/pull writes and competing archive uploads. The PR body acknowledges this window, but it disappears on merge and the checked-in documentation only describes the topology. Acquire both keys for a transition release, or add durable deployment guidance requiring writes to be drained or cut over through a single node before versions are mixed.

jatmn (changes requested):

## Findings

- [P1] Keep advisory-lock sessions out of the shared application pool
  `crates/gitlawb-node/src/git/repo_store.rs:207`
  `acquire_write` now checks out a connection from the same `Db` pool used by every handler and holds it for the full retry loop, Tigris I/O, and write operation. With the default 20-connection pool, one lock holder plus 19 requests contending for that repository consume every connection while the contenders sleep and retry for up to 60 seconds; distinct slow writes can do the same for up to the 600-second git-service timeout. Unrelated database-backed requests then fail their five-second acquire timeout. Use a separate bounded pool for advisory-lock sessions, or otherwise bound lock waiters/write guards while reserving capacity for the application pool.

- [P1] Do not return a session to the pool until its advisory unlock is known to have completed
  `crates/gitlawb-node/src/git/repo_store.rs:469`
  `release` removes the lock-owning connection from `self.conn` before awaiting `pg_advisory_unlock`. If that future is cancelled, the guard's `Drop` sees `None`, so SQLx returns the session to the pool even though PostgreSQL may still hold its session-scoped advisory lock. A later borrower of that session can reentrantly acquire the lock while other sessions remain blocked, defeating the write exclusion. The Tigris-download error cleanup has the same unsafe outcome when its ignored unlock query fails. Keep the connection in the guard through a successful unlock, and close it rather than pool it on cancellation or unlock failure; add a regression test that aborts during unlock.

beardthelion (dismissed):
Re-reviewed at `26c4dfe8`. Both points from my last round landed properly. The rolling-upgrade caveat is now in the `acquire_write` docstring, and the differs-test rewrite pulls real weight: I ran the mutation matrix and dropping either parameter from the hash turns it RED (`key must depend on repo_name` / `key must depend on owner_slug`), where the old both-vary-at-once shape stayed green. The golden constant still catches an algorithm swap. The SHA-256 change itself remains correct.

Holding the lock-owning connection in the guard is the right move and the new `Drop` works: I confirmed that dropping a guard without calling `release()` frees the lock to an independent session, because `close_on_drop` really does end the session in sqlx 0.8.6. What is not right yet is everything that happens once the connection leaves the guard, plus a collision with #79 that will bite whichever of the two merges second.

## Findings

- [P1] Settle one lock-key derivation with #79 before either merges
  `crates/gitlawb-node/src/git/repo_store.rs:496`
  #79 also claims to fix #210, and it hashes `format!("{owner}/{repo}")` folded big-endian while this PR hashes `owner` + `":"` + `repo` folded little-endian. Same repo, different key. Whichever lands second re-keys every repository a second time, breaks the other's golden constant, and makes operators pay the exclusion-gap window twice; #79's legacy fallback is main's `DefaultHasher`, so its dual-key transition would not cover a fleet already running this key either. My call as lead: this PR owns the lock layer, since it is the focused change that closes #210 and it got here first. I am asking #79 to drop its lock work and rebase onto this, so nothing here needs to change except knowing the decision is settled.

- [P1] Stop pinning a pool connection for the whole 60-second acquire wait
  `crates/gitlawb-node/src/git/repo_store.rs:207`
  The connection is now checked out before the retry loop, so a writer that waits out a contended repo holds one of the pool's 20 connections (`config.rs:192`) for up to 60 seconds without doing any work. On the base commit the loop used `fetch_one(&self.pool)` and returned its connection between probes, so this is a regression the fix introduces rather than an existing gap. What makes it more than a tuning question is who can reach it: `create_issue` sits behind `authorize_repo_read` and nothing stronger (`api/issues.rs:46`) before it calls `acquire_write` at `:66`, so any authenticated identity with read access to a repo can occupy a connection for the length of a write. Only the winning attempt needs session affinity, so acquire inside the loop and drop before each sleep; I compiled that change. The other half, every in-flight write pinning a connection for its whole duration, is what jatmn's dedicated-pool ask is really about, and that one still stands.

- [P1] Keep the connection inside the guard across the unlock await, and check the unlock result
  `crates/gitlawb-node/src/git/repo_store.rs:469`
  `self.conn.take()` moves the connection out before the await, which disarms the `Drop` you just added: from that instant the lock is held by a connection nothing will close. Cancellation there, or an unlock that does not actually unlock, returns that connection to the pool with the lock still held. This is worse than a one-off leak because advisory locks are counted per session. I checked it against Postgres directly: two `pg_try_advisory_lock` calls on one session both return true, and after a single `pg_advisory_unlock` the session still holds the lock. So the next writer for that repo, handed the same pooled connection, re-acquires reentrantly and its own `release()` only decrements the count back to one. The lock then never clears until sqlx recycles the connection, which on the 0.8.6 defaults is 10 minutes idle or 30 minutes lifetime, and every writer on every node fails the 60-second retry in the meantime. Note also that `pg_advisory_unlock` reports "you did not hold this" as a `false` return with a WARNING, not an error, so `let _ = ... .execute(...)` cannot see it at either this site or `:267`. Your own error path at `:266` already uses the correct borrow shape, so the two paths just need to agree. Unlocking through `if let Some(ref mut conn) = self.conn`, taking the connection only once the unlock is confirmed, and otherwise warning and leaving it for `Drop` to close, compiles and keeps the existing 20 repo_store tests green.

- [P2] Move the operator caveat out of CHANGELOG.md
  `CHANGELOG.md:3`
  My last round asked for a CHANGELOG line and that was the wrong placement, so this one is on me. That file is generated by release-please (`release-please-config.json`, and every commit touching it is a `chore(main): release`), and main is already at 0.7.0, so a hand-written `## Unreleased` section will sit above the next generated release forever. Keep the docstring, which is the durable half I actually wanted, and put the operator guidance in the operational checklist in `docs/RUN-A-NODE.md`, next to where the signed-peer rolling-upgrade posture is already documented.

- [P3] Drop the Cargo.lock hunk when you rebase
  `Cargo.lock:3303`
  It was legitimate drift correction when you wrote it, but main now carries both halves already: the crates are at 0.5.1 there and `icaptcha-client` already lists `sha2`. After a rebase this hunk is empty churn.

- [P2] Close the same gap in the acquire loop
  `crates/gitlawb-node/src/git/repo_store.rs:216`
  The probe query is the one await where the connection is a bare local with no guard around it. If the future is dropped while that query is in flight, the request may still have reached Postgres and taken the lock, and the connection then returns to the pool holding it, with the same counted-lock consequence as above. If you take the acquire-inside-the-loop change this shrinks to the single winning query, and wrapping that connection so its `Drop` closes the session until the guard takes ownership closes it entirely.

Also still open from jatmn's last round: the two regression tests, a competing-session case proving a second Postgres session is excluded while the guard is held, and one that exercises cancellation during the unlock. The second one is worth writing RED against the current `take()` shape so it proves the fix rather than tracking it.

jatmn (changes requested):
I rechecked this at `30e0dec`. The dedicated `lock_pool`, acquire-inside-the-retry-loop change, and keeping the lock-owning connection across the unlock `.await` all landed correctly since my last round. The SHA-256 key and golden test still look right. What remains is the unlock/return-to-pool contract, lock-pool sizing, and the regression tests reviewers asked for.
## Findings
- [P1] Verify `pg_advisory_unlock` actually released the lock before returning the session to `lock_pool`
  `crates/gitlawb-node/src/git/repo_store.rs:506`
  `release()` still treats a successful `execute()` as proof the session is clean and calls `self.conn.take()`, returning the connection to `lock_pool`. PostgreSQL reports “you did not hold this lock” as a `false` return value, not a query error, so `execute()` cannot distinguish that case. Worse, session-level advisory locks are counted: if this session already held the key (for example because a prior holder returned a lock-bearing connection to the pool and this writer re-acquired reentrantly), a single `pg_advisory_unlock` can return `true` while the session still holds the lock. The next borrower of that pooled connection inherits it, and exclusion can stay broken until sqlx recycles the session. beardthelion called this out on the prior head; it is still present on both the happy path here and the Tigris error cleanup at `:288`. Please unlock through `query_as` into `(bool,)`, treat `false` like a failed unlock (leave the connection for `Drop`/`close_on_drop`), and add a regression test that aborts during unlock so this cannot regress again.
- [P1] Hard-coding `lock_pool` to two connections caps the whole node at two in-flight writes
  `crates/gitlawb-node/src/main.rs:283`
  Each `RepoWriteGuard` now pins a `lock_pool` connection for the full write path (Tigris download, git receive-pack/merge/issue work, Tigris upload, unlock) — up to `git_service_timeout_secs` (600s by default). With `max_connections(2)`, a third concurrent `acquire_write()` anywhere on the node blocks on `lock_pool.acquire()` and, after the 5s `db_acquire_timeout_secs`, fails with `acquiring connection for advisory lock` instead of entering the 60×1s advisory-lock retry loop. That is a new node-wide write denial under modest concurrency, including cross-repo traffic (`repos.rs`, `pulls.rs`, `issues.rs` all call `acquire_write`). Worse, each iteration of the contended-repo retry loop calls `lock_pool.acquire().await?` (`:226`), so a same-repo waiter that is merely waiting on an advisory lock can abort after ~5s when two unrelated writes occupy both slots — the `"could not acquire advisory lock after 60s"` path is unreachable under pool saturation. Please size the pool for expected concurrent writers (config/env), retry pool-acquire failures inside the lock loop, or stop holding a pooled connection for the entire git/Tigris phase once the lock is acquired.
- [P2] The competing-session and cancellation regression tests still do not protect CI
  `crates/gitlawb-node/src/git/repo_store.rs:757`
  beardthelion and I both asked for load-bearing Postgres tests here. The new `competing_session_is_excluded_while_guard_held` and `cancelled_guard_releases_lock_when_dropped` cases are marked `#[ignore]` and `make_store_for_live()` wires `lock_pool` with `connect_lazy("postgres://invalid")` while the `#[sqlx::test]` fixture provisions `pool` against the real test database, so even a manual run cannot pass. The ignore reason (`requires DATABASE_URL`) is also misleading — CI already provisions Postgres in `pr-checks.yml`. Even after fixing the fixture, `cancelled_guard_releases_lock_when_dropped` probes lock availability immediately after `handle.await` while `Drop` only calls `close_on_drop()`, which spawns an async session teardown in sqlx 0.8.6, so the test can pass or fail depending on scheduler timing rather than proving the lock is gone. Please point `lock_pool` at the same database as the test `pool`, drop the `#[ignore]`, wait for the close future (or use synchronous teardown) before probing, and keep the cases in the default `cargo test -p gitlawb-node` path.
- [P2] A cancelled winning lock probe can still poison `lock_pool`
  `crates/gitlawb-node/src/git/repo_store.rs:232`
  Failed probes drop their connection without holding a lock, but the one successful `pg_try_advisory_lock` runs on a bare `PoolConnection` until `RepoWriteGuard` is constructed immediately afterward. If the task is cancelled during that `fetch_one().await` after Postgres took the lock, the connection is returned to `lock_pool` without `close_on_drop`, recreating the counted-lock/reentrancy failure mode the new `Drop` is meant to prevent. Please wrap the winning probe connection in the same “close on drop unless explicitly handed to the guard” RAII used for cancellation safety.
- [P3] `lock_pool` startup uses a single `connect()` with no retry
  `crates/gitlawb-node/src/main.rs:287`
  The main application pool goes through `connect_db_with_retry`, but the advisory-lock pool is created with one `connect().await?` immediately afterward. A transient connection failure or brief `max_connections` pressure after the main pool comes up aborts node startup even though migrations already succeeded. Please reuse the same retry/backoff pattern (or share the established pool with a dedicated advisory-lock semaphore) so startup is not flakier than the rest of the DB bootstrap.
- [P3] Operator connection sizing ignores the two extra `lock_pool` sessions
  `crates/gitlawb-node/src/main.rs:282`
  Each node now opens a second pool to the same `database_url` with `max_connections(2)` on top of `GITLAWB_DB_MAX_CONNECTIONS` (default 20, documented in `.env.example`). Fleet sizing that budgets only the main pool can hit Postgres `max_connections` earlier than expected and contribute to the acquire failures above. Please document the extra headroom in `RUN-A-NODE.md` / `.env.example`, or make the lock-pool size configurable alongside `db_max_connections`.

beardthelion (dismissed):
Re-reviewed at `30e0dec`. The SHA-256 key is still right and its tests still carry weight: I reverted to `DefaultHasher`, flipped the byte order, dropped the `":"` separator, and dropped each parameter in turn, and the golden test went red on all five, green restored. That half closes #210 and I have no findings against it.

The lock-session work around it is a different story, and after four rounds I think the problem is the shape of the PR rather than any one line in it. So this round is a scope call rather than another findings list.

Issue #210 asked for a stable hash. Rounds two through four rebuilt Postgres advisory-lock session affinity on top of it: the guard now owns a `PoolConnection`, there is a `Drop` impl, the probe moved inside the retry loop, and `main.rs` opens a second pool. Every finding since round two, mine and jatmn's both, has been against that rework rather than against the hash, and the two most serious ones on this head were introduced by the fixes for the previous round. A five-line change is at +367/-43, and what is now on the branch is more dangerous than what it replaced.

My call as lead: split it. Keep #215 as the SHA-256 key, its two tests, and the operator note. Take the connection-affinity work to its own PR against a new issue, where it gets a real test fixture and its own review budget. I will file that issue with the reproductions below so the follow-up starts from executable evidence instead of rediscovering it a round at a time. To be clear about what the split costs: the unlock-on-a-different-session problem jatmn opened with in round one is real, and it is pre-existing (`repo_store.rs:388` on `main` unlocks through `&self.pool` while the acquire took a separate checkout). Splitting leaves `main` exactly where it already is on that axis, which is worse than a correct fix and better than this one.

## Findings

- [P1] Split the connection-affinity work out of this PR
  `crates/gitlawb-node/src/git/repo_store.rs:207`
  This is the only ask against #215 itself. Revert to the SHA-256 key change plus `advisory_lock_key_is_stable` and `advisory_lock_key_differs_for_different_inputs`, keep the `acquire_write` docstring and the `RUN-A-NODE.md` row, and drop the `lock_pool`, the `PoolConnection` in the guard, the `Drop` impl and the `main.rs` wiring. The remaining findings below are the evidence for that call and belong to the follow-up PR, not to this one, so please do not fix them here.

- [P1] For the follow-up, not this PR: two lock-pool connections cap the whole node at two concurrent writes
  `crates/gitlawb-node/src/main.rs:283`
  A guard pins a `lock_pool` connection for the entire write, so with `max_connections(2)` the third concurrent write anywhere on the node fails. I ran it against a real Postgres with the pool built exactly as `main.rs` builds it: two guards held on **distinct** repos, and the third `acquire_write` returned `acquiring connection for advisory lock: pool timed out while waiting for an open connection` after 5.0 seconds, where the same call succeeds against an idle pool. Note where it fails, because it makes the retry loop moot: `lock_pool.acquire()` sits inside the loop and propagates with `?`, so a write to a completely uncontended repo dies at one `db_acquire_timeout_secs` on the first iteration and never reaches the 60-second wait the code advertises. `create_issue` reaches `acquire_write` behind `authorize_repo_read` alone, so two slow writes are enough.

- [P1] For the follow-up, not this PR: a cancelled lock probe recycles a lock-bearing session and wedges the repo
  `crates/gitlawb-node/src/git/repo_store.rs:232`
  The winning `pg_try_advisory_lock` runs on a bare `PoolConnection` that nothing closes until `RepoWriteGuard` exists a few lines later. Cancel in that window and Postgres has the lock while the connection goes back to `lock_pool` clean. Driven through the real `acquire_write` with cancellations swept across the measured 187us round trip, 6 of 2000 attempts left the advisory lock held; on a node whose Postgres is a network hop away that window is milliseconds, not microseconds. The follow-on is what makes it a wedge rather than a leak. I reproduced that too: the next writer handed that session re-acquires the same key reentrantly, and because session advisory locks are counted, its own `release()` decrements to one and leaves the lock held. Every writer on every node then fails the 60-second retry until sqlx recycles the session, which on the 0.8.6 defaults is ten minutes idle or thirty minutes lifetime.

- [P2] For the follow-up, not this PR: the two regression tests cannot pass and never run
  `crates/gitlawb-node/src/git/repo_store.rs:749`
  `make_store_for_live` builds `lock_pool` from `connect_lazy("postgres://invalid")` while `acquire_write` now sources its connection from `lock_pool`, so both cases panic in `acquire_write(...).unwrap()` before reaching an assertion. I removed the `#[ignore]` lines and ran them: both fail with `failed to lookup address information`. The ignore reason is misleading in the other direction as well, since `pr-checks.yml:93` runs a plain `cargo test --workspace` and no workflow passes `--ignored`, so they would not run even with the fixture fixed. The good news is that the test logic is sound: I re-ran the competing-session assertion with `lock_pool` pointed at the `#[sqlx::test]` fixture database and it passes and excludes the second session correctly. Build the lock pool from the fixture's own connect options and it becomes load-bearing.

- [P2] For the follow-up, not this PR: `release()` reads a successful query as a successful unlock
  `crates/gitlawb-node/src/git/repo_store.rs:506`
  `pg_advisory_unlock` reports "you did not hold this" as a `false` return with a warning, not as an error, so `execute(...).await` returning `Ok` cannot tell a real release from a no-op, and the connection goes back to the pool either way. I confirmed both halves directly against Postgres: two `pg_try_advisory_lock` calls on one session both return true and a single unlock leaves the session still holding it, and unlocking a key the session never held returns `f` with `WARNING: you don't own a lock of type ExclusiveLock`. Read it through `query_as` into `(bool,)` and treat `false` like a failed unlock, leaving the connection for `Drop`. Same at the Tigris error path, `:288`.

- [P3] Name the release that actually carries the re-key
  `docs/RUN-A-NODE.md:181`
  The row says v0.7.0 re-keys every repo's advisory lock, but `main` is already at 0.7.0, so that release has shipped without this change and an operator reading the row would place the drain at the wrong upgrade. This branch is 36 commits behind, which is why its own manifest still says 0.5.1. Key the warning to the first build carrying the SHA-256 lock, or drop the version and describe it by behavior. Worth fixing in #215 since the row stays.

- [P3] Drop the Cargo.lock hunk when you rebase
  `Cargo.lock:3303`
  Same note as last round. `main` now carries the 0.5.1 crate versions and the `icaptcha-client` `sha2` line already, so after the rebase this hunk is empty churn.

Two smaller notes, neither an ask. The comment at `:250` says the connection drops back to `lock_pool` before the sleep; it is scoped to the loop body, so it is actually held across the sleep and released at the iteration boundary. I checked whether that starves anything and it does not: with a guard held on one repo and a waiter retrying on it, a write to an unrelated repo still succeeded in 591ms. And the branch and current `main` both touch `test_support.rs`, so please rebase onto `main` and let the resolved diff get a look before merge.

The hash change is good work and it has been ready for a while. It should not have to wait on a Postgres connection-lifecycle problem it did not create.

jatmn (changes requested):
I rechecked this at `30e0dec` against live `main` (currently **0.7.0**; this branch is **36 commits behind** and still carries 0.5.1-era manifests). The SHA-256 `advisory_lock_key`, golden test, and per-axis `advisory_lock_key_differs_for_different_inputs` are correct and close the core of #210 — I ran `cargo test -p gitlawb-node advisory_lock_key` and both pass.

The connection-affinity rework in commits 2–4 (`c27b355` through `30e0dec`) is what blocks merge as-is: every lifecycle finding below is **PR-introduced** by that rework, verified in the current head, not reviewer-comment drift. Splitting leaves `main`'s pre-existing unlock-on-a-different-session behavior unchanged (`release()` on base still unlocks through `&self.pool` while acquire used a separate checkout) — worse than a correct fix, better than shipping this bundle.

## Findings

- [P1] Split the connection-affinity work out of this PR
  `crates/gitlawb-node/src/git/repo_store.rs:208`
  Issue #210 asked for a stable hash. Commit `04cb922` is sufficient: SHA-256 `advisory_lock_key`, `advisory_lock_key_is_stable`, `advisory_lock_key_differs_for_different_inputs`, the `acquire_write` rolling-upgrade docstring, and the `RUN-A-NODE.md` operator row (with the version fix below). Revert commits 2–4: the `lock_pool` field, dedicated `PoolConnection` in `RepoWriteGuard`, `Drop`/`close_on_drop`, acquire-inside-the-retry-loop restructuring, and the `main.rs` second pool. The findings below are verified evidence for that split; they belong in a follow-up PR with its own fixture and review budget — please do not fix them here.

- [P1] Hard-coding `lock_pool` to two connections caps the whole node at two in-flight writes *(connection-affinity follow-up, not narrowed #215)*
  `crates/gitlawb-node/src/main.rs:283`
  Introduced by this PR; base probed `&self.pool` (default 20 connections) without pinning a session. Each `RepoWriteGuard` now pins a `lock_pool` connection from lock acquisition through Tigris download, git receive-pack/merge/issue work, Tigris upload, and unlock — up to `git_service_timeout_secs` (600s by default). With `max_connections(2)`, a third concurrent `acquire_write()` anywhere on the node blocks on `lock_pool.acquire()` and, after `db_acquire_timeout_secs` (5s by default), fails with `acquiring connection for advisory lock` instead of entering the 60×1s advisory-lock retry loop. That affects cross-repo traffic (`repos.rs`, `pulls.rs`, `issues.rs` all call `acquire_write`). Worse, `lock_pool.acquire().await?` at `repo_store.rs:226` propagates pool-timeout on the first iteration, so a same-repo waiter merely waiting on advisory-lock contention can abort after ~5s when two unrelated writes occupy both slots — the `"could not acquire advisory lock after 60s"` path is unreachable under pool saturation.

- [P1] Verify `pg_advisory_unlock` actually released the lock before returning the session to `lock_pool` *(connection-affinity follow-up, not narrowed #215)*
  `crates/gitlawb-node/src/git/repo_store.rs:506`
  Base already used `execute()` without reading the boolean, but on `&self.pool` — a different failure shape. This PR adds returning the session to `lock_pool` after `Ok(execute())`, which is unsafe: PostgreSQL reports "you did not hold this lock" as a `false` return value, not a query error. Session-level advisory locks are counted; if this session already held the key (for example because a prior holder returned a lock-bearing connection to the pool and this writer re-acquired reentrantly), a single `pg_advisory_unlock` can return `true` while the session still holds the lock. Unlock through `query_as` into `(bool,)`, treat `false` like a failed unlock (leave the connection for `Drop`/`close_on_drop`), and add a regression test that aborts during unlock. Same gap on the Tigris error cleanup at `:288`.

- [P2] A cancelled winning lock probe can still poison `lock_pool` *(connection-affinity follow-up, not narrowed #215)*
  `crates/gitlawb-node/src/git/repo_store.rs:232`
  PR-introduced. Failed probes drop their connection without holding a lock, but the one successful `pg_try_advisory_lock` runs on a bare `PoolConnection` until `RepoWriteGuard` is constructed at `:240`. If the task is cancelled during that `fetch_one().await` after Postgres took the lock, the connection is returned to `lock_pool` without `close_on_drop`, recreating the counted-lock/reentrancy failure mode the new `Drop` is meant to prevent.

- [P2] The competing-session and cancellation regression tests still do not protect CI *(connection-affinity follow-up, not narrowed #215)*
  `crates/gitlawb-node/src/git/repo_store.rs:757`
  PR-introduced. `make_store_for_live()` wires `lock_pool` with `connect_lazy("postgres://invalid")` while the `#[sqlx::test]` fixture provisions `pool` against the real test database, so even a manual run cannot pass. Both cases are marked `#[ignore]` with a misleading reason — CI already provisions Postgres in `pr-checks.yml` and runs plain `cargo test --workspace` with no `--ignored`. Even after fixing the fixture, `cancelled_guard_releases_lock_when_dropped` probes lock availability immediately after `handle.await` while `Drop` only calls `close_on_drop()`, which spawns async session teardown in sqlx 0.8.6, so the test can pass or fail depending on scheduler timing rather than proving the lock is gone.

- [P3] Name the release that actually carries the re-key *(narrowed #215 — fix before merge)*
  `docs/RUN-A-NODE.md:181`
  Verified against live `main`: release-please manifest is **0.7.0** and that release has shipped without this SHA-256 change. The row says v0.7.0 re-keys every repo's advisory lock, so an operator reading it would plan the drain at the wrong upgrade. Key the warning to the first build that actually carries the SHA-256 lock, or drop the version and describe it by behavior.

- [P3] `lock_pool` startup uses a single `connect()` with no retry *(connection-affinity follow-up, not narrowed #215)*
  `crates/gitlawb-node/src/main.rs:287`
  PR-introduced. The main application pool goes through `connect_db_with_retry`, but the advisory-lock pool is created with one `connect().await?` immediately afterward. A transient connection failure or brief `max_connections` pressure after the main pool comes up aborts node startup even though migrations already succeeded.

- [P3] Operator connection sizing ignores the two extra `lock_pool` sessions *(connection-affinity follow-up, not narrowed #215)*
  `crates/gitlawb-node/src/main.rs:282`
  PR-introduced. Each node opens a second pool to the same `database_url` with `max_connections(2)` on top of `GITLAWB_DB_MAX_CONNECTIONS` (default 20, documented in `.env.example`). Fleet sizing that budgets only the main pool can hit Postgres `max_connections` earlier than expected.

- [P3] Rebase onto current `main` and drop the `Cargo.lock` hunk *(narrowed #215 — fix before merge)*
  `Cargo.lock:3303`
  Branch is 36 commits behind live `main`. After rebase, the 0.5.0→0.5.1 version bumps and `icaptcha-client` `sha2` line are already on `main` — this hunk is empty churn. Both this branch and current `main` touch `test_support.rs`, so the resolved diff needs a look before merge.

Gravirei:
Review feedback addressed. I split this PR down to the narrowed #215 per the P1 finding:

**Now in this PR (narrowed #215):**
- SHA-256 `advisory_lock_key` (stable across toolchains)
- Golden test `advisory_lock_key_is_stable` + one-axis-at-a-time `advisory_lock_key_differs_for_different_inputs`
- `acquire_write` rolling-upgrade re-keying docstring
- `RUN-A-NODE.md` operator row — **version dropped**, describes the re-key by behavior, since the SHA-256 change ships in the first release after this merges (neither 0.7.0 nor 0.7.1 carries it)
- Branch rebased onto current `main` (`fdf716d`); no `Cargo.lock`/`CHANGELOG` churn (0 lines differ from upstream)

**Removed (moved to a follow-up PR):** the entire connection-affinity rework — `lock_pool` field, dedicated `PoolConnection` in `RepoWriteGuard`, `Drop`/`close_on_drop`, acquire-inside-retry-loop restructuring, the `main.rs` second pool, and the two DB regression tests. Base's pre-existing unlock-on-a-different-session behavior is unchanged, per the split rationale. `test_support.rs` is no longer touched by this PR.

Diff vs `main` is now exactly two files: `repo_store.rs` and `docs/RUN-A-NODE.md`.

beardthelion (approved):
Re-reviewed at `89e10d0`. The split asked for at `30e0dec` landed: this head is only the SHA-256 `advisory_lock_key`, the golden + one-axis differs tests, the `acquire_write` rolling-upgrade docstring, and the `RUN-A-NODE.md` operator row. No `lock_pool`, no pinned `PoolConnection`, no `Drop`/`close_on_drop`.

The key itself is still right. I recomputed `SHA-256("did_key_...:hello")[..8]` LE as `-6680856138670956537` (BE differs), then ran the mutation matrix on this head: revert to `DefaultHasher`, flip to `from_be_bytes`, drop the `":"` separator, and drop each parameter in turn. Golden and/or the differs test went RED on all five, GREEN restored. `owner_slug` is colon-sanitized in `local_path` and `validate_repo_name` forbids `:`, so the separator stays injective on production inputs. That closes #210.

One process note, not a finding: #79 still ships a different SHA-256 fold (`format!("{owner}/{repo}")` big-endian). Whichever lands second re-keys the fleet again. Settled earlier: this PR owns the key; #79 drops its lock-key work and rebases onto this.

Not an ask, recorded only: unlock still goes through `&self.pool` on a different checkout than acquire. That is pre-existing on `main` (identical shape at `origin/main`) and lives in #279 / #285. Splitting left that axis alone on purpose. The optional dual-key transition stays out of scope here; #210 already accepted the documented drain / single-node window, which this head carries in the docstring and `RUN-A-NODE.md`.

jatmn (approved):
LGTM

kevincodex1 (approved):

kevincodex1:
@coderabbitai please fix conflicts

jatmn (approved):
LGTM

beardthelion (approved):
Re-reviewed at `c68e1b07`. My previous approval was at `89e10d0f`, so this round covers the main merge and the test-sync commit on top of it.

The merge is clean. Net diff against main is still the same three files, +91/-12, base level with main, and all twelve checks are green on this head rather than an earlier one.

I recomputed the golden key instead of trusting the assert. SHA-256 of `did_key_z6Mkq...:hello` truncated to eight bytes little-endian is `-6680856138670956537`, and big-endian gives `549446935909779619`, so the pinned value fixes byte order as well as algorithm. That matters, because SHA-256 alone would not have frozen the key. I then broke the derivation five ways and each guard failed for its own reason: reverting to `DefaultHasher`, flipping to big-endian, dropping the `':'` separator, and dropping either input from the hash. The two one-axis cases failed on their named assertions rather than generically.

`c68e1b07` is the right call. Put the local copy back on `DefaultHasher` and the deadline test goes red: the key stops colliding, `acquire_write` takes the lock on the first try, and the stall the test is built around never happens. Restored, it passes against Postgres in about three seconds.

On the rolling-upgrade window, I went looking for a deployment that actually reaches it and could not find one. `infra/aws` has no `db_host` input: it is computed at `main.tf:296` as either that stack's own RDS address or the container Postgres on that stack's own instance, and `aws_instance.node` carries no `count`, so one stack is one node with one database. The fly setup is per-app databases federating over libp2p, which is the topology where this lock has no cross-machine effect regardless of the key. Nothing shipped lets two nodes share a Postgres, and no operator doc describes wiring one by hand. The re-key is real and the drain note is worth keeping as cheap insurance, but the concurrent-write window it guards needs a second node on the same database, and that is not a configuration anyone can currently select. So I am not asking for a dual-key transition release here, and I am not filing one as follow-up work either.

Two smaller things I noticed while checking that. #210's reachability argument rests on `purge-spam` contending through `try_lock_repo`, and neither symbol exists in `crates/` any more, so that reasoning is stale as this PR closes the issue. And the docstring's "already shifted across toolchain versions" claims more than the history supports: the SipHash-2-4 to 1-3 change predates `DefaultHasher::new()` stabilizing in 1.13.0, so the algorithm has not observably moved within this API's stable life. The standard library's disclaimer is explicit and the fix is still correct, the sentence just does not need to lean on a shift that has not happened yet. Neither is worth a change on its own.

One optional tidy, not a blocker: the deadline test hand-copies the derivation, and that copy is what drifted here. Making `advisory_lock_key` `pub(crate)` and importing it in the test removes the copy. I checked it compiles, clears clippy with `-D warnings`, and keeps the deadline test green. Fold it in if you want it in this PR, otherwise I will pick it up.

jatmn (approved):
LGTM

beardthelion (approved):
Re-reviewed at `aa68d97a`. My previous approval was at `c68e1b07`, so this round covers the `pub(crate)` change and the removal of the test's hand-copied key derivation.

The refactor is right, and it closes the drift rather than patching it. That test copy had already diverged once when the key moved to SHA-256; re-syncing it at `c68e1b07` left a second copy in place to diverge again, and importing the production function removes the possibility. There is now exactly one derivation in the crate and one production caller.

I re-ran the verification rather than trusting the suite. The golden value recomputes to `-6680856138670956537` independently, and five mutations all go red: reverting to `DefaultHasher`, flipping to big-endian, dropping the `:` separator, and hashing either component alone. The receive-pack deadline test is still load-bearing on this head, since removing the `tokio::time::timeout` wrapper turns it red, so importing the shared key did not quietly break the lock collision that test depends on. CI is green across all 12 checks.

## Findings

- [P3] Pin the colon-free contract on `owner_slug` in the doc comment
  `crates/gitlawb-node/src/git/repo_store.rs:759`
  Domain separation is `owner_slug + ":" + repo_name` with no length prefix, so it is injective only while the first component cannot contain a colon. That holds for the `owner_did.replace([':', '/'], "_")` form `local_path` produces, and it is stated nowhere. Now that the function is `pub(crate)`, a caller passing a raw DID would collide: `("did:key:abc", "x")` and `("did", "key:abc:x")` both hash to `2845113821417416517`. A line in the doc comment plus a `debug_assert!(!owner_slug.contains(':'))` would make the requirement explicit at the boundary this commit widens. I checked that assert compiles and leaves the existing tests green.

Not blocking. Happy for it to land as a follow-up if you would rather hold this PR at its current shape.

jatmn (approved):
LGTM

beardthelion (approved):
Re-reviewed at `b01d17b7`. This round covers the colon-free `owner_slug` contract on `advisory_lock_key` (`debug_assert` plus doc comment); prior rounds already landed SHA-256, the golden and differential tests, the `pub(crate)` export, and the `repos.rs` deadline-test import.

I ran `cargo test -p gitlawb-node git::repo_store::tests::advisory_lock_key_is_stable -- --exact` (pass) and gutted the SHA-256 body back to `DefaultHasher` (RED exit 101: `-1980555221488505497` vs golden `-6680856138670956537`). Dropping the `:` separator or hashing only one axis also REDs the respective tests. Production path is `acquire_write` → `local_path` (replaces `:` and `/` in DIDs) → `advisory_lock_key`; `validate_repo_name` disallows `:` in repo names, so domain separation holds on the live path. Blast-radius pass: only `acquire_write` and tests consume the function.

Not an ask, recorded only: `p2p/mod.rs` still seeds PeerId and gossipsub message IDs from `DefaultHasher` (same defect class, separate issue). The `acquire_write` doc mentions an optional dual-key transition; the circular-remedy guidance still applies if anyone pursues that path.

kevincodex1 (approved):

PR #221, opened 2026-07-19
fix(node): implement reconciliation sweep as durability backstop (#218)

### LINKED ISSUE #218: Implement the reconciliation sweep the replication path already assumes as a durability backstop
## Problem

The codebase justifies tolerating dropped or delayed post-push replication work by pointing at a reconciliation sweep that does not exist. `CONCEPTS.md` frames narrowing the exposure set as producing "a durability gap the reconciliation sweep backstops," and at least seven code comments lean on the same backstop:

- `crates/gitlawb-node/src/state.rs:138` "there is no reconciliation sweep to re-derive a dropped copy"
- `crates/gitlawb-node/src/state.rs:243` "there is no reconciliation sweep, so a *dropped* job would be lost forever"
- `crates/gitlawb-node/src/state.rs:422` "there is no sweep, so it stays lost until a later push re-walks the repo"
- `crates/gitlawb-node/src/api/repos.rs:5316`, `repos.rs:5406`, and several `crates/gitlawb-node/src/git/push_delta.rs` comments say the same.

Grepping the crate confirms there is no function, spawned periodic task, or job that re-walks repos for under-replicated or unsealed objects. The periodic tasks in `main.rs` are peer-count metrics, rate-limiter and proof-ledger cleanup, and gossip bootstrap/peer ping, none of which do this. So "lost forever" is literal, not conservative phrasing.

This matters against the project's stated mission (README): "once code is pushed to the network, it should not disappear because one server went down." Every path that drops a pin or a recovery copy (a mid-drain panic, a node crash or shutdown while the drain task is running, and, until the companion detach issue lands, a client disconnect at the receive-pack tail) violates that promise with no safety net.

## Fix

Implement the reconciliation sweep the design already assumes: a periodic pass that re-derives the set of objects a repo should have replicated (pins) and sealed (encrypted recovery copies) under current visibility rules, and produces the missing ones. Because pinning and sealing are idempotent and content-addressed, the sweep is safe to run repeatedly and only ever fills gaps, never serves wrong content.

Sketch, to be scoped in its own plan:

- Bound the work per pass (a cursor over repos, a cap per run) so the sweep cannot itself become the O(repos) amplification the #174 admission work exists to prevent.
- Reuse the existing fresh-resolution pipeline (`resolve_drain_object_list` / `replication_withheld_set` / `resolve_candidates_for_push`) so the sweep computes the same withheld/exposure sets as the push path, fail-closed.
- Emit a metric for gaps found and filled, so the durability guarantee is observable rather than assumed.

## Notes

- This is the complete durability backstop; the companion detach issue is the cheaper first step that shrinks the disconnect window but leaves the crash/shutdown-mid-drain window open. This issue closes that window.
- On landing, the seven "the reconciliation sweep backstops this" comments become accurate rather than aspirational.

Gravirei:
## Summary

Implements the periodic reconciliation sweep the replication path already assumes as a durability backstop. Previously, every path that drops a pin or recovery copy (mid-drain panic, node crash/seal, client disconnect at the receive-pack tail) resulted in data loss with no safety net.

## Motivation & context

Closes #218

The codebase justified tolerating dropped post-push replication work by pointing at a reconciliation sweep that did not exist. This made "lost forever" literal rather than conservative phrasing, violating the project's stated promise that "once code is pushed to the network, it should not disappear because one server went down."

## Kind of change

- [ ] Bug fix
- [x] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

- **crates/gitlawb-node/src/reconciliation.rs** (new): Periodic sweep that re-derives the set of objects a repo should have pinned/sealed under current visibility rules
  - Runs hourly, capped at 100 repos per pass with a cursor to prevent O(repos) amplification
  - For each announceable repo: full-scans all git objects, applies fail-closed visibility filtering, pins missing objects to IPFS and Pinata, re-seals encrypted recovery copies, and anchors sealed manifests to Arweave
  - Reuses existing fresh-resolution pipeline (list_all_objects, replicable_blob_set, replicable_objects_fail_closed, ipfs_pin::pin_new_objects, pinata::pin_new_objects, encrypted_pin::encrypt_and_pin)
  - Skips non-announceable repos (private / mode A / undetermined)
  - Respects graceful shutdown signal

- **crates/gitlawb-node/src/metrics.rs**: Added gitlawb_reconciliation_gaps_found_total and gitlawb_reconciliation_gaps_filled_total counters

- **crates/gitlawb-node/src/main.rs**: Registered reconciliation module and spawned the background sweep task

## How a reviewer can verify

```sh
cargo check -p gitlawb-node
cargo clippy -p gitlawb-node -- -D warnings
cargo test -p gitlawb-node -- metrics::tests
```

## Before you request review

- [ ] Scope is one logical change; no unrelated churn
- [ ] `cargo test --workspace` passes locally (DB-dependent tests require a running Postgres)
- [ ] New behavior is covered by tests (unit tests cover the building blocks; sweep itself is integration-tested at runtime)
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` is clean
- [ ] Commit titles use Conventional Commits (`fix(...)`)

## Notes for reviewers

The sweep is intentionally conservative per pass (100 repos, hourly) to avoid competing with the push path for resources. The cursor wraps around so every repo is eventually covered. Encrypted pin re-sealing and Arweave manifest anchoring are best-effort (failures are logged and skipped).

## Summary by CodeRabbit

- **New Features**
  - Added hourly background reconciliation sweeps to detect and restore missing replication/pinning content.
  - Reconciliation computes candidate objects and pins missing public objects, skipping quarantined repositories.
  - Supports encrypted recovery/resealing for path-scoped visibility when configured, including best-effort encrypted manifest anchoring.

- **Monitoring**
  - Added Prometheus counters for reconciliation gaps found and filled.
  - Improved reconciliation start and outcome logging.

- **Bug Fixes**
  - Updated IPFS pinning checks to rely on IPFS-specific local pin state.
  - Improved pinned-CID storage behavior to repair/retain local IPFS CID and keep Pinata-only rows retriable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3300,7 +3300,7 @@ dependencies = [
 
 [[package]]
 name = "git-remote-gitlawb"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "anyhow",
  "gitlawb-core",
@@ -3312,7 +3312,7 @@ dependencies = [
 
 [[package]]
 name = "gitlawb-attest"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "base64",
  "ed25519-dalek",
@@ -3329,7 +3329,7 @@ dependencies = [
 
 [[package]]
 name = "gitlawb-core"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "anyhow",
  "base64",
@@ -3356,7 +3356,7 @@ dependencies = [
 
 [[package]]
 name = "gitlawb-node"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "alloy",
  "anyhow",
@@ -3412,7 +3412,7 @@ dependencies = [
 
 [[package]]
 name = "gl"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "alloy",
  "anyhow",
@@ -3824,6 +3824,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
+ "sha2",
  "thiserror 2.0.18",
  "tracing",
 ]
--- a/crates/gitlawb-node/src/db/mod.rs
+++ b/crates/gitlawb-node/src/db/mod.rs
@@ -2159,19 +2159,16 @@ impl Db {
 // ── Pinned CIDs ───────────────────────────────────────────────────────────────
 
 impl Db {
-    pub async fn is_pinned(&self, sha256_hex: &str) -> Result<bool> {
-        let row = sqlx::query("SELECT COUNT(*) as cnt FROM pinned_cids WHERE sha256_hex = $1")
-            .bind(sha256_hex)
-            .fetch_one(&self.pool)
-            .await?;
-        Ok(row.get::<i64, _>("cnt") > 0)
-    }
-
+    /// Record the local IPFS CID for a git object.
+    /// If a Pinata-only row already exists (cid IS NULL), this updates it with
+    /// the real IPFS CID so that `has_ipfs_cid` correctly reflects IPFS state.
     pub async fn record_pinned_cid(&self, sha256_hex: &str, cid: &str) -> Result<()> {
         sqlx::query(
             "INSERT INTO pinned_cids (sha256_hex, cid, pinned_at)
              VALUES ($1, $2, $3)
-             ON CONFLICT(sha256_hex) DO NOTHING",
+             ON CONFLICT(sha256_hex) DO UPDATE SET
+               cid = COALESCE(pinned_cids.cid, EXCLUDED.cid),
+               pinned_at = EXCLUDED.pinned_at",
         )
         .bind(sha256_hex)
         .bind(cid)
@@ -2268,6 +2265,17 @@ impl Db {
             .collect())
     }
 
+    /// Returns true if this object already has a local IPFS CID recorded.
+    pub async fn has_ipfs_cid(&self, sha256_hex: &str) -> Result<bool> {
+        let row = sqlx::query(
+            "SELECT COUNT(*) as cnt FROM pinned_cids WHERE sha256_hex = $1 AND cid IS NOT NULL",
+        )
+        .bind(sha256_hex)
+        .fetch_one(&self.pool)
+        .await?;
+        Ok(row.get::<i64, _>("cnt") > 0)
+    }
+
     /// Returns true if this object already has a Pinata CID recorded.
     pub async fn has_pinata_cid(&self, sha256_hex: &str) -> Result<bool> {
         let row = sqlx::query(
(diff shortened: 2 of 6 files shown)


jatmn (changes requested):

## Findings

- [P1] Preserve reachable blobs for ordinary public repositories
  `crates/gitlawb-node/src/reconciliation.rs:137`
  When a repository has no path-scoped rule (the normal public-repo case), this assigns an empty `allowed` set and then passes it to `replicable_objects_fail_closed`. That helper drops every blob not in `allowed`, so the sweep only re-pins commits and trees; a dropped file blob is never recovered. `replicable_blob_set` already computes the reachable public-blob set for the no-rule case, so use it here as well (or take the equivalent no-withheld path) and add a regression test with a public repo containing a blob.

- [P1] Do not derive replication visibility from raw mirror rows
  `crates/gitlawb-node/src/reconciliation.rs:81`
  `list_all_repos` returns physical mirror rows. Those rows are deliberately stored as public with no visibility rules, while the canonical row can be private or path-restricted; `Db::get_repo` explicitly prefers the canonical row to avoid this bypass. The canonical row is skipped here, but its mirror passes the anonymous root gate and the sweep publishes its commits and trees, disclosing private commit metadata and tree paths. Resolve visibility from the canonical/deduplicated record before doing any external pinning (and retain the physical path only after that decision).

- [P2] Recheck quarantine at the replication write boundary
  `crates/gitlawb-node/src/reconciliation.rs:78`
  The worker reads the quarantine set and repository rows in separate queries, then does no per-repo quarantine check before calling the external pinning backends. If an operator quarantines a repo after the first query, that pass still publishes its objects even though quarantined rows are meant to be ineligible for replication. Fetch only non-quarantined rows atomically or recheck immediately before the external writes.

- [P1] Clamp the batch cursor when the eligible set changes
  `crates/gitlawb-node/src/reconciliation.rs:93`
  The cursor survives between passes but `all` is rebuilt after filtering a mutable repository/quarantine set. For example, a 101-repo pass leaves `cursor = 100`; if deletion or quarantine reduces the next eligible list to 99, this slices `all[100..99]` and panics. Because this is the only spawned worker, reconciliation then stops permanently until a process restart. Reset or clamp the cursor before slicing and cover a shrinking eligible set.

- [P1] Reconcile IPFS and Pinata state independently
  `crates/gitlawb-node/src/reconciliation.rs:169`
  The sole gap test asks only whether a `pinned_cids` row exists, then uses that result to skip both backends. IPFS success records that row even if the subsequent Pinata upload fails; conversely, `record_pinata_cid` can create the row without an IPFS pin. On later passes `repo_gaps` is zero, so the missing backend is never retried, despite Pinata tracking its own `pinata_cid` completion state. Compute each backend's missing set independently (or call each helper with the eligible objects) so partial replication failures are actually repaired.

- [P2] Bound one repository's reconciliation work
  `crates/gitlawb-node/src/reconciliation.rs:132`
  `REPOS_PER_PASS` bounds only the number of repositories. For each selected repo the worker materializes all object IDs and all blob IDs, then reads and uploads every missing object sequentially, with no object/byte/time budget or shutdown observation until the pass completes. A large repository can therefore monopolize the blocking pool and keep the hourly worker running indefinitely, defeating the stated bounded-work and graceful-shutdown behavior. Add per-repo work limits and cooperative cancellation/progress checks.

jatmn (changes requested):

## Findings

- [P1] Continue reconciliation past the per-repository object cap
  `crates/gitlawb-node/src/reconciliation.rs:164`
  `list_all_objects` deterministically returns the whole object set, but the worker permanently truncates it to the first 50,000 entries before the pin helpers discard already-pinned entries. A repository with an eligible missing object after that prefix will revisit and skip the same pinned prefix every hour, so that object is never repaired. Keep resumable/rotating per-repository progress (or select missing objects before applying a continuation-aware cap) and cover an object beyond the cap.

- [P1] Track local-IPFS completion independently of Pinata completion
  `crates/gitlawb-node/src/reconciliation.rs:182`
  Although the new code says the two backends are reconciled independently, `ipfs_pin::pin_new_objects` uses `Db::is_pinned`, which returns true for any `pinned_cids` row. If IPFS fails and the subsequent Pinata upload succeeds, `record_pinata_cid` creates that row; all later IPFS passes skip the object permanently. Represent and query IPFS completion separately, then add the IPFS-failure/Pinata-success recovery case.

- [P1] Retry failed encrypted-manifest anchors
  `crates/gitlawb-node/src/reconciliation.rs:235`
  The Irys anchor is attempted only when `encrypt_and_pin` returns newly sealed blobs. If sealing and recording succeeds but the manifest anchor fails (or the process stops after recording), subsequent sweeps see unchanged recipient tags, return an empty `sealed` list, and never enter the anchor path again. Persist/reconcile anchor completion independently, or re-anchor the existing encrypted rows, so a transient anchor failure does not leave recovery copies undiscoverable forever.

- [P2] Recheck quarantine immediately before publishing reconciliation output
  `crates/gitlawb-node/src/reconciliation.rs:84`
  The deduplicated repository list filters quarantine only at the start of a pass. An operator can quarantine a selected repository while the worker scans it; the pass will still submit public objects, encrypted copies, and possibly an Irys manifest at lines 182-270. Recheck eligibility at the external-write boundary (or select and lock/check it atomically) so quarantine takes effect before replication is published.

- [P2] Make the claimed work and shutdown bounds apply to the whole sweep
  `crates/gitlawb-node/src/reconciliation.rs:129`
  The 50,000-object cap is imposed only after `list_all_objects`, `replicable_blob_set`, and `all_blob_oids` have already materialized/walked the entire repository; phase 2 then walks and encrypts every withheld blob without any cap. The worker also observes `shutdown_rx` only after the full pass, while the IPFS helper has no request timeout. A large repository or a stalled Kubo request can therefore monopolize the worker indefinitely and prevent shutdown. Bound the scans and both phases with continuable progress, request timeouts, and cooperative cancellation.

PR #127, opened 2026-06-30
Release automation

kevincodex1:
## Summary

  Makes release-please actually cut releases (its manifest baseline was broken) and fans every release out to npm, Homebrew, Windows, and the website — so a single merge of the release PR ships all channels.

  ## Motivation & context

  release-please never cut a release here: the manifest said `0.3.8` but the only tag is a manual lightweight `v0.3.9` (PR #12) and `Cargo.toml` is `0.3.9`, so there was no valid baseline (`v0.3.8` doesn't exist) and no release PR was ever opened. On top of that, releases only shipped a Docker image + tarballs — npm/Homebrew were manual and drifting (the standalone `gl-npm` repo and the web `install.sh` pointed at a stale `gitlawb/releases` repo with the wrong archive naming).
  Closes #

  ## Kind of change

  - [ ] Bug fix
  - [x] Feature
  - [ ] Security fix
  - [ ] Docs
  - [x] Tests / CI
  - [ ] Refactor (no behavior change)
  - [ ] Breaking or protocol change (issue required first)

  ## What changed

  No Rust crates are touched — this is CI/release tooling plus a vendored npm wrapper.

  - **fix**: reconcile `.release-please-manifest.json` `0.3.8 → 0.3.9` so release-please diffs `v0.3.9..main` and opens a proper release PR; bump `release-please-action` `v4.3.0 → v5.0.0` (matches the openclaude pipeline).
  - **CI**: add Windows target `x86_64-pc-windows-msvc` to `release-binaries` (ships `.zip`; unix keeps `.tar.gz`) with a portable sha256 helper; Windows builds only `gl` + `git-remote-gitlawb` so a daemon compile failure can't block the release.
  - **npm**: vendor the `@gitlawb/gl` wrapper + 4 platform packages into `npm/` (replaces the standalone `gl-npm` repo); new `npm-publish` job downloads the release tarballs, lays in binaries, stamps versions, and `npm publish --provenance`.
- **Homebrew**: `homebrew-bump` job regenerates `Formula/gl.rb` in `Gitlawb/homebrew-tap` from the release checksums (`brew install gitlawb/tap/gl`).
- **web**: add `install.ps1` (Windows installer); `web-sync` job opens a PR to `Gitlawb/web` syncing `install.sh` + `install.ps1` + `public/version.json`, removing orphaned `public/bin`, and fixing the stale `skill.md` releases link.
- All new jobs are gated on `release_created` and guarded on their secret, so they no-op until configured.
[...]

PR #233, opened 2026-07-22
ci: switch npm publishing to trusted publishing (OIDC) + manual backfill

kevincodex1:
The npm-publish job silently skipped every release since 0.4.x because NPM_TOKEN was never configured — the guard made the job green while @gitlawb/gl sat at 0.3.7 on the registry. Replace token auth entirely:

- npm Trusted Publishing: the job's GitHub OIDC identity (id-token: write) is exchanged with the registry; no long-lived secret to configure, leak, or rotate. Provenance is attested automatically. All five @gitlawb packages exist on the registry now, so OIDC no longer has the first-publish bootstrap problem that forced token auth in #127.
- Fail loud: the secret guard is gone. If a package's trusted publisher isn't configured on npmjs.com, the publish errors visibly instead of green-skipping.
- setup-node 24 + npm >=11.5.1 (OIDC exchange support; older npm falls back to absent token auth with a confusing error).
- workflow_dispatch backfill: publish an already-cut GitHub release (e.g. v0.6.0) to npm by tag — binaries come from the existing release assets, so only the npm-publish job runs.

## Summary by CodeRabbit

* **New Features**
  * Added a manual backfill path to publish an existing GitHub release to npm via a required release tag input.
  * Enhanced npm publishing to support both standard release flows and manual backfills.
  * Enabled tokenless npm publishing with provenance metadata and automatic version/tag resolution.

* **Bug Fixes**
  * Prevented release automation from creating release PRs during manual backfills.
  * Preserved safeguards to avoid duplicate publishing of already-present versions.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,16 @@ on:
   push:
     branches:
       - main
+  # Manual backfill: publish an already-cut release to npm (e.g. after the
+  # npm side of trusted publishing is configured, or a publish was missed).
+  # Only the npm-publish job runs; binaries are downloaded from the existing
+  # GitHub release for the given tag.
+  workflow_dispatch:
+    inputs:
+      npm_backfill_tag:
+        description: "Existing release tag to publish to npm (e.g. v0.6.0)"
+        required: true
+        type: string
 
 concurrency:
   group: release-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


kevincodex1:
@coderabbitai please review again approve if its all good!

kevincodex1:
@coderabbitai please check again theres a latest commit make sure we are good

PR #236, opened 2026-07-22
Fix/gl acceptance findings

kevincodex1:
## Summary

Fixes the four CLI issues surfaced when a fresh agent, given only the published docs, drove gl from zero to a signed push: status not seeing gl init's remote, init's push hint failing on uncommitted repos, cert show not actually verifying, and doctor red-flagging healthy local nodes.

## Motivation & context

A docs-only acceptance test (agent + local harness) reached a verified push but hit exactly these traps; the docs currently carry caveats working around each. Closes #<issue if you file one>.

## Kind of change

- [x] Bug fix
- [x] Feature

## What changed

- `gl` (status.rs): detect_gitlawb_remote scans all remotes — origin first (clones), then `gitlawb` (the name gl init adds), then any gitlawb:// URL.
- `gl` (init.rs): `git init -b main` with plain-init fallback for git <2.28; the closing hint is state-aware — with no commits it prints the add/commit step before the push, using the repo's actual branch name.
- `gl` (cert.rs): `cert show` rebuilds the node's canonical signing payload, derives the Ed25519 key from the certificate's node DID, and prints a VALID/INVALID verdict; `--verify` makes the exit code reflect it. Round-trip, tampered-payload, and malformed-signature unit tests.
- `gl` (doctor.rs): a localhost GITLAWB_NODE passes with an explanatory note instead of failing — the connectivity check still fails loudly when the configured node is unreachable.

## How a reviewer can verify

```sh
cargo test -p gl            # includes cert::tests::verify_signature_round_trip_and_tamper
# functional (local harness): gl init in empty dir → hint shows commit step; branch is main;
# gl status shows the repo; gl doctor passes; gl cert show <repo> <id> --verify prints VALID, exit 0
```

Before you request review

- [x] cargo test --workspace passes locally
- [x] cargo fmt --all and cargo clippy --workspace --all-targets -- -D warnings are clean
- [x] Commit titles use Conventional Commits (feat(...), fix(...), docs(...))
- [x] Docs / .env.example updated if behavior or config changed (or N/A — docs cleanup follows the release, see notes)

Protocol & signing impact

- [x] Touches DID / did:key, Ed25519 / RFC 9421 signatures, UCAN, ref certs, or P2P wire formats
- [ ] Discussed in an issue before implementation — verification only; no wire/format change
- [x] Backward-compatible with existing nodes and previously signed history — cert verification is read-only and mirrors the node's existing signing payload byte-for-byte

Notes for reviewers

Verified functionally against a local node: all four flows exercised end-to-end, including a real VALID verdict on a live certificate (proving the payload reconstruction matches node signing exactly). After this ships in 0.6.1, the docs lose three workaround caveats and skill.md's cert section gets corrected (it documents a cert verify <cert-file> subcommand that has never existed).

**Validation recap:** four per-fix commits; full workspace gate green (fmt, clippy `-D warnings`, all tests incl. DB suites against Postgres); functional verification against a live node — the init hint, main branch, status detection, doctor note, and a real `VALID` Ed25519 verdict with exit 0.

**Also notable:** this is the first PR through your armed merge gate, so expect the four required checks to show as *Required* — worth confirming they do. After merge: release-please opens 0.6.1 → merging it runs the first fully-hands-free multi-channel release → then I do the docs cleanup pass (drop the status/branch caveats, fix the phantom `cert verify <cert-file>` in skill.md, restore "verify with `gl cert show --verify`"). CodeRabbit findings are mine to address when they arrive.

## Summary by CodeRabbit

- **New Features**
  - Added `--verify` to `gl cert show` for offline Ed25519 signature verification with `VALID`/`INVALID`.
  - Added `--expect-node` to `gl cert show` (requires `--verify`) to enforce issuer anchoring.
  - `gl init` now initializes with `main` when available and prints branch/HEAD-aware next steps.

- **Bug Fixes**
  - Improved `gl status` detection of `gitlawb://` remotes across all remotes/URL variants and updated the no-remote message.
  - `gl doctor` now treats loopback `GITLAWB_NODE` values as intentional and detects zsh alias shadowing for `gl`.

- **Documentation**
  - Homebrew caveats and installer notes warn about oh-my-zsh `gl='git pull'` alias shadowing and suggest `unalias gl`.

- **Tests**
  - Added unit tests for certificate verification (canonical payload + failure cases) and `gl doctor` alias/loopback detection.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -596,7 +596,7 @@ jobs:
                 oh-my-zsh's git plugin aliases gl='git pull', which shadows this
                 binary in interactive shells. If \`gl\` prints "fatal: not a git
                 repository", run:
-                  echo 'unalias gl' >> ~/.zshrc && source ~/.zshrc
+                  echo 'unalias gl 2>/dev/null' >> ~/.zshrc && source ~/.zshrc
               CAVEATS
             end
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -591,6 +591,15 @@ jobs:
               bin.install "git-remote-gitlawb"
             end
 
+            def caveats
+              <<~CAVEATS
+                oh-my-zsh's git plugin aliases gl='git pull', which shadows this
+                binary in interactive shells. If \`gl\` prints "fatal: not a git
+                repository", run:
+                  echo 'unalias gl' >> ~/.zshrc && source ~/.zshrc
+              CAVEATS
+            end
+
             test do
               assert_match version.to_s, shell_output("#{bin}/gl --version")
             end
--- a/crates/gl/Cargo.toml
+++ b/crates/gl/Cargo.toml
@@ -13,6 +13,7 @@ path = "src/main.rs"
 [dependencies]
 gitlawb-core = { path = "../gitlawb-core" }
 icaptcha-client = { path = "../icaptcha-client" }
+base64 = { workspace = true }
 tokio = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
--- a/crates/gl/src/cert.rs
+++ b/crates/gl/src/cert.rs
@@ -178,19 +178,29 @@ async fn cmd_show(
         }
     }
 
-    let info: Value = client
-        .get("/")
-        .await?
-        .json()
-        .await
-        .context("failed to fetch node info")?;
-    let current_node_did = info["did"].as_str().unwrap_or("");
-    if current_node_did == node_did {
-        println!("  Issuing node DID matches the node being queried.");
-    } else {
-        println!("  WARNING: Certificate node DID ({node_did}) does not match");
-        println!("           current node DID ({current_node_did}).");
-        println!("           This certificate was issued by a different node.");
+    // Contextual only — the verdict above stands on its own, so a node-info
+    // hiccup here must not turn a successfully displayed certificate into an
+    // error exit.
+    let current_node_did = match client.get("/").await {
+        Ok(resp) => resp
+            .json::<Value>()
+            .await
+            .ok()
+            .and_then(|info| info["did"].as_str().map(str::to_string)),
+        Err(_) => None,
+    };
+    match current_node_did.as_deref() {
+        Some(current) if current == node_did => {
+            println!("  Issuing node DID matches the node being queried.");
+        }
+        Some(current) => {
+            println!("  WARNING: Certificate node DID ({node_did}) does not match");
+            println!("           current node DID ({current}).");
+            println!("           This certificate was issued by a different node.");
+        }
+        None => {
+            println!("  NOTE: could not fetch current node info — skipping node-DID comparison.");
+        }
     }
 
     if require_valid {
--- a/crates/gl/src/cert.rs
+++ b/crates/gl/src/cert.rs
@@ -41,6 +41,9 @@ pub enum CertCmd {
         node: String,
         #[arg(long)]
         dir: Option<PathBuf>,
+        /// Exit non-zero unless the Ed25519 signature verifies
+        #[arg(long)]
+        verify: bool,
     },
 }
 
@@ -52,7 +55,8 @@ pub async fn run(args: CertArgs) -> Result<()> {
             id,
             node,
             dir,
-        } => cmd_show(repo, id, node, dir).await,
+            verify,
+        } => cmd_show(repo, id, node, dir, verify).await,
     }
 }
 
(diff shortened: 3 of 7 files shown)


kevincodex1:
@coderabbitai please review and approve

kevincodex1:
@coderabbitai please make a thorough review here

beardthelion:
Reviewed the new cert verification path and the doctor/init/status fixes against the node's signing code. The core is sound: the Ed25519 verify fails closed (a bad DID, non-base64 signature, wrong length, or wrong key all render INVALID and `--verify` exits non-zero), the canonical payload in `verify_signature` is byte-identical to what `gitlawb-node/src/cert.rs::issue_ref_certificate` signs (same `json!` fields and order, no `preserve_order` anywhere), and all five earlier review findings landed correctly. No shell-injection surface in the git-invoking paths.

One design question, then a few nits.

## Design question: what does `cert show --verify` promise?

The signature is verified against the key embedded in the cert's own `node_did`, and the node-DID comparison below it is contextual-only: it prints a warning on a mismatch but never affects the `--verify` exit, and it reads the "current" DID from the same node that served the cert. For checking a cert from your own configured node that is a fine integrity check.

But the module docs frame certs as "receipts proving a push was accepted," and the code warns when a cert was "issued by a different node," which reads like cross-node receipt verification. Under that reading a self-signed cert passes: a hostile source generates a keypair, sets `node_did` to it, signs the payload, and gets `VALID` + exit 0. Verifying against a key the cert supplies proves internal consistency, not that a node you trust issued it.

If `--verify` is only ever meant for a cert from your own node, this is fine as-is, though a doc note would keep `VALID` from being read as "trusted." If it is meant to verify certs from an untrusted source, it needs an out-of-band anchor: pin or resolve the expected node DID and require `cert.node_did` to equal it, gating the exit on that match. Your call on the intended scope.

## Nits (P3)

- `verify_signature`'s round-trip test signs and verifies with the CLI's own `json!`, so it can't catch canonicalization drift from the node. Interop is correct today, but a future added field or a `preserve_order` bump would make every real cert render INVALID with no test failing. A fixture cert produced by the node's signing path would pin it.
- The printed `unalias gl` remediation (`install.sh`, the brew caveats, `gl doctor`) uses `echo 'unalias gl' >> ~/.zshrc && source ~/.zshrc`, which appends a duplicate line on every run and errors on the bare `unalias gl` when `gl` is not currently aliased. `gl doctor` already uses the `unalias gl 2>/dev/null` form worth matching in the printed advice.
- Minor: the alias-shadow check is zsh-only (an explicit `alias gl=...` in `~/.bashrc` is not flagged), and `is_loopback_url` does not treat `127.0.0.0/8` or a trailing-dot host as loopback. Both are fine to leave; noting for completeness.

Otherwise this is in good shape. The `cert --verify` scope is the only thing I would want settled before it goes in.

kevincodex1:
@coderabbitai make a thorough review her eplease

beardthelion (approved):
Approving. The signature-verification path is the substantive change here and it holds up. I rebuilt the node's canonical signing payload on both sides: `gl` reconstructs the identical 7-field form the node signs, `issued_at` is stored as TEXT so it round-trips byte-exact, and the single-cert response carries every field the verifier reads. `--verify` fails closed on every case I threw at it (bad signature, missing or `"?"` node DID, an unreachable node with no `--expect-node`, a hostile self-signed cert, malformed base64 or DID input), none of which panic. 299 gl tests pass, and the new tamper test goes red when I stub the verifier to always succeed, so it is load-bearing.

Four P3 polish items below, none blocking, all yours to take or leave.

## Findings

- [P3] Qualify the `VALID` line so a scripted reader can't misread it. `crates/gl/src/cert.rs:184` — under `--verify`, `VALID` prints to stdout before the issuer-anchor check can bail, so a valid-signature/untrusted-issuer cert shows `VALID` on stdout while the command exits non-zero. The exit code is correct and authoritative, but a consumer doing `gl cert show ... --verify | grep -q VALID` would be misled. Either defer the line until after the anchor passes, or label it `VALID (signature only)`.

- [P3] Don't suggest `git switch -c main` when `main` already exists. `crates/gl/src/init.rs:213` — the `(None, _)` arm always prints `git switch -c main`, which fails with "a branch named 'main' already exists" on a detached HEAD in a repo that already has `main` (checkout an old commit or tag, then `gl init`). It is a printed hint, not an executed command, so recovery is trivial, but the instruction is wrong in a real scenario. Check `refs/heads/main` first and print `git switch main` when it exists. This is the open CodeRabbit thread.

- [P3] Fall back to an assumed scheme in `is_loopback_url`. `crates/gl/src/doctor.rs:345` — `Url::parse` needs a scheme, so a scheme-less `GITLAWB_NODE=127.0.0.1:7545` returns false and the doctor reports a bare loopback address as an ordinary remote node. The documented `http://127.0.0.1:7545` form is detected correctly and this change fixes the real `localhost.example` false positives, so it is net positive; retrying the parse as `http://{value}` closes the remaining shorthand gap.

- [P3] Match the shell unalias regex to the Rust splitter. `install.sh:202` vs `rc_unaliases_gl` in `crates/gl/src/doctor.rs` — the shell regex requires whitespace or end-of-line after `gl`, so `unalias gl;` is not matched and the install NOTE prints even though `gl` is already unaliased. `rc_unaliases_gl` splits on `;&|` first and handles it, so the two drift. Extending the trailing class to `([[:space:];&|]|$)` realigns them. This is the other open CodeRabbit thread.

PR #245, opened 2026-07-23
test(icaptcha-client): bind new()'s empty-API-key filter (#216)

### LINKED ISSUE #216: icaptcha-client: no test binds new()'s empty-API-key filter, so a Some("") regression would ship green
`IcaptchaCfg::new` (`crates/icaptcha-client/src/lib.rs:204-210`) normalizes an empty
`GITLAWB_ICAPTCHA_API_KEY` to `None` via `.filter(|s| !s.is_empty())` before storing
`api_key`. If that filter regresses, a trusted-origin solve carries an Authorization
header with an empty bearer token on both the challenge and answer requests
(`req.bearer_auth(key)` at lib.rs:310 and lib.rs:335): a malformed credential instead of
an absent one, and the "operator set the var but left it blank" state silently flips from
keyless to key-bearing. No test in the crate drives the empty-string value through
`new()`, so the regression ships green.

Verified by mutation at 7c24cd7 (#213's head; `new()` is identical on main, that head
adds only the #213 test):

- Baseline: 19 lib tests green.
- Filter removed (`std::env::var("GITLAWB_ICAPTCHA_API_KEY").ok()` with no emptiness
  filter): all 19 still green. That is the gap.

Suggested regression test, load-bearing both ways at the same head: RED under the
filter-removal mutation with exactly one failure (`left: Some("")`, `right: None`), and
20 passed 0 failed with the current code. It goes in lib.rs's `#[cfg(test)] mod tests`
so it shares `ICAPTCHA_ENV_LOCK`, same shape as the #213 test:

```rust
#[test]
fn new_treats_empty_api_key_as_absent() {
    let guard = ICAPTCHA_ENV_LOCK.lock().unwrap();
    let prev_key = std::env::var_os("GITLAWB_ICAPTCHA_API_KEY");
    let prev_op = std::env::var_os("GITLAWB_ICAPTCHA_URL");
    std::env::set_var("GITLAWB_ICAPTCHA_API_KEY", "");
    std::env::set_var("GITLAWB_ICAPTCHA_URL", "https://icap.mynode.example");
    let api_key = IcaptchaCfg::new(
        "did:key:zTEST",
        Some("https://icap.mynode.example/v1".into()),
        None,
    )
    .api_key;
    match prev_key {
        Some(v) => std::env::set_var("GITLAWB_ICAPTCHA_API_KEY", v),
        None => std::env::remove_var("GITLAWB_ICAPTCHA_API_KEY"),
    }
    match prev_op {
        Some(v) => std::env::set_var("GITLAWB_ICAPTCHA_URL", v),
        None => std::env::remove_var("GITLAWB_ICAPTCHA_URL"),
    }
    drop(guard);
    assert_eq!(api_key, None, "empty key must not become a bearer header");
}
```

Scope, to stay clear of the in-flight iCaptcha work: #213 covers the key_trusted gate
arms and this deliberately does not touch them; #214's proxy-observer and tripwire
rewrite owns wire-level observation (asserting the bearer actually rides the challenge
and answer requests fits there, not here); #212 is closed by #213. The sibling
empty-string filter on `GITLAWB_ICAPTCHA_URL` (lib.rs:200-202) is deliberately not asked
for: an empty operator already fails `is_https` and drops out inside
`resolve_solver_url`, so a test on it would not be load-bearing. Not blocking #213 or
#214; whoever picks this up may want to wait for #213 to land first since it sits in the
same test module.

beardthelion:
## Summary

Adds a guard test that drives an empty `GITLAWB_ICAPTCHA_API_KEY` through `IcaptchaCfg::new` on a trusted operator origin and asserts it normalizes to `None`, with a contrast arm showing a real key still attaches.

## Motivation & context

Closes #216

`new()` normalizes an empty key to `None` via `.filter(|s| !s.is_empty())` (`crates/icaptcha-client/src/lib.rs:213`). No test exercised the empty-string value, so dropping the filter shipped green and a trusted-origin solve would send an empty bearer token on the challenge and answer requests (`lib.rs:317`, `lib.rs:342`).

## Kind of change

- [ ] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [x] Tests / CI

## What changed

- `crates/icaptcha-client/src/lib.rs`: added `new_normalizes_empty_api_key_to_none_on_trusted_origin`

## How a reviewer can verify

```sh
cargo test -p icaptcha-client
```

Mutation check: remove `.filter(|s| !s.is_empty())` from the `api_key` read in `new()`. The new test fails; the other 21 lib tests stay green.

## Before you request review

- [x] Commit titles use Conventional Commits (`test(...)`)
- [x] Docs / `.env.example` updated if behavior or config changed (N/A: test only)
- [x] Checked existing PRs so this isn't a duplicate (nearest is #213 for #212, which covers the `key_trusted` gate, not the emptiness filter)

## Protocol & signing impact

No protocol impact.

## Notes for reviewers

The test follows the same env-handling pattern as the sibling tests in the file: it saves and restores both env vars and releases `ICAPTCHA_ENV_LOCK` before asserting, so a failure cannot poison the shared lock or cascade into a sibling.

## Summary by CodeRabbit

* **Tests**
  * Added coverage confirming that an empty CAPTCHA API key is treated as unset for trusted origins.
  * Verified that non-empty API keys remain preserved.

### DIFF
--- a/crates/icaptcha-client/src/lib.rs
+++ b/crates/icaptcha-client/src/lib.rs
@@ -720,4 +720,45 @@ mod tests {
 
         assert!(failures.is_empty(), "\n{}", failures.join("\n"));
     }
+
+    #[test]
+    fn new_normalizes_empty_api_key_to_none_on_trusted_origin() {
+        let guard = ICAPTCHA_ENV_LOCK.lock().unwrap();
+        let prev_key = std::env::var_os("GITLAWB_ICAPTCHA_API_KEY");
+        let prev_op = std::env::var_os("GITLAWB_ICAPTCHA_URL");
+
+        // Operator origin configured and resolved as trusted (no advert), so
+        // api_key reaches the env read and the emptiness filter is the only
+        // thing standing between a blank var and an empty bearer token.
+        std::env::set_var("GITLAWB_ICAPTCHA_URL", "https://icap.mynode.example");
+
+        // "Operator set the var but left it blank" must stay keyless: a
+        // Some("") would attach `Authorization: Bearer ` (empty token) to the
+        // challenge and answer requests instead of omitting the header.
+        std::env::set_var("GITLAWB_ICAPTCHA_API_KEY", "");
+        let empty = IcaptchaCfg::new("did:key:zTEST", None, None);
+
+        // Contrast arm: a real key on the same trusted path must survive the
+        // filter, so the filter can't regress to dropping every key.
+        std::env::set_var("GITLAWB_ICAPTCHA_API_KEY", "secret-bearer");
+        let non_empty = IcaptchaCfg::new("did:key:zTEST", None, None);
+
+        // Restore env and release the lock BEFORE asserting, so a failing
+        // assertion can't poison the shared lock or cascade into a sibling.
+        match prev_key {
+            Some(v) => std::env::set_var("GITLAWB_ICAPTCHA_API_KEY", v),
+            None => std::env::remove_var("GITLAWB_ICAPTCHA_API_KEY"),
+        }
+        match prev_op {
+            Some(v) => std::env::set_var("GITLAWB_ICAPTCHA_URL", v),
+            None => std::env::remove_var("GITLAWB_ICAPTCHA_URL"),
+        }
+        drop(guard);
+
+        assert_eq!(
+            empty.api_key, None,
+            "empty GITLAWB_ICAPTCHA_API_KEY must normalize to None"
+        );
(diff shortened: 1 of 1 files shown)


jatmn (approved):
LGTM

beardthelion:
@kevincodex1 ready for merge. jatmn approved the current head (`f35d9a7`), CI is green. Test-only change: binds the empty-API-key filter in `new()` per #216.

kevincodex1 (approved):

PR #246, opened 2026-07-23
fix(icaptcha-client): parse GITLAWB_ICAPTCHA_INSECURE as truthy only (#227)

### LINKED ISSUE #227: Parse GITLAWB_ICAPTCHA_INSECURE value so =0/=false disables the loopback relaxation
`is_https` in `crates/icaptcha-client/src/lib.rs:98` gates its `http`-loopback trust relaxation on `std::env::var_os("GITLAWB_ICAPTCHA_INSECURE").is_some()`, a presence check. An operator who sets `GITLAWB_ICAPTCHA_INSECURE=0` or `=false` intending to disable it still enables the branch: only fully unsetting the variable disables it. Verified against a standalone build: `=0`, `=false`, and empty all yield `is_some() == true`; only unset yields `false`.

The insecure branch is release-reachable (not `#[cfg(test)]`-gated): `is_https` is called on the live path from `resolve_solver_url` (`lib.rs:136`, `:154`).

**Blast radius is bounded, so this is a low-severity footgun, not an exploit.** `resolve_solver_url` still contains a node-advertised loopback URL by two independent gates: the host-allowlist (`allowed(origin_key(a))` at `:154`, which admits only the default `https://icaptcha.gitlawb.com` or the operator's configured `GITLAWB_ICAPTCHA_URL` origin) and the operator-origin key binding (`:166`, the bearer token is attached only when the resolved origin equals the operator origin). So setting the flag alone has no exploitable effect unless the operator also points `GITLAWB_ICAPTCHA_URL` at a loopback origin.

**Repro:** run any `cfg(test)`-off caller with `GITLAWB_ICAPTCHA_INSECURE=0` and observe `is_https("http://127.0.0.1")` returns `true`.

**Fix direction:** parse the value and treat only an explicit truthy set (`"1"`/`"true"`) as enabled, logging when a non-empty non-truthy value is seen; or drop the runtime env flag entirely in favor of a test/example-only injection seam so nothing insecure ships in the release binary.

Notes: this line is pre-existing `main` code (commit `ad7c2b2`, PR #168), not introduced by any open PR. Filing so the value-parse hardening itself has a public tracker; low severity, close if it's not worth pursuing.

Ayush7614:
## Summary
- Treat `GITLAWB_ICAPTCHA_INSECURE` as enabled only for explicit truthy values (`1` / `true`, case-insensitive).
- `=0`, `=false`, empty, and unset leave the loopback HTTP trust relaxation disabled (fixes the presence-check footgun in #227).
- Non-empty non-truthy values log a warning so operators notice misconfiguration.

## Test plan
- [x] `cargo test -p icaptcha-client` — all 23 tests passed
- [x] New regression: `insecure_env_only_truthy_enables_loopback_http` covers `0` / `false` / empty / unset / `1` / `true` / `TRUE`

Fixes #227

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened local loopback HTTP trust to require an explicit `1` or `true` setting, ignoring whitespace and capitalization.
  * Other non-empty values, including `0`, `false`, or arbitrary text, no longer enable insecure behavior and now produce a warning.
  * Untrusted origins no longer receive API keys, while trusted operator connections continue to work as expected.
  * Empty API keys are handled as unavailable rather than being submitted.
  * HTTPS connections remain trusted regardless of the setting.
* **Tests**
  * Added coverage for environment parsing, origin trust, API key handling, and loopback security behavior.


Ayush7614:
@beardthelion — this PR addresses the issue you filed in #227.

Happy to adjust if you’d like a different truthy set or to drop the runtime flag entirely in favor of a test-only seam.

beardthelion (dismissed):
The parse fix is right, and I verified the regression test is load-bearing: reverting `is_https` to the old `var_os(...).is_some()` turns `insecure_env_only_truthy_enables_loopback_http` red on its first assertion, and it also resists a partial regression to "any non-empty value enables". Five things below, the first two worth doing before merge.

## Findings

- [P2] Run `cargo fmt` before this merges
  `crates/icaptcha-client/src/lib.rs:489`
  `cargo fmt --all -- --check` is a required gate in `.github/workflows/pr-checks.yml`, and `cargo fmt -p icaptcha-client --check` exits 1 on this head. The single hunk collapses the `"=1 must enable loopback HTTP"` assert onto one line. Only the quality-signal job has run against the PR so far, so the failure shows up once the full check set does.

- [P2] Pin the trim with a test, or drop it
  `crates/icaptcha-client/src/lib.rs:101`
  `let t = v.trim();` is the only production line in this change that no test holds. I ran the mutations: reverting the guard, swapping the truthy match for `!t.is_empty()`, narrowing to `"1"` only, and flipping `Err(_)` to `true` each turn the suite red, but deleting `.trim()` leaves all 22 lib tests green. One more block in your existing style closes it: `InsecureEnv::with_value(" 1 ")` asserting `is_https("http://127.0.0.1")` is true.

- [P3] Say in the doc comment and the description that the value is trimmed
  `crates/icaptcha-client/src/lib.rs:95`
  Both state the contract as `1` / `true`, case-insensitive, but the value is trimmed first, so `" 1 "` and `"\t1\n"` enable the relaxation too. That behavior is fine; it just is not what the stated contract says.

- [P3] Keep a failing assert from poisoning the env lock
  `crates/icaptcha-client/src/lib.rs:431`
  When I reverted the production line to prove the test bites, two tests failed rather than one: the real assertion, plus `rejects_insecure_advertised_url_on_different_port` panicking with `called Result::unwrap() on an Err value: PoisonError` at `lib.rs:431:49`. So the guard degrades an unrelated test at exactly the moment it does its job. `ICAPTCHA_ENV_LOCK.lock().unwrap_or_else(|e| e.into_inner())` in `with_value` and `unset` avoids that. The pattern predates this PR; the new helpers are where it starts to matter.

- [P3] The new warning does not reach operators by default
  `crates/icaptcha-client/src/lib.rs:106`
  The description says a non-truthy value warns so operators notice the misconfiguration. `gl` builds its subscriber as `EnvFilter::from_default_env().add_directive("gl=info")`, and with `RUST_LOG` unset an `icaptcha_client` WARN is not printed; I saw it only under `RUST_LOG=icaptcha_client=warn`. That gap is pre-existing and already swallows the advertised-URL warning next to it, so either add the directive in `crates/gl/src/main.rs` or soften the claim.

The rest checked out. `is_https` has only the two call sites in `resolve_solver_url`, the one other reader of the variable (`crates/gl/src/http.rs`) sets `"1"` and is unaffected, and every value class that diverges from the old presence check now fails closed, so this is a strict narrowing rather than a behavior swap. Host matching under the relaxation is unchanged.

Ayush7614:
@beardthelion — addressed your review findings:

1. **[P2] fmt** — ran `cargo fmt -p icaptcha-client`; check should be green.
2. **[P2] trim pin** — added `InsecureEnv::with_value(" 1 ")` asserting loopback HTTP is enabled after trim.
3. **[P3] docs** — doc comments now state the value is trimmed (and warn visibility needs `icaptcha_client=warn` / `RUST_LOG`).
4. **[P3] poison lock** — `with_value` / `unset` now use `unwrap_or_else(|e| e.into_inner())` so a failing assert does not cascade.
5. **[P3] warn claim** — softened: documented that the warn is only visible when the `icaptcha_client` target is at WARN+ (did not change `gl`'s default `EnvFilter` here).

`cargo test -p icaptcha-client` — 23 passed locally. Ready for another look.

jatmn (changes requested):

## Findings

- [P2] Do not emit the raw insecure-mode environment value
  `crates/icaptcha-client/src/lib.rs:109`
  The new non-truthy path records the complete trimmed `GITLAWB_ICAPTCHA_INSECURE` value with `value = %t`. That value is arbitrary deployment input, so a mistakenly pasted secret or terminal-control-containing value is written to the tracing sink whenever an HTTP URL is evaluated; `git-remote-gitlawb` enables WARN logging to stderr by default. The warning only needs to report that the setting is invalid—remove the value field or replace it with a bounded, control-safe non-secret classification.

Ayush7614:
@jatmn — addressed your [review](https://github.com/Gitlawb/node/pull/246#pullrequestreview-4780601701):

**[P2] raw env value in warn** — removed `value = %t` from the non-truthy path. The warning now only reports that `GITLAWB_ICAPTCHA_INSECURE` is set but not truthy (expected `1` / `true`); no deployment input is written to the tracing sink.

`cargo test -p icaptcha-client` — 23 passed locally.

jatmn (approved):
@beardthelion LGTM but requires your evaluation

beardthelion (dismissed):
The parse is right and containment holds: with the flag set I drove the host matrix through the real check and only loopback-equivalent hosts are trusted, while `127.0.0.1.evil.com`, `localhost.evil.com`, `169.254.169.254` and `[::1]` are all rejected. `cargo fmt` is clean, the trim is pinned, the doc contract now matches the code, and both the raw-value warning fix and the pure-helper refactor landed. One thing regressed between rounds.

## Findings

- [P2] Cover the line that actually reads the environment
  `crates/icaptcha-client/src/lib.rs:145`
  `icaptcha_insecure_enabled()` is the only production line naming `GITLAWB_ICAPTCHA_INSECURE`, and after the harness was removed nothing exercises it. I mutated it three ways and ran the suite each time: reverting to `std::env::var_os(..).is_some()`, which is exactly the bug #227 is about, 22 passed; hardcoding `true`, so loopback HTTP trust is on unconditionally in release, 22 passed; typo'ing the variable name, so the flag silently never works, 22 passed. At the earlier head the test did bite, because it drove the real env read. The same gap covers the warning: I deleted both call sites of `warn_icaptcha_insecure_non_truthy` and the suite stayed at 22 passed, so the message two reviewers worked on has no guard either. Keep the pure helpers, since they are what fixed the race, and add one serialized test alongside them: take a `Mutex` locked with `unwrap_or_else(|e| e.into_inner())`, save the prior value, assert `icaptcha_insecure_enabled()` is false for unset / `0` / `false` / empty and true for `1` / `true` / `TRUE` / `" 1 "`, then restore. I wrote that and ran it: 23 passed clean, and it fails on all three mutations above. It does not cover the warning emission, and I am not asking you to add a tracing capture harness in this PR.

- [P3] Collapse the space run in the warning
  `crates/icaptcha-client/src/lib.rs:96`
  The literal reads `"...(expected 1 or true);          loopback HTTP trust relaxation remains disabled"`, with ten spaces mid-sentence. It looks like a line-join artifact and it reaches operators verbatim.

Nothing else is outstanding on my side. Unrelated to this PR but worth knowing while you are in here: a test harness bound to `::1` rather than `127.0.0.1` is not trusted even with the flag on, and would quietly fall back to the public default.

Ayush7614:
@beardthelion — addressed your [latest review](https://github.com/Gitlawb/node/pull/246#pullrequestreview-4783669634):

1. **[P2] cover the real env read** — added `icaptcha_insecure_enabled_reads_real_env`: serialized with a poison-recovering `Mutex`, asserts unset / `0` / `false` / empty → false and `1` / `true` / `TRUE` / `" 1 "` → true against the live `std::env::var` path. Pure helpers kept for the race-free parse coverage.
2. **[P3] warning spacing** — collapsed the mid-sentence space run in `warn_icaptcha_insecure_non_truthy`.

`cargo test -p icaptcha-client` — 24 passed locally. Ready for another look.

jatmn (changes requested):

## Findings

- [P3] Collapse the space run in the non-truthy warning
  `crates/icaptcha-client/src/lib.rs:95-98`
  The `tracing::warn!` literal still uses a `\` line continuation, so the emitted message contains a run of spaces between `;` and `loopback` (`...(expected 1 or true);          loopback HTTP trust relaxation remains disabled`). That is the same line-join artifact from the prior review round; the latest commit message claims spacing was collapsed but the continuation indent is still in the string. Please put the sentence on one line (or otherwise join without embedded whitespace) so operators see normal prose.

beardthelion (approved):
The parse is right, and I re-verified it on this head rather than carrying the last round forward. Seven mutations of the changed lines each turn the suite red: reverting `icaptcha_insecure_enabled` to `var_os(..).is_some()`, hardcoding it `true`, typo'ing the variable name, dropping the `.trim()`, widening the match to any non-empty value, narrowing it to exactly `"1"`, and flipping the `NotUnicode` arm to fail open. That is the gap from my last round closed.

I also drove the release-reachable path with a `cfg(test)`-off example through `IcaptchaCfg::new`: `0`, `false`, `FALSE`, `no`, empty and unset all fall back to the public default with the API key withheld, and `1` / `true` / `TRUE` / `" 1 "` all enable. With the flag on, every hostile advert I tried falls back to the operator origin: `127.0.0.1.evil.com`, `localhost.evil.com`, `169.254.169.254`, `[::1]`, `http://user@evil.com`, `https://evil.com`, and a different loopback port. fmt and clippy are clean and the `gl` consumer suite passes.

On the warning spacing: that one is already correct, so please leave it as it is. Rust's `\` line continuation strips the newline *and* the continuation line's leading whitespace, so the emitted message has a single space after the semicolon. I confirmed it twice by execution rather than by eye: compiling the exact literal (length 117, no double space anywhere), and pulling the string back out of the built rlib. Changing it now would be a change with no observable effect.

One thing is still open, and it is mine to land rather than a fifth round for you. `resolve_solver_url` passes `icaptcha_insecure_enabled()` into `resolve_solver_url_with`, and that seam has no coverage in the fail-open direction: hardcoding the argument `true` keeps this crate at 23 passed and `gl` at 11 passed, which is the #227 behaviour restored with nothing going red. The opposite direction is caught, `false` fails two of gl's mockito tests. It pairs with tightening the test lock, since the six resolver tests still read the environment while `EnvGuard` writes it. I ran the suite thirty times and saw no flake, so nothing is wrong today. I will send both as a follow-up.

Thanks for staying with this one through four rounds.

jatmn (approved):
lgtm

beardthelion:
@kevincodex1 LGTM

Issue #242, opened 2026-07-22
Release automation ships a stale Cargo.lock: lock-sync the release PR and build --locked in CI


beardthelion:
Every release-please PR bumps the five workspace crate versions in the Cargo.tomls but never touches Cargo.lock, so each release merge re-creates manifest/lock drift. Current state, verified by execution: main's Cargo.lock records all five workspace crates at 0.5.1 while the manifests say 0.6.0 (and 0.7.0 once #240 merges), and cargo metadata --locked on main exits 101 ("cannot update the lock file because --locked was passed").

Consequences:
- A --locked build from any release tag fails, so the lockfile we ship is dead weight for reproducible builds.
- The release binaries are built from a resolver-fresh dependency set, not the committed lock, which quietly weakens the lock's supply-chain value.
- Any plain cargo build on a drifted checkout rewrites the lock's version fields, so contributors keep hitting mystery-dirty trees (this has already produced repeated hand-sync commits, which fix it only until the next release merges).

Hand syncs are the wrong tool: the drift is reintroduced by automation, so the fix has to live in automation.

Proposed fix:
1. A lock-sync step in the release flow so the release PR itself carries the Cargo.lock bump (cargo update --workspace on the release branch), making the tagged tree internally consistent.
2. Build with --locked in PR CI so any future manifest/lock drift fails loudly at PR time instead of surfacing as dirty trees later.

Sequencing note: the --locked CI gate can only land together with (or after) a lock sync, and a sync commit must postdate the 0.7.0 release merge or it re-breaks immediately.

beardthelion:
Note: #185 was the earlier report of this same drift; closed in favor of this issue, which #243 closes.

PR #248, opened 2026-07-25
fix(node): use readiness for peer liveness

### LINKED ISSUE #176: Peer liveness ping targets /health (constant 200), so a mid-life DB outage doesn't drain peer traffic
Follow-up to #170. Not a regression: #170's startup resilience is a net improvement. This is a gap in how the mesh consumes it.

## Gap

#170 added a DB-probed `/ready` (503 on outage) and gates Fly routing/deploys on it. But peer liveness still pings `/health`, which the full server pins at a constant 200 with no DB check (`server.rs:484`). `ping_peer_health` (`main.rs:962`) and `api/peers.rs:434` both target `/health` and treat any 2xx as alive.

## Evidence (executed)

Built the full router against a state whose DB is unavailable and drove both endpoints:

- `/health` -> 200 OK (ignores DB state)
- `/ready` -> 503 Service Unavailable (reflects DB state)

## Consequence

A DB outage that happens *after* startup leaves `/health` at 200 forever, so peers keep counting the node alive and routing sync/gossip to it while every DB-touching endpoint 503s. #170's drain-on-outage goal holds for the Fly edge (which reads `/ready`) but not the P2P mesh. The commit message's claim that peers "stop counting a DB-less node as alive" is only true during the startup window, where the degraded server 503s `/health`.

## Direction

Point the peer liveness ping at `/ready` (readiness), or fold a DB-state check into the peer-routing decision. Keep `/health` as pure liveness so Fly's kill/restart semantics are unaffected.

MikeTomlin19:
## Summary

Use the database-aware `/ready` endpoint for periodic and manual peer checks, while falling back to `/health` only when an older peer returns 404 for `/ready`. Periodic federation state now requires two consecutive failed samples before marking a peer unreachable; the anonymous manual-ping endpoint is read-only.

## Motivation & context

`/health` intentionally reports process liveness and remains successful during a database outage. Using it for peer reachability lets a node continue treating a peer as healthy even though that peer cannot serve database-backed requests. A single transient readiness failure also should not remove a peer from federated repository results for the next five-minute interval.

Closes #176

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [x] Docs
- [x] Tests / CI

## What changed

- Changed the shared peer probe from `/health` to `/ready`, with a 404-only `/health` fallback for nodes released before `/ready` existed.
- Reused that helper from the manual peer-ping API so both paths have identical readiness and no-redirect behavior.
- Kept the anonymous manual-ping endpoint read-only; only the periodic gossip loop updates persisted federation reachability.
- Required two consecutive periodic failures before marking a peer unreachable, while restoring reachability immediately after a successful sample.
- Added structured logs for readiness failures, legacy fallback, fallback failures, transport failures, and timeout exhaustion.
- Added regression coverage proving the manual handler uses readiness without mutating the federation gate, one transient failure preserves the gate, sustained failure clears it, success restores it, legacy peers remain compatible, and neither request path follows redirects.
- Updated the operator guide and degraded-router comment to distinguish liveness from readiness.

## How a reviewer can verify

```sh
DATABASE_URL=postgres://postgres:postgres@localhost:5432/gitlawb_test cargo test --workspace
cargo fmt --all -- --check
cargo build --release --workspace
```

The focused regressions can also be run with:

```sh
cargo test -p gitlawb-node gossip_ssrf_tests
DATABASE_URL=postgres://postgres:postgres@localhost:5432/gitlawb_test \
  cargo test -p gitlawb-node manual_ping_uses_readiness_without_mutating_federation_gate
```

## Before you request review

## Protocol & signing impact

N/A — this does not change identity, signatures, UCANs, ref certificates, or P2P wire formats.

## Notes for reviewers

The repository's current MSRV lane fails before compilation because newly resolved AWS crates require Rust 1.91.1 while CI installs `1.91` (currently 1.91.0). That existing manifest/lock/dependency drift is tracked in #242 and #243; stable tests, clippy, formatting, and the release build are clean for this change.

## Summary by CodeRabbit

* **New Features**
  * Peer monitoring now uses the database-aware `/ready` endpoint, with fallback to `/health` when unavailable.
* **Bug Fixes**
  * Peer reachability status now requires multiple consecutive failed readiness checks before marking a peer unreachable.
  * Successful checks restore reachability immediately, while stale failure records are removed.
  * Improved handling of redirects, HTTP errors, connection failures, and readiness-to-health fallback timing.
  * Health checks no longer alter persisted reachability status during individual probes.
* **Documentation**
  * Updated the public URL checklist to require both `GET /health` and `GET /ready`.

### DIFF
--- a/crates/gitlawb-node/src/api/peers.rs
+++ b/crates/gitlawb-node/src/api/peers.rs
@@ -444,8 +444,9 @@ pub async fn ping_peer(
         .ok_or_else(|| AppError::RepoNotFound(format!("peer {did} not found")))?;
 
     // Use DB-aware readiness rather than process liveness: `/health` stays 200
-    // during a mid-life database outage, while `/ready` reports 503. The shared
-    // helper also preserves the no-redirect SSRF guard on peer-controlled URLs.
+    // during a mid-life database outage, while `/ready` reports 503. A 404-only
+    // fallback keeps pre-readiness peers compatible. The shared helper also
+    // preserves the no-redirect SSRF guard on peer-controlled URLs.
     let ok = crate::ping_peer_readiness(&state.http_client, &peer.http_url).await;
 
     let _ = state.db.mark_peer_ping(&did, ok).await;
--- a/crates/gitlawb-node/src/api/peers.rs
+++ b/crates/gitlawb-node/src/api/peers.rs
@@ -449,8 +449,9 @@ pub async fn ping_peer(
     // preserves the no-redirect SSRF guard on peer-controlled URLs.
     let ok = crate::ping_peer_readiness(&state.http_client, &peer.http_url).await;
 
-    let _ = state.db.mark_peer_ping(&did, ok).await;
-
+    // This anonymous diagnostic endpoint is intentionally read-only. The
+    // periodic gossip loop owns the persisted federation-reachability gate and
+    // requires consecutive failures before marking a peer unreachable.
     Ok(Json(serde_json::json!({
         "did": did,
         "http_url": peer.http_url,
@@ -557,7 +558,7 @@ mod tests {
     use axum::body::Body;
     use axum::extract::ConnectInfo;
     use axum::http::{header, Method, Request, StatusCode};
-    use axum::routing::post;
+    use axum::routing::{get, post};
     use axum::{middleware, Extension, Router};
     use gitlawb_core::http_sig::sign_request;
     use gitlawb_core::identity::Keypair;
@@ -767,6 +768,74 @@ mod tests {
             .await
             .unwrap();
         did
+    }
+
+    #[sqlx::test]
+    async fn manual_ping_uses_readiness_without_mutating_federation_gate(pool: PgPool) {
+        let mut server = mockito::Server::new_async().await;
+        let ready = server
+            .mock("GET", "/ready")
+            .with_status(503)
+            .expect(1)
+            .create_async()
+            .await;
+        let health = server
+            .mock("GET", "/health")
+            .with_status(200)
+            .expect(0)
+            .create_async()
+            .await;
+
+        let mut state = test_state(pool.clone()).await;
+        state.http_client =
+            Arc::new(crate::build_http_client().expect("production HTTP client should build"));
+        let did = Keypair::generate().did().to_string();
+        let now = chrono::Utc::now().to_rfc3339();
+        sqlx::query(
+            "INSERT INTO peers (did, http_url, last_seen, last_ping_ok, announced_at)
+             VALUES ($1, $2, $3, TRUE, $3)",
+        )
+        .bind(&did)
+        .bind(server.url())
+        .bind(now)
+        .execute(&pool)
+        .await
+        .unwrap();
+
+        let response = Router::new()
+            .route("/api/v1/peers/{did}/ping", get(super::ping_peer))
+            .with_state(state.clone())
+            .oneshot(
+                Request::builder()
+                    .method(Method::GET)
+                    .uri(format!("/api/v1/peers/{did}/ping"))
+                    .body(Body::empty())
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(response.status(), StatusCode::OK);
+        let body = axum::body::to_bytes(response.into_body(), usize::MAX)
+            .await
+            .unwrap();
+        let json: serde_json::Value = serde_json::from_slice(&body).unwrap();
+        assert_eq!(json["reachable"], false);
+        ready.assert_async().await;
+        health.assert_async().await;
+
+        let peer = state
+            .db
+            .list_peers()
+            .await
+            .unwrap()
+            .into_iter()
+            .find(|peer| peer.did == did)
+            .unwrap();
+        assert!(
+            peer.last_ping_ok,
+            "an anonymous manual sample must not rewrite the federation gate"
+        );
     }
 
     #[sqlx::test]
--- a/crates/gitlawb-node/src/api/peers.rs
+++ b/crates/gitlawb-node/src/api/peers.rs
@@ -443,18 +443,10 @@ pub async fn ping_peer(
         .find(|p| p.did == did)
         .ok_or_else(|| AppError::RepoNotFound(format!("peer {did} not found")))?;
 
-    // Async ping
-    let url = format!("{}/health", peer.http_url.trim_end_matches('/'));
-    // Use the shared no-redirect client: bare `reqwest::get` follows redirects,
-    // so a peer could answer with `302 -> http://127.0.0.1/` and turn the ping
-    // into an SSRF probe.
-    let ok = state
-        .http_client
-        .get(&url)
-        .send()
-        .await
-        .map(|r| r.status().is_success())
-        .unwrap_or(false);
+    // Use DB-aware readiness rather than process liveness: `/health` stays 200
+    // during a mid-life database outage, while `/ready` reports 503. The shared
+    // helper also preserves the no-redirect SSRF guard on peer-controlled URLs.
+    let ok = crate::ping_peer_readiness(&state.http_client, &peer.http_url).await;
 
     let _ = state.db.mark_peer_ping(&did, ok).await;
 
--- a/crates/gitlawb-node/src/main.rs
+++ b/crates/gitlawb-node/src/main.rs
@@ -1101,16 +1101,19 @@ async fn sweep_rate_limiters(state: &AppState) {
 /// runs, not a hand-rolled equivalent.
 fn build_http_client() -> reqwest::Result<reqwest::Client> {
     reqwest::Client::builder()
-        .timeout(std::time::Duration::from_secs(10))
+        .timeout(OUTBOUND_HTTP_TIMEOUT)
         .redirect(reqwest::redirect::Policy::none())
         .build()
 }
+
+const OUTBOUND_HTTP_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
 
 /// Ping a peer's DB-aware `/ready` endpoint and report whether it answered 2xx.
 ///
 /// A 404 falls back to `/health` for compatibility with nodes released before
 /// `/ready` existed. Other readiness failures, including 503 during a database
-/// outage, fail closed and must not use the liveness-only endpoint.
+/// outage, fail closed and must not use the liveness-only endpoint. The complete
+/// `/ready`-then-`/health` probe shares one timeout budget.
 ///
 /// Takes the client by reference so callers supply the shared, no-redirect
 /// `state.http_client`. Peer URLs are attacker-influenceable, so a `3xx` to a
@@ -1118,19 +1121,31 @@ fn build_http_client() -> reqwest::Result<reqwest::Client> {
 /// `reqwest::Client::new()`: its default follows redirects and would
 /// reintroduce the SSRF this guards against (#93).
 pub(crate) async fn ping_peer_readiness(client: &reqwest::Client, http_url: &str) -> bool {
-    let base_url = http_url.trim_end_matches('/');
-    let readiness = client.get(format!("{base_url}/ready")).send().await;
+    ping_peer_readiness_with_timeout(client, http_url, OUTBOUND_HTTP_TIMEOUT).await
+}
 
-    match readiness {
-        Ok(response) if response.status().is_success() => true,
-        Ok(response) if response.status() == reqwest::StatusCode::NOT_FOUND => client
-            .get(format!("{base_url}/health"))
-            .send()
-            .await
-            .map(|response| response.status().is_success())
-            .unwrap_or(false),
-        _ => false,
-    }
+async fn ping_peer_readiness_with_timeout(
+    client: &reqwest::Client,
+    http_url: &str,
+    timeout: std::time::Duration,
+) -> bool {
+    let base_url = http_url.trim_end_matches('/');
+    tokio::time::timeout(timeout, async {
+        let readiness = client.get(format!("{base_url}/ready")).send().await;
+
+        match readiness {
+            Ok(response) if response.status().is_success() => true,
+            Ok(response) if response.status() == reqwest::StatusCode::NOT_FOUND => client
+                .get(format!("{base_url}/health"))
+                .send()
+                .await
+                .map(|response| response.status().is_success())
+                .unwrap_or(false),
+            _ => false,
+        }
+    })
+    .await
+    .unwrap_or(false)
 }
 
 fn load_or_create_keypair(config: &Config) -> Result<Keypair> {
@@ -1168,7 +1183,11 @@ fn load_or_create_keypair(config: &Config) -> Result<Keypair> {
 
 #[cfg(test)]
 mod gossip_ssrf_tests {
-    use super::ping_peer_readiness;
+    use super::{ping_peer_readiness, ping_peer_readiness_with_timeout};
+    use axum::{http::StatusCode, routing::get, Router};
+    use std::sync::atomic::{AtomicUsize, Ordering};
+    use std::sync::Arc;
+    use std::time::Duration;
 
     // Build the client exactly as production does (super::build_http_client) so
     // these tests bind the redirect guarantee to the real shared client the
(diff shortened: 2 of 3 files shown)


MikeTomlin19:
For reviewer context, the `needs-tests` signal is a triage false positive rather than missing coverage.

This PR adds regression coverage inside the existing `#[cfg(test)]` module in `crates/gitlawb-node/src/main.rs`, using Tokio's `#[tokio::test]` attribute. In particular, `ping_peer_readiness_ignores_liveness_only_health` proves that `200 /health` plus `503 /ready` is treated as unready and that `/health` is never requested.

The triage workflow currently recognizes newly added literal `#[test]` and `#[cfg(test)]` lines, but not `#[tokio::test]`, so it misses these inline async tests.

Validation completed locally:

- `cargo test -p gitlawb-node gossip_ssrf_tests` — 4 passed
- `cargo test --workspace` against PostgreSQL 16 — passed
- `cargo fmt --all -- --check` — passed
- `cargo clippy --workspace --all-targets -- -D warnings` — passed

jatmn (changes requested):

## Findings

- [P1] Preserve compatibility with peers that do not implement `/ready`
  `crates/gitlawb-node/src/main.rs:985`
  This unconditionally changes the peer wire contract from `/health` to `/ready`. Released v0.5.0 nodes expose `/health` but not `/ready`, and #170 deliberately left the peer probe on `/health` for exactly that rolling-upgrade case. Their `/ready` response is therefore 404, which this helper maps to `false`; both the periodic job and the manual ping persist that value. Since federated repository listing filters on `last_ping_ok`, a healthy v0.5.0 peer disappears from federation after the first probe. Please add a version/capability transition (for example, a 404-only legacy `/health` fallback that still treats a real `/ready` 503 as unready) and cover the mixed-version case, or provide the required coordinated rollout and compatibility window.

- [P2] Update the operator peer-endpoint requirement
  `docs/RUN-A-NODE.md:180`
  The operator guide currently says that serving `/health` is sufficient because peers ping it. After this change peers instead require `/ready`, so an operator following the documented contract can be reported unreachable. Update the guidance to require `/ready` (and retain `/health` only for its liveness purpose).

- [P3] Correct the degraded-router peer-ping comment
  `crates/gitlawb-node/src/main.rs:735`
  This comment still says peer pings use a successful `/health` response. The changed callers now use `/ready`, so the comment gives the wrong rationale for the degraded response and is likely to mislead future outage-routing work.

MikeTomlin19:
Addressed all three findings in `f9b4ce3`:

- `/ready` now falls back to `/health` only on a 404, preserving compatibility with pre-`/ready` nodes while keeping real readiness failures such as 503 fail-closed.
- Added mixed-version coverage (`404 /ready` + `200 /health`) and redirect-safety coverage for the legacy fallback. The existing test still proves `503 /ready` never falls back to a successful `/health`.
- Updated the operator checklist to require both liveness `/health` and DB-aware `/ready`.
- Corrected the degraded-router comment to describe peer readiness probes rather than 2xx `/health` pings.

Validation:

- `cargo test -p gitlawb-node gossip_ssrf_tests` — 6 passed
- `cargo test --workspace` against PostgreSQL 16 — passed
- `cargo fmt --all -- --check` — passed
- `cargo clippy --workspace --all-targets -- -D warnings` — passed

jatmn (changes requested):

## Findings

- [P2] Keep the legacy fallback within one peer-probe deadline
  `crates/gitlawb-node/src/main.rs:988`
  The shared client gives each request a 10-second timeout, but the new 404 fallback starts a second independently timed `/health` request. A legacy (or attacker-controlled public) peer can therefore hold the serial gossip loop for nearly 20 seconds by delaying `/ready` before returning 404 and delaying `/health`; the prior probe had a single 10-second budget. Since the peer table is unbounded and the loop updates one row at a time, enough such rows cause sweeps to overrun the five-minute interval and leave peer reachability stale. Apply one deadline to the complete `/ready`-then-`/health` operation (or pass the remaining budget to the fallback), and cover a delayed-404 case.

jatmn (approved):
@beardthelion LGTM, requires your manual review.

beardthelion (dismissed):
The readiness switch is the right call and all four findings from the earlier rounds landed properly. I confirmed the compatibility story by execution rather than by reading it: an axum router carrying only `/health` and no fallback, which is v0.5.0's shape, answers 404 for `GET /ready`, and the probe falls through to `/health` and returns ready. `/ready` first shipped in v0.5.1, so the peers that arm exists for really do return 404. A 403 and a 500 both fail closed without touching `/health`.

I also mutated each line inside `ping_peer_readiness_with_timeout` one at a time against the seven tests. Pointing the probe at `/health`, widening the fallback trigger past 404, giving the fallback its own timeout budget, letting the client follow redirects, flipping the catch-all to true, and deleting the 404 arm each turn at least one test red. That part is well covered. Four things below.

## Findings

- [P2] Bind the manual ping path to readiness with a test
  `crates/gitlawb-node/src/api/peers.rs:442`
  Reverting this line to the pre-PR inline `/health` probe leaves all seven tests green. Every line inside the helper reddens under mutation, so the harness works and this is a real hole: the handler's readiness behavior is unpinned, and a later edit can quietly return `GET /api/v1/peers/{did}/ping` to liveness-only while CI stays green. A test that drives the handler (or at minimum asserts the call site's behavior) closes it.

- [P2] Do not let a single probe sample drop a peer from federation
  `crates/gitlawb-node/src/main.rs:948-949`
  The probe result is written to `last_ping_ok` unconditionally, and `api/repos.rs:1451` filters the federated fan-out on that flag, so one failed sample removes a peer's repos from `/api/v1/repos/federated` until the next 300s tick. The single-sample write predates this PR; what changes is the set of states that flip it, since a peer whose database hiccups for two seconds now drops out where previously only an unreachable process did. Two consecutive failures before writing false, or one re-probe inside the existing budget, would fix it. Whichever shape you pick, the property to preserve is that a peer survives one transient failure and still drops out on a sustained one.

- [P3] Log the probe outcome
  `crates/gitlawb-node/src/main.rs:1000-1017`
  Five materially different exits collapse into one bool with no `tracing` on any arm: ready, database-degraded 503, legacy 404 then healthy, refused redirect, and budget exhaustion. From the logs an operator cannot tell a peer whose database is down from one that is unreachable, which is the distinction this PR exists to draw. The bootstrap announce in the same function logs both its error and its timeout arm, so this is a gap against the file's own habit. The 503 arm and the legacy-404 downgrade are the two worth recording; the downgrade also gives you the list of peers still on the old build, which is what tells you when the fallback can be dropped.

- [P3] Reconsider the anonymous ping route writing the federation gate
  `crates/gitlawb-node/src/api/peers.rs:444`, route wired at `crates/gitlawb-node/src/server.rs:292`
  `peer_read_routes` carries no auth layer and no per-IP brake, while `sync_trigger_routes` and `peer_write_routes` in the same file each carry one with a comment explaining that outbound fan-out needs it. This handler now drives a database-backed probe on the target and writes the flag that gates federation, from a single unauthenticated sample. Same lineage as the finding above, and the cheaper fix is probably to report the probe result without calling `mark_peer_ping`, leaving the flag owned solely by the gossip loop.

## Not asks, just recording them

The probe URL is built by string concatenation, so a peer that announces `https://host/?x=1` gets probed at `/` rather than `/ready`. I ran it: the `/ready` mock is never requested, `/` answers 200, and the peer reads ready even though `/ready` would have said 503. `is_public_http_url` validates scheme and host only, and `upsert_peer` stores the URL verbatim, so that shape survives announce. It is the same concatenation the pre-PR code used, and a hostile peer controls its own answer regardless, so it only really bites an honestly misconfigured operator. `Url::join` would be the tidy follow-up, separate from this PR.

`README.md:276` and `docs/OSS-READINESS-AUDIT.md:69` still list `/health` alone. Both predate this change and neither claims to state the peer contract, so I am not asking you to touch them here.

beardthelion (dismissed):
Both of my round-three asks landed, and both are now pinned by a test that fails without them. I re-ran the mutation matrix on this head rather than carrying last round's results forward: pointing the probe at `/health` reddens four tests, widening the fallback trigger past 404 reddens one, giving the fallback a fresh budget reddens one, flipping the catch-all to true reddens two, and letting the client follow redirects reddens two. Re-adding `mark_peer_ping` to `ping_peer`, or putting that handler back on an inline `/health` probe, each redden the new handler test.

The new hysteresis is the exception, and it is the one thing I want before merge.

## Findings

- [P2] Pin the gossip loop's use of the hysteresis, not just the helper
  `crates/gitlawb-node/src/main.rs:956`
  I replaced the whole `match peer_ping_db_update(...)` block with the pre-PR `let _ = state.db.mark_peer_ping(&peer.did, ok).await;`, which deletes the feature outright, and everything stayed green: 8 tests in `gossip_ssrf_tests`, 20 in `api::peers::tests`. The unit test covers the pure function and nothing covers its only call site, so this round's work is removable without CI noticing. Same shape as the manual-ping hole from last round, which you closed well. Deleting the `failed_once.retain` prune is green too, so one test can cover both. I built the fix before asking for it: extract the tick body into `async fn gossip_ping_round(db: &crate::db::Db, client: &reqwest::Client, failed_once: &mut HashSet<String>, peers: Vec<crate::db::PeerRecord>)`, call it from the `select!` arm, and drive it from a `#[sqlx::test]` against a mockito peer answering `/ready` 503: after round one `last_ping_ok` is still TRUE, after round two it is FALSE. That is green as written and RED once the hysteresis inside the extracted function is deleted. Seed the peer with a raw INSERT the way your `manual_ping...` test does; `upsert_peer` rejects a loopback `http_url` and mockito binds on 127.0.0.1.

- [P3] Give the gossip interval `MissedTickBehavior::Delay`
  `crates/gitlawb-node/src/main.rs:940`
  The hysteresis is counted in ticks, but its value is wall-clock: the second sample is meant to be 300s after the first. `tokio::time::interval` defaults to `MissedTickBehavior::Burst` in the pinned tokio 1.50, so a sweep that overruns the period is followed by catch-up ticks and the two samples can land seconds apart, which is the mass-failure case the guard exists for. A sequential sweep at up to 10s per peer needs roughly 30 unreachable peers to get there. Adding `interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);` after the constructor is the whole change and it compiles clean. Worst case today is the guard degrading to the old behavior rather than doing something wrong, hence P3.

## Not asks, recorded

- The anonymous ping route still has no per-IP brake, and `/ready` costs the target a database query where `/health` did not. I checked the delta, not the state: the pre-PR handler made the same unauthenticated outbound request and additionally wrote `last_ping_ok`, and the probe now sits inside a single 10s budget covering both legs. This PR moves that surface in the right direction; the brake is a separate question and not something I want in this PR.
- `last_seen` is no longer refreshed on a first-failure tick, since that tick skips the write. Every reader is display or ordering and none gates on freshness. No action.
- A peer alternating ready/unready is never persisted unreachable, because one success clears the set. That is the tradeoff behind the invariant I gave last round: survive one transient failure, drop out on a sustained one. Settled as intended, not open.

Scope is closed after this. The two commits since my last round have never been through CI: `PR Checks` is awaiting approval on both, and the last green run was a1072e86, before the hysteresis and the prune.

MikeTomlin19:
Addressed the final review findings in `32e4578`:

- Extracted one periodic sampling pass into `gossip_ping_round`.
- Added a PostgreSQL-backed test that drives a mock `/ready` 503 peer through the real round function: the first round preserves `last_ping_ok = TRUE`, the second persists `FALSE`, and removed peers are pruned from the bounded failure set.
- Set the five-minute interval's missed-tick behavior to `Delay`.

Validation:

- `cargo test --workspace` against PostgreSQL 16 — passed
- `cargo fmt --all -- --check` — passed
- `cargo clippy --workspace --all-targets -- -D warnings` — passed

beardthelion (approved):
Approving. The premise is pinned now: pointing the probe back at `/health` reddens six tests across both modules, and reverting `gossip_ping_round`'s match to the pre-PR `let _ = db.mark_peer_ping(&peer.did, ok).await;` fails `gossip_ping_round_requires_two_failures_before_persisting_unreachable`. That was my round-four finding and it is closed by execution rather than by reading the new test.

I re-ran the whole matrix on this head instead of carrying earlier results forward. Deleting the hysteresis in `peer_ping_db_update` reddens two tests, deleting the `failed_once.retain` prune reddens one, widening the fallback trigger past 404 reddens two, unbinding the shared `/ready`+`/health` deadline reddens one, letting the client follow redirects reddens two, flipping the transport-error arm to true reddens one, and re-adding `mark_peer_ping(&did, ok)` to the manual handler or putting it back on an inline `/health` probe each redden the handler test.

CI is green on this head across all nine jobs. Your note about the MSRV lane failing on the AWS crate drift is stale: the base is current with main and `MSRV (Rust 1.91)` passes.

## Findings

None blocking.

## Recorded, not asks

- I overstated one thing last round. I wrote that re-adding `mark_peer_ping` to `ping_peer` reddens the handler test. That holds for the faithful revert, which writes the probe result: `mark_peer_ping(&did, ok)` with `/ready` at 503 writes false and the test fails. It does not hold for a write of a hardcoded `true`, which stays green because the test seeds `last_ping_ok = TRUE` and asserts the same value. `mark_peer_ping` also stamps `last_seen` (`db/mod.rs:2106`), so asserting `last_seen` is byte-identical would pin the no-write property in both directions. Not asking for it.
- `gossip_task` has no test, so the cross-tick lifetime of `failed_once` is unpinned: moving `let mut failed_once` (`main.rs:942`) inside the `interval.tick()` arm turns the hysteresis into a no-op and all 29 tests stay green, as does deleting the `MissedTickBehavior::Delay` line. This is the next layer of the shape I asked about last round, and pinning it needs the loop extracted the way the round was. Not asking for that here: the feature is pinned at the pure function and at the round, and the residue is loop scaffolding.
- Dropping the `Some(true)` arm's write leaves all 29 green. A regression there strands a peer as permanently unreachable with no operator path back, since the manual ping is read-only by design now. Recording it, not asking.
- The two-tick guard is spaced in ticks but argued in wall-clock. Probes are serial at up to 10s each and `list_peers` orders by `last_seen DESC` (`db/mod.rs:2118`), so a peer probed last in one round and first in the next can have its samples land seconds apart. That is traced from the code rather than measured. It takes roughly 30 unreachable peers to get there and the cost is one peer missing from `/api/v1/repos/federated` for a cycle, so I am leaving it. Keying first-failure on an `Instant` would make it wall-clock by construction if it ever bites.
- `peer_read_routes` (`server.rs:290`) still carries no auth layer and no per-IP brake, unlike `sync_trigger_routes` and `peer_write_routes` in the same file. What changes here is the shape of the probe, not its cost to us: I checked the pre-PR handler and it already made one outbound request under the same 10s client timeout and additionally wrote `mark_peer_ping`, so our own per-request DB work goes down. What is new is that a peer answering 404 draws a second outbound leg, and the probe lands on `/ready`, which runs a query on the target where `/health` (`server.rs:497`) ran none. Filing that separately rather than growing this PR.
- Not yours: `upsert_peer` (`db/mod.rs:2093`) rewrites `http_url` on conflict and leaves `last_ping_ok` alone, and announce accepts unsigned bodies while `require_signed_peer_writes` defaults false, so a repointed peer URL inherits `reachable=true`. Pre-existing, I will file it.

Scope is closed.

beardthelion:
@jatmn this was force-push rebased onto c83cbc5 after you approved, so the approval state you see is still the one from a1072e86 and GitHub carried it forward on its own.

The rebase itself is clean. The branch's own diff is unchanged, and the head-to-head delta is exactly main's, which I checked over the whole tree rather than just the touched files. Nothing came in with it.

What your approval predates is the commits after a1072e86, and that is the part worth your time. `ping_peer` no longer writes the federation gate at all, so the anonymous endpoint is read-only now. The gossip loop gained hysteresis: `peer_ping_db_update` plus the `gossip_ping_round` extraction in `main.rs`, with a timeout wrapper on the probe. One failed readiness probe no longer marks a peer unreachable, it takes two consecutive failures, so a peer that fails once stays reachable for another five-minute tick. That availability tradeoff is the call I would like a second read on.

CI is green on the rebased head apart from `cargo audit`, which is the ruint advisory hitting every branch right now (#292 is the bump). Nothing in this diff touches dependencies.

jatmn (approved):
LGTM

beardthelion:
@kevincodex1 this is ready to merge. jatmn and I have both approved, and CI is green across the board on b4c9ad7b.

MikeTomlin19:
Rebased onto current `main` (`fdf716d`) with the six-commit PR delta preserved by range-diff. `cargo fmt --all -- --check` passes. The focused node test is blocked locally because this host has Rust 1.91.0 while the rebased lockfile requires 1.91.1; CI will validate on its configured toolchain.

kevincodex1 (approved):

kevincodex1:
approved, kindly rebase to main and fix conflicts

kevincodex1:
@coderabbitai please do a full code review

PR #254, opened 2026-07-27
fix(node): map ipfs/arweave DB outages to 503 via bare ? (#251)

### LINKED ISSUE #251: Database outage renders as 500 internal_error, not the retryable 503, on /ipfs/pins, /ipfs/{cid} and /arweave/anchors
## What happens

On `main` (111cff7), when the database is unreachable, three routes return `500 internal_error` where every sibling route returns `503 db_unavailable`. Observed through the real mounted router against a closed pool, all four requests anonymous and unsigned:

```
/api/v1/arweave/anchors -> 500 {"error":"internal_error","message":"attempted to acquire a connection on a closed pool"}
/api/v1/ipfs/pins       -> 500 {"error":"internal_error","message":"attempted to acquire a connection on a closed pool"}
/api/v1/peers           -> 503 {"error":"db_unavailable","message":"database is temporarily unavailable"}
/api/v1/repos           -> 503 {"error":"db_unavailable","message":"database is temporarily unavailable"}
```

Same process, same pool, same request cycle. The split is per handler, not a deliberate policy.

## Why

The db layer returns `anyhow::Result`, and `impl From<anyhow::Error> for AppError` in `crates/gitlawb-node/src/error.rs` is the only place that downcasts a `sqlx::Error` back out of the anyhow chain into `AppError::Db`. `IntoResponse` then sends `AppError::Db(e)` through `db_unavailable(e)`, which maps the connection-level variants (`PoolTimedOut`, `PoolClosed`, `Io`, `Tls`) to 503.

A handler that writes `.map_err(AppError::Internal)` rather than a bare `?` constructs the `Internal` variant directly, so that downcast never runs and the error takes the unconditional 500 arm.

Sites on main:

- `crates/gitlawb-node/src/api/ipfs.rs:84`
- `crates/gitlawb-node/src/api/ipfs.rs:94`
- `crates/gitlawb-node/src/api/ipfs.rs:224`
- `crates/gitlawb-node/src/api/arweave.rs:33`

## Why it matters

503 is the code that means "temporary, retry", and 500 means "this request is broken". A proxy, an uptime check, or a client retry policy keyed on 503 will not fire for these routes during a database outage, so a recoverable condition reads as a hard failure on exactly the metadata endpoints most likely to be polled. All three routes are unauthenticated (`/api/v1/ipfs/pins` and `/api/v1/arweave/anchors` carry no auth layer, `/ipfs/{cid}` carries only `optional_signature`), so the behavior is observable by anyone.

## Scope

Pre-existing, and not introduced by #247. That PR makes the body of both 500 arms opaque, which is a separate and correct change; the status split survives it, which is why this is its own issue rather than a review comment there.

#134 is currently in flight on `api/ipfs.rs` and `api/arweave.rs`. It rewrites the arweave call site and adds three more `.map_err(AppError::Internal)?` sites, so a fix here should land after it or be coordinated with it, and should cover the sites #134 adds rather than only the four above.

## Fix direction

Use a bare `?` at these call sites so the existing `From<anyhow::Error>` conversion runs and connection-class failures reach the 503 arm. Non-sqlx errors still land in `Internal` through that same conversion, so nothing else changes shape. I have not compiled that change, so this is the direction rather than a verified patch.

A regression test would drive one of these routes against a closed pool and assert 503 plus `db_unavailable`, which is the shape the reproduction above already uses: build the router from a migrated test pool, call `pool.close()`, then issue the request.

Ayush7614:
## Summary
- Replace `.map_err(AppError::Internal)` with bare `?` on `/api/v1/ipfs/pins`, `GET /ipfs/{cid}`, and `/api/v1/arweave/anchors` DB calls.
- Connection-class sqlx failures (`PoolClosed`, etc.) now downcast through `From<anyhow::Error>` into `AppError::Db` and return **503 `db_unavailable`**, matching `/api/v1/repos` and `/api/v1/peers`.
- Non-sqlx errors still become `Internal` via that same conversion.

## Test plan
- [x] `cargo test -p gitlawb-node pool_closed_via_anyhow_from_is_503_db_unavailable` — downcast path
- [x] `list_pins_closed_pool_returns_503_db_unavailable` — closed-pool route (CI / `DATABASE_URL`)
- [x] `list_anchors_closed_pool_returns_503_db_unavailable` — closed-pool route (CI / `DATABASE_URL`)

Fixes #251

## Summary by CodeRabbit

* **Bug Fixes**
  * Database connection failures now return a clear **503 Service Unavailable** response instead of an internal server error.
  * Improved error reporting for Arweave anchor and IPFS pin queries when the database is unavailable.
  * Added coverage to verify consistent database-unavailable responses across affected endpoints.

### DIFF
--- a/crates/gitlawb-node/src/api/arweave.rs
+++ b/crates/gitlawb-node/src/api/arweave.rs
@@ -26,14 +26,61 @@ pub async fn list_anchors(
     Query(q): Query<ListAnchorsQuery>,
 ) -> Result<Json<serde_json::Value>> {
     let limit = q.limit.min(200);
+    // Bare `?` so connection-class sqlx failures downcast to `AppError::Db` and
+    // map to 503 `db_unavailable` (not 500 via `.map_err(AppError::Internal)`) (#251).
     let anchors = state
         .db
         .list_arweave_anchors(q.repo.as_deref(), limit)
-        .await
-        .map_err(crate::error::AppError::Internal)?;
+        .await?;
 
     Ok(Json(serde_json::json!({
         "anchors": anchors,
         "count": anchors.len(),
     })))
+}
+
+#[cfg(test)]
+mod closed_pool_tests {
+    use super::*;
+    use axum::http::{Request, StatusCode};
+    use axum::Router;
+    use serde_json::Value;
+    use sqlx::PgPool;
+    use tower::ServiceExt;
+
+    /// #251: a closed pool on /api/v1/arweave/anchors must be 503 db_unavailable.
+    #[sqlx::test]
+    async fn list_anchors_closed_pool_returns_503_db_unavailable(pool: PgPool) {
+        let state = crate::test_support::test_state(pool.clone()).await;
+        pool.close().await;
+
+        let resp = Router::new()
+            .route("/api/v1/arweave/anchors", axum::routing::get(list_anchors))
+            .with_state(state)
+            .oneshot(
+                Request::builder()
+                    .uri("/api/v1/arweave/anchors")
+                    .body(axum::body::Body::empty())
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(
+            resp.status(),
+            StatusCode::SERVICE_UNAVAILABLE,
+            "closed-pool outage must be retryable 503, not 500"
+        );
+        let bytes = axum::body::to_bytes(resp.into_body(), usize::MAX)
+            .await
+            .expect("read body");
+        let v: Value = serde_json::from_slice(&bytes).expect("json body");
+        assert_eq!(
+            v,
+            serde_json::json!({
+                "error": crate::error::DB_UNAVAILABLE_CODE,
+                "message": crate::error::DB_UNAVAILABLE_MESSAGE,
+            })
+        );
+    }
 }
--- a/crates/gitlawb-node/src/api/ipfs.rs
+++ b/crates/gitlawb-node/src/api/ipfs.rs
@@ -204,7 +204,9 @@ pub async fn get_by_cid(
     .await
     {
         Ok(Ok(repos)) => repos,
-        Ok(Err(e)) => return Err(AppError::Internal(e)),
+        // Bare conversion (not `AppError::Internal`) so connection-class sqlx
+        // failures downcast to `AppError::Db` → 503 `db_unavailable` (#251).
+        Ok(Err(e)) => return Err(e.into()),
         Err(_elapsed) => {
             tracing::warn!(
                 budget_secs = state.config.ipfs_request_budget_secs,
@@ -228,7 +230,8 @@ pub async fn get_by_cid(
     .await
     {
         Ok(Ok(rules)) => rules,
-        Ok(Err(e)) => return Err(AppError::Internal(e)),
+        // Same #251 downcast path as list_all_repos above.
+        Ok(Err(e)) => return Err(e.into()),
         // FAIL CLOSED (security-critical): a timeout on the access-control query must
         // DENY. Returning here — before the loop — means the scan can never fall
         // through and apply an empty rule map, which would serve an unfiltered listing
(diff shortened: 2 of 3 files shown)


Ayush7614:
@beardthelion — this PR addresses #251 that you filed. Happy to rebase/coordinate with #134 if that lands first on the same call sites.

Ayush7614:
@coderabbitai — addressed your [nit](https://github.com/Gitlawb/node/pull/254#pullrequestreview-4787413471):

Added `get_by_cid_closed_pool_returns_503_db_unavailable` — valid sha2-256 CIDv1 + closed pool → exact `503` / `db_unavailable` payload, matching the `list_pins` pattern so both bare-`?` DB sites stay covered.

beardthelion (approved):
Approving. The mechanism is right and the sweep went further than the issue asked: after this, no live `.map_err(AppError::Internal)` remains on a db call anywhere in the crate (the two left in `register.rs` wrap UCAN work, not queries). I checked the downcast actually survives at all four call sites, since that is the part that silently breaks: `list_pinned_cids`, `list_arweave_anchors`, `list_all_repos` and `list_visibility_rules_for_repos` each end on a bare `?` over `fetch_all` with no `.context()` layer, so there is nothing for `downcast::<sqlx::Error>()` to strip.

I verified the tests are load-bearing rather than green by construction. Reverting each production line back to `.map_err(AppError::Internal)` turns the matching test red with `left: 500 / right: 503`, and all three pass unmutated on `7cba310`. The `get_by_cid` test you added on the last push covers the `list_all_repos` conversion specifically: revert `ipfs.rs:80` and it fails at `ipfs.rs:294`.

Two notes, neither an ask, and one small thing worth doing.

## Findings

- [P3] Worth adding to the PR body: 500 responses on these routes now carry `db_error` where they carried `internal_error`
  `crates/gitlawb-node/src/error.rs:150`
  For a server-reported query error the status stays 500 and the message text is byte-identical, but the code string changes, and the description currently reads as though only the 503 case moves. I drove a dropped-column error through `/api/v1/arweave/anchors` to confirm: `{"error":"db_error","message":"error returned from database: column \"cid\" does not exist"}`. Nothing in the repo branches on that string, and the new code is what the sibling routes already emit for the same condition, so this is an alignment rather than a regression. Description fix only, no code change and no need for another push.

- [P3] Coverage boundary on `get_by_cid`, recorded so it is not mistaken for full coverage later. Not an ask.
  `crates/gitlawb-node/src/api/ipfs.rs:86`
  The new test pins the first converted call site but not the second. With `ipfs.rs:80` restored and only `ipfs.rs:86` reverted, the test still passes, because a closed pool fails `list_all_repos` first and the second call is never reached. No test can isolate it with this injection, and the conversion there is the same one-line pattern, so I would leave it as is.

- [P2] `db_unavailable` cannot see server-reported unavailability. Separate defect, not an ask on this PR.
  `crates/gitlawb-node/src/error.rs:65`
  The classifier matches only `PoolTimedOut | PoolClosed | Io | Tls`, so no SQLSTATE-bearing error can reach the 503 arm at all, which the probe above confirms. Meanwhile `main.rs:685` already downcasts `sqlx::Error::Database(db)` and switches on code prefixes, treating everything outside `28`/`3D` as transient, so the startup path and the request path currently disagree about the same condition. Which variant a real restart or connection-slot exhaustion emits is a direction I have not run, which is the other reason it does not belong in your change.

## Merge order

Taking your offer: this goes first and #134 rebases onto it. #134 adds five `.map_err(AppError::Internal)?` sites in these same two files, which would put the 500 behavior straight back. I will raise those in that PR's round, so nothing there is on you.

Not yours, but relevant to whoever merges: `PR Checks` had never run on either head (fork approval pending), so the green on this PR was two triage checks rather than the suite. I have approved the run and it is going now.

jatmn (approved):
LGTM

beardthelion:
@kevincodex1 this one is ready to merge. Approved by me on 07-27 and by jatmn on 07-29, both against the current head 7cba310, so neither review is stale. CI is green across all 12 checks and it merges clean.

kevincodex1 (approved):

kevincodex1:
thanks for your contribution @Ayush7614 please rebase to main and kindly fix conflicts

Ayush7614:
@kevincodex1 rebased onto current `main` and resolved conflicts on head `3143404`.

Main had grown budget-timeout wrappers around `get_by_cid`'s DB calls; kept those and switched the sqlx error arms to `Into<AppError>` (instead of `AppError::Internal`) so closed-pool / connection failures still map to **503 `db_unavailable`**. `list_pins` + `/arweave/anchors` use bare `?` again, with closed-pool regressions restored.

Ready for another look / merge when CI is green.

Ayush7614:
cc: @kevincodex1

kevincodex1 (approved):

kevincodex1:
hello @Ayush7614 kindly do rebase again and fix conflicts

Ayush7614:
@kevincodex1 — rebased again onto current `main` (`241b366`, includes merged #247).

**Conflict resolution:** `error.rs` kept all #226 opaque-body tests and added the #251 `pool_closed_via_anyhow_from_is_503_db_unavailable` downcast test. Dropped the separate LEDGER commit — main already carries those #248 peer-readiness fixture writers from #247.

**Head:** single commit `c29a91a` on `fix/251-db-outage-503-ipfs-arweave`. `get_by_cid` still uses main's budget-timeout wrappers with `Into<AppError>` on sqlx failures; `list_pins` and `/arweave/anchors` use bare `?`.

Ready for merge when CI is green.

beardthelion (approved):
Re-reviewed on `c29a91af`, since the earlier approvals all predate today's two rebases.

The mechanism is right and the fail-closed ordering on the access-control query is intact: the timeout arm still returns before the loop, so a rule-map failure can never fall through to an unfiltered scan. Checked the conversion end to end, `db_unavailable` covers PoolTimedOut, PoolClosed, Io and Tls, and query or schema errors still land on the opaque 500 `db_error` body rather than being reclassified as retryable. Three of the four changed arms are pinned by the new tests, each verified by reverting the production line and watching the test go red.

## Findings

- [P2] Pin the visibility-rules error arm
  `crates/gitlawb-node/src/api/ipfs.rs:234`
  The closed-pool tests cannot reach this arm, since the pool is shut before `list_all_repos` and the handler returns at the first query. Reverting this line to `AppError::Internal(e)` leaves all three new tests green, so the one arm on the access-control query is unpinned. A test that seeds a repo (needed, the db function short-circuits on an empty id list), drops `visibility_rules`, then drives `GET /ipfs/{cid}` does pin it: assert 500 with body `db_error`, which goes red with `internal_error` if the arm regresses. Not a blocker for merge.

- [P3] Update the Internal-arm comment that still names this handler
  `crates/gitlawb-node/src/error.rs:190`
  The comment gives `GET /ipfs/{cid}` as the example of a handler reaching the Internal arm via `.map_err(AppError::Internal)`. This PR removes both of those call sites, so the example now points at a handler that no longer does it. `error.rs` is already in the diff.

PR #255, opened 2026-07-27
fix(node): opaque GraphQL DB error messages (#250)

### LINKED ISSUE #250: Unauthenticated POST /graphql leaks raw sqlx/Postgres error text in query resolver errors (sibling of #226/#106)
An unauthenticated `POST /graphql` returns raw sqlx/Postgres error text in the response body when any query resolver's database call fails. Same class as #226 and #106, on a surface neither of them covers, and one that does not route through `AppError` at all.

### Mechanism

Every resolver on `QueryRoot` maps database failures straight into the GraphQL error message:

```rust
// crates/gitlawb-node/src/graphql/query.rs:17
let repos = db
    .list_all_repos_deduped()
    .await
    .map_err(|e| async_graphql::Error::new(e.to_string()))?;
```

The db layer propagates sqlx through `anyhow` with a bare `?` and no `.context()`, so `e.to_string()` is the sqlx `Error` Display verbatim. Six sites in `query.rs` do this: lines 17, 30, 74, 89, 119, 128.

Because this path builds an `async_graphql::Error` directly, it bypasses `AppError` and its `IntoResponse` entirely. Opaque-ifying the `AppError` arms does not touch it.

### Reachability

`/graphql` is mounted with `optional_signature` only:

```rust
// crates/gitlawb-node/src/server.rs:61-66
let graphql_routes = Router::new()
    .route("/graphql", get(graphql_playground).post(graphql_handler))
    .layer(middleware::from_fn(auth::optional_signature))
```

and `optional_signature` (`crates/gitlawb-node/src/auth/mod.rs:251-258`) only enforces when signature headers are present, passing a header-less request straight through to `next.run(request)`. So an anonymous `POST /graphql` reaches every query resolver.

### Reproduction

Executed against a real Postgres on `main` (111cff7), driving the schema with no `AuthenticatedDid` in context, which is how this repo's existing anonymous-GraphQL guards model an unauthenticated caller (`graphql_repos_hides_private_from_anonymous`, #97):

```rust
sqlx::query("ALTER TABLE repos DROP COLUMN is_public").execute(&pool).await.unwrap();
let resp = schema.execute("{ repos { name ownerDid } }").await;
```

Observed response:

```json
{"data":null,"errors":[{"message":"error returned from database: column \"is_public\" does not exist","locations":[{"line":1,"column":3}],"path":["repos"]}]}
```

Any query- or connection-level error on those calls returns Postgres schema, constraint, or connection detail to an anonymous caller.

### Scope

`query.rs` is the anonymous surface and the reason this is filed. `mutation.rs` has the same six-site pattern, but each mutation calls `require_signer` before touching the database, so it needs a valid signature first. Identities here are permissionless, so that is a lower tier rather than a real barrier, and it is worth fixing in the same pass. `subscription.rs` has no such sites.

### Relationship to the siblings

#226 names `get_by_cid`'s two `AppError::Internal` arms and asks for opaque `AppError::Internal` bodies. #247 does that and also opaque-ifies the `Db` arm, which is the right fix for what #226 describes. Neither reaches this surface, so #226 closing does not close this. Worth noting that #226's own notes predict that opaque-ifying `AppError::Internal` resolves the class; that is not true for GraphQL, which never routes through `AppError`.

#106 is the same shape on the smart-HTTP git endpoints via `AppError::Git`, also still open.

### Fix

Stop passing the raw error into the client-visible message. Log the real error server-side and return a generic string, mirroring what #247 lands for `AppError`:

```rust
.map_err(|e| {
    tracing::error!(error = %format!("{e:#}"), "graphql query failed");
    async_graphql::Error::new("a database error occurred")
})?
```

A shared helper is preferable to twelve call sites repeating it. Use `{e:#}` rather than `%e` in the log so the anyhow cause chain survives, the way `api/repos.rs:217` already does.

### Severity

Calibrated against the siblings rather than asserted: #226 (raw DB error text, unauthenticated) is sev:high, #106 (filesystem path) is sev:low. This one leaks DB error text to an unauthenticated caller, so it matches #226.

Ayush7614:
## Summary
- Add shared `graphql_db_err` helper that logs `{e:#}` server-side and returns a generic `"a database error occurred"` client message.
- Route all 6 `query.rs` and 6 `mutation.rs` DB `.map_err` sites through it.
- Closes the unauthenticated `POST /graphql` sqlx/Postgres leak described in #250 (bypasses `AppError`, so #226/#247 do not cover it).

## Test plan
- [x] `cargo test -p gitlawb-node graphql_db_err_is_opaque` — helper rejects raw schema text
- [x] `repos_query_db_error_message_is_opaque` — anonymous `{ repos { ... } }` after a schema break returns only the opaque message (CI / `DATABASE_URL`)

Fixes #250

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized GraphQL error handling so database and schema details are no longer exposed.
  * Preserved actionable business error messages while routing unexpected database and application failures to safe, consistent messages.
  * Improved task listing by clamping `limit` to `0..=200`; negative values return no tasks and larger values are capped at 200.
* **Tests**
  * Added coverage for database error masking, safe business messages, and task-list limit handling.

### DIFF
--- a/crates/gitlawb-node/src/graphql/mod.rs
+++ b/crates/gitlawb-node/src/graphql/mod.rs
@@ -14,16 +14,47 @@ use subscription::SubscriptionRoot;
 
 pub type GitlawbSchema = Schema<QueryRoot, MutationRoot, SubscriptionRoot>;
 
-/// Client-facing message for GraphQL resolver DB failures. The real error is
-/// logged server-side; never put sqlx/anyhow detail in the GraphQL `errors`
-/// array (#250 — sibling of #226, which only covers `AppError`).
+/// Client-facing message for GraphQL resolver failures that wrap a real
+/// `sqlx::Error`. The real error is logged server-side; never put sqlx/Postgres
+/// detail in the GraphQL `errors` array (#250).
+///
+/// Kept as its own constant on this PR's base (main still renders
+/// `AppError::Db` with `e.to_string()`). If/when #247's `DB_ERROR_MESSAGE`
+/// lands, fold this into that shared constant.
 pub const GRAPHQL_DB_ERROR_MESSAGE: &str = "a database error occurred";
 
-/// Map a database/`anyhow` failure to an opaque GraphQL error: log the full
-/// cause chain (`{e:#}` for anyhow) and return a generic client message.
-pub(crate) fn graphql_db_err(e: impl std::fmt::Display) -> async_graphql::Error {
-    tracing::error!(error = %format!("{e:#}"), "graphql database error");
-    async_graphql::Error::new(GRAPHQL_DB_ERROR_MESSAGE)
+fn anyhow_has_sqlx(e: &anyhow::Error) -> bool {
+    e.chain().any(|c| c.downcast_ref::<sqlx::Error>().is_some())
+}
+
+/// Map an `anyhow` failure from the db layer to a GraphQL error.
+///
+/// - Real DB faults (`sqlx::Error` anywhere in the chain) → opaque client
+///   message + `error!` log with the full `{e:#}` cause chain.
+/// - Application/business errors (e.g. claim race, not-in-claimed-state) →
+///   keep the actionable message; log at `warn!` so they are not mistaken for
+///   infrastructure failures (#250 review).
+pub(crate) fn graphql_db_err(e: anyhow::Error) -> async_graphql::Error {
+    if anyhow_has_sqlx(&e) {
+        tracing::error!(error = %format!("{e:#}"), "graphql database error");
+        async_graphql::Error::new(GRAPHQL_DB_ERROR_MESSAGE)
+    } else {
+        tracing::warn!(error = %format!("{e:#}"), "graphql application error");
+        async_graphql::Error::new(e.to_string())
+    }
+}
+
+/// Map an `AppError` from a shared collector (e.g. ref-update feed) to a
+/// GraphQL error. DB variants are opaque; other variants stay opaque too
+/// because `AppError::Internal` can still carry raw detail on this base.
+pub(crate) fn graphql_app_err(e: crate::error::AppError) -> async_graphql::Error {
+    match e {
+        crate::error::AppError::Db(sql) => graphql_db_err(sql.into()),
+        other => {
+            tracing::error!(error = %other, "graphql error");
+            async_graphql::Error::new(GRAPHQL_DB_ERROR_MESSAGE)
+        }
+    }
 }
 
 pub fn build_schema(
@@ -43,11 +74,18 @@ mod tests {
     use super::*;
 
     #[test]
-    fn graphql_db_err_is_opaque() {
+    fn graphql_db_err_opaques_sqlx_chain() {
         let leak = "error returned from database: column \"is_public\" does not exist";
-        let err = graphql_db_err(anyhow::anyhow!("{leak}"));
+        let err = graphql_db_err(anyhow::Error::from(sqlx::Error::Protocol(leak.into())));
         assert_eq!(err.message, GRAPHQL_DB_ERROR_MESSAGE);
         assert!(!err.message.contains("is_public"));
         assert!(!err.message.contains(leak));
+    }
+
+    #[test]
+    fn graphql_db_err_keeps_business_message() {
+        let msg = "task not claimable: not found or already claimed";
+        let err = graphql_db_err(anyhow::anyhow!("{msg}"));
+        assert_eq!(err.message, msg);
     }
 }
--- a/crates/gitlawb-node/src/graphql/mod.rs
+++ b/crates/gitlawb-node/src/graphql/mod.rs
@@ -47,10 +47,9 @@ pub(crate) fn graphql_db_err(e: anyhow::Error) -> async_graphql::Error {
 /// Map an `AppError` from a shared collector (e.g. ref-update feed) to a
 /// GraphQL error.
 ///
-/// - `Db` → opaque via [`graphql_db_err`] (sqlx detail stays in logs).
-/// - `Internal` → opaque (may carry raw anyhow/sqlx text on this base).
-/// - Other variants already carry safe, actionable messages → surface them
-///   and log at `warn!` (same posture as business errors in `graphql_db_err`).
+/// Fail closed: only explicitly curated variants surface their `Display`
+/// text. Unnamed variants (including `Git`, which may embed on-disk paths)
+/// render opaque so a future addition cannot leak by default (#255 review).
 pub(crate) fn graphql_app_err(e: crate::error::AppError) -> async_graphql::Error {
     match e {
         crate::error::AppError::Db(sql) => graphql_db_err(sql.into()),
@@ -58,9 +57,21 @@ pub(crate) fn graphql_app_err(e: crate::error::AppError) -> async_graphql::Error
             tracing::error!(error = %format!("{err:#}"), "graphql internal error");
             async_graphql::Error::new(GRAPHQL_DB_ERROR_MESSAGE)
         }
+        // Curated client-safe variants — `Display` is intentional API text.
+        safe @ (crate::error::AppError::RepoNotFound(_)
+        | crate::error::AppError::RepoExists(_)
+        | crate::error::AppError::NotFound(_)
+        | crate::error::AppError::Unauthorized(_)
+        | crate::error::AppError::Forbidden(_)
+        | crate::error::AppError::BadRequest(_)
+        | crate::error::AppError::TooManyRequests(_)
+        | crate::error::AppError::Incomplete(_)) => {
+            tracing::warn!(error = %safe, "graphql application error");
+            async_graphql::Error::new(safe.to_string())
+        }
         other => {
-            tracing::warn!(error = %other, "graphql application error");
-            async_graphql::Error::new(other.to_string())
+            tracing::error!(error = %other, "graphql unclassified AppError (opaque)");
+            async_graphql::Error::new(GRAPHQL_DB_ERROR_MESSAGE)
         }
     }
 }
@@ -84,10 +95,15 @@ mod tests {
     #[test]
     fn graphql_db_err_opaques_sqlx_chain() {
         let leak = "error returned from database: column \"is_public\" does not exist";
-        let err = graphql_db_err(anyhow::Error::from(sqlx::Error::Protocol(leak.into())));
+        // Context layer must not hide sqlx from the chain walk (db helpers
+        // wrap with `.context(...)` in several places).
+        let err = graphql_db_err(
+            anyhow::Error::from(sqlx::Error::Protocol(leak.into())).context("loading repos"),
+        );
         assert_eq!(err.message, GRAPHQL_DB_ERROR_MESSAGE);
         assert!(!err.message.contains("is_public"));
         assert!(!err.message.contains(leak));
+        assert!(!err.message.contains("loading repos"));
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


Ayush7614:
@beardthelion — this PR addresses #250 that you filed. GraphQL never went through `AppError`, so the opaque helper is separate from #247 as you noted.

beardthelion (dismissed):
The approach is right and the suppression is genuinely complete inside the module. I checked the part that usually goes wrong: `Error::new` sets `source: None`, and both `source` fields carry `#[serde(skip)]`, so nothing can reach the wire even by accident, and `{e:#}` does render the full anyhow chain into the log rather than trading a client leak for an operator blindspot. I also checked the sink a grep for the old `map_err` shape cannot see, since async-graphql auto-converts any `Display` into an error message on a bare `?`. No resolver uses one today, so the sweep holds.

Two things block it, and one of them is not style.

## Findings

- [P1] Drop the redundant closures, CI cannot pass as written
  `crates/gitlawb-node/src/graphql/query.rs:17`
  All twelve sites are written `.map_err(|e| crate::graphql::graphql_db_err(e))`, which trips `clippy::redundant_closure`, and `.github/workflows/pr-checks.yml:49` runs `cargo clippy --workspace --all-targets -- -D warnings`. I ran clippy on this head and got twelve of them. `.map_err(crate::graphql::graphql_db_err)` is enough: I applied it to the six in query.rs and the count went to six, with `cargo check` clean.

- [P1] Only send real DB faults through the opaque helper
  `crates/gitlawb-node/src/graphql/mutation.rs:79`
  `claim_task` returns `anyhow!("task not claimable: not found or already claimed")` (`db/mod.rs:2619`) and `finish_task` returns `anyhow!("task not found or not in claimed state")` (`db/mod.rs:2641`). Neither ever contained sqlx text, so neither was leaking. With this change a signed agent that loses a claim race is told `a database error occurred`, which is wrong and gives it nothing to act on, and losing a claim race is the ordinary path in a task queue rather than an edge case. Downcasting to `sqlx::Error` at the call site, or returning a typed variant from the db layer, keeps those two messages while still closing the leak.

- [P2] Cover the other eleven sites, mutations first
  `crates/gitlawb-node/src/graphql/mutation.rs:57`
  The `#[sqlx::test]` makes exactly one line load-bearing. `list_all_repos_deduped` selects `d.is_public` (`db/mod.rs:1215`), so dropping that column fails the first call and `?` returns before `query.rs:30` ever runs; I confirmed by reverting each line separately, and only `query.rs:17` goes red. The unit test in mod.rs calls the helper directly, so it stays green even with every call site reverted. `mutation.rs:175` already has a test module using `test_support::test_state` and `Request::new(q).data(AuthenticatedDid(...))`, so the mutation-side test has both a home and a pattern to copy.

- [P2] Clamp the tasks limit before it reaches the query
  `crates/gitlawb-node/src/graphql/query.rs:112`
  `limit: i64` goes unclamped into `LIMIT $3` (`db/mod.rs:2572`), and Postgres rejects a negative limit outright (`2201W: LIMIT must not be negative`). So an anonymous `{ tasks(limit: -1) { id } }` is a deterministic failure that now writes an error-level log line on every request, where before it only returned one. The REST feed already clamps at `api/events.rs:129-134`. Worth considering `warn!` rather than `error!` for client-caused failures generally.

- [P3] Share one constant with #247 instead of adding a second
  `crates/gitlawb-node/src/graphql/mod.rs:20`
  #247 defines `DB_ERROR_MESSAGE: &str = "a database error occurred"` at `error.rs:94`, the same string, and this adds a second constant with its own test and a different log event name for the same concept. Whichever lands second should fold into the first. The `#226` reference in the comment above also does not hold on this base: `error.rs:150` still returns `e.to_string()` for `AppError::Db`, so a reader would conclude the leak class closes when this merges, and it does not.

Not asks, just so they are on the record. `api/tasks.rs:138-143` still puts `e.to_string()` into the REST body on an anonymous route, which is the same class as #250 but untouched here, so it wants its own issue rather than scope creep into this one. And async-graphql has a `custom-error-conversion` feature that removes the blanket `Display` conversion, which would make a future bare `?` in a resolver a compile error instead of a silent re-leak. That is the durable version of this fix if it appeals later, not something to do now.

One thing that is not yours: `PR Checks` has never run on this head, so the green you see is a triage check rather than the suite. I have approved the run, which means it will now go red on the clippy item above rather than staying silent about it.

Ayush7614:
@beardthelion — addressed your review (and the clippy red):

1. **[P1] redundant closures** — all anyhow sites are now `.map_err(crate::graphql::graphql_db_err)`; the `AppError` collector site uses `.map_err(crate::graphql::graphql_app_err)`. `clippy -D warnings` clean locally.
2. **[P1] sqlx-only opaque** — helper opaques only when `sqlx::Error` is in the anyhow chain; claim/finish business messages (`task not claimable…`, etc.) are kept and logged at `warn!`. Unit test + mutation assert cover this.
3. **[P2] more coverage** — `create_task_db_error_message_is_opaque` (mutation), claim business-message assert, `tasks_negative_limit_clamped`.
4. **[P2] clamp tasks limit** — `limit.clamp(0, 200)` before SQL so `tasks(limit: -1)` cannot trip Postgres 2201W.
5. **[P3] constant / comment** — noted that `GRAPHQL_DB_ERROR_MESSAGE` should fold into #247’s `DB_ERROR_MESSAGE` when that lands; comment no longer claims #226 already closed the class on this base.

Ready for another look.

Ayush7614:
@coderabbitai thanks — addressed in 88a436a:

1. **`graphql_app_err`**: only `Db` / `Internal` stay opaque. Other `AppError` variants (e.g. `NotFound`, `BadRequest`) now surface their safe `to_string()` messages and log at `warn!`.
2. Unit tests cover opaque Db/Internal and pass-through for safe variants.
3. Documented the `tasks` `limit` clamp in the GraphQL schema (`desc`).

@beardthelion FYI — this was the remaining CodeRabbit major on #255.

beardthelion (dismissed):
Every round-one finding landed, and I re-ran the checks rather than re-reading them. Clippy on this head is exit 0 with zero `redundant_closure`, down from the twelve I counted last time. The sqlx-vs-business split does keep `claim_task`'s message: I reverted each converted call site in turn and watched which tests move. The clamp is in and the negative case is genuinely held. On the question the PR exists to answer, I could not find a reachable path that still puts sqlx text on the wire: every db function these resolvers touch is sqlx plus `?`, the only non-sqlx errors in that set are your two intentional business strings at `db/mod.rs:2619` and `2641`, and no db-layer code stringifies a sqlx error out of downcastability.

Four things left, none of them a live leak.

## Findings

- [P2] Make `graphql_app_err` opaque by default instead of verbatim by default
  `crates/gitlawb-node/src/graphql/mod.rs:60`
  The `other` arm returns `other.to_string()` for every variant not named above it, so the client-safe set is whatever happens to be left over rather than a decision. `AppError::Git(String)` sits in that leftover set and is built straight from raw git errors at ten sites (`api/issues.rs:68`, `api/pulls.rs:155`, `api/encrypted.rs:66`), several of them over `&disk_path`, so its payload carries on-disk paths. Nothing reaches it today, and I checked rather than assumed: `collect_visible_ref_updates` is the only caller, and its errors all arrive through `impl From<anyhow::Error> for AppError` (`error.rs:79-86`), which can only yield `Db` or `Internal`. The invariant worth holding is that a variant nobody has classified renders opaque, so adding one later fails closed. Naming the eight curated variants explicitly and sending everything else to `GRAPHQL_DB_ERROR_MESSAGE` compiles, passes `clippy --all-targets -- -D warnings`, and leaves all 18 graphql tests green including your `graphql_app_err_keeps_safe_variant_messages`.

- [P2] Cover all twelve call sites with one source scrape rather than one test each
  `crates/gitlawb-node/src/graphql/mod.rs:71`
  I reverted each converted site to the old mapping and ran the suite. Only `query.rs:17` and `mutation.rs:57` go red; reverting the other ten (`query.rs:30,74,89,127,136` and `mutation.rs:79,111,121,152,163`) leaves 18 of 18 passing, and `fail_task` has no GraphQL test at all. My round-one ask for the other eleven was the wrong shape, so please ignore it: ten more integration tests would be slow and would still miss the next resolver someone adds. The crate already scrapes source for exactly this class in `mod authz_guard` (`api/mod.rs:137`). A test asserting that every `.map_err(` in query.rs and mutation.rs routes through one of the two helpers, or discards its argument, covers all twelve and everything added later. I wrote that test, confirmed it green as-is, and confirmed it red when a raw mapping returns at `mutation.rs:163`.

- [P3] Give `graphql_db_err_opaques_sqlx_chain` an actual chain. This is an ask, and a small one.
  `crates/gitlawb-node/src/graphql/mod.rs:84`
  The input is a bare `sqlx::Error` at the chain root, so the test never exercises the walk its name describes. I replaced `e.chain().any(...)` with a top-level `e.downcast_ref::<sqlx::Error>()` and all 18 tests stayed green. The walk is the load-bearing half, since `db/mod.rs` wraps in `.context()` in several places and a context layer is what would hide the sqlx error from a top-level check. One case built as `anyhow::Error::from(sqlx::Error::Protocol(..)).context("loading repos")` pins it.

- [P3] Pin the 200 in the tasks clamp. This is an ask, and a small one.
  `crates/gitlawb-node/src/graphql/query.rs:123`
  `tasks_negative_limit_clamped` is load-bearing for the lower bound: remove the clamp and it goes red on Postgres 2201W. The ceiling is not held by anything, though. I changed `clamp(0, 200)` to `clamp(0, i64::MAX)` and the suite stayed green, so the "Max 200" you put in the field description is currently a promise with no test behind it, and it is also a new client-visible cap. Seeding 201 rows and asserting `tasks(limit: 5000)` returns 200 covers both.

Three things that are not asks, just for the record. `PR Checks` has never run on 88a436a, so the green you see is the triage check again rather than the suite; I ran clippy against this head myself, which is how I know the closure item is settled, but CI has not validated the rest. `AgentTaskType` exposes `ucan_token` to anonymous callers on this same endpoint (`graphql/types.rs:23`), which predates this PR and wants its own issue rather than scope creep here. And `api/tasks.rs` still puts `e.to_string()` into the REST body, the same class as #250 and still untouched, as noted last round.

Ayush7614:
@beardthelion addressed your second-round findings:

1. **[P2] fail-closed `graphql_app_err`** — eight curated safe variants (`RepoNotFound`, `RepoExists`, `NotFound`, `Unauthorized`, `Forbidden`, `BadRequest`, `TooManyRequests`, `Incomplete`) surface their messages; `Git` / anything else (plus `Db`/`Internal`) stay opaque. Test covers a path-bearing `Git` payload.
2. **[P2] source scrape** — `every_graphql_map_err_uses_opaque_helpers` asserts every `.map_err(` in `query.rs` / `mutation.rs` routes through `graphql_db_err` / `graphql_app_err` or discards (`|_|`).
3. **[P3] real sqlx chain** — `graphql_db_err_opaques_sqlx_chain` now wraps with `.context("loading repos")`.
4. **[P3] pin 200** — `tasks_limit_ceiling_clamped_to_200` seeds 201 rows and asserts `tasks(limit: 5000)` returns 200.

Ready for another look.

jatmn (approved):
LGTM

beardthelion (approved):
The leak is closed, and I checked it by execution rather than by reading. Reverting `graphql_db_err` to a leaking closure turns the completeness test red naming the property it protects, gutting `anyhow_has_sqlx` turns four tests red including the chain case, and the graphql suite is 21 passed on this head. Every ask from the earlier rounds landed; I verified each against the code rather than the replies.

## Findings

- **[P2] Scrape the sink, not just `.map_err(`** `crates/gitlawb-node/src/graphql/mod.rs:174`

  async-graphql ships a blanket `impl<T: Display> From<T> for Error`, so `db.foo().await?` converts straight into a client-visible error carrying `e.to_string()`, with no `.map_err(` on the line for the scan to find. Confirmed by execution: swapping `.map_err(graphql_db_err)?` for a bare `?` in `tasks` leaves the whole graphql suite green, while the same swap in `repos` leaks `column "is_public" does not exist` end to end. What catches it today is a per-resolver DB-fault test, and only `repos` and `create_task` have one. A second rule in the same loop rejecting a bare `.await?` in these two files closes it. There are no such occurrences today, so it stays quiet until someone writes the bypass.

- **[P3] Add `subscription.rs` to the scrape, and a floor on sites checked** `crates/gitlawb-node/src/graphql/mod.rs:169`

  The file list is hardcoded to `query.rs` and `mutation.rs`, but `SubscriptionRoot` is mounted in the schema at line 15. It has no fallible path today so nothing leaks, though a resolver added there would sit outside the scan. Asserting a minimum number of matched sites would also make "scraped nothing" fail instead of passing green.

- **[P3] Carry an error code in `extensions`** `crates/gitlawb-node/src/graphql/mod.rs:37`

  REST returns `db_unavailable` with a retryable signal where GraphQL now returns a single opaque string, so a client cannot separate a pool timeout from a query bug. No in-repo consumer breaks, so this is a follow-up rather than something to hold the PR for.

None of these is a live leak. Every error path in the module routes through the helpers today, including the six direct `Error::new` calls, which all take string literals.

Two notes that are not asks. #275 touches the same lines in `mutation.rs`, so expect a rebase conflict there. And `cargo audit` is red on ruint 1.17.2 for RUSTSEC-2026-0220, which is hitting every branch right now and has nothing to do with this diff; #292 is the bump.

@kevincodex1 this is ready from my side once the ruint bump clears the audit job.

beardthelion:
@kevincodex1 this is ready to merge now. The earlier approve still stands from jatmn and me; the ruint audit hold I called out is gone with #292. Rebased onto main, CI green on 726ba336.

kevincodex1 (approved):
lgtm

Issue #260, opened 2026-07-27
gl init's create-repo path loses the status when the error body is not JSON


beardthelion:
Follow-up from the #253 phase-1 work.

`gl init`'s create-repo call parses the response body before checking the status, and it parses with `.context(...)?` rather than a fallback, so a non-JSON error body becomes a hard parse error and the real status never reaches the user (`crates/gl/src/init.rs:145-148`):

```rust
let repo_status = resp.status();
let repo_result: Value = resp.json().await.context("invalid JSON from create repo")?;

if !repo_status.is_success() {
```

Reproduced against a mock returning `429` with `content-type: text/html`:

```
invalid JSON from create repo: error decoding response body: expected value at line 1 column 1
```

The 429 is lost. Any non-JSON denial does this: the per-IP flood brake, a proxy error page, a CDN block page.

## Correcting two things I got wrong in the original text

**The ordering is not forced.** I first wrote that this site could not be converted because its tolerated-failure branch (`repo_already_exists`) needs the parsed body, and the shared `json_or_denial` helper cannot express that. The second half is true: `json_or_denial` (`crates/gl/src/http.rs:371-392`) bails on any non-success and never hands back the parsed body, so it genuinely cannot serve a call site that wants to inspect a tolerated failure. But that does not force parsing *before* the status check. `repo_result` is only read on the failure path (`init.rs:154-155`); the success path never touches it. Checking the status first and parsing inside the `if !repo_status.is_success()` block is a pure reordering, no restructuring needed.

**"Every other gl write call site checks status first" was false.** Seventeen other POST/PUT/DELETE sites have the same capture-parse-then-check shape: `bounty.rs:192,292,317,347,372`, `mcp.rs:1304`, `mirror.rs:108`, `protect.rs:111,143`, `quickstart.rs:119`, `repo.rs:420,452,659`, `star.rs:78,102`, `visibility.rs:126,159`.

They are not equivalent, which is the actual distinction worth drawing. They parse with `unwrap_or_default()`, so a non-JSON body degrades to `Null` and the site still reports the status, for example `star failed (429): unknown error`. Ugly, but the status survives. `init.rs` is the only site that turns a non-JSON error body into a hard error and loses the status entirely.

So the bug is narrower than the original wording implied: this is about one call site swallowing a status, not about a shape that was eliminated everywhere else.

## Fix

[...]

Issue #259, opened 2026-07-27
Signature ledger: nonce-arm identity keying is unpinned, and the ledger-full 429 carries no Retry-After


beardthelion:
Two small follow-ups from the #253 phase-1 work.

**The nonce arm's identity keying is not covered by a test.** `ledger_key`'s nonce branch hashes the resolved public key rather than the wire DID, so two multibase spellings of one key cannot reuse a nonce. That is asserted in a doc comment, but reverting that one line to the wire DID leaves the whole suite green: the tests that exercise re-spelling either sign without a nonce or drive only the per-identity cap. The property does hold (the cap and single-use are both tested), the specific line just is not pinned. A test that signs one nonce under two spellings and asserts the second is a 409 would close it.

**The ledger-full 429 has no `Retry-After`.** `ledger_rejection` sets `X-Gitlawb-Error` and a message saying to retry shortly, but nothing tells the caller when. The node can compute it: the oldest live row's `expires_at` for that identity is exactly when a slot frees. `gl` now turns this into a hard error, so the user is told to wait with no idea how long. Emitting `Retry-After` would also let a client back off correctly instead of guessing.

Issue #258, opened 2026-07-27
GITLAWB_REQUIRE_SIGNATURE_NONCE checks nonce length, not uniqueness


beardthelion:
Follow-up from the #253 phase-1 work.

The flag rejects an absent or too-short nonce (currently a 16-character floor). Length is the only property a verifier can observe, so that is a deliberate limit, not an oversight. The gap is what the code claims about it: the doc comment computes a birthday-collision bound, and that figure only holds if the nonce came from a CSPRNG. A client sending a constant `nonce="0000000000000000"` passes the check and then collides with itself on every request for the retention window.

Impact is self-inflicted: that client gets one accepted mutation per window and 409s after, and no other identity is affected. It is not exploitable against anyone else, which is why it did not block phase 1.

Two things worth doing:
- reword the doc comment so it states the floor rejects the degenerate empty and short cases, and either drop the birthday figure or condition it explicitly on a CSPRNG nonce
- consider whether the flag should additionally reject a nonce the node has seen before *from the same identity within the window*, which the ledger already knows and which is closer to what the flag's name promises

Issue #257, opened 2026-07-27
GraphQL mutations are not covered by the spent-signature ledger


beardthelion:
Follow-up from the #253 phase-1 work (PR to follow).

`/graphql` is layered with `optional_signature`, not the `add_auth_layers` stack, so `consume_signature` never runs for it. `MutationRoot` exposes `create_task`, `claim_task`, `complete_task` and `fail_task`, which are the same mutations the REST surface offers and which the ledger does cover there.

So after phase 1 a captured signature over a GraphQL mutation is still replayable, while the same mutation over REST is not. The asymmetry is the problem: a client that gets a 409 on REST can reach the identical side effect through `/graphql`.

Two ways out, both larger than the phase-1 change:
- extend the ledger to the signed half of `/graphql`, which means `optional_signature` publishing the same verified identity and a charge point that only fires when a signature was actually presented, or
- move those mutations behind `add_auth_layers` and require signatures on them.

The first keeps unsigned GraphQL queries working, which is why `optional_signature` is there in the first place. The second is simpler but is a breaking change for any unsigned mutation caller.

PR #261, opened 2026-07-27
fix(node): make a verified signature single-use on mutation routes (#253 phase 1)

beardthelion:
Phase 1 of #253. **Deliberately not "closes"**: #253 also covers host binding, and this PR does not do that, so it must not auto-close on merge. See Scope below.

A captured RFC 9421 signature is currently a bearer credential. The node verifies it and keeps no record that it did, so the same signed request can be delivered again and applied again. Five probes against `main` at 111cff7 confirmed it: one signature POSTed to `/api/v1/tasks` three times produced three `201`s and three rows, the same signature was accepted under three different `Host` headers, a `created` 250s in the future was accepted, and four whitespace variants of the `Signature` header all verified.

This makes a signature single-use on mutation routes and tightens the acceptance window.

## What changed

`sign_request` now emits a 128-bit `nonce` in the `Signature-Input` parameter tail. The tail is what `@signature-params` is built from, so the nonce is covered by the signature without touching the covered-component list, and a verifier that predates it keeps working because it rebuilds `@signature-params` from the received header text.

`check_created` is one-sided. It accepted `(now - created).abs() > 300`, which let a signer pre-date `created` and hold a signature valid for twice the intended span. Forward skew is now 30s; the 300s backward budget is unchanged, because the verifier buffers the whole body before this check and a large pack spends its upload time inside that budget.

A new migration adds `consumed_signatures`, and a `consume_signature` layer charges every verified signature against it. The key is `(identity, nonce)` when the signer sent one and the signing-string hash otherwise, domain-separated. The fallback is what makes whitespace padding useless: those variants are distinct header bytes that reconstruct to one signing string, so keying on the reconstruction collapses them to one entry. Keying on the header would not have.

Layer order is load-bearing. Axum runs the last layer first, so the ledger is listed first and runs last: a request with a good signature but a rejected UCAN must not burn its key on the way to being denied. A test drives the real `build_router` and fails if the two layers are swapped.

Reads are skipped by method. `write_routes` chains PUT/DELETE/GET on the visibility path and `gl` drives that GET signed, so ledgering it would put a database write on the signed read path.

Failure is closed in both directions: a missing verified identity is a 500 rather than a pass-through, so a wrong layer order cannot silently delete the defence, and a ledger error is a 503.

`GITLAWB_REQUIRE_SIGNATURE_NONCE` (default false) closes the hash fallback once every client emits a nonce. It is enforced in the ledger layer only, not in `require_signature`, which also backs `optional_signature`: enforcing there would reject nonce-less signatures on authenticated reads and `/graphql` for un-upgraded clients.

`GITLAWB_SIGNED_WRITE_RATE_LIMIT` (default 600/IP/hour, `0` disables) puts a per-IP brake in front of the five write groups that had none. Without it, any freshly minted keypair could force a durable ledger row on a repo it does not own or that does not exist: the request verifies, charges a row, then 404s. Measured at ~428 bytes on disk and ~983 bytes of WAL per request.

`gl` fails loudly on every denial instead of handing back a response that pretty-prints like success, and the node message is capped and sanitised before it reaches a terminal. The denial codes live in one enum in `gitlawb-core` that the client matches exhaustively, so adding a code without handling it in the client is a compile error.

## Scope, stated plainly

**GraphQL mutations are not covered** (#257). `/graphql` carries `optional_signature`, not the auth stack, and `MutationRoot` exposes `create_task`, `claim_task`, `complete_task` and `fail_task`. A captured signature over a GraphQL mutation stays replayable after this PR. Closing it means either extending the ledger to the signed half of `/graphql` or moving those mutations behind the auth stack, both larger than this change.

**Cross-node replay stays open**, which is the other half of #253 and why this PR does not close it. Nothing binds a signature to a host, so one capture is still accepted once by each node. Binding `@authority` is a bidirectional flag day (an old client fails on a new node and a new client fails on an old node, because the verifier builds the signing string from the client's component list), so it is separate work. This PR closes single-node replay, not the issue.

**Peer routes are only covered when signed peer writes are on.** With `GITLAWB_REQUIRE_SIGNED_PEER_WRITES=false`, which is the default, `announce` and `sync/notify` take `optional_signature` and are verified but never spent.

**Denial detection is header-only.** `gl` keys on `X-Gitlawb-Error`. A proxy that strips unknown `X-` headers returns the denial to the caller as a plain response. Reading the code from the body instead would mean rebuilding the response, which needs a dependency this branch should not carry.

## Known issues not fixed here

All of these are filed, so please do not open a PR against them without checking the issue first.

- **#258** The nonce check enforces length, not entropy. A client sending a constant 16-character nonce passes it and then collides with itself for the retention window. Self-inflicted only, no effect on other identities. The doc comment's collision figure assumes a CSPRNG and should say so.
- **#259** The nonce branch keying on the resolved key rather than the wire DID is asserted in a doc comment but not covered by a test: reverting that line leaves the suite green. The property holds (the cap and single-use are both tested), the specific line is not pinned. Same issue covers the missing `Retry-After` on the ledger-full 429, which tells a client to retry without telling it when.
- **#260** `gl init`'s create-repo path parses the body before checking the status, and parses with `.context(...)?` rather than a fallback, so a non-JSON error body (the flood brake, a proxy error page) becomes a parse error and the status is lost. It is the only site that loses the status this way; the other capture-parse-then-check sites use `unwrap_or_default()`, so the status still surfaces there.
- **#96** The federation retry is a mitigation, not a cure. It rides out a transient rejection, but a wide multi-ref push that trips a peer's per-identity cap needs the retention window to age out, which no in-process budget covers. Batching the fan-out into one multi-ref request is the structural fix, is a wire-format change, and is already tracked there. Refs that fail now surface in one summary line instead of one warning per ref.

## Operational notes

Migration version 16 is deliberate, not the next free number. #173 currently numbers four migrations 11 through 14 and sits behind main, so on rebase they shift to 12 through 15. Whichever of the two merged second previously had its migration silently skipped, because the runner keyed on the version integer alone and never compared the recorded name. It now compares the name and refuses to start on a mismatch, naming both sides and the remedy, and warns about recorded versions this build does not define so an orphan row is visible before it becomes a boot failure. **#173 and this PR need to agree on ordering before either merges.**

`gitlawb_signature_ledger_total{outcome}` counts all eight outcomes of the layer, so the series sum is the traffic the layer saw. Note the brake rejects upstream of the layer, so its 429s appear in no series here.

One thing worth confirming before deploy: whether the production nodes share a Postgres. If they do the ledger is globally correct; if not it is per-node, which is the assumption the fail-closed behaviour is written against. I could not determine it from the repo.

## Verification

1,109 tests, `cargo fmt --all` and `cargo clippy --workspace --all-targets -- -D warnings` clean. Every guard here was proven load-bearing by reverting the exact production line it protects and observing the test go red: the layer order, the GET skip, the fail-closed paths, the ledger's atomic insert, the per-identity cap, the sweep cutoff, the nonce validation, the rate brake, the retry gating, the migration name check, and the compile-time denial-code guard. The two acceptance tests filed with the issue are green with their `#[ignore]` removed.

## Summary by CodeRabbit

* **Security**
  * Added replay protection, configurable signature nonces, per-IP signed-write limits, and stronger ownership enforcement.
  * Improved peer authorization and standardized denial responses.

* **Reliability**
  * Added resilient peer-sync retries, readiness checks, migration safeguards, and maintenance recovery.
  * Improved concurrency controls and IPFS operation limits.

* **Bug Fixes**
  * CLI commands now handle rate limits, denials, malformed responses, and oversized or incomplete data consistently.
  * Improved validation of Git advertisements and error messages.

* **Documentation**
  * Expanded setup, configuration, security limitations, workflows, and roadmap guidance.


beardthelion:
@jatmn if you have time for this one, the parts most worth your attention are the three where a mistake would be invisible rather than loud.

**Layer ordering.** The ledger is listed first in `add_auth_layers` so it runs last, after the UCAN check. If it ran earlier, a request with a good signature but a rejected UCAN would burn its ledger key on the way to being denied, and the client could not retry those exact bytes. There is a test driving the real `build_router` that fails if the two are swapped, but this is the property I most expect a later refactor to undo quietly, so a second pair of eyes on the ordering argument itself is worth more than on the test.

**The GET and HEAD skip in `consume_signature`.** This is the deliberate hole in coverage: signed reads are not ledgered, because `write_routes` chains PUT, DELETE and GET on the visibility path and `gl` drives that GET signed. I convinced myself no mutation is reachable by those methods, including the HEAD that axum derives from a GET handler, but that is exactly the kind of thing I would rather have checked by someone who did not write it.

**The two fail-closed paths.** A missing verified identity is a 500 and a ledger error is a 503. Both exist so the defense cannot silently disappear: if either one passed the request through instead, the guard would vanish while every test stayed green. Worth trying to drive them to pass through.

More generally, the useful attack surface here is a permissionless one. Any caller can mint a fresh `did:key`, so anything that assumes an identity is scarce is probably wrong. The per-identity cap is the obvious example, and I would not be surprised if there is a way around it I have not thought of.

On the scope limits in the description (GraphQL mutations still replayable, cross-node replay still open, peer routes unledgered in the default config): those are decisions rather than oversights, and each has its reasoning in the body, so please do not feel you need to re-argue them. That said, if while you are in there you find that one of them is worse than I have assumed, or reachable in a way I have not described, I would genuinely like to know. Filed as #257 and the open half of #253 if it is easier to comment there.

Happy to walk through any of it if something reads oddly.

jatmn (changes requested):

## Findings

- [P2] Route MCP error handling through the capped, sanitized denial helpers
  `crates/gl/src/mcp.rs:1303-1317`
  `json_ok` still calls `resp.json()` with no byte cap and interpolates the node's raw `message`/`error` into tool errors without `sanitize_node_msg`. Every other denial-aware path this PR adds in `crates/gl/src/http.rs` (`read_body_capped`, `json_or_denial`, `send_signed` → `into_error`) caps hostile bodies and strips terminal control/bidi characters before they reach a caller. MCP is explicitly in scope for this change — the PR adds `json_ok` to stop 4xx/5xx bodies from deserializing as successful tool results — but left this helper on the old pattern. A malicious or compromised node can force an unbounded allocation on a failed tool call and smuggle terminal control sequences into MCP-visible errors.

- [P2] Cap `git_refs` denial bodies before buffering
  `crates/gl/src/mcp.rs:805-811`
  `git_refs` calls `resp.bytes().await?` before checking `status`, so a non-2xx response is fully buffered even when the call will fail. The status check itself is correct (the test at `mcp::tests::test_git_refs_denial_errors_not_empty_refs` covers that), but the read is uncapped. Please read through `read_body_capped` for denials the same way `sync.rs` and `json_or_denial` do, and only pass the bounded body to `parse_info_refs` on success.

- [P2] Emit `X-Gitlawb-Error` on the receive-pack `info/refs` rate limit
  `crates/gitlawb-node/src/error.rs:137-167`, `crates/gitlawb-node/src/api/repos.rs:556-563`
  This PR makes flood-brake 429s machine-readable by giving `rate_limit::too_many_requests()` both a JSON `error: rate_limited` body and an `X-Gitlawb-Error` header, and it changes `AppError::TooManyRequests` to use the same wire code in the JSON body. But `AppError::IntoResponse` still omits the header. `git_info_refs` applies the shared `push_rate_limiter` inline and returns `AppError::TooManyRequests`; the following `git-receive-pack` POST hits the same bucket through `rate_limit_by_ip` → `too_many_requests()`, which does carry the header. `gl` classifies denials from the header only, so the two steps of a push can return different 429 shapes from one limiter. Please route this path through the same response helper as the middleware brake, or add the header in `AppError::TooManyRequests`'s `IntoResponse`.

## Follow-up (not blocking merge)

- [P3] Serialize the per-identity ledger cap check under concurrency
  `crates/gitlawb-node/src/db/mod.rs:1530-1559`
  The `live` COUNT + conditional INSERT can both pass when two distinct signatures from the same identity arrive in parallel at `n == 511`, so the 512-row ceiling can be exceeded by roughly the in-flight request count. Replay atomicity is fine (`sig_hash` PK); this is a separate TOCTOU on the cap. Worth hardening if you need a strict ceiling under burst concurrency, but the per-IP brake bounds how fast rows accumulate and overshoot is bounded by parallelism, not unbounded.

- [P3] Finish the `json_or_denial` migration on PR/issue comment writes
  `crates/gl/src/pr.rs:437-450`, `crates/gl/src/pr.rs:473-486`, `crates/gl/src/issue.rs:311-324`
  Sibling commands in the same files were migrated; these three were not. `send_signed` still catches every `NodeDenial` with a header before they run, so this is polish for the proxy-strips-header / non-JSON-body cases you already document in `http.rs`, not a silent-success gap.

- [P3] Finish the `gl init` create-repo status/body ordering
  `crates/gl/src/init.rs:145-148`
  Tracked in #260. Signature ledger denials are caught by `send_signed`; the gap is non-JSON error bodies surfacing as parse failures. Fine to land here or in the follow-up you already filed.

- [P3] Move the retry-gate rustdoc onto `classify_notify_status`
  `crates/gitlawb-node/src/api/repos.rs:816-843`
  Docs-only. CodeRabbit's split is right: keep the one-line result summary on `NotifyAttempt` and move the classification rationale onto the function.

## Author hotspot check

I spent extra time on the three areas you called out to me:

**Layer ordering** — The `add_auth_layers` stack is correct for Axum's last-listed-runs-first semantics: `require_signature` → `require_ucan_chain` → `consume_signature` → handler. `send_full_auth` through the real `build_router` plus `a_ucan_rejection_does_not_burn_the_signature` would catch a UCAN/ledger swap. The ordering argument in `server.rs` matches the code.

**GET/HEAD skip** — The only GET behind `add_auth_layers` is `visibility::list_visibility` on the chained PUT/DELETE/GET route. That handler only reads and returns rules; HEAD is axum-derived from the same GET handler. I did not find a mutation reachable through GET or HEAD on a ledgered router.

**Fail-closed paths** — Missing `SignatureIdentity` returns 500 `signature_identity_missing`; ledger DB errors return 503 `signature_ledger_unavailable`. Neither path calls `next.run()`. Metrics tests cover both outcomes.

**Permissionless cap** — Cap counting on `key_fingerprint` (not wire DID spelling) looks correct for the stated model. A concurrent TOCTOU at the cap boundary is possible (see follow-up above) but is bounded by request parallelism and the per-IP signed-write brake, not a permissionless bypass of the cap scheme itself.

I did not find the documented scope limits (GraphQL mutation replay #257, cross-node replay, default-config unledgered peer routes) to be worse or more reachable than you described. Post-ledger signature spend on handler/per-DID/iCaptcha rejection is an explicit documented tradeoff, not a defect.

## CodeRabbit

The two actionable inline comments map to the P2 MCP finding and the P3 doc split above. The `init.rs` nit is in follow-up and #260.

## Merge gate

Mergeable, no conflicts. All required CI checks green on head `2f4995bae0d0cfaf052194371a0e3931a503ed1a`. Coordinate migration v16 ordering with #173 before either PR lands — that is an operational sequencing note, not a code defect in this diff.

jatmn (changes requested):

## Findings

- [P2] Route MCP error handling through the capped, sanitized denial helpers
  `crates/gl/src/mcp.rs:1303-1317`
  This PR introduces `json_ok` and migrates dozens of MCP call sites to it, and adds `read_body_capped` and `sanitize_node_msg` in `crates/gl/src/http.rs` for exactly this class of problem. `json_ok` itself still calls `resp.json()` with no byte cap and interpolates the node's raw `message`/`error` into tool errors without `sanitize_node_msg`. That is an incomplete migration within this diff, not pre-existing drift: base had scattered `resp.json()` calls and no cap/sanitize helpers at all. A malicious or compromised node can force an unbounded allocation on a failed tool call and smuggle terminal control sequences into MCP-visible errors.

## Follow-up (not blocking merge)

- [P3] Emit `X-Gitlawb-Error` on the receive-pack `info/refs` rate limit
  `crates/gitlawb-node/src/error.rs:137-167`, `crates/gitlawb-node/src/api/repos.rs:556-563`
  Commit `467d46d` correctly gave the flood brake `X-Gitlawb-Error: rate_limited` with a matching JSON body, and aligned `AppError::TooManyRequests` to the same JSON `error` code. `AppError::IntoResponse` still omits the header. On base neither the middleware brake nor `AppError::TooManyRequests` carried that header, so this is not a regression from `main` — it is a half-finished piece of the wire-contract work this PR started. Impact is narrower than it first looked: `git-remote-gitlawb` (the actual `git push` `info/refs` client) keys on status and body text, not the header, and both 429 paths now return JSON with `error: rate_limited`. The gap matters for header-keyed tooling (`gl` `signature_rejection`, peer `classify_notify_status`) and for making every `rate_limited` 429 indistinguishable on the wire. Routing this path through `too_many_requests()` or adding the header in `AppError::TooManyRequests`'s `IntoResponse` would close it.

- [P3] Cap `git_refs` denial bodies before buffering
  `crates/gl/src/mcp.rs:805-811`
  `git_refs` still calls `resp.bytes().await?` before checking `status`, so a non-2xx response is fully buffered even when the call will fail. That uncapped read predates this PR; base had no status check at all, so denial bodies could parse as an empty ref list. This PR fixed the serious bug by rejecting non-success statuses (`test_git_refs_denial_errors_not_empty_refs`). Applying `read_body_capped` on the denial branch would match the new helpers elsewhere — polish, not the same class as the silent-success defect.

- [P3] Serialize the per-identity ledger cap check under concurrency
  `crates/gitlawb-node/src/db/mod.rs:1552-1561`
  The `live` COUNT + conditional INSERT can both pass when two distinct signatures from the same identity arrive in parallel at `n == 511`, so the 512-row ceiling can be exceeded by roughly the in-flight request count. Replay atomicity is fine (`sig_hash` PK); this is a separate TOCTOU on the cap. Worth hardening if you need a strict ceiling under burst concurrency, but the per-IP brake bounds how fast rows accumulate and overshoot is bounded by parallelism, not unbounded.

- [P3] Finish the `json_or_denial` migration on PR/issue comment writes
  `crates/gl/src/pr.rs:437-450`, `crates/gl/src/pr.rs:473-486`, `crates/gl/src/issue.rs:311-324`
  Sibling commands in the same files were migrated; these three were not. `send_signed` still catches every `NodeDenial` with a header before they run, so this is polish for the proxy-strips-header / non-JSON-body cases you already document in `http.rs`, not a silent-success gap.

- [P3] Finish the `gl init` create-repo status/body ordering
  `crates/gl/src/init.rs:145-148`
  Tracked in #260. Signature ledger denials are caught by `send_signed`; the gap is non-JSON error bodies surfacing as parse failures. Fine to land here or in the follow-up you already filed.

- [P3] Move the retry-gate rustdoc onto `classify_notify_status`
  `crates/gitlawb-node/src/api/repos.rs:816-843`
  Docs-only. CodeRabbit's split is right: keep the one-line result summary on `NotifyAttempt` and move the classification rationale onto the function.

- [P3] Update `require_signature` module docs for the nonce parameter
  `crates/gitlawb-node/src/auth/mod.rs:73-78`
  The rustdoc still shows `Signature-Input` without `;nonce=...` even though `sign_request` now always emits one. Small doc drift against the wire format this PR ships.

## Rechecked on head `467d46d`

**`467d46d` fixes I verified:** the signed-write flood brake now carries `X-Gitlawb-Error: rate_limited` with a matching JSON body, and peer `/sync/notify` retries gate on that code (and `signature_ledger_unavailable`) rather than retrying every 429/503. The `gl` migration for profile/repo/fork and the status-before-parse pattern on most write commands look correct on the paths I rechecked.

**Severity recheck:** I compared base `fdf716dd` to head for the three items from the prior round. Only the MCP `json_ok` gap is PR-owned incomplete work with no pre-existing equivalent. The `git_refs` uncapped denial read is pre-existing hardening debt; the status-check fix in this PR addresses the worse silent-success bug. The missing `X-Gitlawb-Error` on `AppError::TooManyRequests` is a new wire-contract inconsistency this PR introduced by fully fixing the middleware path but not the inline handler path — real, but low impact on `git push` because `git-remote-gitlawb` does not read that header.

**Author hotspot check (unchanged on this head):**

**Layer ordering** — Still correct for Axum's last-listed-runs-first semantics: `require_signature` → `require_ucan_chain` → `consume_signature` → handler. `send_full_auth` through the real `build_router` plus `a_ucan_rejection_does_not_burn_the_signature` would catch a UCAN/ledger swap.

**GET/HEAD skip** — The only GET behind `add_auth_layers` is `visibility::list_visibility` on the chained PUT/DELETE/GET route. That handler only reads and returns rules; HEAD is axum-derived from the same GET handler. I did not find a mutation reachable through GET or HEAD on a ledgered router.

**Fail-closed paths** — Missing `SignatureIdentity` returns 500 `signature_identity_missing`; ledger DB errors return 503 `signature_ledger_unavailable`. Neither path calls `next.run()`. Metrics tests cover both outcomes.

**Permissionless cap** — Cap counting on `key_fingerprint` (not wire DID spelling) looks correct for the stated model. A concurrent TOCTOU at the cap boundary is possible (see follow-up above) but is bounded by request parallelism and the per-IP signed-write brake, not a permissionless bypass of the cap scheme itself.

I did not find the documented scope limits (GraphQL mutation replay #257, cross-node replay, default-config unledgered peer routes) to be worse or more reachable than you described. Post-ledger signature spend on handler/per-DID/iCaptcha rejection remains an explicit documented tradeoff, not a defect.

## CodeRabbit

The actionable inline MCP comment still maps to the P2 `json_ok` finding. The `init.rs` nit and the `classify_notify_status` doc split remain in follow-up above (#260 and docs-only, respectively). The `git_refs` cap note from CodeRabbit aligns with the P3 follow-up here, not a merge blocker.

## Merge gate

Mergeable, no conflicts. All required CI checks green on head `467d46deb6d09e30ad6026702d45c627cefc7037`. Coordinate migration v16 ordering with #173 before either PR lands — that is an operational sequencing note, not a code defect in this diff.

beardthelion:
Heads-up while this is open, since it touches `http_sig.rs`: `HttpSignature::parse` has a live panic on attacker-controlled input, now filed as #348.

`crates/gitlawb-core/src/http_sig.rs:60-67` finds `(` and `)` independently and then slices `&rest[open + 1..close]`, so a `Signature-Input` where `)` precedes `(` reverses the range and panics. Every other malformed shape in that function returns a typed `Error` that the caller renders as a 400; this one class walks past it. It is reachable pre-auth, because `auth/mod.rs:88` parses before any key is resolved and `optional_signature` delegates whenever signature headers are merely present.

This PR's diff does not touch that indexing (it adds `parse_exposes_nonce_and_tolerates_its_absence` around `parse` but leaves the `find`/slice lines alone), so the panic survives the merge either way. Flagging it here only so it is visible to whoever lands first: #306 and #261 both touch this file, and the fix in #348 is a single `open < close` comparison, so it is worth deciding deliberately whether it rides along with one of these or goes separately rather than having all three collide.

beardthelion:
Pushed `f37f29d8` onto `test/replay-guards-253` after merging current main.

**[P2] `json_ok`:** denial path now uses `read_body_capped` (64 KiB) and `sanitize_node_msg` before bailing (`crates/gl/src/mcp.rs`). `cargo test -p gl test_json_ok_denial_sanitizes_node_message` green on this head.

Also on this head (follow-up hardening from the review pass, carried through the main merge):

- **`git_refs`:** capped body read; malformed 200 advertisements fail closed instead of empty success (`mcp.rs`).
- **`X-Gitlawb-Error: rate_limited`** on `AppError::TooManyRequests` (alongside main's `Retry-After` on retryable 503s).
- **Ledger cap TOCTOU:** per-identity `pg_advisory_xact_lock` so concurrent spends cannot overshoot the cap (`db/mod.rs`).
- **Peer notify:** shared retry sleep budget across the fan-out (`repos.rs`).

**Checked:** pre-push `cargo clippy --locked --workspace --all-targets -D warnings` green; `cargo test --workspace --locked` 1164/1165 on the merge tree (`ipfs_pin::discovery_record_elapsed_leaves_neither_row` failed once under full parallel load, green on isolated rerun).

Ready for another look.

jatmn (changes requested):

## Findings

- [P1] Reject capped or failed git ref advertisements instead of parsing their prefix
  `crates/gl/src/mcp.rs:810`
  `read_body_capped` deliberately reports `truncated` and `read_failed`, but `git_refs` discards both flags and validates/parses only the bytes it happened to receive. A large valid pkt-line advertisement can end exactly at the 256 KiB cap (or a response can fail after a complete prefix), so `validate_git_advertisement` accepts the prefix and the MCP tool returns a successful but incomplete ref list. Agents can then make branch, merge, or release decisions as though omitted refs do not exist.

  The root cause is treating a bounded/error-prone transport read as if it had reached a clean EOF. Make completeness part of the protocol boundary: fail `git_refs` before validation whenever `truncated` or `read_failed` is set, with a distinct actionable error. Add regression coverage for a syntactically valid packet prefix at the cap and for a body that fails after such a prefix; do not simply relax or remove the response cap.

- [P2] Keep smart-HTTP advertisements as raw bytes
  `crates/gl/src/mcp.rs:811`
  The new capped reader decodes the advertisement with `String::from_utf8_lossy`, then this converts it back with `as_bytes()` for pkt-line parsing. Git ref names can contain non-UTF-8 bytes, and that round trip changes both those bytes and pkt-line length accounting, so valid repositories can fail validation or have their advertised refs corrupted. The base implementation consumed raw response bytes, making this a protocol regression introduced by the hardening path.

  The root cause is sharing a text-oriented denial helper with a byte-oriented Git protocol consumer. Introduce a capped raw-byte read result (including the same completion flags) for smart-HTTP payloads, and derive sanitized text only in paths that display node-supplied errors. Add a non-UTF-8 ref-name advertisement test so this boundary cannot silently revert.

- [P2] Do not throttle the signed visibility read with the signed-write bucket
  `crates/gitlawb-node/src/server.rs:213`
  The new signed-write IP limiter wraps the mixed-method `write_routes` router, which also serves `GET /api/v1/repos/{owner}/{repo}/visibility`. `rate_limit_by_ip` does not skip reads, while `gl visibility list` deliberately signs that owner-only GET. Consequently normal reads consume the write quota and can return 429 after writes (or other reads) from the same IP, contrary to the write-only limiter contract; a shared NAT makes the unintended coupling particularly visible.

  The root cause is classifying a router by its dominant purpose instead of by the methods it actually dispatches. Either split the signed visibility GET onto a read router outside this limiter, or make this specific limiter bypass GET/HEAD just as `consume_signature` does. Retain the limiter for every mutation and add a router-level test proving that an exhausted signed-write bucket rejects a PUT/DELETE but still serves the authenticated visibility GET.

- [P2] Reject an empty successful git advertisement
  `crates/gl/src/mcp.rs:1338`
  `validate_git_advertisement` returns `Ok(())` for an empty body, so a 200 response with no pkt-line data still becomes a successful empty ref list. This change claims malformed 200 advertisements fail closed and adds this validator for that purpose, but even an initialized empty bare repository emits a non-empty upload-pack advertisement (capabilities plus pkt-line framing). A proxy or broken node can therefore make an unavailable ref set look empty.

  The root cause is that the validator only checks malformed data once a packet loop starts; its zero-iteration path is implicitly accepted. Reject empty input before the loop and add an explicit 200-empty-body regression test alongside the existing fake-JSON response test. Keep legitimate empty-repository advertisements valid rather than conflating an empty ref set with an empty protocol response.

PR #262, opened 2026-07-28
fix(node): bound the request body the signature middleware buffers before it verifies

beardthelion:
`require_signature` buffers the whole request body with an unlimited `body.collect()` before it parses a single header, and nothing upstream caps it.

On the git write routes the pack-size `RequestBodyLimitLayer` (2 GB by default, 500 MB in production) is applied to the inner router while `add_auth_layers` wraps it. A later `.layer()` is the outer service, so the middleware runs first and sees the raw body; the limit never applies to it. The other signed groups carry no limit layer at all and rely on axum's 2 MB `DefaultBodyLimit`, which the extractors enforce through a request extension this middleware never consults, so that default is bypassed as well.

Measured against the real router, with `GITLAWB_MAX_PACK_BYTES` set to 64 KiB and a 4 MiB body on an unsigned push:

```
limit=65536  status=401 Unauthorized  bytes_drained=4194304
```

The full body is read, then the request is rejected for having no signature. No DID, no signature and no repo are needed to get there. In front of it sits only a per-IP arrivals-per-hour counter, and there is no timeout layer and no connection cap on `axum::serve`.

The reach is wider than the git routes: `optional_signature` forwards into the same code whenever a caller merely presents signature headers, so `/graphql`, `/ipfs/{cid}` and the other optional-signature groups have it too. Fifteen route groups in total.

## The fix

The collect now runs through `http_body_util::Limited`, so reading stops at the limit rather than after the fact, and an over-limit request gets 413 instead of the 400 reserved for a genuinely unreadable body.

The limit is per route group. `add_auth_layers` takes it and hands it to the middleware as state; the git groups keep `max_pack_bytes` so large pushes are unaffected, and the rest get a constant equal to axum's default, which is already what their extractors enforce. That makes the change behavior-preserving on the response path and purely a memory fix. Signature verification, the content-digest check, clock skew, and the request reconstruction are untouched.

## Verification

The regression test asserts bytes actually polled out of a streaming body rather than the status code, because a status assertion passes whether or not the body was drained.

There is also an end-to-end pair driving a real push: a source repo with a 300 KiB blob and the byte-for-byte receive-pack body captured off the wire from an actual `git push`. Under the limit it asserts the ref moved and the tree resolves, not merely a 200. That distinction is load-bearing, since a truncated body still returns 200 with the failure inside report-status; only the on-disk assertion catches it. Running the whole suite under a truncation mutation leaves exactly one test failing. Over the limit it asserts 413 with neither the ref nor the commit present, so a rejected push cannot half-write.

Each guard was checked by reverting it: neutralizing the limit turns the drain tests red, handing every group the git limit turns the non-git test red, and returning 400 instead of 413 turns the status test red.

518 tests pass, fmt and clippy are clean.

## Residual

This bounds per-request buffering only. It does not bound how many such requests are in flight, or how long each may take to send its body, so total resident memory is still proportional to concurrent connections times the limit.

To be precise about what does exist, since the first version of this paragraph was wrong: the Fly configs do cap connections (`hard_limit = 500`), and they do set `idle_timeout = 120`. Neither helps. The connection cap is sized above what a 1024 MB machine can hold, and a body dripping a byte every few seconds is never idle. There is no request-duration or body-read timeout in process, and `rate_limit_by_ip` counts arrivals per window rather than bounding concurrency.

Filed as #263 with the measurements.

## Summary by CodeRabbit

* **New Features**
  * Added configurable request-body limits for signed and optionally signed requests.
  * Added support for properly signed Git operations within the configured pack-size limit.
  * Added per-IP rate limiting for IPFS content retrieval.

* **Bug Fixes**
  * Oversized requests now return a clear `413 Payload Too Large` response.
  * Improved distinction between oversized and unreadable request bodies.
  * Rejected oversized Git operations no longer modify repository data.


jatmn (changes requested):

## Merge and checks

- Mergeability: `MERGEABLE`; GitHub merge state is `BLOCKED` (policy/review gate, not a conflict).
- Required PR checks on head `e8393fdc8008da2a07dd81da73b00d8f7616e2ba`: fmt + clippy, test (stable), test (beta), build --release, cargo audit, MSRV, Docker smoke, and the rest of the PR Checks workflow are green. `Analyze (swift)` was still pending on the CodeQL workflow at review time; it is unrelated to this Rust-only diff.

## What I verified

- The core fix is sound: per-group `SignatureBodyLimit` wired through `add_auth_layers` and `optional_signature`, `http_body_util::Limited` on pre-auth `collect()`, and `413 body_too_large` for over-limit streams close the unbounded pre-auth drain on both git and non-git signed surfaces.
- Production wiring matches the stated intent: non-git signed groups use `SIGNED_BODY_LIMIT` (2_097_152, matching axum’s default), git write/read groups use `pack_limit`.
- All eight new body-limit tests passed locally, including the real receive-pack e2e fixtures. `cargo clippy -p gitlawb-node -- -D warnings` is clean.

## Out of scope / follow-up (not blockers)

- Residual memory still scales with concurrent connections × per-route limit; read timeout and global in-flight cap are appropriately deferred.
- Header-before-buffer fast reject for incomplete signature attempts (one header present, or missing headers on `require_signature` routes) is real hardening in the same area but pre-existing ordering this PR does not worsen; worth a separate follow-up if desired.
- Overlap with #261 (replay ledger) will need merge coordination but does not conflict with this limit wiring.

jatmn (approved):
LGTM

beardthelion:
@kevincodex1 LGTM

PR #267, opened 2026-07-28
fix(ci): make the docker dependency cache work and unbreak Dockerfile.bins

### LINKED ISSUE #265: ci(docker): the Dockerfile dependency-cache layer is a no-op, every image build is cold
The dependency-cache layer in `Dockerfile` has been a no-op since the workspace grew past four crates, so every image build compiles the full dependency graph from scratch.

## What is wrong

`Dockerfile:12-24` copies four crate manifests before priming the dependency cache:

```dockerfile
COPY Cargo.toml Cargo.lock ./
COPY crates/gitlawb-core/Cargo.toml crates/gitlawb-core/
COPY crates/gitlawb-node/Cargo.toml crates/gitlawb-node/
COPY crates/gl/Cargo.toml crates/gl/
COPY crates/git-remote-gitlawb/Cargo.toml crates/git-remote-gitlawb/
```

The workspace has six members. `Cargo.toml` also lists `crates/gitlawb-attest` and `crates/icaptcha-client`, and `crates/gl/Cargo.toml:15` takes a path dependency on the latter:

```toml
icaptcha-client = { path = "../icaptcha-client" }
```

So cargo fails while loading the workspace, before it ever resolves dependencies or reads `Cargo.lock`. The `|| true` at the end of the layer swallows the error and the build proceeds as if the cache had been primed.

## Reproduction

Built the first 24 lines of `Dockerfile` unchanged (docker 29.1.3), with a diagnostic step appended:

```
--- target/release/deps rlib count ---
0
--- registry cache size ---
no registry cache
--- rerun the layer command without || true ---
error: failed to load manifest for workspace member `/build/crates/gl`
referenced by workspace at `/build/Cargo.toml`

Caused by:
  failed to load manifest for dependency `icaptcha-client`

Caused by:
  failed to read `/build/crates/icaptcha-client/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
exit=101
```

Zero compiled rlibs and no registry cache: the layer downloads nothing and builds nothing.

## Impact

Every build of the main image is a cold build of the whole dependency graph, including the `docker-build` smoke test in `pr-checks.yml` (30-minute timeout) and the `docker` job that pushes to ghcr on release. The real build at `Dockerfile:37` still produces a correct image, so this is wasted CI time rather than a correctness bug. It also silently defeats the layer-caching intent stated in the comment at line 18, so a reader would reasonably assume dependencies are cached when they are not.

## Fix

Copy all six member manifests (and create the matching dummy `src` entries) so the layer can resolve, and consider dropping the `|| true` that hid this. A trailing `|| true` on a cache-priming step means the next crate added to the workspace breaks it silently again; if the layer is expected to succeed, letting it fail loudly is what would have surfaced this at the time.

Noticed while adding `--locked` to the shipping builds in #243. Filing separately rather than widening that PR.

## The fix above is verified, not just proposed

Built the same layer with all six manifests copied, matching dummy `src` entries added, and the `|| true` dropped:

```
--- rlib count after the fixed cache layer ---
619
--- registry cache ---
508M	/usr/local/cargo/registry
```

619 compiled rlibs and a 508 MB registry cache, against 0 and none today. The layer also succeeds without `|| true`, so the guard is not load-bearing once the manifest list is complete.

### LINKED ISSUE #266: ci(docker): Dockerfile.bins pins rust:1.85 below the 1.91 MSRV, so it cannot build
`Dockerfile.bins` pins a Rust image below the workspace MSRV, so it cannot build. Nothing in CI builds it, so no gate reports this.

## What is wrong

`Dockerfile.bins:6`:

```dockerfile
FROM rust:1.85-bookworm AS builder
```

`Cargo.toml:21` declares `rust-version = "1.91"`, and all six crates inherit it via `rust-version.workspace = true`. An unmet `rust-version` is a hard cargo error, not a warning, so the build fails before compiling anything.

## Reproduction

`docker build -f Dockerfile.bins --build-arg TARGET=aarch64-unknown-linux-musl .` (docker 29.1.3):

```
error: rustc 1.85.1 is not supported by the following packages:
  gl@0.7.0 requires rustc 1.91
  gitlawb-core@0.7.0 requires rustc 1.91
  git-remote-gitlawb@0.7.0 requires rustc 1.91
  icaptcha-client@0.4.0 requires rustc 1.91
  alloy-*@1.7.3 requires rustc 1.91
  ...
The command '/bin/sh -c cargo build --release --locked --target "$TARGET" -p gl -p git-remote-gitlawb' returned a non-zero code: 101
```

Same failure with `--locked` removed, so the flag added in #243 is not the cause; the image has been unbuildable since the MSRV moved to 1.91.

## Impact

`scripts/build-bins.sh` uses this file for the two Linux musl targets (`build-bins.sh:29`), so the linux-x86_64 and linux-arm64 paths of that script are broken. The darwin-arm64 path builds natively and is unaffected. Released binaries are not affected either: `release.yml` builds those itself on matching-arch runners.

No workflow references `Dockerfile.bins`, which is why this has stayed quiet. Grepping `.github/` for it returns only the `CODEOWNERS` entry.

## Fix

Bump the base image to a tag at or above the MSRV (`rust:1.91-bookworm`, matching what the main `Dockerfile` already uses). Worth deciding at the same time whether this file should be kept: if `scripts/build-bins.sh` is still a supported path, a CI job that builds it would stop the next MSRV bump from breaking it silently, and if it is not, deleting both is the smaller change.

Noticed while adding `--locked` to the shipping builds in #243. Filing separately rather than widening that PR.

## The fix above is verified, not just proposed

Rebuilt this file unchanged except for `FROM rust:1.91-bookworm`, same `TARGET=aarch64-unknown-linux-musl`: it clears the MSRV error and completes, producing a 158 MB image. So the base-image bump is sufficient on its own; nothing else in the file needs to change to make it build.

beardthelion:
Two container builds that do not work, both found while adding `--locked` to the shipping builds in #243.

Closes #265. Closes #266.

## The dependency-cache layer never cached anything

`Dockerfile` primed its dependency cache from four crate manifests, but the workspace has six members and `crates/gl` takes a path dependency on `icaptcha-client`. Cargo loads the whole workspace before resolving anything, so the layer aborted on the missing manifest and `|| true` swallowed it. Every image build has been a cold build of the full dependency graph since `icaptcha-client` was added, including the `docker-build` smoke test on every PR.

Copying all six manifests fixes it. Dropping the `|| true` is the part that keeps it fixed: the guard is what let this sit unnoticed, and this layer has no reason to tolerate failure.

Before, building the first 24 lines unchanged:

```
--- target/release/deps rlib count ---
0
--- registry cache size ---
no registry cache
```

After:

```
--- rlib count after the fixed cache layer ---
619
--- registry cache ---
508M	/usr/local/cargo/registry
```

## Dockerfile.bins could not build at all

It pinned `rust:1.85-bookworm` against a workspace `rust-version = "1.91"`. An unmet MSRV is a hard cargo error, so `scripts/build-bins.sh` has been broken for both Linux musl targets. Nothing in CI builds this file, which is why it stayed quiet. Released binaries are unaffected: `release.yml` builds those on matching-arch runners.

Bumping the base image to `rust:1.91-bookworm`, matching the main `Dockerfile`, is the whole fix. Verified by building it: it now completes and produces a 158 MB image.

## Verification

Real `docker build` runs, not reasoning about the files:

- The full `gitlawb-node` image builds end to end with the fixed cache layer.
- Removing one member manifest from the fixed layer now **fails the build** with exit 101 (`failed to read /build/crates/icaptcha-client/Cargo.toml`) instead of continuing silently, so the removed `|| true` is load-bearing and the next crate added to the workspace cannot break this quietly again.
- `Dockerfile.bins` with the new base image builds to completion for `aarch64-unknown-linux-musl`; on the old base it fails with `error: rustc 1.85.1 is not supported by the following packages`. It fails identically with and without `--locked`, so #243's flag was never the cause.

One consequence worth naming: until now the cache layer produced no artifacts at all, so the dummy-binary guard at `Dockerfile:26-38` (mtime bump, binary removal, fingerprint removal) was never actually exercised. It is now. I checked the built image rather than assuming it held: it ships a 36 MB `gitlawb-node` reporting `0.7.0` and a 12 MB `gl` reporting `0.7.0`, not the `fn main() {}` stubs.

## Base

Stacked on `ci/242-lock-sync-release` (#243), which touches the same `Dockerfile` line. GitHub will retarget this to `main` when #243 merges. Nothing here depends on #243 semantically, only textually.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,20 +8,33 @@ FROM rust:1.91-bookworm AS builder
 
 WORKDIR /build
 
-# Cache dependencies first for faster rebuilds
+# Cache dependencies first for faster rebuilds.
+# EVERY workspace member's manifest must be copied here, not just the ones this
+# image ships: cargo loads the whole workspace before it resolves anything, so a
+# missing member's Cargo.toml (or a path dependency pointing at one) aborts the
+# layer before a single crate is fetched.
 COPY Cargo.toml Cargo.lock ./
 COPY crates/gitlawb-core/Cargo.toml crates/gitlawb-core/
 COPY crates/gitlawb-node/Cargo.toml crates/gitlawb-node/
 COPY crates/gl/Cargo.toml crates/gl/
 COPY crates/git-remote-gitlawb/Cargo.toml crates/git-remote-gitlawb/
+COPY crates/gitlawb-attest/Cargo.toml crates/gitlawb-attest/
+COPY crates/icaptcha-client/Cargo.toml crates/icaptcha-client/
 
-# Fetch deps (this layer caches until Cargo.{toml,lock} change)
-RUN mkdir -p crates/gitlawb-core/src crates/gitlawb-node/src crates/gl/src crates/git-remote-gitlawb/src && \
+# Fetch deps (this layer caches until Cargo.{toml,lock} change).
+# No `|| true`: this layer is expected to succeed, and swallowing its exit code
+# is what let it sit dead from the moment gl gained a path dependency on
+# icaptcha-client. A failure here should stop the build, not silently turn every
+# image build into a cold one.
+RUN mkdir -p crates/gitlawb-core/src crates/gitlawb-node/src crates/gl/src \
+        crates/git-remote-gitlawb/src crates/gitlawb-attest/src crates/icaptcha-client/src && \
     echo 'fn main() {}' > crates/gitlawb-node/src/main.rs && \
     echo 'fn main() {}' > crates/gl/src/main.rs && \
     echo 'fn main() {}' > crates/git-remote-gitlawb/src/main.rs && \
     echo '' > crates/gitlawb-core/src/lib.rs && \
-    cargo build --release --locked -p gitlawb-node -p gl -p git-remote-gitlawb || true
+    echo '' > crates/gitlawb-attest/src/lib.rs && \
+    echo '' > crates/icaptcha-client/src/lib.rs && \
+    cargo build --release --locked -p gitlawb-node -p gl -p git-remote-gitlawb
 
 # Now copy real sources and build for real.
 # Force-bump mtimes so cargo's fingerprint check rebuilds — without this,
--- a/Dockerfile.bins
+++ b/Dockerfile.bins
@@ -3,7 +3,10 @@
 # Used by scripts/build-bins.sh — not used for node deployment.
 ARG TARGET=x86_64-unknown-linux-musl
 
-FROM rust:1.85-bookworm AS builder
+# Must stay at or above the workspace `rust-version` in Cargo.toml. An unmet
+# MSRV is a hard cargo error, and nothing in CI builds this file, so a stale
+# base image here fails only when someone runs scripts/build-bins.sh.
+FROM rust:1.91-bookworm AS builder
 ARG TARGET
 
 WORKDIR /build


jatmn (approved):
LGTM

kevincodex1 (approved):

Issue #273, opened 2026-07-29
Announce does not bind a DID to its first-seen key, so any caller can rewrite any peer's http_url


beardthelion:
`upsert_peer` has no notion of who owns a DID's row. Any caller can rewrite any peer's `http_url`, because the conflict branch is an unconditional update and the announce route accepts unsigned callers by default. #270 proposes resetting `last_ping_ok` when the URL changes, which limits the blast radius on one consumer. This issue is the underlying question: should a DID be bound to the key that first announced it, so the rewrite is rejected rather than merely defanged?

Split out of #270 so the contained mitigation there can land without waiting on this decision.

## Why the mitigation is not sufficient on its own

The reset-on-change fix keys on `last_ping_ok`, but three of the four consumers of `http_url` never read that flag:

- `sync.rs:204` resolves a queued sync item's origin URL by DID and hands it to `clone_repo`/`fetch_repo` as the git remote. No reachability filter.
- `api/repos.rs:1350` fans out the post-receive announce to every peer with a non-empty `http_url`. No reachability filter.
- `api/resolve.rs:24` serves `GET /api/v1/resolve/{did}` from the peer table and returns the stored URL as the answer for that DID. No reachability filter, and the route is public (`server.rs:381`).
- `api/repos.rs:1451` is the only one that filters on `last_ping_ok`, and it is the one #270's fix addresses.

So a repointed row keeps steering the sync worker, the notify fan-out, and public DID resolution regardless of what the flag says.

## The decision

Options as I see them:

1. **First-writer-wins on the DID.** Record the announcing key on insert; later announces for that DID must carry a signature from the same key. Closes the rewrite outright, works whether or not `require_signed_peer_writes` is on, and does not depend on any downstream consumer reading a flag. Cost is a migration plus a story for peers whose rows predate the column, and a documented recovery path for a genuine key rotation.
2. **Flip `require_signed_peer_writes` to true by default.** Simpler, but it only requires *a* valid signature matching the announced DID, which is what an attacker announcing their own DID already has. It closes the hijack of someone else's row and nothing else, and it is a breaking change for peers that have not upgraded, which is why it is off today.
3. **Ship #270's reset and accept the residual.** Cheapest, and leaves the three unfiltered consumers exposed.

My read is that 1 is the actual fix and 3 is a stopgap worth landing first, but this is a federation-model call rather than a bug with an obvious patch, which is why it is its own issue.

## Related

- #270, the announce hijack and the contained mitigation.
[...]

PR #274, opened 2026-07-29
fix(node): stop a peer-supplied repo slug from escaping repos_dir (#272)

### LINKED ISSUE #272: Unvalidated repo slug on the unsigned notify route escapes repos_dir in the sync worker's clone path
Two unauthenticated requests put an attacker-chosen `repo` string into `sync_queue`, and the sync worker joins that string straight into a filesystem path with no validation. `a//tmp/x` resolves to the absolute `/tmp/x.git`, outside `repos_dir` entirely.

Split out of #270. That issue covers the announce/`upsert_peer` hijack; this one is the unvalidated slug, which is reachable without any hijack.

## Mechanism

`notify_sync` (`crates/gitlawb-node/src/api/peers.rs:355-388`) accepts unsigned callers by default (`require_signed_peer_writes` is false at `config.rs:48`, so `server.rs:318` mounts the route under `optional_signature`). Its only gate is membership:

```rust
let known = peers.iter().any(|p| p.did == req.node_did);
```

An attacker satisfies that with one unsigned announce of their own DID. `req.repo` is then passed through verbatim: `enqueue_sync` (`db/mod.rs:1615`) is a bare INSERT with no validation, and `process_batch` (`sync.rs:225-239`) derives the clone target from it:

```rust
let (owner_short, repo_name) = item.repo.split_once('/')?;
let local_path = config.repos_dir.join(owner_short).join(format!("{repo_name}.git"));
```

`PathBuf::join` does not normalize, and an absolute second component replaces the accumulated path.

Note that `last_ping_ok` is irrelevant here. `sync.rs:204` resolves the origin URL from the peer table with no reachability filter, so a brand-new attacker DID (which inserts with `last_ping_ok = FALSE`) works immediately.

## Verified by execution

Against `origin/main` at 111cff7, a throwaway `#[sqlx::test]` drove the mounted router with no auth extension: unsigned announce of a generated DID returned 200, then an unsigned notify naming `repo = "a//tmp/gitlawb-probe"` returned 200, and the row landed in `sync_queue`:

```
announce=200 OK  notify=200 OK
queued=[("a//tmp/gitlawb-probe", "did:key:z6MkfkpfS5pYKSs7RKxdUPWRVeH2NXe2ss11azRyRHCnXPYr")]
peer=("https://attacker.example.com", false)
```

The path derivation, run separately against the same expression:

```
       z6Mkfoo/bar  ->  /srv/repos/z6Mkfoo/bar.git
    ../../etc/evil  ->  /srv/repos/../../etc/evil.git
       a//etc/evil  ->  /etc/evil.git
```

Not executed: the worker actually consuming that row and running the clone. `clone_repo` needs a reachable origin, and `is_public_http_url` rejects loopback, so a local end-to-end run needs a public host. The two handler hops and the path derivation are executed; the worker's consumption of the queued row is read from `sync.rs:189-275` and not run.

## Why it matters

The clone target is `git clone --mirror <attacker url> <attacker-influenced path>`, running as the node user. Writing outside `repos_dir` is the first problem. The second is that even a well-formed slug naming an existing mirrored repo is unowned: nothing checks that `node_did` is that repo's actual origin, so any known peer can claim any repo slug. `fetch_repo` (`sync.rs:646`) then does `git remote set-url origin <that peer's url>` followed by `fetch --prune` on a `--mirror` clone, whose refspec is `+refs/*:refs/*`, a forced overwrite plus prune of every ref.

## Fix direction

Three separable pieces, roughly in order of cost:

1. Validate `repo` at the trust boundary in `notify_sync`, and again in `process_batch` before the join. Exactly one `/`, both halves non-empty, no `.`/`..` components, and a character allowlist matching what `create_repo` already accepts.
2. Build the path with `git::store::repo_disk_path` rather than a hand-rolled `join`, so the sanitizing convention has one definition instead of two. Worth checking whether the two agree today for legitimate slugs; `repo_disk_path` replaces `:` and `/`, and the sync path does not.
3. Check that `node_did` is the recorded origin for the repo before enqueueing, so a known peer cannot claim a slug that is not theirs.

beardthelion:
Closes #272.

An unauthenticated caller could plant a git mirror at an arbitrary path on disk.

## The bug

`POST /api/v1/sync/notify` accepts unsigned callers by default: `require_signed_peer_writes` is false, so `server.rs` mounts the peer-write routes under `optional_signature` rather than the auth layers. Its only gate is that `node_did` appear in the peers table, and one unsigned announce of your own DID satisfies that. `req.repo` then reached `enqueue_sync` verbatim, and `process_batch` split it on the first `/` and joined both halves onto `repos_dir`. `PathBuf::join` does not normalize, and an absolute second component replaces the accumulated path, so `a//tmp/x` resolved to `/tmp/x.git`.

Reproduced end to end before any of this was written: unsigned announce 200, unsigned notify 200, then the worker cloned an attacker-controlled origin to `/tmp/gitlawb-escape-probe.git` with `repos_dir` left empty and the peer's `last_ping_ok` FALSE. The sync path never reads that flag, so nothing about peer reachability was in the way.

## The fix

Four layers, each catching something the others cannot:

| Layer | Where | What only it stops |
|---|---|---|
| slug validation | `repo_store::validate_repo_slug` | `./hello`, which canonicalizes back inside `repos_dir` and so passes containment |
| notify boundary | `api/peers.rs` | keeps the row out of `sync_queue` and out of `received_ref_updates` |
| worker re-check | `sync.rs` `process_batch` | rows queued before this landed, plus the gossip and trigger writers |
| canonical containment | `repo_store::path_within_root` | a symlink at the owner directory or at the `.git` leaf, which no character rule can see |

The validator reuses `validate_owner_did` and `validate_repo_name` rather than adding a second character rule to the crate. It adds two rules of its own: an owner half may not begin with `.` or `-` (those validators have no leading-character rule, and an owner of `.` puts a peer-controlled mirror at the `repos_dir` root), and the owner half is bounded by NAME_MAX at 255 rather than the DID column's 256.

`Containment` is three-valued rather than a bool, and that distinction is load-bearing. `Outside` is a deterministic verdict, so the row is retired. `IoError` means the filesystem could not answer, so the row stays pending and is retried, because marking it failed is terminal and a transient EACCES would otherwise permanently retire a legitimate repo. The predicate is pure; `create_dir_all` lives at the call site so the admin purge path in #196 can ask a containment question without mutating anything.

## Verification

549 tests, clippy clean under `-D warnings`, `cargo fmt --check` clean.

Every guard was mutation-tested: removed, observed red on a named case, restored. Worth stating plainly because five tests here were green before the guard they named existed, and were rewritten once that was found:

- `../hello` does not exercise the owner delegation; the leading-character rule catches it first. `a..b/hello` does.
- None of the three extra-separator cases isolate the separator count; each is caught by an earlier rule. `z6Mkfoo/hello/extra` does.
- The malformed-slug worker rows ended `failed` regardless when the remote was unreachable, because the clone simply failed. The fixture now serves every composed URL for real.
- The overlong-owner test asserted 257, one past the bound, so it passed for free. It now pins 256 rejected and 255 accepted.
- One root-skip probe used `symlink_metadata` on a path that was never created, which is `Err` for every user, so its guard could never fire.

The two escape tests assert the end-to-end property rather than isolating a layer: the slug rule and containment are each independently sufficient for that input, so removing either alone leaves them green and only removing both turns them red, with the mirror observed written outside `repos_dir`. Per-layer isolation lives in the `./hello` case and the two symlink tests. The comments say so.

An independent adversarial pass through a different model family returned no findings against the final state.

## Deliberately not in this PR

- Requiring `node_did` to be the repo's recorded origin. That needs an answer to what happens on the first sync of a repo this node has never seen, which depends on the DID key-binding decision in #273.
- Unifying the three path conventions (`RepoStore::local_path`, the `sync.rs` join, `repo_disk_path`). It relocates every existing mirror and orphans the `disk_path` already recorded on mirror rows, so it needs its own change with a migration.
- An attempt cap on the deferred arm. Bounding retries needs an `attempts` column on `sync_queue`, which is a migration. The new `deferred` and `rejected` metric labels are what make a stalled queue visible in the meantime.

## Known residuals

A well-formed slug naming a repo this node already mirrors still lets a known peer drive a fetch into it, repointing `remote.origin` and force-overwriting refs through the mirror refspec. That survives every guard here because the slug is well formed and the path stays inside `repos_dir`. It closes with the origin-ownership check above.

The window between the containment check and the git invocation is not atomic. Exploiting it needs local write access inside `repos_dir`, so it sits outside the unauthenticated threat model this addresses.

The gossip and trigger writers still reach `sync_queue` without an ingress check. The worker gate covers the path-write consequence; the ref-update feed remains poisonable with arbitrary strings, which is pre-existing.

Note for whoever rebases: `repo_store.rs` is also touched by #196, additively and in the same area. Whichever lands second resolves a small textual conflict there. Nothing else in this branch overlaps it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened sync/notify handling by rejecting malformed repository slugs and preventing traversal and symlink-based escape attempts.
  * Ensured invalid requests are neither queued nor recorded, while valid syncs continue to mirror correctly.
  * Improved mirror placement safety using stronger containment checks with safe handling of filesystem resolution issues.
* **New Features**
  * Added tracking for sync dequeue attempts to improve retry scheduling and ordering.
* **Documentation**
  * Updated sync metrics documentation to include newly documented `rejected` and `deferred` statuses.
* **Tests**
  * Expanded end-to-end and database tests covering slug validation, escape scenarios, and queue retry ordering.

### DIFF
--- a/crates/gitlawb-node/src/api/peers.rs
+++ b/crates/gitlawb-node/src/api/peers.rs
@@ -381,6 +381,14 @@ pub async fn notify_sync(
         )));
     }
 
+    // #272: the slug reaches a PathBuf::join in the sync worker, where an extra
+    // separator or a traversal segment places the mirror outside repos_dir. Reject
+    // it here, before the queue row and the ref-update row are written. The message
+    // names the field and nothing about where repos live on disk.
+    if let Err(e) = crate::git::repo_store::validate_repo_slug(&req.repo) {
+        return Err(AppError::BadRequest(format!("invalid repo field: {e}")));
+    }
+
     state
         .db
         .enqueue_sync(&req.repo, &req.node_did, &req.ref_name, &req.new_sha, None)
@@ -584,7 +592,9 @@ mod tests {
         state.config = Arc::new(cfg);
     }
 
-    const NOTIFY_BODY: &str = r#"{"repo":"demo","ref_name":"refs/heads/main","new_sha":"0000000000000000000000000000000000000000","node_did":"PEER_DID"}"#;
+    // The slug must be a well-formed owner/name (#272); these brake tests are about
+    // the rate limiter, so the body carries a valid slug and reaches the brake.
+    const NOTIFY_BODY: &str = r#"{"repo":"z6Mkfoo/hello","ref_name":"refs/heads/main","new_sha":"0000000000000000000000000000000000000000","node_did":"PEER_DID"}"#;
 
     // ── trigger: mandatory signature ──────────────────────────────────────────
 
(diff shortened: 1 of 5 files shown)


jatmn (changes requested):

## Findings

- [P2] Do not let deterministic directory failures pin the whole sync queue
  `crates/gitlawb-node/src/sync.rs:255`
  A dangling owner symlink, symlink loop, or regular file at `repos_dir/<owner>` makes `create_dir_all` fail with errors such as `AlreadyExists`/`FilesystemLoop`, but only three error kinds are treated as permanent. The row remains `pending`; because `dequeue_pending_syncs(10)` is oldest-first, ten such rows are retried every poll and permanently starve every healthy item behind them. This is a regression from the prior clone/fetch path, which terminally failed these rows. Classify deterministic filesystem states as terminal (after a non-following containment check) or add retry scheduling/queue progress so deferred rows cannot monopolize the batch.

beardthelion:
Confirmed, with one correction to the mechanism and one addition.

All three states you listed come back as `AlreadyExists`, not `FilesystemLoop`: `mkdir` returns EEXIST before the loop is ever traversed. I checked with a standalone probe against a real filesystem, and a dangling symlink, a regular file, and a symlink loop all report `AlreadyExists` (a symlink pointing at a directory outside the root returns `Ok` and is caught downstream as `Outside`, which is the intended path). None of them were in the permanent set, so your starvation analysis follows. The regression claim holds as well: the `create_dir_all` block is new in this PR, and before it the same filesystem state failed the clone and landed in the terminal `Err` arm.

Classifying error kinds does not close the class, though. An owner directory that exists but denies traversal makes `create_dir_all` return `Ok`, because `mkdir` gets EEXIST and the `is_dir()` fallback succeeds, so it never reaches that match at all. It stalls one line later in `path_within_root`, where `symlink_metadata` returns `PermissionDenied` and lands in `Containment::IoError`, deferring identically. That one is per-repo rather than a whole-`repos_dir` outage, so it starves healthy rows the same way while looking like an ordinary transient failure.

So both halves:

- `AlreadyExists` is now permanent. It only surfaces when something that is not a directory occupies the path, since a concurrent `mkdir` resolves to `Ok` through the `is_dir()` fallback, and it clears only when an operator removes the entry or, for a dangling link, its target appears. Retrying cannot change the answer. This restores the pre-PR terminal behavior.
- `dequeue_pending_syncs` now selects and stamps in one statement, ordering by `COALESCE(attempted_at, enqueued_at)` over a new nullable column. Any deferral rotates to the back whatever its cause, which also means a deferral branch added later cannot reintroduce this by forgetting to mark the row.

Each new test was observed failing first: a regular file and a dangling symlink at `repos_dir/<owner>` reaching `failed` (both sat at `pending` before), a healthy row landing behind 10 stuck rows and again behind 25, and an upgrade-path test that simulates an existing node rather than a fresh database. The larger-than-batch case is there because the batch-sized one passes under a fix that only rotates a single full window. The two existing pending-not-failed tests are untouched and still green, as are the containment tests. 558 pass.

One thing I would rather state than have found later: this removes the starvation, not the residual cost. N permanently deferring rows still consume N dequeue slots per rotation and delay a healthy row by up to `ceil(N/10)` polls, which works out to about five minutes at N=100. That last figure is arithmetic from the batch size of 10 and the 30s interval rather than a measurement; the cases I actually ran are N=10 and N=25. Either way it is bounded progress instead of indefinite starvation, and acceptable at current queue sizes, but it is not zero. Bounding retries properly needs the `attempts` column, which stays out of this PR along with the backoff filter that would pair with it.

Two smaller notes on the diff. The migration is version 17 rather than 12 because two open branches already claim into that range and the runner keys the applied set on the version integer alone; the reservation comment above it records which branches and why. And `enqueued_at` is deliberately left alone so backlog age stays measurable, which is why the scheduling key is a separate column rather than a reuse of that one.

jatmn (changes requested):
My prior queue-starvation finding is fixed on this head: `AlreadyExists` at the owner path is terminal again, and `attempted_at` stamping rotates deferred rows so a full batch of stuck items no longer starves healthy repos behind them.

Thanks for the contribution. I do not see any additional actionable issues from my review.

## Prior feedback

- [P2] Deterministic owner-path failures pinning the sync queue — **resolved** on `fbc5868`. `create_dir_all` `AlreadyExists` at `repos_dir/<owner>` is terminal again (`crates/gitlawb-node/src/sync.rs:274-287`), and `dequeue_pending_syncs` stamps `attempted_at` so deferred rows rotate (`crates/gitlawb-node/src/db/mod.rs:1689-1694`). The starvation tests pin that premise before asserting yield.

## Reviewed and rejected

I looked at these again on the current head and am recording them here so they do not get re-opened as new blockers.

- **`path_within_root` / `repos_extra` prefix trap** — not a bug. `Path::starts_with` is component-wise; `/var/repos_extra/...` does not match `/var/repos`. Canonical containment still blocks the symlink cases this PR tests.
- **Gossip / `trigger_sync` lack ingress slug validation** — real asymmetry, but an acknowledged residual. The worker re-validates before any path join or git call; this PR's notify boundary is additive, not incomplete coverage of the stated threat model.
- **`list_peers` error marks the dequeued batch `failed`** — pre-existing on base; not introduced by this diff. Worth a follow-up if you want transient DB blips to defer rather than retire, but not a regression from #272.
- **Empty `http_url` returns `Some("")` from `resolve_origin_url`** — pre-existing inconsistency with `trigger_sync`'s empty-URL skip; not introduced here.
- **`upsert_mirror_repo` error ignored before `mark_sync_done`** — pre-existing; unchanged by this PR.
- **Migration v17 manual reservation / missing dequeue index** — operational and performance nits, not correctness defects on this head. The upgrade-path test covers the happy path; the reservation comment documents the collision risk.

beardthelion:
@jatmn thanks for the re-review. Your latest review on `fbc5868` records the P2 as resolved and no additional actionable issues, but it came in as `request-changes`, so the PR still reads as blocked and CI-green-plus-CHANGES_REQUESTED will hold up the merge.

Could you re-submit as an approval? Nothing has changed on the head since your pass, and the two CodeRabbit nits you already triaged (v17 version reservation, the partial index for the dequeue `ORDER BY`) I'm leaving as-is for the same reason you gave: operational and performance, not correctness on this diff.

jatmn (approved):
Oops, miss click.
LGTM

kevincodex1 (approved):
nice catch!

Issue #268, opened 2026-07-28
Unauthenticated task reads expose agent-task UCAN tokens, payloads, and private-repo IDs on both GraphQL and REST


beardthelion:
The agent-task read surfaces have no authorization gate on either transport, so any unauthenticated caller can enumerate the whole task queue and read each task's `ucan_token`, `payload`, and `repo_id`, including tasks scoped to private repos.

Two surfaces, same root cause:

- GraphQL `tasks` and `task(id:)` (`crates/gitlawb-node/src/graphql/query.rs:108` and `:123`) never call `require_signer`, unlike every mutation in `graphql/mutation.rs`. `AgentTaskType` (`graphql/types.rs:15`) is a plain `SimpleObject` and its `ucan_token: Option<String>` field carries no skip or guard.
- REST `GET /api/v1/tasks` and `GET /api/v1/tasks/{id}` (`api/tasks.rs:130` and `:149`) take no `AuthenticatedDid` extension, and `server.rs:80-82` registers them as `task_read_routes` deliberately outside `add_auth_layers` (the sibling write routes at `server.rs:70-77` do get it). `task_to_json` puts `"ucan_token": t.ucan_token` straight into the body at `api/tasks.rs:83`.

## Reproduction

Seeding one task with a sentinel token and payload and driving `{ tasks { id ucanToken payload delegatorDid assigneeDid } }` with no identity returns:

```json
{"tasks":[{"id":"task-sentinel-1","ucanToken":"SENTINEL-UCAN-DO-NOT-LEAK","payload":"SENTINEL-PAYLOAD-DO-NOT-LEAK","delegatorDid":"did:key:z6MkDelegator","assigneeDid":null}]}
```

`task(id:)` returns the same fields. Driving the identical query three ways, anonymous, signed as an unrelated third-party DID, and signed as the delegator, produces byte-identical responses with an empty `errors` array, so there is no gate to bypass rather than a gate that is too loose.

A second case seeds a private repo and a task scoped to it. An anonymous `{ tasks { repoId kind capability delegatorDid assigneeDid } }` returns:

```json
{"tasks":[{"id":"task-private-1","repoId":"private-r1","kind":"deploy-prod","capability":"repo:write","delegatorDid":"did:key:z6MkDelegator","assigneeDid":"did:key:z6MkStranger","ucanToken":"...","payload":"..."}]}
```

[...]

PR #275, opened 2026-07-29
fix(node): honor pre-assigned task assignee on claim

Ayush7614:
## Summary
- `claim_task` previously updated any `pending` row and overwrote `assignee_did`, so a stranger could steal a task pre-assigned to someone else and receive its `ucan_token` / payload.
- Claim now admits only the reserved assignee (DID-normalized) or first-claimer on open (`NULL`) tasks; the UPDATE re-checks the assignee slot to close the race.
- `finish_task` now binds the stored assignee in SQL so complete/fail cannot win a check-then-act race after the handler gate.

## Why (direct PR)
Real authz gap on the agent-task surface (REST + GraphQL + DB). Not tracked by an existing issue/PR (distinct from #268 anonymous UCAN *reads*).

## Test plan
- [x] `claim_task_honors_preassigned_assignee` (REST) — thief gets 409, no UCAN; reserved assignee claims and receives UCAN
- [x] `claim_task_open_still_first_claimer_wins` — open pending tasks still claimable
- [x] `claim_task_honors_preassigned_assignee` (GraphQL) — same steal/admit behavior
- [x] Existing `complete_task_authorizes_assignee_only` / GraphQL complete tests still pass
- [x] `cargo clippy -p gitlawb-node --all-targets -- -D warnings` clean
- [x] CI `test (stable)` with Postgres

## Summary by CodeRabbit

* **Bug Fixes**
  * Task claims now correctly reject attempts to claim tasks reserved for another assignee.
  * Unauthorized users can no longer complete or fail tasks assigned to someone else.
  * Reserved tasks preserve assignment details, while unassigned or blank-assignee tasks remain claimable.
  * Failed or unauthorized actions no longer expose task access tokens or alter task state.
  * Assignment formatting is handled consistently during authorization and filtering.
  * Task activity events accurately identify the persisted assignee.
  * Task creation responses now reflect the task as stored.

### DIFF
--- a/crates/git-remote-gitlawb/tests/real_git_fetch.rs
+++ b/crates/git-remote-gitlawb/tests/real_git_fetch.rs
@@ -128,19 +128,6 @@ impl Drop for Shim {
     }
 }
 
-/// Wait until the shim accept loop is polling. Without this, Windows CI can
-/// race: `git fetch` connects before the spawned thread enters `accept()`.
-fn wait_for_shim_ready(addr: std::net::SocketAddr) {
-    let deadline = Instant::now() + Duration::from_secs(10);
-    while Instant::now() < deadline {
-        if TcpStream::connect_timeout(&addr, Duration::from_millis(200)).is_ok() {
-            return;
-        }
-        std::thread::sleep(Duration::from_millis(25));
-    }
-    panic!("shim at {addr} did not accept connections within 10s");
-}
-
 /// Start a minimal smart-HTTP shim serving `repo` on 127.0.0.1. Handles the
 /// upload-pack advertisement (GET) and pack negotiation (POST), counting POSTs.
 fn start_shim(repo: PathBuf, mode: ShimMode) -> Shim {
@@ -166,8 +153,6 @@ fn start_shim(repo: PathBuf, mode: ShimMode) -> Shim {
             }
         }
     });
-
-    wait_for_shim_ready(addr);
 
     Shim {
         base_url,
@@ -262,26 +247,6 @@ fn write_response(mut stream: TcpStream, status: &str, content_type: &str, body:
 /// Run `git fetch` in `clone` through the helper, with a hard timeout so a
 /// regression to the deadlock fails fast instead of hanging the suite.
 fn fetch_with_helper(clone: &Path, node_url: &str) -> (bool, std::process::Output) {
-    let mut last = fetch_with_helper_once(clone, node_url);
-    for attempt in 1..3u32 {
-        if last.0 && last.1.status.success() {
-            return last;
-        }
-        let stderr = String::from_utf8_lossy(&last.1.stderr);
-        let transient = stderr.contains("10053")
-            || stderr.contains("connection error")
-            || stderr.contains("connection was aborted")
-            || stderr.contains("error sending request");
-        if !transient {
-            return last;
-        }
-        std::thread::sleep(Duration::from_millis(250 * attempt as u64));
-        last = fetch_with_helper_once(clone, node_url);
-    }
-    last
-}
-
-fn fetch_with_helper_once(clone: &Path, node_url: &str) -> (bool, std::process::Output) {
     let helper_bin = PathBuf::from(env!("CARGO_BIN_EXE_git-remote-gitlawb"));
     let helper_dir = helper_bin.parent().unwrap().to_path_buf();
     let path_env = match std::env::var_os("PATH") {
--- a/crates/git-remote-gitlawb/tests/real_git_fetch.rs
+++ b/crates/git-remote-gitlawb/tests/real_git_fetch.rs
@@ -262,6 +262,26 @@ fn write_response(mut stream: TcpStream, status: &str, content_type: &str, body:
 /// Run `git fetch` in `clone` through the helper, with a hard timeout so a
 /// regression to the deadlock fails fast instead of hanging the suite.
 fn fetch_with_helper(clone: &Path, node_url: &str) -> (bool, std::process::Output) {
+    let mut last = fetch_with_helper_once(clone, node_url);
+    for attempt in 1..3u32 {
+        if last.0 && last.1.status.success() {
+            return last;
+        }
+        let stderr = String::from_utf8_lossy(&last.1.stderr);
+        let transient = stderr.contains("10053")
+            || stderr.contains("connection error")
+            || stderr.contains("connection was aborted")
+            || stderr.contains("error sending request");
+        if !transient {
+            return last;
+        }
+        std::thread::sleep(Duration::from_millis(250 * attempt as u64));
+        last = fetch_with_helper_once(clone, node_url);
+    }
+    last
+}
+
+fn fetch_with_helper_once(clone: &Path, node_url: &str) -> (bool, std::process::Output) {
     let helper_bin = PathBuf::from(env!("CARGO_BIN_EXE_git-remote-gitlawb"));
     let helper_dir = helper_bin.parent().unwrap().to_path_buf();
     let path_env = match std::env::var_os("PATH") {
--- a/crates/git-remote-gitlawb/tests/real_git_fetch.rs
+++ b/crates/git-remote-gitlawb/tests/real_git_fetch.rs
@@ -107,6 +107,11 @@ enum ShimMode {
 struct Shim {
     base_url: String,
     posts: Arc<AtomicUsize>,
+    /// Advertisement (GET /info/refs) count. One `git fetch` performs exactly
+    /// one advertisement, so tests assert this is EXACTLY one: any fetch-level
+    /// retry that comes back re-advertises and goes red on the spot, which the
+    /// lower-bound POST assertions alone cannot see (#275 round 6).
+    gets: Arc<AtomicUsize>,
     stop: Arc<AtomicBool>,
     handle: Option<std::thread::JoinHandle<()>>,
 }
@@ -136,15 +141,17 @@ fn start_shim(repo: PathBuf, mode: ShimMode) -> Shim {
     let addr = listener.local_addr().unwrap();
     let base_url = format!("http://{addr}");
     let posts = Arc::new(AtomicUsize::new(0));
+    let gets = Arc::new(AtomicUsize::new(0));
     let stop = Arc::new(AtomicBool::new(false));
 
     let posts_t = posts.clone();
+    let gets_t = gets.clone();
     let stop_t = stop.clone();
     let handle = std::thread::spawn(move || {
         while !stop_t.load(Ordering::SeqCst) {
             match listener.accept() {
                 Ok((stream, _)) => {
-                    handle_conn(stream, &repo, mode, &posts_t);
+                    handle_conn(stream, &repo, mode, &posts_t, &gets_t);
                 }
                 Err(ref e) if e.kind() == std::io::ErrorKind::WouldBlock => {
                     std::thread::sleep(Duration::from_millis(5));
@@ -157,6 +164,7 @@ fn start_shim(repo: PathBuf, mode: ShimMode) -> Shim {
     Shim {
         base_url,
         posts,
+        gets,
         stop,
         handle: Some(handle),
     }
@@ -182,7 +190,13 @@ fn normalize_accepted_stream(stream: &TcpStream) {
     stream.set_read_timeout(Some(Duration::from_secs(30))).ok();
 }
 
-fn handle_conn(stream: TcpStream, repo: &Path, mode: ShimMode, posts: &AtomicUsize) {
+fn handle_conn(
+    stream: TcpStream,
+    repo: &Path,
+    mode: ShimMode,
+    posts: &AtomicUsize,
+    gets: &AtomicUsize,
+) {
     normalize_accepted_stream(&stream);
     let mut reader = BufReader::new(stream);
 
@@ -219,6 +233,7 @@ fn handle_conn(stream: TcpStream, repo: &Path, mode: ShimMode, posts: &AtomicUsi
     }
 
     let (content_type, payload) = if method == "GET" && target.contains("/info/refs") {
+        gets.fetch_add(1, Ordering::SeqCst);
         // v0 advertisement, wrapped exactly as the node's info_refs does.
         let adv = upload_pack(repo, true, b"");
         let mut wrapped = pkt(b"# service=git-upload-pack\n");
@@ -935,6 +950,7 @@ fn shim_answers_a_connection_that_arrives_non_blocking() {
 
     let server = std::thread::spawn(move || {
         let posts = AtomicUsize::new(0);
+        let gets = AtomicUsize::new(0);
         let stream = loop {
             match listener.accept() {
                 Ok((s, _)) => break s,
@@ -949,7 +965,7 @@ fn shim_answers_a_connection_that_arrives_non_blocking() {
         // the assertions below run on every platform.
         stream.set_nonblocking(true).unwrap();
         accepted_tx.send(()).unwrap();
-        handle_conn(stream, &repo, ShimMode::Normal, &posts);
+        handle_conn(stream, &repo, ShimMode::Normal, &posts, &gets);
     });
 
     let mut client = TcpStream::connect(addr).unwrap();
@@ -1012,6 +1028,17 @@ fn real_git_multi_round_fetch_completes() {
     assert!(
         posts >= 2,
         "fixture did not force multi-round negotiation (observed {posts} POST(s)); the bridging path was not exercised"
+    );
+    // Exactly one advertisement: the POST bound above is a floor, so a
+    // reintroduced fetch-level retry could satisfy it by running a second full
+    // fetch instead of multi-round negotiation. Re-advertising is what a
+    // second fetch cannot avoid, so this is the committed guard on the
+    // retry's absence.
+    let gets = shim.gets.load(Ordering::SeqCst);
+    assert_eq!(
+        gets, 1,
+        "expected exactly one advertisement GET per fetch; {gets} means a \
+         fetch-level retry is back"
     );
 
     // The fetched tip is present and the clone's object graph is intact.
@@ -1103,6 +1130,17 @@ fn real_git_withheld_shaped_first_post() {
     assert!(
         completed,
         "helper hung on a withheld-shaped response (should forward-and-terminate, not deadlock). stderr:\n{stderr}"
+    );
+
+    // Exactly one advertisement, pass or fail: the POST assertions below are
+    // floors, so a reintroduced fetch-level retry could mask the withheld
+    // shape behind a second full fetch. A retry cannot avoid re-advertising,
+    // so this is the committed guard on its absence.
+    let gets = shim.gets.load(Ordering::SeqCst);
+    assert_eq!(
+        gets, 1,
+        "expected exactly one advertisement GET per fetch; {gets} means a \
+         fetch-level retry is back"
     );
 
     if out.status.success() {
--- a/crates/git-remote-gitlawb/tests/real_git_fetch.rs
+++ b/crates/git-remote-gitlawb/tests/real_git_fetch.rs
@@ -128,6 +128,19 @@ impl Drop for Shim {
     }
 }
 
+/// Wait until the shim accept loop is polling. Without this, Windows CI can
+/// race: `git fetch` connects before the spawned thread enters `accept()`.
+fn wait_for_shim_ready(addr: std::net::SocketAddr) {
+    let deadline = Instant::now() + Duration::from_secs(10);
+    while Instant::now() < deadline {
+        if TcpStream::connect_timeout(&addr, Duration::from_millis(200)).is_ok() {
+            return;
+        }
+        std::thread::sleep(Duration::from_millis(25));
+    }
+    panic!("shim at {addr} did not accept connections within 10s");
+}
+
 /// Start a minimal smart-HTTP shim serving `repo` on 127.0.0.1. Handles the
 /// upload-pack advertisement (GET) and pack negotiation (POST), counting POSTs.
 fn start_shim(repo: PathBuf, mode: ShimMode) -> Shim {
@@ -153,6 +166,8 @@ fn start_shim(repo: PathBuf, mode: ShimMode) -> Shim {
             }
         }
     });
+
+    wait_for_shim_ready(addr);
 
     Shim {
         base_url,
--- a/crates/gitlawb-node/src/api/tasks.rs
+++ b/crates/gitlawb-node/src/api/tasks.rs
@@ -117,7 +117,7 @@ pub async fn create_task(
         updated_at: now,
         deadline: body.deadline,
     };
-    state.db.create_task(&task).await.map_err(|e| {
+    let task = state.db.create_task(&task).await.map_err(|e| {
         (
             StatusCode::INTERNAL_SERVER_ERROR,
             Json(json!({ "error": e.to_string() })),
@@ -175,16 +175,28 @@ pub async fn claim_task(
         return Err(forbidden("assignee_did must be the authenticated signer"));
     }
     let task = state.db.claim_task(&id, &auth.0).await.map_err(|e| {
+        if e.downcast_ref::<crate::db::TaskReservedForOtherAssignee>()
+            .is_some()
+        {
+            return forbidden("task not claimable: reserved for another assignee");
+        }
         (
             StatusCode::CONFLICT,
             Json(json!({ "error": e.to_string() })),
         )
     })?;
+    let by_did = task
+        .assignee_did
+        .as_deref()
+        .map(crate::db::trim_assignee_did)
+        .filter(|s| !s.is_empty())
+        .map(str::to_string)
+        .unwrap_or_else(|| auth.0.clone());
     let _ = state.task_event_tx.send(TaskEventBroadcast {
         task_id: id,
         old_status: "pending".to_string(),
         new_status: "claimed".to_string(),
-        by_did: auth.0,
+        by_did,
         at: Utc::now().to_rfc3339(),
     });
     Ok(Json(task_to_json(&task)))
@@ -219,14 +231,14 @@ pub async fn complete_task(
         })?;
     if !crate::api::did_matches(
         &auth.0,
-        existing.assignee_did.as_deref().unwrap_or_default(),
+        crate::db::trim_assignee_did(existing.assignee_did.as_deref().unwrap_or_default()),
     ) {
         return Err(forbidden("only the task assignee can complete it"));
     }
     let by_did = auth.0;
     let task = state
         .db
-        .finish_task(&id, "completed", body.result.as_deref())
+        .finish_task(&id, "completed", body.result.as_deref(), &by_did)
         .await
         .map_err(|e| {
             (
@@ -234,6 +246,13 @@ pub async fn complete_task(
                 Json(json!({ "error": e.to_string() })),
             )
         })?;
+    let by_did = task
+        .assignee_did
+        .as_deref()
+        .map(crate::db::trim_assignee_did)
+        .filter(|s| !s.is_empty())
+        .map(str::to_string)
+        .unwrap_or(by_did);
     let _ = state.task_event_tx.send(TaskEventBroadcast {
         task_id: id,
         old_status: "claimed".to_string(),
@@ -272,7 +291,7 @@ pub async fn fail_task(
         })?;
     if !crate::api::did_matches(
         &auth.0,
-        existing.assignee_did.as_deref().unwrap_or_default(),
+        crate::db::trim_assignee_did(existing.assignee_did.as_deref().unwrap_or_default()),
     ) {
         return Err(forbidden("only the task assignee can fail it"));
     }
@@ -280,7 +299,7 @@ pub async fn fail_task(
     let reason = body.reason.unwrap_or_default();
     let task = state
         .db
-        .finish_task(&id, "failed", Some(&reason))
+        .finish_task(&id, "failed", Some(&reason), &by_did)
         .await
         .map_err(|e| {
             (
@@ -288,6 +307,13 @@ pub async fn fail_task(
                 Json(json!({ "error": e.to_string() })),
             )
         })?;
+    let by_did = task
+        .assignee_did
+        .as_deref()
+        .map(crate::db::trim_assignee_did)
+        .filter(|s| !s.is_empty())
+        .map(str::to_string)
+        .unwrap_or(by_did);
     let _ = state.task_event_tx.send(TaskEventBroadcast {
         task_id: id,
         old_status: "claimed".to_string(),
--- a/crates/gitlawb-node/src/db/mod.rs
+++ b/crates/gitlawb-node/src/db/mod.rs
@@ -2815,28 +2815,54 @@ impl Db {
 
 // ── Agent Tasks ───────────────────────────────────────────────────────────────
 
+/// Permanent authorization denial: the task is reserved for a different assignee.
+/// Handlers downcast this to return 403 (vs 409 for a lost claim race).
+#[derive(Debug, thiserror::Error)]
+#[error("task not claimable: reserved for another assignee")]
+pub struct TaskReservedForOtherAssignee;
+
+/// ASCII whitespace treated as blank for assignee slots — must stay in sync with
+/// `BTRIM(assignee_did, E' \t\n\r')` in `claim_task`'s SQL open-slot check.
+const ASSIGNEE_BLANK: &[char] = &[' ', '\t', '\n', '\r'];
+
+fn assignee_slot_blank(s: &str) -> bool {
+    s.trim_matches(ASSIGNEE_BLANK).is_empty()
+}
+
+/// Trim assignee DIDs for auth matching (same ASCII blank set as SQL/open checks).
+pub fn trim_assignee_did(s: &str) -> &str {
+    s.trim_matches(ASSIGNEE_BLANK)
+}
+
 impl Db {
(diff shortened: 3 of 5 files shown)


beardthelion (dismissed):
The claim gate holds up. I reverted the `did_matches` reservation check and both new tests went RED, so the premise is covered. Findings are about the denial's shape and about what the race half of the change actually proves.

Ignore the `needs-tests` label: the triage job looks for an added `#[test]` or `#[cfg(test)]` line and does not know about `#[sqlx::test]`, which is how this repo writes DB tests. Your three tests are there. That regex is ours to fix, not yours, and #277 does it.

One scope note, since it affects how much your tests can claim: the anonymous read of `ucan_token` through `GET /api/v1/tasks/{id}` and the GraphQL `task` query is tracked in #268 and is not in scope here. Your UCAN assertions bind the claim path, which is the right thing for this PR to do, but the token is readable without claiming anything until #268 lands.

## Findings

- [P2] Return 403 for a reserved-task claim, not 409
  `crates/gitlawb-node/src/api/tasks.rs:177`
  The reservation denial at `db/mod.rs:2624` is an `anyhow` error, and the handler maps every `claim_task` error to `CONFLICT`. Driving a thief through the route gives `409 {"error":"task not claimable: reserved for another assignee"}`, while the signer-binding denial three lines above uses `forbidden()` and returns `403 {"error":"forbidden","message":...}`. Same concept, two statuses and two body shapes, and a permanent authorization denial is now indistinguishable from a lost race, so a retrying client never stops. I verified a fix: a marker error type on the reserved branch plus a `downcast_ref` in the handler gives 403 for the steal and leaves an ordinary already-claimed race at `409 {"error":"task not claimable: not found or already claimed"}`. The assertion at `test_support.rs:504` flips with it; that was the only test that changed.

- [P2] Add a direct test for the `finish_task` assignee gate
  `crates/gitlawb-node/src/db/mod.rs:2670`
  Both handlers already reject a non-assignee with 403 before calling in, so nothing committed reaches the new DB-layer check. I deleted the whole gate (the `did_matches` branch, the `AND assignee_did=$5`, and the bind) and all six task tests stayed green, which leaves the signature change unproven. A direct `db.finish_task(id, "completed", None, stranger)` asserting `Err` and that the row stays `claimed` covers it; I wrote that test and confirmed it goes RED with the gate removed and GREEN with it in place.

- [P2] Keep the stored DID form when claiming a reserved task
  `crates/gitlawb-node/src/db/mod.rs:2635`
  The UPDATE writes `assignee_did=$2`, the caller-presented form, even when the row already held a different-but-matching one. Executed: a reservation stored as `zBAREKEY...` claimed as `did:key:zBAREKEY...` succeeds and rewrites the row to the `did:key:` form. `list_tasks` filters on exact `assignee_did=$1` (`db/mod.rs:2572`, `2589`), so the delegator who reserved the task by the bare key stops finding it afterwards. `SET assignee_did=COALESCE(assignee_did, $2)` fixes it; I ran it and confirmed a reserved task keeps its stored form while an open task still takes the claimer's DID.

- [P3] Treat a blank reservation as open
  `crates/gitlawb-node/src/db/mod.rs:2624`
  `create_task` binds `assignee_did` from the body without validation, so `""` reaches the gate as `Some("")`, enters the reserved branch and matches nobody. Executed: every claimer gets `task not claimable: reserved for another assignee`, permanently. Filtering blank at the gate (`.as_deref().filter(|r| !r.trim().is_empty())`) makes it claimable again; verified.

- [P3] Narrow the pre-check to the columns the decision needs
  `crates/gitlawb-node/src/db/mod.rs:2615`
  `claim_task` now loads the full row, including `payload`, `result` and `ucan_token`, before rejecting a caller who is not the reservation holder. The old single-statement UPDATE rejected without returning anything, so a permissionless caller can replay a denial and make the node read the whole row each time. Bounded by the default body limit rather than unbounded, hence P3, but a `SELECT status, assignee_did` would keep the denial path cheap.

Not an ask, recorded only: the doc comment and the PR body both credit the new assignee clause in the UPDATE with closing the race. `assignee_did` has one writer, `claim_task`'s own UPDATE, which requires `status='pending'`, so the pre-existing status predicate already closes that window. I removed the new clause and all six task tests stayed green. It is reasonable defense-in-depth against a future writer; the wording is what overstates it.

Ayush7614:
@beardthelion addressed your review findings:

1. **[P2] 403 for reserved claim** — `TaskReservedForOtherAssignee` marker; REST handler downcasts to `forbidden()` (403). Ordinary claim races stay 409. Test assertion flipped to `FORBIDDEN`.
2. **[P2] `finish_task` DB gate test** — `finish_task_rejects_non_assignee_at_db` (stranger Err + row stays claimed; assignee completes).
3. **[P2] Keep stored DID form** — `COALESCE(NULLIF(BTRIM(assignee_did), ''), $2)` on claim; `claim_task_keeps_stored_assignee_did_form` covers bare→full claim.
4. **[P3] Blank reservation = open** — blank/whitespace treated as open; `claim_task_blank_reservation_is_open`.
5. **[P3] Narrow pre-check** — `SELECT status, assignee_did` only before the gate.

Also softened the race wording to defense-in-depth (status predicate remains the primary window closer). Ready for another look.

jatmn (changes requested):
## Findings

- [P2] Required PR checks did not run on the new head
  GitHub Actions run [30586459176](https://github.com/Gitlawb/node/actions/runs/30586459176) for `cdbee5a0` completed with `action_required` and zero jobs (`test (stable)`, `fmt + clippy`, `build --release`, etc. are absent from `statusCheckRollup`). Only triage and CodeRabbit report success. Please have a maintainer approve/trigger the fork workflow and confirm green checks on this head before merge.

- [P3] Align Rust and SQL whitespace handling for blank/open reservations
  `crates/gitlawb-node/src/db/mod.rs:2638`, `2651-2652`
  The `""` case is fixed, but Rust `trim().is_empty()` and PostgreSQL `BTRIM()` still disagree on tab/newline-only `assignee_did` values (`"\t"`, `"\n"`). Rust treats them as open (`reserved_exact = None`); SQL still sees an occupied slot, so claim always returns 409 and the row stays `pending`. Apply the same blank definition in Rust and SQL (or normalize whitespace-only values to `NULL` on create). Also trim consistently in `finish_task` `did_matches` (`db/mod.rs:2690`) to match the claim gate (`2640`) so tab-padded stored values cannot claim then fail complete/fail.

- [P3] Emit `TaskEventBroadcast.by_did` from the persisted assignee
  `crates/gitlawb-node/src/api/tasks.rs:188`, `242`, `296`; `crates/gitlawb-node/src/graphql/mutation.rs:80`, `122`, `164`
  The COALESCE fix keeps bare-key `assignee_did` in the row while broadcasts still send the signer's full `did:key:…` form. That skew is new with this PR (previously claim overwrote to caller form). Subscribers or filters doing exact-string correlation with `list_tasks` will miss events. Please set `by_did` from the returned task's `assignee_did` on claim, complete, and fail.

- [P3] Match GraphQL steal test parity with REST
  `crates/gitlawb-node/src/graphql/mutation.rs:326`
  REST reloads the row after a failed stranger claim and asserts `status == "pending"` and unchanged `assignee_did` (`test_support.rs:302-304`). The GraphQL test only checks error strings and UCAN absence. Please add the same `get_task` assertions.

Ayush7614:
@jatmn @beardthelion addressed the latest review on head `139b8da`:

1. **Whitespace Rust vs SQL** — blank assignee slots now use the same definition in Rust (`trim_matches` on space/tab/LF/CR) and SQL (`BTRIM(..., E' \t\n\r')`). Whitespace-only values are normalized to `NULL` on `create_task`. `finish_task` / GraphQL complete+fail trim the same way so tab-padded stored DIDs can finish after a successful claim.
2. **`TaskEventBroadcast.by_did`** — claim/complete/fail (REST + GraphQL) now emit the **persisted** `task.assignee_did` (fallback to signer only if missing).
3. **GraphQL steal parity** — after a failed stranger `claimTask`, assert via `get_task` that status stays `pending` and `assignee_did` is unchanged.

**CI (P2):** fork PR Checks still need a maintainer to approve workflows on the latest head — please approve when you can so the suite can run against this commit.

jatmn (changes requested):

## Findings

- [P1] Run and satisfy the required checks on this head
  `.github/workflows/pr-checks.yml` / current head `139b8da`
  The only check runs attached to this head are the two triage jobs and CodeRabbit; the Rust test, fmt/clippy, release-build, audit, MSRV, and Docker checks have not run. The fork workflow needs maintainer approval/triggering before merge. In addition, `cargo fmt --all -- --check` currently fails on this submitted head (including the new task changes), so the format job will fail once it runs. Please format the patch and get the full required workflow green on this head.

- [P2] Preserve the exact nonblank reservation when claiming
  `crates/gitlawb-node/src/db/mod.rs:2664`
  The authorization and race predicate bind `$4` to the exact stored `assignee_did`, but the update stores `BTRIM(assignee_did, ...)` instead. A reservation such as `" did:key:z...\t"` is accepted for its matching signer and then silently rewritten without the surrounding whitespace. That contradicts the new stored-form preservation contract and makes exact `list_tasks(..., assignee_did=<original>)` filters lose the task after claim. Use the exact reserved value for the reserved branch, and apply the blank normalization only when filling an open slot; add a padded-reservation regression test.

- [P2] Return the normalized assignee value from task creation
  `crates/gitlawb-node/src/db/mod.rs:2543`, `crates/gitlawb-node/src/api/tasks.rs:126`, `crates/gitlawb-node/src/graphql/mutation.rs:58`
  `create_task` now persists an all-whitespace `assignee_did` as SQL `NULL`, but both REST and GraphQL serialize the original in-memory `AgentTask`. Consequently, a create response for `assignee_did: " \t"` says the task is reserved, while an immediate GET/list reports it as open and another agent can claim it. Normalize the returned task (or return the persisted row) in both paths and cover both response contracts.

beardthelion (dismissed):
The claim gate still holds on `139b8dac`. Reverting the reserved `did_matches` deny still reds both steal tests. jatmn's two code points are true on this head: the claim SET re-derives through `BTRIM` instead of keeping the reserved form, and create returns the input assignee after the DB has nulled a blank. `cargo fmt --all -- --check` fails too, and every hunk is in files this PR touches.

One scope note carried from last round: the anonymous read of `ucan_token` through `GET /api/v1/tasks/{id}` and the GraphQL `task` query is still open (#268). The new "UCAN must not leak on a failed steal" assertions only inspect the 403 envelope, which never carried task fields, so they pass while the token stays readable without claiming. Keep the claim-path assertions; the leak framing still overclaims.

## Findings

- [P2] Keep the exact nonblank reservation on claim, and trim REST finish if you do
  `crates/gitlawb-node/src/db/mod.rs:2664`
  Same ask as jatmn and CodeRabbit. The SET is still `COALESCE(NULLIF(BTRIM(assignee_did, E' \t\n\r'), ''), $2)`, so a padded reservation is rewritten on claim and an exact list filter loses it. The docstring at `:2630` still says the reserved form is kept via `COALESCE`. Do not land `COALESCE($4, $2)` alone: REST complete/fail still compare untrimmed while GraphQL trims, and that couple creates a cross-surface 403 for the real assignee. Related: reserve bare, claim as `did:key:<same>`, then list by the claimer's full DID returns 0 rows while the stored bare form returns 1, because `list_tasks` (`:2582`) is plain equality while claim went through `did_matches`. Either keep `BTRIM` and fix the docstring, or keep the stored form and trim REST on the same ASCII blank set. Add a padded-reservation regression, and a bare-then-full list regression, either way.

- [P2] Return the normalized assignee from create
  `crates/gitlawb-node/src/api/tasks.rs:120`
  Same as jatmn. `Db::create_task` nulls a whitespace-only assignee; REST still returns `task_to_json(&task)` and GraphQL `create_task` (`mutation.rs:55`) still returns `AgentTaskType::from(task)`, so create can show `""` while the row is NULL/open. Normalize or re-read before serializing; cover both surfaces.

- [P2] Pin the named guards with fail-on-remove proofs
  `crates/gitlawb-node/src/db/mod.rs:2633`
  The claim UPDATE assignee re-check, `create_task`'s blank filter, and `finish_task`'s `AND assignee_did=$5` can each be removed with the task suite staying green (the finish case because the Rust gate returns first). Related: `claim_task_blank_reservation_is_open` seeds `Some("")` through `create_task`, which now stores NULL, so it never reaches either blank branch it is named for. Seed the row directly, and add proofs that fail when each of those three guards is removed.

- [P2] Format the touched files
  `crates/gitlawb-node/src/api/tasks.rs`
  `cargo fmt --all -- --check` exits 1 on this head. The six diffs are in `api/tasks.rs`, `db/mod.rs`, and `graphql/mutation.rs`, so this is yours to clear before the next push.

- [P3] Surface the reserved deny distinctly in GraphQL `claimTask`
  `crates/gitlawb-node/src/graphql/mutation.rs:77`
  REST now returns 403 for a reserved steal and 409 for a lost race. `claimTask` still maps every db error through `e.to_string()` with no downcast, so a GraphQL client cannot tell those apart without parsing the message. That is also why the new test has to assert on `contains("reserved") || contains("not claimable")`.

One process note, not a finding: `PR Checks` is `action_required` with zero jobs on this fork head. That is mine to approve, not yours to clear. The fmt finding above is the part you can fix on your own.

Ayush7614:
@beardthelion @jatmn addressed [review 4831516189](https://github.com/Gitlawb/node/pull/275#pullrequestreview-4831516189) on head `62172cf`:

1. **Keep exact nonblank reservation** — claim SET is now `COALESCE($4, $2)` (no BTRIM rewrite). Padded reservation regression + bare→full claim with list-by-bare / list-by-full coverage.
2. **Trim REST finish** — complete/fail use `trim_assignee_did` (same ASCII blank set as claim/SQL). Padded claim can finish.
3. **Normalize create responses** — `create_task` returns the persisted shape (whitespace-only → `None`); REST + GraphQL serialize that.
4. **Fail-on-remove proofs** — blank/tab blank seeded via raw INSERT (claim SQL open path); `create_task_normalizes_whitespace_assignee`; `finish_task_sql_assignee_predicate_is_load_bearing` for `AND assignee_did=$5`.
5. **`cargo fmt`** — touched files formatted.
6. **GraphQL reserved deny** — `claimTask` downcasts `TaskReservedForOtherAssignee` to a distinct `"reserved for another assignee"` message (test asserts `contains("reserved")`).

UCAN anonymous-read framing note acknowledged — still #268; claim-path assertions kept.

**CI:** please approve fork PR Checks on this head when you can.

beardthelion (dismissed):
The claim gate still holds on `62172cf`. Gutting the reserved `did_matches` deny reds both steal tests again (REST stranger lands 200). The r2 behavioral asks landed: claim SET is `COALESCE($4, $2)`, padded reservations survive and can finish after `trim_assignee_did`, create returns the normalized row (REST create with `" \t\n "` returns `assignee_did: null`), GraphQL surfaces the reserved deny distinctly, and `cargo fmt --check` is clean. `create_task_normalizes_whitespace_assignee` goes RED when the blank filter is removed, and the padded test goes RED when SET is reverted to BTRIM.

Two named guards from the r2 pin-guards ask are still unpinned. Replacing the claim UPDATE assignee slot re-check with `AND (true)` leaves all ten filtered task tests green. Removing `AND assignee_did=$5` from `Db::finish_task` leaves both finish tests green: the Rust gate returns first, and `finish_task_sql_assignee_predicate_is_load_bearing` runs a parallel UPDATE that never calls production.

One scope note carried again: anonymous `GET /api/v1/tasks/{id}` and the GraphQL `task` query still serve `ucan_token` (#268). Keep the claim-path assertions; the deny-envelope "must not leak" framing still overclaims.

## Findings

- [P2] Pin the claim UPDATE assignee re-check with a fail-on-remove proof
  `crates/gitlawb-node/src/db/mod.rs:2673`
  The docstring still calls the `AND (($4 IS NULL AND open) OR assignee_did = $4)` clause defense-in-depth against a writer racing the pre-check. I deleted that clause (left `AND (true)`) and the claim/finish/create task filter suite stayed 10/10 green, including both steal tests (the Rust reserved deny fires first). Add a proof that fails when that production predicate is removed.

- [P2] Pin `Db::finish_task`'s `AND assignee_did=$5`, not a parallel query
  `crates/gitlawb-node/src/test_support.rs:807`
  `finish_task_sql_assignee_predicate_is_load_bearing` runs its own `UPDATE ... AND assignee_did=$4` with a mismatched bind and asserts zero rows. That stays green after I remove `AND assignee_did=$5` from `db/mod.rs:2720`, because the test never calls `Db::finish_task` and the stranger path is stopped by the Rust `did_matches` gate at `:2713`. Replace it with a proof that fails when the production clause is gone.

Not an ask, recorded only: collision band is EXPECT-REBASE (mechanical overlap on `db/mod.rs`); expect a rebase, not a redone review of the claim gate.

jatmn (changes requested):

## Findings

- [P2] Pin the claim UPDATE assignee re-check with a fail-on-remove proof
  `crates/gitlawb-node/src/db/mod.rs:2673`
  The re-check is still documented as defense-in-depth against a writer changing the reservation after the pre-check, but no current test reaches that production predicate. Replacing the whole condition with `AND true` leaves the task filter suite green because the existing steal tests stop at the earlier Rust denial. Add a production-path race (or focused equivalent) that changes the pending row's assignee slot after the pre-read and asserts that this claim cannot update it.

- [P2] Make the claimed fail-on-remove test exercise `finish_task`
  `crates/gitlawb-node/src/test_support.rs:807`
  This test still executes a second hand-written `UPDATE` rather than `Db::finish_task`, so it remains green if `AND assignee_did=$5` is removed from the production update. The other new DB test also stops at the Rust `did_matches` check, leaving the check-then-act protection untested. Please test the production path with an assignee change between its read and update (for example through a controlled concurrent update/trigger) and assert that it leaves the task claimed; otherwise the new race guard can be deleted without a regression signal.

Ayush7614:
@beardthelion @jatmn — rebased onto current `main` and squashed to a single commit (`3c126b7`). All 13 `PR Checks` jobs are green on this head.

**R3 pin-guards (your latest round on `62172cf`)**

1. **Claim UPDATE assignee re-check** — replaced the hand-written parallel `UPDATE` with `claim_task_update_assignee_slot_recheck_is_load_bearing`: a second connection holds `FOR UPDATE` on the pending row, changes `assignee_did` after the pre-read, then the production `Db::claim_task` must return `TaskNotClaimable` (0 rows updated). Removing the `AND (($4 IS NULL AND open) OR assignee_did = $4)` predicate should make this go red.

2. **`finish_task` `AND assignee_did=$5`** — replaced `finish_task_sql_assignee_predicate_is_load_bearing` (which never called production) with the same `FOR UPDATE` race pattern against `Db::finish_task`: hog swaps the stored assignee between read and update; production finish must leave the task `claimed`. Removing `AND assignee_did=$5` should make this go red.

**Prior rounds (still on this head)**

- Claim SET is `COALESCE($4, $2)` (exact reserved form preserved); padded-reservation + bare/full list regressions retained.
- `create_task` returns the normalized persisted row on both REST and GraphQL.
- GraphQL `claimTask` surfaces the reserved deny distinctly (not a generic db string).
- `cargo fmt --check` clean.

**Scope note (unchanged):** anonymous `GET /api/v1/tasks/{id}` / GraphQL `task` still serve `ucan_token` (#268); claim-path deny-envelope assertions kept as-is.

Ready for another look.

jatmn (changes requested):

## Findings

- [P1] Rebase onto current `main` and resolve the conflict
  `crates/gitlawb-node/src/db/mod.rs:7035`
  GitHub currently reports this head as `CONFLICTING` / `DIRTY`. The three-way merge conflicts in the peer-writer guard ledger, so this cannot merge as submitted. Please rebase, resolve the conflict while preserving the current-main security changes, and rerun the checks; the resolved head needs a fresh review.

beardthelion (dismissed):
Both round-3 asks are closed, and I checked by execution rather than by reading the replies. Deleting the claim UPDATE's assignee-slot predicate now reds `claim_task_update_assignee_slot_recheck_is_load_bearing`; neutering `finish_task`'s `AND assignee_did=$5` reds `finish_task_sql_assignee_predicate_is_load_bearing`. Last round both of those deletions left the suite green. I re-ran the full named-guard set on this head, six mutations, each red on its own assertion message: the reserved deny on REST and GraphQL, those two predicates, `create_task`'s blank-to-NULL filter, and the `COALESCE($4, $2)` exact-form SET. Baseline is 21 passed.

The blocker is the conflict jatmn already flagged, and it resolves more easily than it looks. Both peer test functions already existed at your base while the ledger there omitted them, so your addition repaired a guard that was failing at that base; main fixed the same thing in `af821e09` with slightly different wording. It is a duplicate add, so take main's version of the doc table and the LEDGER rows and keep the rest of your diff intact. None of your task changes conflict. #247 landed on main today and touched only the error and events modules, so the rebase has no gate to lose on your surfaces, and your GraphQL path already picked up `graphql_db_err` correctly.

Three small things to fold into that push.

## Findings

- [P3] Trim the assignee before putting it in the task event
  `crates/gitlawb-node/src/api/tasks.rs:188`
  I drove the REST claim handler with a reservation stored as `" zKEY\t"` and read the broadcast: `by_did` came out as `" zKEY\t"` while the authenticated caller was `did:key:zKEY`. A field named `by_did` should not carry padding or a bare key. `trim_assignee_did` is already in scope. Same one-liner at the other five sites.

- [P3] Make the claim race assertion provable only through the SQL re-check
  `crates/gitlawb-node/src/test_support.rs:1159`
  The test asserts `contains("not claimable")` and both denials say exactly that: the Rust pre-check is `"task not claimable: reserved for another assignee"`, the SQL miss is `"task not claimable: not found or already claimed"`. If the background writer commits before the pre-read, the test passes through the pre-check and proves nothing about the predicate it is named for. It does exercise that predicate today, since removing it goes red, but the assertion should be what guarantees it. Assert the specific message, or that the error is not a `TaskReservedForOtherAssignee`. The finish test does not have this problem: `"not in claimed state"` cannot match its Rust gate's message.

- [P3] Cover the other blank characters
  `crates/gitlawb-node/src/test_support.rs:830`
  `ASSIGNEE_BLANK` and the SQL `BTRIM(assignee_did, E' \t\n\r')` are the same four characters today, and a comment says they must stay that way. The SQL blank branch is only ever seeded with `''` and a tab, so space, LF and CR never reach it. Extending the tab test to the remaining three is the cheapest guard against those two definitions drifting.

Nothing else holds this up. Once the rebase is clean I will re-run the guard set on the resolved head.

Ayush7614:
@beardthelion @jatmn — rebased onto current `main` (`241b366`, includes merged #247) and pushed `1676eb0`.

**Conflict:** took main's peer-writer LEDGER wording (duplicate of our prior add from #248 fixtures).

**Your latest P3 round:**
1. **`by_did` trim** — claim/complete/fail broadcasts now use `trim_assignee_did` on the persisted assignee (REST + GraphQL, all six sites).
2. **Claim race assertion** — `claim_task_update_assignee_slot_recheck_is_load_bearing` now requires `"not found or already claimed"` and asserts the error is *not* `TaskReservedForOtherAssignee` (proves the SQL UPDATE predicate, not the Rust pre-check).
3. **ASCII blank coverage** — merged blank/tab tests into `claim_task_ascii_blank_reservations_are_open` covering `ASSIGNEE_BLANK` (`""`, space, tab, LF, CR).

Prior pin-guards and behavioral fixes remain on this head. Ready for re-review when CI is green.

Ayush7614:
Pushed `1e6b065` to fix the red **test (windows, non-blocking)** job.

The failure was a race on Windows CI: `real_git_multi_round_fetch_completes` and `real_git_withheld_shaped_first_post` connected before the in-test HTTP shim entered its accept loop (`connection aborted`, os error 10053, 0 POSTs). Added a readiness probe after `start_shim` spawns.

All other required checks were already green on `1676eb0`; re-running the full suite on this head.

jatmn (changes requested):

## Findings

- [P2] Do not retry the stateful withheld-first-POST scenario against the same shim
  `crates/git-remote-gitlawb/tests/real_git_fetch.rs:264`
  `fetch_with_helper` now retries every matching connection error, but `real_git_withheld_shaped_first_post` reuses one `ShimMode::WithheldFirstPost` instance. If the first fetch reaches its first POST and then fails with one of those transient diagnostics, that POST consumes the shim's one special NAK+full-pack response. The retry is served ordinary `upload-pack`; if it succeeds, the test takes its success branch and falsely certifies the withheld-first-POST behavior without a successful fetch ever exercising that behavior. Keep retries out of this stateful case, or recreate/reset the shim and fixture for each attempt.

- [P2] Synchronize with the accept-loop thread instead of the listener backlog
  `crates/git-remote-gitlawb/tests/real_git_fetch.rs:131`
  `TcpListener::bind` has already put the socket in the listening state before the accept-loop thread is spawned, so `TcpStream::connect_timeout` succeeds once the OS queues the TCP handshake; it does not require the spawned thread to have reached `accept()`. `wait_for_shim_ready` can therefore return while the scheduling race described in its comment remains, leaving the Windows flake unsynchronized. Have the spawned thread send an explicit readiness acknowledgement (or complete a protocol-level probe) before `start_shim` returns.

beardthelion (dismissed):
The task work is done, and I checked it by execution on this head rather than by reading the replies. All three of last round's asks landed: `trim_assignee_did` is applied at the `by_did` sites, the race test now asserts the error is not a `TaskReservedForOtherAssignee` so the SQL predicate is what denied it, and the blank test seeds empty, space, tab, LF and CR. I re-ran the guard set on `0484740`, one gut at a time. Removing the reserved deny in `claim_task` reds both steal tests; deleting the claim UPDATE's assignee-slot predicate reds the race guard; neutering `finish_task`'s `AND assignee_did = $5` reds its twin; dropping `create_task`'s blank-to-NULL filter reds the normalization test; rewriting the `COALESCE($4, $2)` SET to a BTRIM reds the padded-reservation test. Baseline is 22 passed. The rebase came out the way I asked and the duplicate peer-writer ledger rows are gone.

What holds this up is the Windows CI work, and the answer is to take it out rather than patch it. That covers both of jatmn's asks on this head. Separately, the open CodeRabbit thread on `.cursor/rules/rtk-token-savings.mdc` points at a file that is no longer in the diff, so you can resolve it.

## Findings

- [P2] Revert `real_git_fetch.rs` to main and let #312 fix the Windows lane
  `crates/git-remote-gitlawb/tests/real_git_fetch.rs:262`
  The retry keys on `10053`, and that is the signature of a real defect rather than a flake: on Windows the accepted socket inherits the listener's non-blocking state, so `read_line` returns `WouldBlock`, the shim drops the connection without writing a response, and the request never reaches the branch that counts POSTs. #312 fixes that at the source and is approved and mergeable, so this retries a live bug into silence. It also lets the withheld test certify a shape it never sent. `posts` is shim-global and the withheld response fires once, so I set that arm to never fire, which is exactly the state a retry leaves behind, and the test passed through its success branch with no note printed. `wait_for_shim_ready` does not close the gap either, since a listener that is bound completes a connect with `accept()` never called. The same three hunks are byte-identical in #276. Your base blob for this file matches main, so the revert is clean, and the Windows job is non-blocking, so nothing here goes red.

Ayush7614:
Re: the last review round — the Windows CI workaround is reverted on this head (`6813340`).

Both workaround commits (`1e6b065` wait-for-shim, `0484740` retry-on-10053) are taken back out of `crates/git-remote-gitlawb/tests/real_git_fetch.rs`, restoring the file to the base blob that matches main. #312 fixed the root cause (`normalize_accepted_stream`) at the source and is merged, so the Windows lane is now carried by the real fix rather than retried into silence.

Verified on this head:
- `cargo test -p git-remote-gitlawb --test real_git_fetch` — 6 passed, 4 fixture-ignored
- `cargo fmt --all -- --check` — clean
- `cargo clippy --all-targets -- -D warnings` (both crates) — clean
- Task/claim suite against Postgres — 22 passed, incl. the guarded proofs (`claim_task_update_assignee_slot_recheck_is_load_bearing`, `finish_task_sql_assignee_predicate_is_load_bearing`, padded-reservation, ascii-blank)

The remaining `gitlawb-node` failures in a full-suite run are pre-existing timing/deadline flakes (`run_bounded_git`, store, visibility_pack) — confirmed identical at the base commit.

@beardthelion @jatmn — the round-4 asks are closed; requesting re-review on the new head. CodeRabbit thread is on a file no longer in the diff (GitHub rejects API-resolving it), but it is non-blocking as you noted.

beardthelion (dismissed):
Last round's asks landed, and I checked them by execution on this head rather than by reading the replies. The retry is gone (`grep -c retry` returns 0), the readiness probe is gone, and the withheld test's outcome guards are intact. All 12 checks are green, and I confirmed the Windows result at the step level rather than the job level, since that lane runs `continue-on-error: true` and would report SUCCESS over a failed step. Step 5 genuinely passed.

Two asks before this merges.

## Findings

- [P2] Rebase onto current `main` so the revert's premise holds at this head
  `crates/git-remote-gitlawb/tests/real_git_fetch.rs`
  The commit message credits #312 with carrying the Windows lane, but this head does not contain #312. The file here is blob `26d88f27`, byte-identical to the pre-#312 base: `normalize_accepted_stream` appears 0 times against 10 on main, and the `stream.set_nonblocking(false)` that main carries at line 181 is absent. `git merge-base --is-ancestor 10e0840d 68133409` returns NO, and the base is 18 commits behind. The shipped state is fine, since the three-way merge resolves this file to `72fdd5c2`, main's exact blob. The head is what worries me: its Windows run passed on the unmitigated shape, so the green tells us the race did not fire on that run, not that the fix is present. Rebase or cherry-pick #312 and let the lane run against the code it is credited to.

- [P3] Pin the retry's absence with a committed guard
  `crates/git-remote-gitlawb/tests/real_git_fetch.rs:249`
  Nothing goes red if a fetch-level retry comes back. The shim counts POSTs only, incrementing at line 209 on the POST branch, and both assertions are lower bounds: `posts >= 2` at 778 and `posts >= 1` at 892. A second full fetch satisfies both while masking the withheld shape the test exists to observe. A GET counter asserted to be exactly one per test would make a reintroduced retry fail on the spot.

One housekeeping item, not an ask: the CodeRabbit thread on `.cursor/rules/rtk-token-savings.mdc` is still open, but that file is absent at this head and never appeared in the PR's net diff, so the work behind it is done and the thread can be resolved.

euxaristia:
@coderabbitai full review

Ayush7614:
@beardthelion @jatmn — requesting re-review on the current head `8fa81f4`.

**What changed since your 8/10 and 8/15 reviews:**

1. **[P2] Rebase onto current `main`** — addressed. `ef167bb` merged `origin/main` into the branch; `normalize_accepted_stream` and `stream.set_nonblocking(false)` from #312 are present on this head. `git merge-base --is-ancestor 10e0840d 8fa81f47` returns YES now. Windows lane runs against the credited code.

2. **[P3] Pin retry's absence with an exact GET count** — addressed. `8fa81f4` (`test(git-remote-gitlawb): pin the fetch retry's absence with an exact GET count`) counts `GET /info/refs` in the shim and asserts exactly one per fetch in both the multi-round and withheld tests. A reintroduced fetch-level retry must re-advertise, so the count trips the guard.

3. **CodeRabbit thread on `.cursor/rules/rtk-token-savings.mdc`** — noted as housekeeping. File is absent at this head (not in net diff), thread can be resolved.

4. **jatmn's P2 (withheld shim reuse / accept-loop sync)** — `fetch_with_helper` retries were removed and the shim no longer accepts retried connections against the same stateful instance; the readiness probe is gone. `grep -c retry crates/git-remote-gitlawb/tests/real_git_fetch.rs` returns 0.

All 13 `PR Checks` jobs green on `8fa81f4` (incl. test windows). The task-claim subject itself (`1676eb0`) is unchanged across these review rounds. Ready for re-review.

jatmn (approved):
@beardthelion lgtm

beardthelion (approved):
Round 7 on `8fa81f47`: approve. Both asks from my 8/15 round landed and I checked them by execution rather than by reading the replies. `10e0840d` is now an ancestor of this head and `normalize_accepted_stream` appears 10 times against 0 before, so the Windows lane ran against the code the revert was credited to. The PR's own premise still holds on the merged head: replacing the reserved-assignee `did_matches` deny in `Db::claim_task` with `if false` reds both `claim_task_honors_preassigned_assignee` tests on the stranger-steal assertion, and dropping the SQL slot predicate to `OR FALSE` reds them as well. CI is 12/12 here.

Two P3s on the new guard. Neither blocks the merge.

## Findings

- [P3] Move the GET assert below the pre-POST guard
  `crates/git-remote-gitlawb/tests/real_git_fetch.rs:1139`
  The `gets == 1` assert runs before the `if out.status.success()` split, so it fires ahead of the `posts >= 1` guard at :1164. Pointing the fetch at a dead port to force a pre-POST failure fails the test at :1140 with "0 means a fetch-level retry is back", which is wrong for 0 and hides the guard whose message names the actual cause. Ordering it after the `posts >= 1` check keeps the diagnostic honest.

- [P3] The guard does not red on the retry it was written to pin
  `crates/git-remote-gitlawb/tests/real_git_fetch.rs:1139`
  I cherry-picked `04847406` back in and the suite stayed green, 8 passed. That retry only fires when stderr matches a transient marker (`10053`, `connection error`, `connection was aborted`, `error sending request`), and the withheld failure is `fatal: git fetch-pack: expected ACK/NAK`, so it returns after one attempt and `gets` stays 1. Forcing the transient check to `true` does red it at `left: 2, right: 1`, so the counter itself works; it never sees the shape that was actually removed. A counter on fetch invocations rather than on advertisements would pin what the commit message claims.

Housekeeping: the CodeRabbit thread on `.cursor/rules/rtk-token-savings.mdc` is still open and that file is absent at this head, so it can be resolved.

beardthelion:
@kevincodex1 LGTM

Issue #135, opened 2026-06-30
GET /ipfs/{cid} serves tree/commit objects of withheld subtrees, leaking structure get_tree protects (KTD3)


beardthelion:
## Summary

`GET /ipfs/{cid}` (`get_by_cid`) serves tree, commit, and tag objects unconditionally under path-scoped visibility rules, disclosing the structure of a withheld subtree (every child filename and child object id) to a caller who is denied that subtree. Blob content stays protected; this is a metadata/structure leak, not a content leak.

This is the KTD3 behavior that #126 explicitly scoped out, surfaced during review of #133 (which correctly fixes the blob-content half of the same endpoint). Filing so the divergence is tracked rather than buried in a PR thread.

## Where

`crates/gitlawb-node/src/api/ipfs.rs` (the per-object gate):

```rust
let path_scoped = has_path_scoped_rule(rules);
if path_scoped && obj_type == "blob" {
    // ... allowed-set membership check; non-blobs skip it
}
```

Under a path-scoped rule the allowed-set gate applies only to `obj_type == "blob"`. A `tree`/`commit`/`tag` object falls through and is served. A git tree object's raw body is `<mode> <filename>\0<raw-oid>` per entry, so fetching the tree CID of a withheld subtree returns every child filename and child oid in cleartext, recursively. The leaf blobs themselves still require allowed-set membership, so their content is protected.

## Why it matters

This is the same confidentiality boundary that `get_tree` (`crates/gitlawb-node/src/api/repos.rs:444`, the N3 fix) deliberately defends: it gates on the *requested* subtree path precisely so "a caller denied a withheld subtree can still enumerate its names/SHAs" cannot happen. The two read surfaces over the same git data disagree: the REST tree-read path protects subtree names/oids, the IPFS CID path does not.

Bounded by the attacker needing the tree object's sha256 CID to request it. That is not free, but CID-enumeration surfaces exist (see #121), so it should not be relied on as the only barrier.

## Scope / not a regression

Pre-existing. The pre-#133 deny-set (`withheld_blob_oids`) also enumerated only blob oids, so trees were never withheld there either. #133 preserves the behavior (and the merged test at `test_support.rs` asserts "tree object is served to anon (KTD3)"). This issue is about closing the divergence, not about #133.

## Suggested direction

[...]

PR #276, opened 2026-07-29
fix(node): close quarantine bypass on encrypted blobs and visibility routes

Ayush7614:
## Summary
- Encrypted-blob handlers (`list` / `get` / `replicate`) used `visibility_check` alone after a bare `get_repo`, skipping the quarantine short-circuit in `authorize_repo_read`. A **public-but-quarantined** mirror still exposed its encrypted blob index (and would serve envelopes).
- Visibility handlers (`set_visibility` / `remove_visibility` / `list_visibility` / `withheld_paths`) did the same bare `get_repo` (+ `visibility_check` or `require_owner`) and were reachable while quarantined — an anon caller learned a public-but-quarantined repo's existence and private-subtree layout, and a caller matching the row's `owner_did` learned every rule.
- Fix: all seven routes go through `authorize_repo_read` (`crates/gitlawb-node/src/api/encrypted.rs`, `crates/gitlawb-node/src/api/visibility.rs`) so a quarantined repo is an opaque `404` there too, checked before any owner/visibility branch. The completeness fence in `crates/gitlawb-node/src/api/mod.rs` now pins both `authorize_repo_read(` and `require_owner(` per visibility handler. CID serving (`list_all_repos` / `api/ipfs.rs`) is already handled on `main`'s paged scan via per-row quarantine flags, so this head carries no DB change.

## Why (direct PR)
Real quarantine invariant break on two surfaces. Not covered by existing issues/PRs (#110/#126 are visibility/CID gating; #135 is withheld-subtree trees — different). CID surface now out of scope for this head; title/body narrowed per review.

## Test plan
- [x] `encrypted_blobs_quarantined_repo_opaque_404` — anon + owner (full `did:key` and bare) 404 on `encrypted-blobs` / `replicate` / `encrypted-blob/{oid}` while quarantined, no CID/oid leak; released repo lists again
- [x] `encrypted_blobs_quarantined_mirror_admission_opaque_404` — same for a `upsert_mirror_repo` slash-form row
- [x] `withheld_paths_and_list_visibility_quarantine_opaque` — anon and both owner forms 404 on `withheld-paths` and `list_visibility` plus `set`/`remove` while quarantined; released serves globs again
- [x] Existing quarantine suite green locally (`cargo test -p gitlawb-node quarantine`)
- [x] Clippy `-D warnings` clean on this head; `fmt --check` clean
- [ ] CI `PR Checks` (`fmt+clippy`, `test (stable)` with Postgres) on this exact head — fork PR awaiting approval to run (previous green run on `bcc7fb4`)

## Notes on review follow-up
- `withheld-paths` now correctly returns opaque `404` via `authorize_repo_read` (same as encrypted discovery). As noted by reviewers, `crates/gl/src/clone.rs:265` treats any `404` on `fetch_withheld` as "old node, no endpoint" and falls back to empty globs before the git advertisement 404s — that is a separate capability-discovery fix (version/capabilities negotiation) to be tracked outside this node PR, not a reason to change the repo-scoped $404.

### DIFF
--- a/crates/gitlawb-node/src/api/encrypted.rs
+++ b/crates/gitlawb-node/src/api/encrypted.rs
@@ -6,29 +6,24 @@ use axum::Json;
 use crate::auth::AuthenticatedDid;
 use crate::error::{AppError, Result};
 use crate::state::AppState;
-use crate::visibility::{visibility_check, Decision};
 
 /// GET /api/v1/repos/{owner}/{repo}/encrypted-blobs
 /// Returns [{oid, cid}] for every encrypted blob in the repo, to any caller who
 /// can read the repo. Not recipient-scoped: recipient identities are not stored,
 /// so access control here is repo readability and decryption is gated by the
 /// envelope crypto (only a real recipient can open an envelope).
+///
+/// Quarantined repos are opaque 404 via [`crate::api::authorize_repo_read`] —
+/// same as issues/changelogs — so a public-but-quarantined mirror cannot leak
+/// its encrypted blob index.
 pub async fn list_encrypted_blobs(
     State(state): State<AppState>,
     auth: Option<Extension<AuthenticatedDid>>,
     Path((owner, repo)): Path<(String, String)>,
 ) -> Result<Json<serde_json::Value>> {
-    let record = state
-        .db
-        .get_repo(&owner, &repo)
-        .await?
-        .ok_or_else(|| AppError::RepoNotFound(format!("{owner}/{repo}")))?;
     let caller = auth.as_ref().map(|e| e.0 .0.as_str());
-    let rules = state.db.list_visibility_rules(&record.id).await?;
-    if visibility_check(&rules, record.is_public, &record.owner_did, caller, "/") == Decision::Deny
-    {
-        return Err(AppError::RepoNotFound(format!("{owner}/{repo}")));
-    }
+    let (record, _rules) =
+        crate::api::authorize_repo_read(&state, &owner, &repo, caller, "/").await?;
     let rows = state.db.list_all_encrypted_blobs(&record.id).await?;
     let blobs: Vec<_> = rows
         .into_iter()
@@ -45,17 +40,9 @@ pub async fn get_encrypted_blob(
     auth: Option<Extension<AuthenticatedDid>>,
     Path((owner, repo, oid)): Path<(String, String, String)>,
 ) -> Result<Vec<u8>> {
-    let record = state
-        .db
-        .get_repo(&owner, &repo)
-        .await?
-        .ok_or_else(|| AppError::RepoNotFound(format!("{owner}/{repo}")))?;
     let caller = auth.as_ref().map(|e| e.0 .0.as_str());
-    let rules = state.db.list_visibility_rules(&record.id).await?;
-    if visibility_check(&rules, record.is_public, &record.owner_did, caller, "/") == Decision::Deny
-    {
-        return Err(AppError::RepoNotFound(format!("{owner}/{repo}/{oid}")));
-    }
+    let (record, _rules) =
+        crate::api::authorize_repo_read(&state, &owner, &repo, caller, "/").await?;
     let cid = state
         .db
         .encrypted_blob_cid(&record.id, &oid)
@@ -81,17 +68,9 @@ pub async fn replicate_encrypted_blobs(
     auth: Option<Extension<AuthenticatedDid>>,
     Path((owner, repo)): Path<(String, String)>,
 ) -> Result<Json<serde_json::Value>> {
-    let record = state
-        .db
-        .get_repo(&owner, &repo)
-        .await?
-        .ok_or_else(|| AppError::RepoNotFound(format!("{owner}/{repo}")))?;
     let caller = auth.as_ref().map(|e| e.0 .0.as_str());
-    let rules = state.db.list_visibility_rules(&record.id).await?;
-    if visibility_check(&rules, record.is_public, &record.owner_did, caller, "/") == Decision::Deny
-    {
-        return Err(AppError::RepoNotFound(format!("{owner}/{repo}")));
-    }
+    let (record, _rules) =
+        crate::api::authorize_repo_read(&state, &owner, &repo, caller, "/").await?;
     let rows = state.db.list_all_encrypted_blobs(&record.id).await?;
     let blobs: Vec<_> = rows
         .into_iter()
--- a/crates/gitlawb-node/src/api/mod.rs
+++ b/crates/gitlawb-node/src/api/mod.rs
@@ -242,13 +242,16 @@ mod authz_guard {
             // PRE-GATED — already owner-gated, in-scope group; guard the gate itself
             (protect, "protect_branch", "did_matches("),
             (protect, "unprotect_branch", "did_matches("),
+            (visibility, "set_visibility", "authorize_repo_read("),
             (visibility, "set_visibility", "require_owner("),
+            (visibility, "remove_visibility", "authorize_repo_read("),
             (visibility, "remove_visibility", "require_owner("),
+            (visibility, "list_visibility", "authorize_repo_read("),
             (visibility, "list_visibility", "require_owner("),
         ];
 
-        // The visibility rows prove require_owner is CALLED; this proves the helper
-        // itself does DID-safe matching, not a raw/trailing-segment compare.
+        // Rows above prove each visibility handler calls authorize_repo_read and
+        // require_owner; this proves the helper itself does DID-safe matching.
         assert!(
             fn_body(visibility, "require_owner").contains("did_matches("),
             "visibility::require_owner must use did_matches for DID-safe owner matching"
--- a/crates/gitlawb-node/src/api/visibility.rs
+++ b/crates/gitlawb-node/src/api/visibility.rs
@@ -85,11 +85,10 @@ pub async fn set_visibility(
     Path((owner, repo)): Path<(String, String)>,
     Json(req): Json<SetVisibilityRequest>,
 ) -> Result<(StatusCode, Json<serde_json::Value>)> {
-    let record = state
-        .db
-        .get_repo(&owner, &repo)
-        .await?
-        .ok_or_else(|| AppError::RepoNotFound(format!("{owner}/{repo}")))?;
+    // Quarantine first (via authorize_repo_read), then owner — same posture as
+    // list_visibility so a quarantined repo cannot be mutated while reads 404.
+    let (record, _rules) =
+        crate::api::authorize_repo_read(&state, &owner, &repo, Some(&auth.0), "/").await?;
     require_owner(&record, &auth.0)?;
     validate_path_glob(&req.path_glob)?;
 
@@ -141,11 +140,8 @@ pub async fn remove_visibility(
     Path((owner, repo)): Path<(String, String)>,
     Json(req): Json<RemoveVisibilityRequest>,
 ) -> Result<Json<serde_json::Value>> {
-    let record = state
-        .db
-        .get_repo(&owner, &repo)
-        .await?
-        .ok_or_else(|| AppError::RepoNotFound(format!("{owner}/{repo}")))?;
+    let (record, _rules) =
+        crate::api::authorize_repo_read(&state, &owner, &repo, Some(&auth.0), "/").await?;
     require_owner(&record, &auth.0)?;
 
     state
@@ -171,14 +167,12 @@ pub async fn list_visibility(
     Extension(auth): Extension<AuthenticatedDid>,
     Path((owner, repo)): Path<(String, String)>,
 ) -> Result<Json<serde_json::Value>> {
-    let record = state
-        .db
-        .get_repo(&owner, &repo)
-        .await?
-        .ok_or_else(|| AppError::RepoNotFound(format!("{owner}/{repo}")))?;
+    // Quarantine first (via authorize_repo_read), then owner — a quarantined
+    // mirror must be opaque even to a caller matching owner_did.
+    let (record, rules) =
+        crate::api::authorize_repo_read(&state, &owner, &repo, Some(&auth.0), "/").await?;
     require_owner(&record, &auth.0)?;
 
-    let rules = state.db.list_visibility_rules(&record.id).await?;
     let rules_json: Vec<_> = rules
         .into_iter()
         .map(|r| {
@@ -205,28 +199,18 @@ pub async fn list_visibility(
 /// denied one (`reinclude`), so a clean-clone client can sparse-exclude the
 /// denied subtrees while re-including the allowed nested paths. Unlike
 /// `list_visibility` this is not owner-gated and never exposes reader_dids.
+///
+/// Quarantined repos are opaque 404 via [`crate::api::authorize_repo_read`] —
+/// same posture as encrypted-blob discovery — so admission and private-subtree
+/// layout are not disclosed to anon, owner, or peers.
 pub async fn withheld_paths(
     State(state): State<AppState>,
     auth: Option<Extension<AuthenticatedDid>>,
     Path((owner, repo)): Path<(String, String)>,
 ) -> Result<Json<serde_json::Value>> {
-    let record = state
-        .db
-        .get_repo(&owner, &repo)
-        .await?
-        .ok_or_else(|| AppError::RepoNotFound(format!("{owner}/{repo}")))?;
-
-    let rules = state.db.list_visibility_rules(&record.id).await?;
     let caller = auth.as_ref().map(|e| e.0 .0.as_str());
-
-    // Whole-repo read gate: a caller who cannot read "/" gets repo-not-found,
-    // matching the git read endpoints, so this never discloses a private repo's
-    // existence or its path layout to an unauthorized caller.
-    if crate::visibility::visibility_check(&rules, record.is_public, &record.owner_did, caller, "/")
-        == crate::visibility::Decision::Deny
-    {
-        return Err(AppError::RepoNotFound(format!("{owner}/{repo}")));
-    }
+    let (record, rules) =
+        crate::api::authorize_repo_read(&state, &owner, &repo, caller, "/").await?;
 
     let withheld =
         crate::visibility::withheld_globs(&rules, record.is_public, &record.owner_did, caller);
--- a/crates/gitlawb-node/src/test_support.rs
+++ b/crates/gitlawb-node/src/test_support.rs
@@ -14612,6 +14612,310 @@ mod tests {
         assert_eq!(resp.status(), StatusCode::NOT_FOUND);
     }
 
+    /// Public-but-quarantined repos must not expose encrypted blob indexes
+    /// (discovery or replicate). Previously these handlers used
+    /// `visibility_check` alone and skipped the quarantine short-circuit in
+    /// `authorize_repo_read`.
+    #[sqlx::test]
+    async fn encrypted_blobs_quarantined_repo_opaque_404(pool: PgPool) {
+        let state = test_state(pool).await;
+        let owner = "did:key:zENCQUAROWNERAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
+        let short = owner.split(':').next_back().unwrap();
+        let mut repo = seed_private_repo(owner, "enc-quar");
+        repo.is_public = true;
+        let repo_id = repo.id.clone();
+        state.db.create_repo(&repo).await.unwrap();
+        state
+            .db
+            .record_encrypted_blob(&repo_id, "deadbeef", "bafybeiquarantinedcid", "")
+            .await
+            .unwrap();
+        state.db.set_repo_quarantine(&repo_id, true).await.unwrap();
+
+        let router = crate::server::build_router(state.clone());
+        for suffix in [
+            "encrypted-blobs",
+            "encrypted-blobs/replicate",
+            "encrypted-blob/deadbeef",
+        ] {
+            let path = format!("/api/v1/repos/{short}/enc-quar/{suffix}");
+            let resp = router.clone().oneshot(anon_get(&path)).await.unwrap();
+            assert_eq!(
+                resp.status(),
+                StatusCode::NOT_FOUND,
+                "quarantined public repo must 404 on {path}"
+            );
+            let body = axum::body::to_bytes(resp.into_body(), usize::MAX)
+                .await
+                .unwrap();
+            let text = String::from_utf8_lossy(&body);
+            assert!(
+                !text.contains("bafybeiquarantinedcid") && !text.contains("deadbeef"),
+                "blob index must not leak on quarantine 404 for {path}: {text}"
+            );
+        }
+
+        // Owner (full did:key and bare key) must also 404 — quarantine is not a
+        // visibility deny that the owner short-circuit can bypass.
+        for caller in [owner, short] {
+            for suffix in [
+                "encrypted-blobs",
+                "encrypted-blobs/replicate",
+                "encrypted-blob/deadbeef",
+            ] {
+                let path = format!("/api/v1/repos/{short}/enc-quar/{suffix}");
+                let resp = router
+                    .clone()
+                    .oneshot(signed_request_as(caller, Method::GET, &path, Body::empty()))
+                    .await
+                    .unwrap();
+                assert_eq!(
+                    resp.status(),
+                    StatusCode::NOT_FOUND,
+                    "owner form {caller} must not read quarantined {path}"
+                );
+            }
+        }
+
+        // Control: clear quarantine → all three discovery surfaces admit again.
+        // `encrypted-blob/{oid}` may 500 without a live IPFS node after the gate
+        // opens; assert it is not the quarantine 404.
+        state.db.set_repo_quarantine(&repo_id, false).await.unwrap();
+        for suffix in ["encrypted-blobs", "encrypted-blobs/replicate"] {
+            let path = format!("/api/v1/repos/{short}/enc-quar/{suffix}");
+            let resp = router.clone().oneshot(anon_get(&path)).await.unwrap();
+            assert_eq!(resp.status(), StatusCode::OK, "released must admit {path}");
+            let body = axum::body::to_bytes(resp.into_body(), usize::MAX)
+                .await
+                .unwrap();
+            let text = String::from_utf8_lossy(&body);
+            assert!(
+                text.contains("bafybeiquarantinedcid"),
+                "released repo must expose blob index on {path}: {text}"
+            );
+        }
+        let get_path = format!("/api/v1/repos/{short}/enc-quar/encrypted-blob/deadbeef");
+        let resp = router.oneshot(anon_get(&get_path)).await.unwrap();
+        assert_ne!(
+            resp.status(),
+            StatusCode::NOT_FOUND,
+            "released get must clear the quarantine 404 (may 5xx without IPFS)"
+        );
+    }
+
+    /// Mirror-admission path (`upsert_mirror_repo` + slash-form id) must also
+    /// opaque-404 encrypted discovery while quarantined.
+    #[sqlx::test]
+    async fn encrypted_blobs_quarantined_mirror_admission_opaque_404(pool: PgPool) {
+        let state = test_state(pool).await;
+        let short = "z6MkEncMirrorAdmitAAAAAAAAAAAAAAAAAAAA";
+        state
+            .db
+            .upsert_mirror_repo(short, "enc-mirror", "/tmp/enc-mirror", None, true)
+            .await
+            .unwrap();
+        let repo_id = format!("{short}/enc-mirror");
+        state
+            .db
+            .record_encrypted_blob(&repo_id, "aabbccdd", "bafybeimirrorcid", "")
+            .await
+            .unwrap();
+
+        let router = crate::server::build_router(state.clone());
+        for suffix in [
+            "encrypted-blobs",
+            "encrypted-blobs/replicate",
+            "encrypted-blob/aabbccdd",
+        ] {
+            let path = format!("/api/v1/repos/{short}/enc-mirror/{suffix}");
+            let resp = router.clone().oneshot(anon_get(&path)).await.unwrap();
+            assert_eq!(
+                resp.status(),
+                StatusCode::NOT_FOUND,
+                "anon must 404 quarantined mirror {path}"
+            );
+            let body = axum::body::to_bytes(resp.into_body(), usize::MAX)
+                .await
+                .unwrap();
+            let text = String::from_utf8_lossy(&body);
+            assert!(
+                !text.contains("bafybeimirrorcid") && !text.contains("aabbccdd"),
+                "blob index must not leak on quarantine 404 for {path}: {text}"
+            );
+
+            let resp = router
+                .clone()
+                .oneshot(signed_request_as(short, Method::GET, &path, Body::empty()))
+                .await
+                .unwrap();
+            assert_eq!(
+                resp.status(),
+                StatusCode::NOT_FOUND,
+                "bare-key mirror owner must not read quarantined {path}"
+            );
+        }
+
+        state.db.set_repo_quarantine(&repo_id, false).await.unwrap();
+        for suffix in ["encrypted-blobs", "encrypted-blobs/replicate"] {
+            let path = format!("/api/v1/repos/{short}/enc-mirror/{suffix}");
+            let resp = router.clone().oneshot(anon_get(&path)).await.unwrap();
+            assert_eq!(resp.status(), StatusCode::OK, "released must admit {path}");
+        }
+        let get_path = format!("/api/v1/repos/{short}/enc-mirror/encrypted-blob/aabbccdd");
+        let resp = router.oneshot(anon_get(&get_path)).await.unwrap();
+        assert_ne!(
+            resp.status(),
+            StatusCode::NOT_FOUND,
+            "released get must clear the quarantine 404 (may 5xx without IPFS)"
+        );
+    }
+
+    /// `withheld-paths` and `list_visibility` must share the quarantine gate.
+    #[sqlx::test]
+    async fn withheld_paths_and_list_visibility_quarantine_opaque(pool: PgPool) {
+        use crate::db::VisibilityMode;
+
+        let state = test_state(pool).await;
+        let owner = "did:key:zVISQUAROWNERAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
+        let short = owner.split(':').next_back().unwrap();
+        let mut repo = seed_private_repo(owner, "vis-quar");
+        repo.is_public = true;
+        let repo_id = repo.id.clone();
+        state.db.create_repo(&repo).await.unwrap();
+        state
+            .db
+            .set_visibility_rule(&repo_id, "/secret/**", VisibilityMode::B, &[], owner)
+            .await
+            .unwrap();
+        state.db.set_repo_quarantine(&repo_id, true).await.unwrap();
+
+        let router = crate::server::build_router(state.clone());
+        let withheld = format!("/api/v1/repos/{short}/vis-quar/withheld-paths");
+        let resp = router.clone().oneshot(anon_get(&withheld)).await.unwrap();
+        assert_eq!(
+            resp.status(),
+            StatusCode::NOT_FOUND,
+            "anon withheld-paths must 404 while quarantined"
+        );
+        let body = axum::body::to_bytes(resp.into_body(), usize::MAX)
+            .await
+            .unwrap();
+        let text = String::from_utf8_lossy(&body);
+        assert!(
+            !text.contains("/secret"),
+            "withheld layout must not leak: {text}"
+        );
+
+        for caller in [owner, short] {
+            let resp = Router::new()
+                .route(
+                    "/api/v1/repos/{owner}/{repo}/withheld-paths",
+                    axum::routing::get(crate::api::visibility::withheld_paths),
+                )
+                .layer(axum::middleware::from_fn(crate::auth::optional_signature))
+                .with_state(state.clone())
+                .oneshot(signed_request_as(
+                    caller,
+                    Method::GET,
+                    &withheld,
+                    Body::empty(),
+                ))
+                .await
+                .unwrap();
+            assert_eq!(
+                resp.status(),
+                StatusCode::NOT_FOUND,
+                "owner form {caller} withheld-paths must 404 while quarantined"
+            );
+
+            let vis = format!("/api/v1/repos/{short}/vis-quar/visibility");
+            let resp = Router::new()
+                .route(
+                    "/api/v1/repos/{owner}/{repo}/visibility",
+                    axum::routing::get(crate::api::visibility::list_visibility),
+                )
+                .with_state(state.clone())
+                .oneshot(signed_request_as(caller, Method::GET, &vis, Body::empty()))
+                .await
+                .unwrap();
+            assert_eq!(
+                resp.status(),
+                StatusCode::NOT_FOUND,
+                "owner form {caller} list_visibility must 404 while quarantined"
+            );
+
+            // PUT/DELETE must also opaque-404 while quarantined (no rule mutation).
+            let put_body = Body::from(r#"{"path_glob":"/extra/**","reader_dids":[]}"#);
+            let resp = Router::new()
+                .route(
+                    "/api/v1/repos/{owner}/{repo}/visibility",
+                    axum::routing::put(crate::api::visibility::set_visibility),
+                )
+                .with_state(state.clone())
+                .oneshot(signed_request_as(caller, Method::PUT, &vis, put_body))
+                .await
+                .unwrap();
+            assert_eq!(
+                resp.status(),
+                StatusCode::NOT_FOUND,
+                "owner form {caller} set_visibility must 404 while quarantined"
+            );
+
+            let del_body = Body::from(r#"{"path_glob":"/secret/**"}"#);
+            let resp = Router::new()
+                .route(
+                    "/api/v1/repos/{owner}/{repo}/visibility",
+                    axum::routing::delete(crate::api::visibility::remove_visibility),
+                )
+                .with_state(state.clone())
+                .oneshot(signed_request_as(caller, Method::DELETE, &vis, del_body))
+                .await
+                .unwrap();
+            assert_eq!(
+                resp.status(),
+                StatusCode::NOT_FOUND,
+                "owner form {caller} remove_visibility must 404 while quarantined"
+            );
+        }
+
+        state.db.set_repo_quarantine(&repo_id, false).await.unwrap();
+        let resp = router.clone().oneshot(anon_get(&withheld)).await.unwrap();
+        assert_eq!(resp.status(), StatusCode::OK);
+        let body = axum::body::to_bytes(resp.into_body(), usize::MAX)
+            .await
+            .unwrap();
+        let text = String::from_utf8_lossy(&body);
+        assert!(
+            text.contains("/secret"),
+            "released withheld-paths must return globs: {text}"
+        );
+
+        // Same release control for list_visibility (owner-gated after gate opens).
+        let vis = format!("/api/v1/repos/{short}/vis-quar/visibility");
+        let resp = Router::new()
+            .route(
+                "/api/v1/repos/{owner}/{repo}/visibility",
+                axum::routing::get(crate::api::visibility::list_visibility),
+            )
+            .with_state(state.clone())
+            .oneshot(signed_request_as(owner, Method::GET, &vis, Body::empty()))
+            .await
+            .unwrap();
+        assert_eq!(
+            resp.status(),
+            StatusCode::OK,
+            "released list_visibility must admit owner"
+        );
+        let body = axum::body::to_bytes(resp.into_body(), usize::MAX)
+            .await
+            .unwrap();
+        let text = String::from_utf8_lossy(&body);
+        assert!(
+            text.contains("/secret"),
+            "released list_visibility must return rules: {text}"
+        );
+    }
+
     #[sqlx::test]
     async fn repo_gate_public_repo_anon_read_admitted(pool: PgPool) {
         struct DirGuard(std::path::PathBuf);


beardthelion (dismissed):
The two production changes are correct and both are load-bearing. I reverted each one and watched the matching test go red: dropping the `WHERE quarantined = FALSE` makes `get_by_cid_skips_quarantined_public_repo` serve the blob (200, `public bytes`), and restoring the old `encrypted.rs` gate makes the discovery route hand back the blob index. The ordering the fix depends on is right too, since `authorize_repo_read` checks quarantine at `api/mod.rs:54` before `visibility_check` at `:58`, and I confirmed by driving the handler that a quarantined repo now 404s for its own owner in both the full `did:key:` and bare-key forms.

What holds the round is the route next door.

## Findings

- [P1] Gate `withheld-paths` on quarantine as well
  `crates/gitlawb-node/src/api/visibility.rs:213`
  This handler is mounted in the same anonymously-reachable group as the three routes you fixed (`server.rs:430-449`) and still does the bare `get_repo` plus `visibility_check` that the PR removed from `encrypted.rs`. I drove it against a public-but-quarantined repo: it returns 200 with `{"withheld":["/secret/**"],"reinclude":[]}` while `encrypted-blobs` on that same repo correctly returns 404. So an anonymous caller still learns the repo is admitted here and learns its private-subtree layout, and because the owner short-circuit sits inside `visibility_check`, the row's claimed owner gets the same answer. It also has a cross-node consequence: a peer's `withheld-paths` answer picks the mirror mode in `sync.rs`. Swapping the resolve for `let (record, rules) = crate::api::authorize_repo_read(&state, &owner, &repo, caller, "/").await?;` is the same shape as your `encrypted.rs` change; I applied exactly that and ran it, and the quarantined request became 404 while a released repo still returns its globs.

- [P2] Add an owner-caller regression in both DID forms
  `crates/gitlawb-node/src/test_support.rs:4233`, `:4290`
  Both new tests drive only `anon_get` / `cid_anon`. The code is right, and I verified that separately, but an anonymous pass cannot distinguish "dropped before the gate" from "denied by the gate", and those come apart the moment someone moves the quarantine check below `visibility_check` or turns the SQL predicate into an in-loop skip after the visibility branch. The invariant this PR implements is that a quarantined repo is withheld from every reader including its owner, and the tree already has that shape for the feed collector in `feed_quarantined_mirror_withheld_from_owner` and `feed_quarantined_full_did_repo_withheld_from_owner`. Mirror those here: one owner-caller case per surface, in the full `did:key:` and bare-key forms.

- [P2] Gate `list_visibility` on quarantine
  `crates/gitlawb-node/src/api/visibility.rs:174`
  Same bare `get_repo` followed by `require_owner`, no quarantine drop, so the caller matching the row's `owner_did` gets 200 with every rule's `path_glob`, `mode`, `reader_dids`, and `created_by`. On a quarantined mirror that `owner_did` came off the wire with the row, so "the owner" here is whoever asserted the DID. Narrower reach than the one above since it needs that DID, which is why it is not P1, but it is the same invariant and the same one-line fix.

- [P3] Extend the release control to all three encrypted routes
  `crates/gitlawb-node/src/test_support.rs:4272`
  The deny loop covers `encrypted-blobs`, `encrypted-blobs/replicate`, and `encrypted-blob/{oid}`, but the release control only re-requests the first. The other two have no must-not-over-drop side, so a change that 404s them even after release would still pass. Worth noting for the third leg specifically: I reordered the loop and confirmed that on revert it fails at 500 rather than 200, because the handler gets past the gate and then dies in `ipfs_pin::cat`. It detects the regression, it just is not demonstrating the leak its name claims.

One process note, not a finding: the body ticks "CI `test (stable)` with Postgres", but the suite has not run on this head. The workflow is sitting at `action_required` awaiting approval, which is the normal shape for a fork PR. I approved it so it runs. I also ran the quarantine suite locally against your head and it is 16 green.

jatmn (changes requested):

## What checks out

The two production changes in this diff are correct and load-bearing for the **single physical row** cases they target:

- Routing the three encrypted handlers through `authorize_repo_read` closes the quarantine bypass that used bare `get_repo` + `visibility_check`. I reverted this hunk locally and the new deny test goes red.
- Adding `WHERE quarantined = FALSE` to `list_all_repos` closes the CID serve bypass for a quarantined public row with no unquarantined sibling. I reverted this hunk locally and `get_by_cid_skips_quarantined_public_repo` serves the blob again.

CI on head `bcc7fb4` is green (`fmt + clippy`, `test (stable)`, MSRV, cargo audit, Docker smoke). None of the findings below are regressions this PR introduced; they are sibling gaps, an incomplete multi-row closure on the CID path, or test holes.

## Findings

- [P1] Gate `withheld-paths` on quarantine (pre-existing sibling gap)
  `crates/gitlawb-node/src/api/visibility.rs:208`
  This handler is mounted beside the three routes you fixed (`server.rs:430-449`) and still uses the **old** `encrypted.rs` pattern: bare `get_repo` plus `visibility_check`, with no `is_repo_quarantined` / `authorize_repo_read` short-circuit. On a public-but-quarantined repo, `visibility_check` allows `"/"` for anonymous callers when there is no matching deny rule, so the handler returns **200** with `withheld` / `reinclude` instead of the opaque **404** that `authorize_repo_read` would return. That discloses both admission and private-subtree layout. The owner short-circuit inside `visibility_check` (`visibility.rs:79-83`) means the row's claimed owner gets the same answer. Downstream, `sync.rs:245-256` and `gl` clone's `fetch_withheld` use this response to choose promisor mirror mode and sparse-clone withhold sets, so the leak is not local to discovery. Please resolve through `authorize_repo_read` the same way `encrypted.rs` now does (or an equivalent quarantine-first gate).

- [P1] `get_by_cid` still serves via an unquarantined mirror twin while slug routes 404 (incomplete CID fix / comment overclaim)
  `crates/gitlawb-node/src/api/ipfs.rs:81-116`, `crates/gitlawb-node/src/db/mod.rs:1234-1238`
  The new `list_all_repos` filter drops only rows with `quarantined = TRUE` on **that physical row**. When a node holds both a quarantined canonical row (UUID `id`) and a non-quarantined mirror row (`{owner}/{name}`) for the same logical repo — a shape the tree explicitly supports (`get_repo_prefers_canonical_over_mirror`, `authorize_repo_read_denies_private_canonical_even_with_public_mirror`; mirror upsert sets `quarantined = false` whenever `get_repo` already finds a canonical row at `sync.rs:290-309`) — slug endpoints 404 through `authorize_repo_read` because `get_repo` prefers the quarantined canonical, but `get_by_cid` still iterates the surviving mirror row, passes `visibility_check` (mirror is hardcoded public with no rules), and serves the object. That contradicts the comment claiming the same fail-closed posture as `authorize_repo_read` (`ipfs.rs:43-44`). The feed collector already guards this class by consulting `list_quarantined_repos` and dropping any row naming a quarantined logical repo (`events.rs:66-103`); `get_by_cid` needs the same logical-repo quarantine fold before serving. This is **not** the separately tracked stale-public-mirror bypass (#124, explicitly deferred at `ipfs.rs:55-57`).

- [P2] Gate `list_visibility` on quarantine (pre-existing sibling gap)
  `crates/gitlawb-node/src/api/visibility.rs:174`
  This owner-authenticated route still does `get_repo` + `require_owner` with no quarantine drop. A caller matching the row's `owner_did` gets **200** with every rule's `path_glob`, `mode`, `reader_dids`, and `created_by` even while the mirror is quarantined. Narrower reach than `withheld-paths` because it needs that DID, but it violates the same "withheld from every reader" invariant stated in `events.rs:58-65`. Same one-line shape as the `encrypted.rs` fix.

- [P2] Add owner-caller regression tests for the new surfaces (test gap)
  `crates/gitlawb-node/src/test_support.rs:4233`, `:4290`
  Both new HTTP tests drive only anonymous callers (`anon_get` / `cid_anon`). That cannot distinguish "quarantine checked before visibility" from "denied later by visibility", and those diverge if someone moves the quarantine check below `visibility_check` or relies only on the SQL filter in `list_all_repos`. The tree already has owner-withheld coverage for feeds (`feed_quarantined_mirror_withheld_from_owner`, `feed_quarantined_full_did_repo_withheld_from_owner`). Please mirror that here: one owner-caller case per surface (encrypted discovery and CID serve), in both bare-key and full `did:key:` owner forms.

- [P3] Extend the encrypted release control to all three routes (test gap)
  `crates/gitlawb-node/src/test_support.rs:4272`
  The deny loop covers `encrypted-blobs`, `encrypted-blobs/replicate`, and `encrypted-blob/{oid}`, but the post-release control only re-requests `encrypted-blobs`. A regression that kept replicate/get at **404** after release would still pass. Please assert **200** (or the appropriate success shape) on all three after `set_repo_quarantine(..., false)`.

- [P3] Exercise mirror-admission row shape and dual-row CID deny in HTTP tests (test gap)
  `crates/gitlawb-node/src/test_support.rs:4233`, `:4290`; `crates/gitlawb-node/src/db/mod.rs:4623`
  The new integration tests quarantine via `create_repo` + `set_repo_quarantine` on a UUID `id`. Production mirror admission uses `upsert_mirror_repo` with a slash-form `id` (`owner/name`) and sets `quarantined` only on first insert. The DB unit test `list_all_repos_excludes_quarantined` covers the SQL filter for mirror rows only. No HTTP test drives the admission path or the canonical-quarantined / mirror-unquarantined coexistence case that the P1 CID finding above depends on.

## Reviewer follow-up

`beardthelion`'s CHANGES_REQUESTED review on head `bcc7fb4` covers the `withheld-paths` gate, `list_visibility` gate, owner-caller tests, and partial release control. None of those are addressed on the current head. Reconciliation also surfaced the dual-row `get_by_cid` bypass above, which is in scope for this PR's CID serve change and comment but was not in that review.

## Out of scope for this round

The following sibling surfaces were inspected and are real, but tracked or deferred separately and not counted as blockers on this PR: stale-public mirror CID serve (#124, `ipfs.rs:55-57`), ungated global `list_pins` / arweave anchors (#121), GraphQL/REST listing when a mirror row survives dedup, and owner mutations (merge/close/webhooks) on quarantined repos.

Ayush7614:
@beardthelion @jatmn addressed the review findings on this head:

**P1**
1. **`withheld-paths`** — now uses `authorize_repo_read` (same shape as encrypted).
2. **`get_by_cid` dual-row** — after `list_all_repos`, fold in `list_quarantined_repos` and skip any candidate that shares owner+name with a quarantined row (feed-collector pattern), so a public mirror twin cannot serve while the canonical is quarantined.

**P2**
3. **`list_visibility`** — `authorize_repo_read` before `require_owner`.
4. **Owner-caller regressions** — encrypted discovery + CID serve covered for full `did:key:` and bare-key owner forms.

**P3**
5. **Release control** — list + replicate assert 200 + CID; get asserts not-404 after release (IPFS may still 5xx).
6. **Mirror admission + dual-row HTTP tests** — `encrypted_blobs_quarantined_mirror_admission_opaque_404` and `get_by_cid_skips_mirror_twin_of_quarantined_canonical`.

Ready for another look.

Ayush7614:
@coderabbitai addressed your review on `b555276`:

1. **Major (CID fold)** — quarantine fold now only matches **quarantined canonical** rows (`!id.contains('/')`). A quarantined mirror no longer suppresses a healthy canonical under the same owner+name. Added `get_by_cid_serves_canonical_when_only_mirror_quarantined`.
2. **Nit (`list_visibility`)** — reuse rules from `authorize_repo_read` instead of re-querying.

@beardthelion @jatmn FYI — this tightens the dual-row fold from the prior round without changing the quarantined-canonical + healthy-mirror deny case.

jatmn (changes requested):

## GitHub state

- **Mergeability:** `MERGEABLE`, `mergeStateStatus: BLOCKED` (my prior `CHANGES_REQUESTED` on `bcc7fb4`, plus beardthelion's on the same head).
- **Checks on head `fb14bd2`:** only `Quality-signal triage` and `CodeRabbit` succeeded. `PR Checks` (`fmt + clippy`, `test (stable)`, MSRV, cargo audit, Docker smoke) is `action_required` on this head and on `b555276` — the suite has not run since your reviewer-response commits. The PR body still ticks CI `test (stable)` with Postgres as done.

## What checks out

Your production changes on this head address the core bypasses from my first round. I rechecked after pushing back on my own draft — the fold is **not** broken in production for the standard `did:key` dual-row case:

- Encrypted discovery/fetch/replicate, `withheld-paths`, and `list_visibility` route through `authorize_repo_read`, so a public-but-quarantined canonical row opaque-404s for anon, owner, and peers.
- `list_all_repos` excludes `quarantined = TRUE` rows, closing the single-row CID serve bypass.
- The CID logical-repo fold (quarantined canonical only, via `!id.contains('/')` + `did_matches`) matches `get_repo` preference: a quarantined mirror no longer suppresses a healthy canonical (`get_by_cid_serves_canonical_when_only_mirror_quarantined`).
- **Fold verification:** with mirror git seeded at the production acquire path (`/tmp/{bare_short}/dual-quar.git`), the handler 404s with the fold present and serves `public bytes` with the fold removed. The stock test passes in both cases because it never populates that path — a test-layout issue, not a production leak on this head.
- CodeRabbit nit from `b555276` is fixed: `list_visibility` reuses rules from `authorize_repo_read` (no double fetch).
- Locally, I ran `cargo test -p gitlawb-node quarantine` — 20 tests, all green on head `fb14bd2`.

## Findings

- [P1] `get_by_cid_skips_mirror_twin_of_quarantined_canonical` does not exercise the logical-repo fold
  `crates/gitlawb-node/src/test_support.rs:4513-4578`
  This is the regression test for the dual-row CID bypass beardthelion and I blocked on. It seeds git at `/tmp/{slug}/{name}.git` where `slug = owner.replace([':', '/'], "_")` (`seed_cid_repos`, line 2953), but the mirror twin row uses bare `owner_did = short`, so `repo_store::acquire` resolves to `/tmp/{short}/{name}.git` (`repo_store.rs:262-266`) — a path the fixture never populates. `upsert_mirror_repo`'s `disk_path` is not consulted by `get_by_cid`. With the fold removed, `object_type` misses on the mirror row and the handler 404s anyway, so the test passes vacuously. In production, sync stores mirror bare repos at `{repos_dir}/{owner_short}/{name}.git` (`sync.rs:313-319`), where objects exist; I confirmed the fold blocks serve when that path is populated and leaks when the fold is removed. Please align the fixture with production disk layout (e.g. clone to `/tmp/{short}/dual-quar.git`) so removing the fold makes this test go red.

- [P2] Required CI has not run on the reviewer-response heads
  PR metadata / GitHub Actions
  Your comment on `fb14bd2` and the PR test plan claim CI `test (stable)` with Postgres is done, but `PR Checks` is `action_required` on `fb14bd2` and `b555276` (only `bcc7fb4` has a completed green `PR Checks` run). Please get `PR Checks` green on the current head before merge; do not rely on the first-commit run for the two follow-up fix commits.

- [P3] Mirror-admission encrypted test still covers only list discovery
  `crates/gitlawb-node/src/test_support.rs:4318-4359`
  `encrypted_blobs_quarantined_mirror_admission_opaque_404` exercises only `encrypted-blobs` (anon + bare-key owner). beardthelion's P3 request also named replicate and get. Worth extending to `encrypted-blobs/replicate` and `encrypted-blob/{oid}` if you want full mirror-admission coverage; not a production gap on this head.

- [P3] Owner-caller regression still missing for encrypted replicate/get
  `crates/gitlawb-node/src/test_support.rs:4271-4288`
  The owner loop in `encrypted_blobs_quarantined_repo_opaque_404` only hits `encrypted-blobs`. Replicate and get are anon-only in the deny loop. Low-risk test hole — production handlers share the same `authorize_repo_read` gate.

- [P3] `list_visibility` release path untested after quarantine clear
  `crates/gitlawb-node/src/test_support.rs:4436-4447`
  `withheld_paths_and_list_visibility_quarantine_opaque` re-checks `withheld-paths` after release but not `list_visibility`. Minor test completeness gap.

## Follow-up on my prior review

My `CHANGES_REQUESTED` items on `bcc7fb4` are **addressed in production code** on `fb14bd2` for the did:key single-row and dual-row cases: `withheld-paths`, `list_visibility`, CID canonical fold, owner-caller coverage on encrypted list and CID serve, expanded release controls on list/replicate, and mirror-admission / dual-row HTTP tests. beardthelion's overlapping items land the same way.

**What's still holding this head:** the dual-row CID test is not load-bearing (P1 above — test quality, not a production bypass on `fb14bd2`), and `PR Checks` has not run on your fix commits (P2). The P3 items are incomplete test coverage from the prior round, not new production defects.

CodeRabbit findings on `b555276` (canonical-only fold scope, `list_visibility` rules reuse) are fixed on `fb14bd2`.

**I rechecked and am dropping these from this round** (drift / pre-existing / out of scope):

- Truncated-owner / `did:gitlawb` dual-row geometry via `did_matches` vs `ref_update_row_names_repo` — pre-existing `get_repo` owner-key behavior; same semantics as `authorize_repo_read`; not introduced or worsened by this diff.
- Visibility `PUT`/`DELETE` during quarantine — pre-existing; I counted owner mutations on quarantined repos out of scope in my first review.
- TOCTOU between `list_all_repos` and `list_quarantined_repos`, per-request `list_quarantined_repos` cost — same class as other non-transactional gates in the tree; not PR-specific actionable defects.
- `list_visibility` 403 → 404 for authenticated non-owners — intentional opaque-404 posture from routing through `authorize_repo_read` before `require_owner`.

## Out of scope for this round

Stale-public mirror CID serve (#124), ungated global `list_pins` / arweave anchors (#121), GraphQL/REST listing when a mirror row survives dedup, and owner mutations on quarantined repos outside visibility (merge/close/webhooks) remain separately tracked. I'm not counting them as blockers here unless this PR regresses them.

Ayush7614:
@jatmn @beardthelion addressed the latest review on head `c167be8`:

1. **[P1] Dual-row CID fixture** — `get_by_cid_skips_mirror_twin_of_quarantined_canonical` now clones objects to `/tmp/{short}/dual-quar.git` (the path `repo_store::acquire` uses for bare-key mirror rows) and points `upsert_mirror_repo` disk_path there. Removing the fold should make this test go red.
2. **[P3] Mirror encrypted coverage** — `encrypted_blobs_quarantined_mirror_admission_opaque_404` now covers list + replicate + get (anon + bare-key owner), with release controls.
3. **[P3] Owner-caller encrypted** — quarantined canonical owner deny now covers replicate/get as well as list.
4. **[P3] `list_visibility` release** — after quarantine clear, also asserts owner `list_visibility` returns 200 with the rule glob.

**CI (P2):** fork PR Checks still need maintainer approval on this head — please approve when you can.

jatmn (changes requested):

## Findings

- [P2] Run the required PR checks on the current head
  GitHub Actions / current head `c167be80dbe2e7caef7682d4418a7386ba10a9b4`
  The current head has only the successful `Quality-signal triage` and CodeRabbit statuses; the required `PR Checks` workflow is `action_required` and has no run for the reviewer-response commits. Please obtain a green run for this head before merging, since the latest commit changes the regression fixtures that protect the quarantine/CID behavior.

Ayush7614:
@jatmn @beardthelion on head `0cc91df`:

Prior code/test findings from the dual-row fixture round remain addressed on `c167be8`. This follow-up clears the **cargo audit** red from the approved `PR Checks` run: bumped **ruint 1.17.2 → 1.20.0** for [RUSTSEC-2026-0220](https://rustsec.org/advisories/RUSTSEC-2026-0220) (fixable lock pin; matches repo policy of fixing in Cargo.lock rather than ignoring).

Please re-approve / re-run `PR Checks` on this head so audit can go green with the rest of the suite.

beardthelion (dismissed):
Prior-round production asks hold on 0cc91dff (lock-only tip on c167be80). Encrypted list/get/replicate, withheld-paths, and list_visibility go through authorize_repo_read; list_all_repos drops quarantined rows; the canonical-only CID fold is load-bearing (removing it serves public bytes via the mirror twin; widening it to mirrors 404s a healthy canonical). Owner full-DID and bare-key denials still land. The dual-row fixture path fix from the last round is real.

What this round catches is a pin the fold made vacuous, and a sibling gap on the same /visibility path this PR partially gated.

## Findings

- [P2] Pin list_all_repos WHERE with a quarantined mirror-only CID HTTP test
  `crates/gitlawb-node/src/test_support.rs:4503`
  get_by_cid_skips_quarantined_public_repo seeds a quarantined canonical. After the dual-row fold landed, gutting WHERE quarantined = FALSE leaves that test green because the fold skips the row when it reappears in list_all_repos. The SQL filter is still the only drop for a quarantined public mirror with no canonical sibling: I seeded that shape, removed WHERE, and got 200 with public bytes; restoring WHERE 404s. list_all_repos_excludes_quarantined still reds on gut, but it never drives the HTTP serve path. Add a CID test for quarantined mirror-only (objects at /tmp/{short}/{name}.git) that fails when WHERE is removed. Keep the existing canonical and dual-row cases.

- [P2] Gate set_visibility and remove_visibility through quarantine before require_owner
  `crates/gitlawb-node/src/api/visibility.rs:82`
  This PR routes GET list_visibility through authorize_repo_read so a quarantined repo is opaque even to owner_did. PUT and DELETE on the same path still do bare get_repo + require_owner. I drove owner PUT during quarantine: 201 Created, while GET list_visibility on that row 404s. That discloses admission and lets the owner rewrite path_glob/reader_dids while reads are opaque. Same shape as list_visibility: authorize_repo_read first, then require_owner. Cover both verbs with an owner-caller quarantine deny (full DID and bare key). Also pin list_visibility's authorize_repo_read marker in the completeness fence (api/mod.rs still lists only require_owner for that handler).

One process note, not a finding: your 0cc91dff lock bump matches open #292 (RUSTSEC-2026-0220 / ruint). After #292 merges, rebase and drop the parallel lock commit if it is still duplicated.

Not an ask, recorded only: get_by_cid's docstring still cites #124 as open; that issue closed via #141. The deny-body !contains("public bytes") asserts on CID 404s are framing debt (opaque envelopes never carry object bytes); the release 200 witnesses are the load-bearing half.

jatmn (changes requested):

## Findings

- [P2] Run the required PR checks on the current head
  GitHub Actions / current head `0cc91dff2c83fc809948c83eb30b4c3d02a7256e`
  The current head has only `Quality-signal triage` and CodeRabbit statuses; the required `PR Checks` workflow is `action_required` and has no run for the reviewer-response commits or the lockfile update. The changed quarantine paths and their new regression fixtures therefore have not received the repository's required fmt/clippy/test/audit validation. Please obtain a green `PR Checks` run for this exact head before merging.

- [P2] Complete the quarantine posture across the `/visibility` route
  `crates/gitlawb-node/src/api/visibility.rs:82`
  `PUT` and `DELETE` retain their pre-existing `get_repo` plus `require_owner` flow, while this PR changes `GET /visibility` to make a quarantined repository opaque even to its owner. A signed owner can therefore create or remove visibility rules and receive a successful response while `GET` returns 404. Complete the route's quarantine posture by checking before the owner gate on both mutations, with full-DID and bare-key denial tests that prove no write persists.

- [P2] Make the new CID regression test exercise the SQL quarantine filter end-to-end
  `crates/gitlawb-node/src/test_support.rs:4503`
  The new DB test covers `WHERE quarantined = FALSE`, but `get_by_cid_skips_quarantined_public_repo` quarantines a canonical row. If that predicate is removed, the new canonical-quarantine fold in `get_by_cid` still skips the row, so the HTTP regression remains green and does not witness the SQL filter's serve-path effect. Add a quarantined mirror-only fixture with its object at the bare-owner acquisition path (`/tmp/{short}/{repo}.git`); that shape has no canonical row for the fold to discard and must 404 only because the SQL filter excludes it.

Ayush7614:
@beardthelion @jatmn — rebased onto current `main` (resolved `ipfs.rs` merge: kept main's walk-budget/admission + this PR's canonical-only quarantine fold) and squashed to a single commit (`3303fb3`). All 13 `PR Checks` jobs are green on this head.

**Your latest round (`0cc91df`)**

1. **`set_visibility` / `remove_visibility` quarantine posture** — both mutations now go through `authorize_repo_read` before `require_owner`, matching `list_visibility` / `withheld-paths`. Owner PUT/DELETE during quarantine return opaque 404 (full `did:key:` and bare-key forms) in `withheld_paths_and_list_visibility_quarantine_opaque`.

2. **Pin `list_all_repos` `WHERE quarantined = FALSE` on the HTTP serve path** — added `get_by_cid_skips_quarantined_mirror_only_repo`: quarantined mirror-only row (no canonical sibling), objects at `/tmp/{short}/mirror-only.git`. Gutting the SQL filter should serve `public bytes`; with it, 404.

3. **Authz fence** — `set_visibility`, `remove_visibility`, and `list_visibility` now assert `authorize_repo_read(` in `api/mod.rs` (owner gate rows retained).

**Prior rounds (still on this head)**

- Encrypted list/replicate/get, `withheld-paths`, and `list_visibility` through `authorize_repo_read`.
- Canonical-only CID fold (`!id.contains('/')`) + dual-row fixture at production acquire path (`/tmp/{short}/dual-quar.git`).
- Owner-caller regressions, expanded release controls, mirror-admission encrypted coverage.

**Note:** the parallel `ruint` lock bump from `0cc91df` was dropped during the main rebase/squash (upstream already carries the advisory fix).

Ready for another look.

jatmn (changes requested):

## Findings

- [P1] Rebase this conflicting branch before merge
  `crates/gitlawb-node/src/db/mod.rs:6977`
  GitHub reports this head as `CONFLICTING`, and it is not descended from the PR base: its merge base is `225644a`, while the declared base is `241b366`. Resolving the merge currently conflicts in this file. The base-to-head diff also shows older versions of the signature and opaque-error code, but those are stale-base drift rather than changes authored by this branch. Rebase onto current `main`, resolve the conflict, and rerun the full review and validation on the resulting diff.

- [P2] Preserve the owner-gate check in the visibility completeness fence
  `crates/gitlawb-node/src/api/mod.rs:241`
  These rows now only assert `authorize_repo_read`, while the separate assertion only proves that the helper itself uses `did_matches`; it does not prove that each visibility handler calls `require_owner`. If that call is removed from GET, PUT, or DELETE, the guard stays green and any admitted reader can inspect or mutate visibility rules. Assert both the quarantine/read gate and `require_owner(` for each visibility handler.

- [P2] Bound the canonical-quarantine matching before the CID visit cap
  `crates/gitlawb-node/src/api/ipfs.rs:354`
  Every row returned by the unbounded `list_all_repos()` scans every quarantined canonical row before `ipfs_max_repo_visits` or any deadline check can run. Thus an unauthenticated CID request performs O(servable repos × quarantined canonical repos) synchronous comparisons while holding the scarce walk permit, so a large mirror/quarantine set can starve the service despite the request budget. Precompute normalized `(owner, name)` canonical keys in a `HashSet` (or filter in SQL) and make the per-repo test O(1).

Ayush7614:
@beardthelion @jatmn — rebased onto current `main` (`241b366`) and pushed `896aff0`.

**Conflict:** took main's peer-writer LEDGER wording (duplicate add).

**Your latest round:**
1. **Authz fence** — `set_visibility` / `remove_visibility` / `list_visibility` rows now assert both `authorize_repo_read(` and `require_owner(`.
2. **CID quarantine fold perf** — precompute `HashSet<(owner_key, name)>` via `quarantine_owner_key` so per-repo checks are O(1) before the visit cap.
3. **Mirror-only CID fixture** — `get_by_cid_skips_quarantined_mirror_only_repo` seeds directly at `/tmp/{short}/mirror-only.git` (no self-clone).

Prior quarantine gates, dual-row fold, visibility PUT/DELETE, and HTTP regressions remain on this head. Ready for re-review when CI is green.

Ayush7614:
Pushed `9544559` (same Windows CI fix as #275): wait for the `real_git_fetch` shim accept loop before driving `git fetch`, fixing the flaky `connection aborted` (10053) failures on **test (windows, non-blocking)**.

All other required checks were already green on `896aff0`; re-running the full suite on this head.

beardthelion (dismissed):
The quarantine change itself is in good shape, and both asks from the last round have landed. What blocks this round is what else the branch now carries: three of its four commits are not this PR's subject.

## Findings

- [P1] Drop the assignee and Windows commits from this branch
  `crates/gitlawb-node/src/api/tasks.rs:1`
  `e5a84688`'s `crates/` tree is byte-identical to #275's head `1676eb0e`, and #275 is open with changes requested. `95445594` and `013bd856` are a second attempt at the Windows shim race that #312 already fixes at the root, in the same file. Merging #276 as it stands lands all three without their own reviews finishing, and it would make #312 conflict for no reason. Rebase down to `896aff0b` alone.

- [P1] Remove the committed Cursor rule
  `.cursor/rules/rtk-token-savings.mdc:1`
  The file's entire content is `/Users/ayushkumar/.cursor/rules/rtk-token-savings.mdc`, an absolute path from your machine. It rode in on `e5a84688`, so it needs removing from #275 too. It also breaks checkout for anyone whose working tree keeps its own ignored `.cursor/` directory, since git writes straight through it.

- [P2] The Windows retry lets `posts >= 2` pass without multi-round negotiation
  `crates/git-remote-gitlawb/tests/real_git_fetch.rs:813`
  `shim.posts` is created once in `start_shim` and counts cumulatively, and `fetch_with_helper` now retries up to three times against that same shim. Two aborted single-POST attempts therefore satisfy the `posts >= 2` lower bound, which is the one assertion proving the bridging path was exercised at all. The deadlock signature is still caught, since a timeout returns `completed == false` and the transient predicate does not fire on it, so this is narrower than it looks. Whatever shape the fix takes, the count the assertion reads has to come from the attempt that actually succeeded. This goes away if the commits move to #312, which is why it sits under the ask above rather than beside it.

- [P3] Delete the unreachable arm in `quarantine_owner_key` and pin its equivalence
  `crates/gitlawb-node/src/api/ipfs.rs:97`
  The second and third match arms return the same value, so the middle guard never changes the result. The more useful half: the fold's correctness now rests on this key agreeing with `did_matches` for every owner spelling, and nothing tests that. I ran it over 289 pairs including `did:web`/`did:gitlawb` collisions and nested `did:key:did:key:` with no disagreement, but that belongs in a unit test beside the helper.

What checks out on `896aff0b`: I gutted each of the four production lines one at a time and watched the matching test fail, then restored. The `WHERE quarantined = FALSE` filter, the CID canonical fold, the encrypted-blob gate and the `set_visibility` gate are all load-bearing. The completeness fence now pins both markers per visibility handler, and removing the owner gate from `set_visibility` turns it red. The fold is a hash lookup instead of a scan per candidate. `list_all_repos` has exactly one production caller, so the new `WHERE` cannot over-withhold on a listing surface, and `authorize_repo_read` is a strict superset of the check it replaced in `encrypted.rs` and `visibility.rs`, so the tightening can only withhold quarantined repos. CodeRabbit's `test_support.rs:4877` note is a false positive: `slug` and `short` resolve to different directories and the test passes.

jatmn (changes requested):

## Findings

- [P1] Drop the unrelated assignee and Windows commits from this branch
  `crates/gitlawb-node/src/api/tasks.rs:1`
  The quarantine/CID implementation is `896aff0b`, but the current base-to-head range also includes `e5a8468`'s task-claim behavior and `9544559`/`013bd856`'s Windows fetch-test changes. Those are separate subjects with open reviews (#275 and #312), so this is current branch-scope drift rather than a change required by the quarantine fix. Rebase this branch down to the quarantine work and keep the other changes in their own reviews.

- [P1] Remove the committed author-local Cursor rule
  `.cursor/rules/rtk-token-savings.mdc:1`
  The path is a tracked symlink (`120000`) whose complete target is `/Users/ayushkumar/.cursor/rules/rtk-token-savings.mdc`.  It is dangling in a fresh checkout and can conflict with a contributor's own tooling directory, so it cannot provide a usable repository rule.  Remove it from this branch (and from the task branch that introduced it) or replace it with actual repository-owned content.

- [P2] Keep the successful fetch's POST count separate from retried attempts
  `crates/git-remote-gitlawb/tests/real_git_fetch.rs:262`
  `fetch_with_helper` now retries against the same `Shim`, but `shim.posts` is cumulative. Two transient attempts that each reach one POST, followed by a successful one-POST fetch, satisfy the `posts >= 2` assertion at the multi-round test even though the successful attempt never exercised multi-round negotiation. Reset or snapshot the counter for each attempt and assert the delta for the successful attempt; alternatively, move this unrelated retry change out with the other Windows work.

euxaristia:
@coderabbitai full review

Ayush7614:
@beardthelion @jatmn — requesting re-review on the current head `30e6856`.

**What changed since your 8/10 reviews:**

1. **[P1] Drop the unrelated commits from this branch** — addressed. Branch now carries only the quarantine/CID subject (`30e6856`) on top of `bfc44f9` (Kevin's merge of main). `e5a8468` (assignee subject), `9544559`, `013bd856` (Windows workarounds), and the cursor symlink are NOT on this head — `git ls-tree -r origin/fix/quarantine-encrypted-and-ipfs-cid | grep cursor` returns nothing, and `crates/gitlawb-node/src/api/tasks.rs` is not in the diff.

2. **[P2] Per-attempt POST counter / fetch_with_helper retry** — moot: `fetch_with_helper` no longer carries the retry-with-same-shim shape that confused the `posts >= 2` floor. That change was on #275; this branch never had it.

3. **[P3] Pin `quarantine_owner_key` equivalence with `did_matches`** — addressed. `test(ipfs): quarantine_owner_key_agrees_with_did_matches` lives at `crates/gitlawb-node/src/api/ipfs.rs:2993` and asserts the fold key agrees with `crate::api::did_matches` across `["", "zABC", "zXYZ", "did:key:zABC", "did:key:zXYZ", "did:key:", "did:key:did:key:zABC", "did:web:zABC", "did:gitlawb:zABC", "did:web:did:key:zABC"]` — every spelling the CID fold can see.

4. **Unreachable-arm note on `quarantine_owner_key`** — keeping both arms: `Some(k) if !k.contains(':') => k.to_string()` normalizes a clean `did:key:base58`, and `_ => owner_did.to_string()` preserves everything else (no prefix, nested `did:key:` inside, empty-after-strip `did:key:`). They return the same `owner_did.to_string()` expression by coincidence of the normalization contract — the distinction in the match is which key the fold produces, not the output expression. Happy to collapse to `let key = match owner_did.strip_prefix("did:key:") { Some(k) if !k.contains(':') => k, _ => return owner_did.to_string() };` if you want it tighter, but the behavior is load-bearing on the test at line 2993.

All 13 `PR Checks` jobs green on `30e6856` (incl. test windows). Ready for re-review.

jatmn (changes requested):

## Merge readiness

- Narrow the PR title and description to the current diff. This head no longer changes CID serving or `list_all_repos`, and the CID/DB tests listed in the test plan are already present in the merge base. The implementation is now a focused encrypted-blob and visibility-route quarantine fix; keeping the older claims will mislead release notes and future reviewers.

## Findings

- [P3] Do not treat a quarantined `withheld-paths` denial as an unsupported endpoint
  `crates/gitlawb-node/src/api/visibility.rs:212`
  Routing this endpoint through `authorize_repo_read` correctly changes a public quarantined repo from `200` to the required opaque `404`. The root cause is that `gl::clone::fetch_withheld` uses the same `404` as a capability signal: it assumes every `404` means an old node does not implement `withheld-paths`, returns empty globs, and prints a normal full-clone start. For a current node denying the request, it then reaches the separately gated Git advertisement and fails later with a less useful error.

  Keep the opaque `404` on the repo-scoped endpoint—changing it to an authorization-specific status would disclose repository state. Instead, separate capability discovery from authorization: advertise `withheld-paths` support through a non-repository-scoped version/capabilities response (or another backward-compatible negotiation path), and only use the empty-glob fallback when that capability is genuinely absent. Once a node is known to support the endpoint, propagate its `404` as a denied/unavailable clone before creating the destination. Add a regression that exercises a quarantined public repo through `gl clone` and asserts it reports denial rather than an unsupported endpoint or normal clone start.

beardthelion (changes requested):
The quarantine gate is correct and load-bearing on this head. I gutted the `is_repo_quarantined` short-circuit in `authorize_repo_read` (`crates/gitlawb-node/src/api/mod.rs:54`) and all three new tests went red on the 200-vs-404 discovery assertion; restored, and the quarantine suite is green again on the pristine head. Removing `require_owner` from `set_visibility` reddens both the completeness fence and the behavioral suite, so the gate is pinned by behavior and not only by the source scrape.

Two things I checked specifically. The owner legs drive both the full `did:key` and the bare-key form across all three route suffixes, which is the form-collision case this surface has gotten wrong before. And the readers of both protected data classes are gated: `list_all_encrypted_blobs` has three callers, two behind `authorize_repo_read` and one the background sync worker at `sync.rs:592`; every `list_visibility_rules` caller is either inside the gate itself, carries its own quarantine check (`repos.rs:621`, `git_upload_pack` at `:1587`), or is a background task.

One note on the deny-body assertions, since they look decorative on their own: they are paired with positive witnesses on the granted 200 after release (`test_support.rs:14693`, `:14889`), so the value is witnessed on a path that can actually carry it. That is the right construction, no change wanted.

What holds the round is the description, not the code.

## Findings

- [P2] Narrow the title and body to what this head changes
  `crates/gitlawb-node/src/db/mod.rs`
  The title still says "and CID serve" and summary bullets 2 and 3 describe a `list_all_repos` quarantine filter, but the diff at `30e6856` is four files: `encrypted.rs`, `mod.rs`, `visibility.rs`, `test_support.rs`. Neither `db/mod.rs` nor `api/ipfs.rs` is in it. The test plan also checks off `get_by_cid_skips_quarantined_public_repo` and `list_all_repos_excludes_quarantined`, which no longer exist at this head. The title becomes the squash subject, so as written it lands in main's history describing work the commit does not contain. Drop the CID half from both; the encrypted-blob and visibility claims are accurate and should stay.

I also confirmed jatmn's point about `fetch_withheld` in `crates/gl/src/clone.rs:265`: a quarantined repo's opaque 404 is read as "old node, no endpoint" and the clone starts with empty globs. The git advertisement 404s the repo immediately after, so it is a worse error message rather than a leak. Theirs to carry, not a second ask from me.

Coverage note, not an ask: the mirror-twin and dual-row CID cases have no direct regression on main or this head. The per-row quarantine flag in `gate_and_serve` handles them structurally, so this is worth a follow-up against main's paged scan rather than anything on this branch.

Ayush7614:
Narrowed title/body per review (pushed no new code commit, metadata-only edit):

**P2 – Title/body still said CID serve `crates/gitlawb-node/src/db/mod.rs` `beardthelion`:** Done. Head `30e6856` touches only `encrypted.rs` + `visibility.rs` + `api/mod.rs` fence + `test_support.rs` — no `db/mod.rs` or `api/ipfs.rs` at this diff, and CID quarantine is already handled on `main`'s paged scan via per-row flags. Title is now `fix(node): close quarantine bypass on encrypted blobs and visibility routes` and body drops the `list_all_repos` / `GET /ipfs/{cid}` bullets and the two CID test-plan items that no longer apply. Release-notes scope is now encrypted-blob + visibility only.

**On `withheld-paths` 404 vs `gl clone` capability discovery `jatmn P3` / `crates/gitlawb-node/src/api/visibility.rs:212`:** Acknowledge — `withheld_paths` now correctly goes through `authorize_repo_read` and returns the same opaque `404` as the other quarantine-closed routes (verified load-bearing: gutting the quarantine check in `authorize_repo_read` reddens all three new quarantine tests). Changing it to a non-404 would disclose existence. The `gl::clone::fetch_withheld` fallback that reads `404` as "old node, no endpoint" is a separate negotiation fix (advertise via version/capabilities, then propagate a known-node `404` as denial before `git clone`), tracked outside this PR per your note — no code change here, body now calls it out.

No code diff in this update; `30e6856` already has the production gates and owner-form (`did:key` + bare) coverage. CI `PR Checks` for this exact head is still `action_required` (fork approval) — prior green was on `bcc7fb4`; will need approval to run on `30e6856`.

jatmn (changes requested):
The quarantine gates themselves are in good shape on this head, and the earlier branch-scope and title/body problems are resolved. Two client-facing edges remain before this is ready.

## Overall guidance

The repeated review rounds are not evidence that the central fix is fundamentally wrong. The server change is focused and its authorization behavior is now well covered. The recurring problem is that each round has tended to validate the changed handler in isolation, while the meaning of its response status is part of a cross-component contract with `gl` and the node sync worker. Moving a handler behind `authorize_repo_read` does more than add a guard: it adds opaque 404 as a possible outcome for callers that previously received 200, including owners and public-repo callers when quarantine is active. Every in-repository consumer of that handler therefore needs to be checked for what it believes 404 means.

That is the common root of both findings below:

- `withheld-paths` loses the distinction between “this old node has no such route” and “this current node has the route but the repo is unavailable.”
- encrypted-blob discovery loses the distinction between “the request succeeded and there are zero blobs” and “the inventory could not be obtained.”

Please address these as one contract-reconciliation pass rather than as two isolated line edits. For each changed endpoint, enumerate the outcomes the server can now produce—successful data, successful empty data, opaque repo-unavailable 404, other HTTP failures, transport failure, and malformed success bodies—and preserve those distinctions until the consuming operation can make an honest final decision. Then exercise the actual consumer, not only the server handler. For this PR, the bounded completion surface is:

1. Keep `authorize_repo_read` and the opaque quarantine 404s exactly as they are; do not weaken the security fix or introduce an authorization-specific response that reveals repo existence.
2. Update `gl clone` so the `withheld-paths` compatibility fallback is taken only when endpoint absence is actually established, while a repo denial from a node known to support the endpoint stops before destination setup.
3. Update encrypted discovery so a 200 empty list remains a clean empty result, but HTTP/transport failure retains an unavailable or incomplete signal through B3 fallback and final reporting. Individual blob failures may remain best-effort.
4. Add consumer-level regressions for both distinctions: legacy endpoint absence versus current-node opaque 404, and successful empty encrypted inventory versus unavailable inventory after the filtered clone phase.
5. Search the direct consumers of these exact endpoints once after the fix (`gl::clone::{fetch_withheld,recover_encrypted_blobs}` and `gitlawb_node::sync::{fetch_withheld,replicate_encrypted_blobs}`) to confirm the new result representation does not turn a safe mirror fallback into a full-clone downgrade or make best-effort replication fatal.

This is not a request for a general capability framework, a clone redesign, changes to unrelated 404 handling, or broad cleanup of every best-effort network call. A small implementation that preserves these endpoint outcomes and proves them at the existing consumer boundaries is sufficient. Keeping the response contract explicit should also prevent another round where fixing one symptom reveals the same ambiguity at the next consumer.

## Findings

- [P3] Preserve discovery failures through encrypted recovery
  `crates/gitlawb-node/src/api/encrypted.rs:26`
  This new gate creates a real mid-clone transition that the client currently cannot represent. If a repo is quarantined after `gl clone` has fetched `withheld-paths` and completed the filtered Git transfer, the signed `/encrypted-blobs` request now returns the required opaque 404. At `crates/gl/src/clone.rs:320-325`, however, `recover_encrypted_blobs` maps every non-success response—and every request error—to `Ok(vec![])`, which is also the value for a successful `{"blobs":[]}` response. With no successful B3 recovery, `run` therefore reaches `Done` even though recovery inventory was unavailable and authorized withheld files may be absent.

  The root cause is the loss of outcome information at discovery: “nothing exists,” “the node denied/unavailable,” and “the request failed” are collapsed into one successful empty result. Keep those outcomes distinct through the B3 fallback and final reporting so an unresolved denial is surfaced as a warning, incomplete result, or error rather than silent success. Please preserve the server's opaque 404, legitimate empty-list success, and best-effort per-blob continuation; this does not require every individual recovery failure to abort the clone. A regression should cover a non-success discovery after the filtered clone, plus a 200 empty-list control.

- [P3] Do not classify the new `withheld-paths` denial as an old node
  `crates/gitlawb-node/src/api/visibility.rs:213`
  A quarantined public repo now correctly returns the opaque 404 here, but `crates/gl/src/clone.rs:263-265` treats every 404 as proof that the node predates `withheld-paths`. It returns empty globs, prints the normal full-clone message, and starts destination setup; only the separately gated Git advertisement later fails with another opaque 404. The security boundary holds, but a current-node denial has been converted into a legacy capability fallback and a misleading later transport failure.

  The root cause is using a repo-scoped opaque response as capability discovery: on a current node the same 404 intentionally covers missing, unreadable, and quarantined repositories, so that response cannot by itself prove the route is absent. Establish endpoint support independently—or preserve equivalent knowledge—before taking the legacy empty-glob fallback; once support is known, surface the repo as unavailable before clone side effects. Keep the opaque 404 and compatibility with genuinely older nodes. A regression should distinguish “known supporting node returns 404” from “legacy node lacks the route”; the specific negotiation mechanism is left open.

beardthelion (changes requested):
The quarantine routing is in good shape on this head, and the title/body narrowing landed. Seven handlers now go through `authorize_repo_read`, the premise tests revert to RED when the short-circuit is removed, and the authz fences pass. What blocks merge is the cross-component contract: moving these handlers behind opaque 404 adds outcomes `gl clone` currently cannot distinguish.

## Findings

- [P3] Do not classify the new `withheld-paths` denial as an old node
`crates/gl/src/clone.rs:265`
`withheld_paths` correctly returns opaque 404 via `authorize_repo_read` on a quarantined repo. `fetch_withheld` maps every `404 | 501` to `Ok((vec![], vec![]))`, which is the legacy "endpoint absent" path. Clone then prints a normal full-clone message and runs `setup_partial_clone` with empty globs; git fetch fails later with another opaque 404. The security boundary holds, but a current-node denial is being read as legacy capability absence. Establish endpoint support before taking the empty-glob fallback, or preserve equivalent knowledge another way. Keep the opaque 404 on the server. A consumer regression should distinguish "known supporting node returns 404" from "legacy node lacks the route."

- [P3] Preserve discovery failures through encrypted recovery
`crates/gl/src/clone.rs:325`
After filtered clone, `recover_encrypted_blobs` calls signed `/encrypted-blobs`. On quarantine the server now correctly 404s. The client maps every non-success response and transport error to `Ok(vec![])`, identical to a successful `{"blobs":[]}`. `run` can reach `Done` with missing withheld files and no warning. Keep legitimate empty-list success distinct from HTTP/transport failure through the B3 fallback and final reporting. Per-blob failures can stay best-effort. A regression should cover non-success discovery after filtered clone plus a 200 empty-list control.

Not an ask, recorded only: the sync worker's `fetch_withheld` already treats non-success as unknown (`None`, promisor-preservation path) rather than legacy-empty; `replicate_encrypted_blobs` is intentionally best-effort. Git smart HTTP and IPFS `gate_and_serve` already quarantine-check before visibility.

One process note, not a finding: expect rebase conflicts with #285 and several others on `api/mod.rs`; mechanical only.

PR #286, opened 2026-07-30
fix(release): bump icaptcha-client with the rest of the workspace

beardthelion:
`icaptcha-client` has been sitting at `0.4.0` since it was created, while the other five workspace crates walked to `0.7.1`. Surfaced while reviewing the 0.7.1 release PR, which bumped five of six annotated crates.

## Cause

release-please rewrites a crate's version line only when both halves hold: the line carries an `# x-release-please-version` annotation, and the manifest is listed under `extra-files` in `release-please-config.json`.

`release-please-config.json` was last written in #130, when the workspace had exactly the five crates it lists. #138 added `crates/icaptcha-client` as a sixth member with the annotation already on its version line, set to `0.4.0` (the workspace version that day), and did not touch the config. The config has been modified twice in the repo's history and never since #130. So every release since has rewritten five files and skipped the sixth.

## What this did and did not break

Nothing functional, which is worth stating plainly since the version has been wrong for four releases:

- Every intra-workspace path dependency is bare `{ path = "..." }` with no version requirement, so no resolution constraint could go unsatisfied.
- No crate is published to a registry. There is no `cargo publish` in any workflow or script.
- `icaptcha-client` is named in no workflow, script, or Dockerfile. It ships only compiled into `gl`.
- It reads no `CARGO_PKG_VERSION` of its own. The 11 uses across the workspace are all in `gl`, `git-remote-gitlawb`, and `gitlawb-node`, so the user-agent in `crates/gl/src/http.rs:28` correctly reports `gl`'s version.

The actual cost was `Cargo.toml:13-19`, which asserts that every crate's version line "carries an `# x-release-please-version` annotation and is listed under `extra-files`". That was false for one of six, and it is the comment the next person adding a crate reads and copies.

## Changes

Adding the missing `extra-files` entry makes that comment true again with no edit to it.

The gate is the part worth reviewing. `scripts/check-release-versioning.sh` derives its required set from `cargo metadata`, so a newly added crate reds CI whether or not its author knows the gate exists. This mirrors `check-gitlawb-core-deps.sh` and its exhaustive-by-construction property; a hand-maintained list would have the same blind spot that caused this. It covers all three release-silent shapes: a member absent from `extra-files`, a member whose version line lost the annotation, and an `extra-files` path that is no longer a member.

## Verification

Each arm was run in both directions:

| Arm | Injection | Result |
| --- | --- | --- |
| Member absent from `extra-files` | none, the live defect | exit 1, names `icaptcha-client` |
| Member missing the annotation | stripped from `gitlawb-attest` | exit 1, annotation message |
| `extra-files` path not a member | added `crates/deleted-crate` | exit 1, orphan message |
| Annotation outside `[package]` | decoy `version` under `[package.metadata.foo]`, real one unannotated | exit 1, annotation message |
| Must not fire | fix applied | exit 0, "6 workspace members" |
| Must not fire | annotated `[package]` plus a second annotated line | exit 0 |
| Must not fire | `[package]` not the first table | exit 0 |

The first arm is the one that matters: it went red against the real defect before the fix and green after, so it is load-bearing rather than vacuously passing. The `[package]` scoping arm was added in 251faf3 after review: a manifest-wide match for the annotation is satisfied by an annotated `version` under any other table, so a crate could leave the version cargo actually reads unannotated, pass this gate, and still freeze at release time. That arm was red before 251faf3 and green after, with the three must-not-fire rows holding in both directions. All injections were reverted. The script restores `Cargo.lock` on exit like its precedent, leaves the tree clean, and runs from repo subdirectories.

## One thing to decide

The next release will move `icaptcha-client` from `0.4.0` to the current workspace version in a single jump. Nothing reads that version, so it is cosmetic, but it will appear in a release diff and is better as a deliberate call than a surprise.

## Summary by CodeRabbit

* **Chores**
  * Added automated release-versioning checks to validate workspace package coverage and version annotations.
  * Updated release configuration to include the captcha client package.
  * Added safeguards that report missing, outdated, or invalid package entries during pull request checks.
  * Improved release readiness by preventing incomplete package version tracking from being merged.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -283,3 +283,35 @@ jobs:
 
       - name: Check gitlawb-core dependency allowlist
         run: bash scripts/check-gitlawb-core-deps.sh
+
+  # release-please only rewrites a crate's version when the line carries the
+  # `# x-release-please-version` annotation AND the manifest is listed under
+  # extra-files in release-please-config.json. Miss either half and the crate
+  # is skipped silently at release time, freezing at its creation version:
+  # icaptcha-client was added as a workspace member after the config was last
+  # written and sat at 0.4.0 through four releases before anyone noticed.
+  # Exhaustive-by-construction, like the allowlist gate above: the required set
+  # is derived from `cargo metadata`, so a newly added crate reds CI whether or
+  # not its author knew this gate existed.
+  release-versioning:
+    name: release versioning coverage
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+
+      - name: Set up Rust toolchain
+        uses: dtolnay/rust-toolchain@b3b07ba8b418998c39fb20f53e8b695cdcc8de1b # stable
+        with:
+          toolchain: stable
+
+      - name: Cache cargo
+        uses: Swatinem/rust-cache@98c8021b550208e191a6a3145459bfc9fb29c4c0 # v2.8.0
+        with:
+          key: release-versioning
+
+      - name: Check every workspace crate is release-managed
+        run: bash scripts/check-release-versioning.sh
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -12,7 +12,8 @@
         { "type": "generic", "path": "crates/gitlawb-node/Cargo.toml" },
         { "type": "generic", "path": "crates/gl/Cargo.toml" },
         { "type": "generic", "path": "crates/git-remote-gitlawb/Cargo.toml" },
-        { "type": "generic", "path": "crates/gitlawb-attest/Cargo.toml" }
+        { "type": "generic", "path": "crates/gitlawb-attest/Cargo.toml" },
+        { "type": "generic", "path": "crates/icaptcha-client/Cargo.toml" }
       ]
     }
   }
--- a/scripts/check-release-versioning.sh
+++ b/scripts/check-release-versioning.sh
@@ -57,10 +57,19 @@ orphaned="$(comm -13 <(printf '%s\n' "$members") <(printf '%s\n' "$listed"))"
 
 # Second half of the invariant: being listed does nothing without the marker,
 # because the generic updater keys on the annotation to find the line.
+#
+# Scoped to the [package] table on purpose. A manifest-wide match is satisfied by
+# an annotated `version` under any other table, so a crate could leave the version
+# cargo actually reads unannotated, pass this gate, and still freeze at release
+# time while release-please rewrote the decoy line instead.
 unmarked=""
 while IFS= read -r manifest; do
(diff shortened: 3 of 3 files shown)


jatmn (changes requested):
The core fix looks good — adding `icaptcha-client` to `extra-files` and gating membership/annotation coverage addresses the real bug. I have one low-risk hardening suggestion, not a merge blocker.

## Findings

- [P3] Tighten the annotation check so it validates `[package].version`, not any `^version` line in the manifest
  `scripts/check-release-versioning.sh:63`
  The gate uses `grep -qE '^version = ".*" # x-release-please-version'` over the whole manifest. A workspace member could leave `[package].version` unannotated and add a decoy annotated line elsewhere — for example under `[package.metadata.foo]` — and the script would exit 0 while Cargo still reads the unannotated package version. release-please would rewrite the decoy line, not the real one, so the crate could freeze at release time while CI stays green. I reproduced this locally; nothing in the repo does it today, and the actual icaptcha-client failure mode this PR fixes is the missing `extra-files` entry, not this path. If you want the gate fully bulletproof, scope the check to the `[package]` section (or otherwise require the annotated `version` field Cargo actually uses). Fine to land as-is and follow up.

kevincodex1:
@beardthelion please address review feedback

beardthelion:
Good catch, and it reproduces. Fixed in 251faf3.

I built a scratch two-crate workspace with the shape you described: `decoy` carries `version = "0.4.0"` unannotated under `[package]`, plus an annotated `version = "0.7.1" # x-release-please-version` under `[package.metadata.foo]`. The gate printed "release versioning coverage: OK (2 workspace members, all annotated and listed)" and exited 0, while `cargo metadata` on the same fixture reported `decoy 0.4.0`. So it is not only that the grep is section-blind, it is that the line cargo reads and the line release-please would rewrite are different lines, which is the freeze this gate exists to catch.

The fix scopes the match to the `[package]` table by tracking the current TOML header in awk rather than matching the whole file.

Ran the same 7-case matrix against both scripts:

| case | want | before | after |
| --- | --- | --- | --- |
| all annotated and listed | 0 | 0 | 0 |
| decoy annotation outside `[package]` | 1 | **0** | 1 |
| no annotation anywhere | 1 | 1 | 1 |
| member missing from `extra-files` | 1 | 1 | 1 |
| orphaned `extra-files` entry | 1 | 1 | 1 |
| annotated `[package]` plus an extra annotated line | 0 | 0 | 0 |
| `[package]` not the first table | 0 | 0 | 0 |

6/7 before with only your case red, 7/7 after, so the change is load-bearing on that property rather than a passing rewrite. The last two rows are the must-not-fire direction: a manifest that is legitimately annotated stays green even with a second annotated line elsewhere, and `[package]` does not have to be the first table. The real workspace still reports all 6 members annotated and listed.

One portability note since it moved from grep to awk: the job runs on `ubuntu-latest`, where `awk` is mawk, and the matrix was run under mawk.

I also added the new arm to the verification table in the PR description.

jatmn (approved):
LGTM

kevincodex1 (approved):

PR #292, opened 2026-07-30
chore(deps): bump ruint to 1.20.0 for RUSTSEC-2026-0220

beardthelion:
`cargo audit` fails on `main`, so every open PR inherits a red check for a problem that is not in its diff.

`ruint` 1.17.2 carries [RUSTSEC-2026-0220](https://rustsec.org/advisories/RUSTSEC-2026-0220): Uint shift operations report incorrect overflow flags and truncate shift amounts. It reaches us transitively through `alloy-primitives`, which `alloy` pulls into both `gitlawb-node` and `gl` for the on-chain operator client.

Our own `U256` use in `operator.rs` is `pow`, `mul`, `div`, and comparison rather than shifts, so the direct exposure looks limited. I did not audit alloy's internals, and the fix is a semver-compatible bump, so this is not worth reasoning about further.

Lockfile only, no manifest change.

## The cost, since it is not free

The bump adds six arkworks crates (`ark-ff`, `ark-ff-asm`, `ark-ff-macros`, `ark-serialize`, `ark-serialize-derive`, `ark-std`), taking the tree from 715 to 721 dependencies. That arrives through `ruint`'s own feature set rather than anything we asked for. Flagging it rather than passing over it, since a dependency-surface increase is exactly the kind of thing a lockfile-only diff hides.

None of it reaches `gitlawb-core`: the dependency purity check still passes at 83 allowlisted normal deps.

## Verification

- `cargo audit` exits 0 (was: 1 vulnerability)
- `cargo build --locked --release -p gitlawb-node`
- 564 node tests, 83 core tests
- `scripts/check-gitlawb-core-deps.sh` OK

The three remaining advisories are unmaintained-crate warnings (`core2`, `derivative`, `paste`, `proc-macro-error2`), which do not fail the check and are out of scope here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -771,6 +771,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "ark-ff"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f7a806ac6c8307b929df4645776290a50ee2aac754ad09d8bdf73391309e43af"
+dependencies = [
+ "ark-ff-asm 0.6.0",
+ "ark-ff-macros 0.6.0",
+ "ark-serialize 0.6.0",
+ "ark-std 0.6.0",
+ "digest 0.10.7",
+ "educe",
+ "num-bigint",
+ "num-traits",
+ "zeroize",
+]
+
+[[package]]
 name = "ark-ff-asm"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


beardthelion:
@kevincodex1 ready when you are. Lockfile only, and it clears `cargo audit` for RUSTSEC-2026-0220 in `ruint`.

Worth knowing that the other open PRs are stale-green rather than unaffected: main and their head branches all carry ruint 1.17.2, and their audit checks passed on runs from 07-28 to 07-30, before the advisory landed in the fetched database. They go red on their next push.

The tradeoff to look at before merging is the six arkworks crates the bump drags in (715 to 721 deps). Those come from ruint's own features rather than anything we asked for. None of them reach gitlawb-core, so the purity check still passes at 83 allowlisted deps.

jatmn (approved):
LGTM

kevincodex1 (approved):

PR #298, opened 2026-08-01
fix: strict DID matching in get_profile

euxaristia:
## Summary

`get_profile` used a broad SQL suffix match (`did LIKE '%:' || $1`) when resolving profiles by short id. That can match a different DID method that happens to share the same trailing id, so the wrong profile row can be returned (or the query can fail when multiple rows match).

This aligns profile lookup with the existing `get_repo` / `normalize_owner_key` contract:

- strip `did:key:` only when the remainder is a bare key id
- leave other DID methods intact
- match via a SQL `CASE` expression (`PROFILE_DID_CASE_SQL`) instead of a wildcard suffix

## Test plan

- [x] Added `get_profile_does_not_match_non_key_did` (sqlx): bare short id and `did:key:` resolve the key-form profile; a non-key DID with the same suffix resolves only itself
- [ ] CI: `cargo test -p gitlawb-node`
- [ ] CI: fmt / clippy

## Notes

Same pattern as the owner-key hardening already applied to repository lookup. Code-only change; no migration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profile lookup accuracy for key-based identifiers.
  * Prevented profiles with key-based and non-key identifiers from being incorrectly treated as the same profile.
  * Ensured profile updates and CID changes work correctly through supported identifier aliases.
  * Added coverage to verify distinct profile matching behavior.

* **Tests**
  * Improved cross-platform test reliability by limiting Unix-specific filesystem checks to supported systems.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -771,6 +771,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "ark-ff"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f7a806ac6c8307b929df4645776290a50ee2aac754ad09d8bdf73391309e43af"
+dependencies = [
+ "ark-ff-asm 0.6.0",
+ "ark-ff-macros 0.6.0",
+ "ark-serialize 0.6.0",
+ "ark-std 0.6.0",
+ "digest 0.10.7",
+ "educe",
+ "num-bigint",
+ "num-traits",
+ "zeroize",
+]
+
+[[package]]
 name = "ark-ff-asm"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -795,6 +812,16 @@ name = "ark-ff-asm"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "62945a2f7e6de02a31fe400aa489f0e0f5b2502e69f95f853adb82a96c7a6b60"
+dependencies = [
+ "quote",
+ "syn 2.0.117",
+]
+
+[[package]]
+name = "ark-ff-asm"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1479009684adc073dff49a1025d3a7065b317a9ead25aaaca38cdc70058ba8a2"
 dependencies = [
  "quote",
  "syn 2.0.117",
@@ -839,6 +866,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "ark-ff-macros"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4a0691ed21ef00ef89c1e9bda832eba493dda3ec2f8d892fb25b705f73f06bb8"
+dependencies = [
+ "num-bigint",
+ "num-traits",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+]
+
+[[package]]
 name = "ark-serialize"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


euxaristia:
🦁

euxaristia:
Addressed CodeRabbit's finding in 9836c3a:

- `upsert_profile` now updates by `existing.did` (the stored row identity) after `get_profile` resolves a bare / `did:key:` alias, and returns that stored form.
- `set_profile_cid` uses the same `PROFILE_DID_CASE_SQL` match so a full `did:key:` form updates a bare-stored profile.
- Added `upsert_profile_updates_via_did_key_alias` covering both write paths.

@coderabbitai full review

euxaristia:
@beardthelion pls review 💪

beardthelion (dismissed):
The get_profile CASE + normalize_owner_key contract matches get_repo: bare and did:key alias, non-key methods stay distinct. I reverted get_profile to the old LIKE matcher against a did:gitlawb-only row and bare short resolved that row; with CASE it returns None. CodeRabbit's write-path ask landed: gutting the existing.did UPDATE bind leaves display_name at "before", and gutting set_profile_cid's CASE leaves profile_cid unset. Both shipped sqlx tests are green on this head after those checks.

## Findings

- [P2] Pin the cross-method deny with a lone non-key fixture
  `crates/gitlawb-node/src/db/mod.rs:4698`
  get_profile_does_not_match_non_key_did seeds both a bare row and a did:gitlawb twin, then asserts bare short returns the bare row. Under the old LIKE matcher Postgres can still return the bare row first, so that assertion can stay green while the bug the PR names is live. Add a case with only the non-key row present and assert get_profile(bare) is None (and keep the dual-row / did:key alias coverage you already have).

- [P2] Pin PROFILE_DID_CASE_SQL with the owner-key boundary matrix
  `crates/gitlawb-node/src/db/mod.rs:922`
  normalize_owner_key_matches_sql_case only exercises OWNER_KEY_CASE_SQL. PROFILE_DID_CASE_SQL is a hand-renamed twin on column did, and a drift there would reopen the same class silently. Run the same boundary VALUES list through PROFILE_DID_CASE_SQL (alias the column did) and assert Rust normalize_owner_key agrees.

Not an ask, recorded only: get_repo still carries ORDER BY / LIMIT 1 for mirror preference and get_profile does not. Auth cannot plant bare+did:key dual rows today, so I am not asking for a preference clause on this round. agents.rs short-id resolve stays method-blind; profile GET does not go through it.

euxaristia:
Addressed review findings in b4016dc:

- \get_profile_does_not_match_non_key_did\ now seeds a lone non-key DID profile first to assert \get_profile(bare)\ returns \None\ before seeding the bare key profile row.
- Added \profile_did_case_sql_matches_normalize_owner_key\ to pin \PROFILE_DID_CASE_SQL\ across the full owner-key boundary matrix.
- Added \#[cfg(unix)]\ target gates to Unix-specific symlink and permission test helpers.

@coderabbitai full review

beardthelion (approved):
Both prior asks landed on `b4016dc` and are load-bearing. Reverting `get_profile` to the old `LIKE` matcher reds the lone non-key fixture. Drifting `PROFILE_DID_CASE_SQL` reds the new boundary matrix. Gutting the `existing.did` UPDATE bind and `set_profile_cid`'s CASE still red the alias write test. CI is green on this head and the base is current with main.

## Findings

(none)

Not an ask, recorded only: `agent_profiles` still keys the raw `did` PK, so bare and `did:key:` siblings can coexist and `fetch_optional` stays plan-order under that state. Auth writes only the `did:key` form, so the API cannot plant that pair today. Same residual as last round. Repos already carry an expression index for the twin CASE; profiles do not. Table is small enough that I am not asking for the migration on this PR.

kevincodex1 (approved):
lgtm nice catch

PR #304, opened 2026-08-08
feat(node): add a commit status reporting surface

beardthelion:
## Summary

Adds a commit status surface so an external CI system can report a verdict against a commit and readers can see it, including a rolled-up state on a pull request head. The node executes nothing: this is the reporting half only.

## Motivation & context

A repo here can be pushed to, reviewed, and merged, but there is nowhere to record whether the code is any good, so anyone who wants build results keeps a GitHub mirror. Every mirror is a reason a GitHub outage still reaches our users.

The gap was narrower than it looked. A completed push already fires an outbound `push` webhook whose payload mirrors GitHub's shape, so the trigger half existed and the reporting half did not.

I looked at running CI on the node and decided against it. Gitea, GitLab, sourcehut and Radicle all put the executor out of process, GitHub's own docs say self-hosted runners should almost never serve public repos, and every major free CI provider gated access in 2021 after cryptomining abuse. A node here is structurally the "public repo, operator's machine" case. That decision is settled, not open for this PR.

## Kind of change

- [x] Feature

## What changed

All in `gitlawb-node`.

- Migration v24: an append-only `status_claims` table ordered by a database-assigned sequence, a nullable `head_commit` on `pull_requests`, and `repo_push_events`.
- `POST /api/v1/repos/{owner}/{repo}/statuses/{sha}`, owner-gated, validated, rate-limited, with three write caps.
- `GET .../commits/{sha}/status` and `GET .../pulls/{number}/status`, both behind the existing repo read gate.
- `GET .../push-events`, a cursor-paged catch-up surface so a checker that missed a webhook can still find the commit.
- `head_commit` maintained on push and frozen at close or merge, so the rollup has an honest target.
- `require_signature` now attaches the verified RFC 9421 material as a request extension, so the write path can persist what a signature actually covered.

Version 24 rather than 18: `pr-173` and `fix/issue-135-ipfs-cid-tree-gate` both claim through 23, and the runner keys only on version without comparing the recorded name, so a collision is a silent full skip rather than an error.

Three design calls worth stating plainly rather than leaving to be inferred. The wire shape follows GitHub's commit status contract, which buys existing status-consuming clients and settles an inconsistency we already had at the push webhook. Writes are append-only claims and the visible status is a projection over them, which is what keeps a later move to signed attestations a substrate swap. Each claim stores the signature and the bytes it covered, because a claim nobody can re-verify after the request is gone is not history that substrate could adopt.

## How a reviewer can verify

```sh
cargo test --workspace
```

The tests worth reading first are the ones that pin the security properties rather than the happy path:

```sh
cargo test -p gitlawb-node --bin gitlawb-node non_owner_on_quarantined_repo_is_indistinguishable_from_missing
cargo test -p gitlawb-node --bin gitlawb-node a_replayed_success_cannot_overturn_the_later_failure
cargo test -p gitlawb-node --bin gitlawb-node capped_insert_holds_the_bound_against_concurrent_writers
cargo test -p gitlawb-node --bin gitlawb-node tightening_visibility_hides_existing_claims_from_anon
cargo test -p gitlawb-node --bin gitlawb-node stored_claim_re_verifies_and_a_tampered_row_does_not
```

## Before you request review

## Protocol & signing impact

The only signing-path change left here is additive and internal: `require_signature` attaches the material it already computed to the request extensions, so a handler can persist what a signature covered. No wire format, no verification behavior, nothing a client can observe.

This branch previously also made `Content-Digest` mandatory and refused duplicated covered components. Those are a protocol change, they are not needed by this feature, and they wanted their own scrutiny, so they now live in #306 behind #305. This branch has been rebased to drop them and depends on nothing there.

## Notes for reviewers

Known limitations, stated up front.

**No first producer is named yet.** Nothing in this PR runs CI, and no specific system is committed to reporting into it. It can merge complete and still not answer the question that motivated it until someone points a checker at it.

**The rollup fallback only helps going forward.** When a pull request has no stored head, it resolves from `repo_push_events`, which only carries pushes that land after this deploys. A pull request whose branch was last pushed before then stays unresolved until it is pushed again. There is no backfill.

**Two contracts are not enforced by a guard yet.** No verdict data reaches gossip, the GraphQL broadcast channel, or anchoring, and I checked that by reading rather than assuming, but the source-scrape test that would keep a future edit from adding one is deferred. So is a reference signing shim.

**No `gl` or MCP coverage.** The signing path already exists in the client, so this is a missing wrapper rather than a reachability problem, but the status domain has none of the roughly thirty tools the other domains have.

Two behaviours I would rather you agree with than discover. A replayed byte-identical write is idempotent and returns the original claim instead of erroring, which also means two genuinely separate reports with identical bodies inside the same clock second collapse into one, since RFC 9421 signs `created` at second granularity. And the pull request rollup persists a resolved head during an unauthenticated GET, bounded in SQL so it fires at most once per pull request.

`head_commit` now appears on existing pull request responses through the record's serialization. Additive, and the head is already derivable by anyone who can list refs, but it is a response shape change rather than an invisible one.

## Summary by CodeRabbit

* **New Features**
  * Added commit status creation, retrieval, aggregation, and pull-request status views.
  * Added signed status submissions with replay protection and rate limits.
  * Added repository push-event history with scoped, opaque cursor pagination.
  * Added visibility-aware status and event access controls.

* **Bug Fixes**
  * Pull requests now record the commit actually merged.
  * Pushes synchronize open pull-request heads and capture branch updates.
  * Push-event records now reflect only successfully accepted Git references.
  * Large pushes are validated and processed reliably in batches.
  * Improved handling of branch deletions and unresolved pull-request heads.


beardthelion:
Worked all seven. Five landed in `5445dda` and `61033af`, two declined with evidence below. Each finding was checked against the code before I acted on it, and one of them turned out to be right about the defect but wrong about its size.

## The seq keyset skip

Real, and the sharpest of the set. I confirmed the window is open rather than assuming it: `acquire_write` releases its advisory lock at `api/repos.rs:952`, while `record_push_events` runs at `:1026`, outside it. So two pushes to one repo serialize through receive-pack and then race on the insert exactly as described.

Taken the second remedy, with one addition. The insert now runs in an explicit transaction that takes a per-repo advisory lock, so the lock releases at commit and no writer can allocate a sequence until the previous row is visible. The wait is bounded (`SET LOCAL lock_timeout`, three attempts at one second, under the pool's five second acquire timeout) because the lock sits on a pooled connection on the inline push path, and an unbounded wait would turn one busy repo into pool exhaustion for unrelated requests.

That bound needed the caller fixed too. `record_push_events` warned and continued, so any error the lock introduced would have dropped those events permanently while the push had already returned success, reaching the same outcome by a different route. It now retries once and then logs at error level naming the repo, the refs, and the SHAs.

Worth recording how the defect is pinned, because the obvious test does not work. A barrier-driven concurrency test passes against the unfixed code: before this change the write was a single autocommit statement, so allocate and commit were one round trip and the interleaving window was never open to a caller of the public API. The reproduction owns both transaction boundaries directly and observes the strand: A allocates, B allocates and commits, the walk from zero leaves the cursor past A, A commits, and the next page comes back empty with both rows committed.

I also corrected the doc comment, which claimed the sequence "cannot disagree with the order the rows actually became visible". True for a single writer, and the sentence that made this invisible.

## The body retention, with a correction

Real, and I have made the change, but the finding overstates it and I would rather say so than quietly ship a fix under a claim that does not hold.

The part I can show: `git_receive_pack` takes its own `body: Bytes` and moves it into `smart_http::receive_pack` at `api/repos.rs:947`. So the whole pack is pinned across the push by the handler's own parameter, independently of the middleware, and that predates this branch. Removing the extension handle does not lower peak memory on a push; the finding's conclusion about concurrent pushes exhausting memory holds either way and is not something this change addresses.

On top of that, my reading of the extractor is that the request is consumed to produce `body`, which would end the extension's handle well before the handler returns rather than at its end. I have not measured that, so treat it as a direction rather than a result.

So the change is still worth making and is in: the field is optional now and populated only behind a marker the persisting route group applies. What it buys is removing the second handle across the remaining chain, and removing the retention outright on small-body signed routes. Lowering peak memory on a push needs a streaming receive-pack, which is separate work.

The failure mode worth guarding here is layer order: a marker applied inside the auth layers is never seen by the middleware that reads it, so the body silently stops being captured while every test stays green. The guard for that moves the layer inside rather than deleting it, and the production-router test still fails. The handler also refuses an absent body rather than storing an empty column.

## The three smaller ones

The stale cursor comment and both fixture comments are corrected. There was a fourth next to them, calling the timestamp format load-bearing for a cursor that no longer reads it; same root cause, so it went with them. The two comments in `repos.rs` that correctly document `(created_at ASC, id ASC)` for mirror-dedup are untouched.

The scrape helper advances by a character now. You are right that it is not reachable today, since every current anchor is ASCII, but five guards share that helper and one that silently reports "not found" is worse than one that fails loudly. Both directions are guarded, so the fix did not reintroduce the zero-position collapse the byte offset originally prevented.

## Declined: the two thread-local findings

`api/repos.rs:3101` and `db/mod.rs:1132` both rest on `#[sqlx::test]` using a multi-thread Tokio runtime. It does not.

`sqlx-macros-core 0.8.6`, the version pinned in this lockfile, expands the test body to `::sqlx::test_block_on` (`src/test_attr.rs:73`), and that runs on a runtime built with `runtime::Builder::new_current_thread()` (`src/lib.rs:66`). The test future cannot migrate between workers, so the thread-local counter cannot split.

The suggested remedy also points the wrong way for this suite. There is a process-global `AtomicUsize` counter elsewhere in this branch, `BRANCH_RESOLVES`, and it needs an explicit guard on every test that touches it precisely because `#[sqlx::test]` gives each test its own database while a global counter is shared, so an unguarded test inflates the count a guarded test asserts on. Moving the push-write counter to that shape would import that problem rather than solve one.

If the runtime flavour ever changes, the concern becomes real, so the new concurrency work deliberately uses `tokio::spawn` with a `tokio::sync::Barrier` rather than `std::thread::spawn`, which keeps every writer on the runtime thread.

Full workspace suite is green at 1165.

jatmn (changes requested):

## Findings

- [P1] Require a body-bound Content-Digest before persisting signed provenance
  `crates/gitlawb-node/src/auth/mod.rs:212`
  The status handler now stores the request body as evidence of the signed verdict, but `require_signature` substitutes an empty string when `Content-Digest` is absent and only compares the digest when the header exists. An owner can therefore sign the method/path plus an empty digest, omit the header, and submit any status body; the resulting claim asserts provenance that cannot re-verify the stored body. Require and verify the digest on this route (or make this PR depend on the digest-enforcement change) before recording the claim.

- [P1] Apply push side effects only to refs that receive-pack accepted
  `crates/gitlawb-node/src/api/repos.rs:1017`
  `smart_http::receive_pack` regards a zero process exit as success and forwards the protocol response untouched, while receive-pack reports individual rejected refs as `ng <ref> ...` pkt-lines even with that exit status. The new code then updates PR heads and writes poll events for every ref parsed from the request. A rejected non-fast-forward, hook, or unpack failure can consequently publish an uninstalled SHA and make the PR rollup target it. Parse the report-status and use only accepted refs (or derive the updates from the resulting repository state) for these side effects.

- [P1] Keep PR-head writes ordered with the git write lock
  `crates/gitlawb-node/src/api/repos.rs:952`
  The repository lock is released before `update_open_pr_heads` runs. If pushes A then B are accepted for the same branch, B can reach the unconditional `UPDATE pull_requests SET head_commit = v.sha` first and A can overwrite it afterwards. Since a non-null stored head bypasses the fallback, the rollup remains pinned to A even though the branch points at B. Serialize this update with the receive-pack order, or make the update conditional on an ordering/version that cannot move the head backwards.

- [P1] Make push-event cursors opaque and repository/node-bound
  `crates/gitlawb-node/src/api/events.rs:349`
  Every non-negative integer is accepted and passed directly to a per-repo `seq > cursor` query. A high value issued by another repo/node, or retained across a restore, returns an empty 200 and is echoed, permanently skipping this repo's history while the subscriber believes it is caught up. The returned value is also the table-global `BIGSERIAL`, so gaps let a reader measure activity in other repositories, including private ones. Use an opaque scoped cursor, or validate that it was issued for this repository and reject an invalid/ahead value visibly.

- [P2] Enforce the ref-count limit while parsing the receive-pack request
  `crates/gitlawb-node/src/api/repos.rs:874`
  The 10,000-ref bound is checked only after `parse_ref_updates` has scanned the entire request and allocated three strings for every valid pkt-line. Git routes allow a pack body up to the configured 2 GB, so a signed caller can force the allocation and CPU work that the new cap is meant to prevent before receiving the 400. Stop parsing as soon as the cap is exceeded (or otherwise bound the parser) before retaining each update.

- [P2] Apply the status body-size limit before signature middleware buffers it
  `crates/gitlawb-node/src/api/status/mod.rs:176`
  The advertised 8 KiB request bound is checked only after `require_signature` has collected and hashed the full body; the route has no matching transport body limit. A large signed request is therefore allocated before authorization and only then rejected by this handler, defeating the stated bound under concurrent requests. Add a small `RequestBodyLimitLayer` outside the auth middleware (or make the middleware collection bounded) for this route.

jatmn (changes requested):

## Merge readiness

  `crates/gitlawb-node/src/api/repos.rs:980`
  This head is based on `fdf716d`, while live `main` is `bfc44f9`; an actual merge has content conflicts in `api/pulls.rs`, `api/repos.rs`, `db/mod.rs`, and `test_support.rs`. Those include the receive-pack and persistence paths this feature changes, so resolve the rebase and have the resolved diff reviewed rather than merging the stale branch.

- [P1] Resolve the failing CodeQL status before merge
  `.github/workflows`
  The check rollup has a failed aggregate `CodeQL` status, even though its four named Analyze jobs succeeded, and the available check data provides no diagnostic to attribute it. Re-run or inspect that failed status after the rebase and do not merge until the required security check is green or its failure is explained.

## Findings

- [P1] Use the accepted-ref set for every post-receive effect
  `crates/gitlawb-node/src/api/repos.rs:1027`
  The receive-pack report is parsed into `accepted` at `:980-993`, but that value is used only for the new PR-head and poll-event writes. The rest of the success path switches back to the request-declared `ref_updates`: `record_push` selects its first SHA at `:1027`, certificates iterate it at `:1043`, webhooks at `:1078`, and later pin/replication/announcement work derives its tips from it. In a mixed push such as `ng refs/heads/main` plus `ok refs/heads/third`, the node therefore signs, sends, and may replicate metadata asserting that the rejected `main` SHA landed.

  The root cause is two competing representations of the push transition after receive-pack: the declared request set and the verified result set. Make the verified accepted set the single source of truth for every post-receive side effect (including trust attribution, certificates, webhooks, pin candidates, and announcements). Keep the declared set only for pre-receive validation and operations that specifically need the client's proposed update.

- [P1] Do not treat an absent or unreadable report-status as proof that all refs landed
  `crates/gitlawb-node/src/api/repos.rs:1759`
  `accepted_ref_updates` returns `declared.to_vec()` whenever `parse_report_status` returns `None`. That happens when a client does not negotiate `report-status`, and also when the parser cannot decode the response. A zero process exit only says that `receive-pack` ran; it does not say each ref landed. Consequently, an unreported non-fast-forward or hook rejection can be recorded by the new PR-head and catch-up writers as an installed commit. The current comment calls this an availability tradeoff, but it violates the new surface's central integrity property: its persisted transition can name a SHA the repository does not contain.

  The root cause is using “unknown” as though it meant “accepted.” Preserve the availability goal by deriving acceptance from a trustworthy source: require/verify report-status for the effects that need per-ref truth, or read the resulting refs under the existing repository lock before publishing state. If neither source can establish a ref's result, do not persist or announce it as a successful transition.

- [P1] Snapshot the source head when closing a pull request
  `crates/gitlawb-node/src/api/pulls.rs:301`
  New PRs always start with `head_commit = NULL` (`:56-73`), and `close_pr` only changes status. The only fallback that can populate that field is deliberately restricted to open PRs in `rollup_head`. Thus, if the source branch was pushed after this feature deployed but before the PR was created, then the PR is closed before its status endpoint is first read, its existing `repo_push_events` row is never consulted; later pushes also skip closed PRs. The closed PR permanently returns `head_resolved: false`, despite the source head being available in the new push-event history.

  The root cause is that the lifecycle has a create/push/read fallback but no close-time capture: the status transition disables the only recovery path before recording the value that must be frozen. Resolve the source head while the PR is still open and persist it atomically with (or immediately before) the close transition. Keep the existing rule that subsequent pushes cannot move a closed or merged PR's stored head.

PR #306, opened 2026-08-08
fix(node): require Content-Digest on a signed request

### LINKED ISSUE #305: Signed requests are accepted without a Content-Digest header
## What

A signed request with no `Content-Digest` header is currently accepted, and its body is never checked against any signature.

`require_signature` reads the header with a fallback to the empty string:

```rust
.unwrap_or("")
```

When the header is absent, the covered `content-digest` component is empty on both the signing side and the verifying side, so the Ed25519 verification passes. The body comparison is then skipped for want of a header. The result is that a request can carry a body its signature never covered, while still presenting as a valid signed request.

Separately, `HttpSignature::parse` accepts a repeated component identifier in `Signature-Input`. RFC 9421 section 2.1 says an identifier appears at most once. A repeat conveys nothing, but it does repeat a line in the signing string, so the size of what a verifier builds is set by how many times the caller chose to write the same name.

## Why it matters

Every write to a node authenticates this way, so the guarantee a signature is supposed to give (this identity sent this body) does not hold for a caller who simply omits a header.

## Proposed change

Refuse both. Return 400 with a `missing_content_digest` code when the header is absent, which also makes the body comparison unconditional in practice, and reject a duplicated covered component at parse time.

## Compatibility

No conforming client regresses. `sign_request` computes the digest unconditionally and returns it in `SignedHeaders`, and all eight production call sites attach the header: `gl/src/http.rs` (three), `git-remote-gitlawb/src/main.rs` (two), `gitlawb-node/src/main.rs`, `sync.rs`, and `api/repos.rs`. The only paths that omit it are git-remote's unsigned ones, which carry no signature headers at all and are refused earlier by the missing-headers branch.

Both changes refuse things a conforming client never sends, so previously signed history and existing nodes are unaffected.

Raising this per CONTRIBUTING, which asks protocol-level changes to start as an issue. The change is small and I have it written with tests; I will open the PR against this issue.

beardthelion:
## Summary

Refuses a signed request that carries no `Content-Digest` header, and refuses a duplicated covered component in `Signature-Input`.

## Motivation & context

Closes #305

The header was read with a fallback to the empty string. With the header absent, the covered `content-digest` is empty on both the signing and verifying sides, so verification passes and the body comparison is then skipped for want of a header. A request could carry a body its signature never covered while presenting as validly signed. Every write to a node authenticates this way, so the guarantee a signature is meant to give did not hold for a caller who simply left a header off.

The parser change is smaller: RFC 9421 section 2.1 says a component identifier appears at most once, and a repeat conveys nothing while repeating a line in the signing string a verifier has to build.

## Kind of change

- [x] Security fix

## What changed

- `gitlawb-node`: `require_signature` returns 400 with a `missing_content_digest` code when the header is absent, and the body comparison is now unconditional.
- `gitlawb-core`: `HttpSignature::parse` rejects a repeated covered component.

## How a reviewer can verify

```sh
cargo test --workspace
cargo test -p gitlawb-node --bin gitlawb-node signed_request_without_content_digest_is_rejected
cargo test -p gitlawb-core parse_rejects_duplicate_components
```

The first test signs `content-digest` as the empty string and then sends no header, which is the exact shape that slipped through, and asserts the request is refused before any handler runs. Both tests were checked against an injected revert of the production line they cover, so a regression turns them red rather than leaving them green.

## Before you request review

## Protocol & signing impact

No conforming client regresses. `sign_request` computes the digest unconditionally and returns it in `SignedHeaders`, and all eight production call sites attach the header: `gl/src/http.rs` (three), `git-remote-gitlawb/src/main.rs` (two), `gitlawb-node/src/main.rs`, `sync.rs`, and `api/repos.rs`. The only paths that omit it are git-remote's unsigned ones, which carry no signature headers at all and are refused earlier. Both changes refuse things a conforming client never sends, so previously signed history is unaffected.

## Notes for reviewers

This started life inside #304, the commit status surface, because I found it while making stored status claims re-verifiable. It does not belong there: that feature does not need it, this is a protocol change, and the two want different scrutiny. #304 has been rebased to drop it and now depends on nothing here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP signature validation by rejecting duplicate, empty, or parameterized signed components.
  * Requests using signature authentication must now include a valid `Content-Digest` header.
  * Content digests are consistently checked against request bodies.
  * Invalid or missing digest headers now return a clear client error.

### DIFF
--- a/crates/gitlawb-core/src/http_sig.rs
+++ b/crates/gitlawb-core/src/http_sig.rs
@@ -14,7 +14,7 @@
 use base64::{engine::general_purpose::STANDARD, Engine};
 use chrono::Utc;
 use sha2::{Digest, Sha256};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use crate::did::Did;
 use crate::identity::Keypair;
@@ -67,21 +67,19 @@ impl HttpSignature {
         let components_str = &rest[open + 1..close];
         let params_str = &rest[close + 1..]; // starts with ';'
 
-        // "\"@method\" \"@path\" \"content-digest\"" → ["@method", "@path", "content-digest"]
-        let components: Vec<String> = components_str
-            .split_whitespace()
-            .map(|s| s.trim_matches('"').to_string())
-            .collect();
-
-        // RFC 9421 §2.1: a component identifier appears at most once. A repeat
-        // says nothing extra and only lengthens the signing string a verifier
-        // has to build, so it is refused rather than folded.
-        for (i, c) in components.iter().enumerate() {
-            if components[i + 1..].contains(c) {
+        // RFC 9421 §2.1: a component identifier appears at most once. Gitlawb
+        // only supports bare names; parameters are rejected so reordered
+        // equivalent forms cannot bypass duplicate detection as distinct strings.
+        let mut components = Vec::new();
+        let mut seen = HashSet::new();
+        for token in components_str.split_whitespace() {
+            let name = parse_covered_component(token)?;
+            if !seen.insert(name.clone()) {
                 return Err(Error::HttpSignature(format!(
-                    "duplicate covered component '{c}' in Signature-Input"
+                    "duplicate covered component '{name}' in Signature-Input"
                 )));
             }
+            components.push(name);
         }
 
         let params = parse_params(params_str)?;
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P1] Make duplicate-component validation linear
  `crates/gitlawb-core/src/http_sig.rs:79`
  This scans the remaining component list for every component. An unauthenticated caller can send a `Signature-Input` with many distinct quoted components, so the new check performs quadratic string comparisons before timestamp, key, required-component, or signature validation rejects it. This server uses Axum/Hyper's default HTTP/1 configuration and does not set a tighter header buffer limit (Hyper permits roughly 417 KiB by default), enough for tens of thousands of short component tokens and billions of comparisons. `optional_signature` routes invoke this path whenever either signature header is present, including public read routes without an IP limiter, allowing repeated header-only requests to monopolize worker CPU. Track seen components in a `HashSet` (or enforce a strict component count) and add coverage for a long distinct-component list.

beardthelion:
Heads-up while this is open, since it touches `http_sig.rs`: `HttpSignature::parse` has a live panic on attacker-controlled input, now filed as #348.

`crates/gitlawb-core/src/http_sig.rs:60-67` finds `(` and `)` independently and then slices `&rest[open + 1..close]`, so a `Signature-Input` where `)` precedes `(` reverses the range and panics. Every other malformed shape in that function returns a typed `Error` that the caller renders as a 400; this one class walks past it. It is reachable pre-auth, because `auth/mod.rs:88` parses before any key is resolved and `optional_signature` delegates whenever signature headers are merely present.

This PR's diff does not touch that indexing (it adds `parse_rejects_duplicate_components` around `parse` but leaves the `find`/slice lines alone), so the panic survives the merge either way. Flagging it here only so it is visible to whoever lands first: #306 and #261 both touch this file, and the fix in #348 is a single `open < close` comparison, so it is worth deciding deliberately whether it rides along with one of these or goes separately rather than having all three collide.

beardthelion:
Addressed both review items on 25721692.

Duplicate-component validation now tracks seen bare names in a HashSet instead of scanning the tail slice per entry. Revert-check: restoring the old loop made `parse_many_distinct_components_completes_quickly` fail at 2.84s on a 20k-component header; the HashSet path finishes in under 0.1s.

Parameterized component identifiers are rejected at parse (`component parameters are not supported on '…'`). Gitlawb only supports bare `@method`, `@path`, and `content-digest` end to end, so this closes the reordered-parameter duplicate bypass without accepting forms `missing_components()` cannot handle. Added tests for parameterized and reordered-parameter inputs.

`cargo test -p gitlawb-core http_sig::` (19/19) and `signed_request_without_content_digest_is_rejected` still green.

jatmn (approved):
LGTM

beardthelion:
@kevincodex1 LGTM

Issue #308, opened 2026-08-10
Use strict Ed25519 signature verification


euxaristia:
Use verify_strict from ed25519-dalek for identity signature verification in gitlawb-core, matching the crate's recommended RFC 8032 strict verification mode.

beardthelion:
Fixed by #309, which switched identity and attestation verification to verify_strict. Closing.

PR #309, opened 2026-08-10
fix(core): enforce strict RFC 8032 Ed25519 signature verification

euxaristia:
## Summary
Use the strict verification path from `ed25519-dalek` for Ed25519 identity signatures in `crates/gitlawb-core/src/identity.rs`, as recommended by the crate for RFC 8032 compliance.

## Changes
- `crates/gitlawb-core/src/identity.rs`: use `verify_strict` instead of `verify` for signature verification.

## Test plan
- `cargo test -p gitlawb-core` passes.

Refs #308

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened signature verification to reject forged signatures, invalid public keys, and malleable signatures.
  * Added regression coverage to help prevent similar security issues in the future.

### DIFF
--- a/crates/gitlawb-attest/src/attestation.rs
+++ b/crates/gitlawb-attest/src/attestation.rs
@@ -16,7 +16,7 @@
 //! by exact match.
 
 use base64::{engine::general_purpose::URL_SAFE_NO_PAD as B64U, Engine};
-use ed25519_dalek::{Signature, Signer, SigningKey, Verifier, VerifyingKey};
+use ed25519_dalek::{Signature, Signer, SigningKey, VerifyingKey};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
 use crate::error::{Error, Result};
@@ -111,7 +111,7 @@ impl Attestation {
             .try_into()
             .map_err(|_| Error::Signature("signature must be 64 bytes".to_string()))?;
         let sig = Signature::from_bytes(&sig_bytes);
-        vk.verify(&bytes, &sig)
+        vk.verify_strict(&bytes, &sig)
             .map_err(|e| Error::Signature(format!("ed25519: {e}")))?;
 
         Ok(vk)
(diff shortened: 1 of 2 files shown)


kevincodex1 (approved):

beardthelion (dismissed):
The identity switch is correct and load-bearing. `identity::verify` now calls `verify_strict`; reverting that one line reds the new weak-key regression with the named panic, and valid dalek signatures still verify. HTTP Signature, UCAN, auth, gl cert verify, and git-remote all go through that helper, so they inherit the rejection. `icaptcha` was already on `verify_strict`.

## Findings

- [P2] Switch attestation verification to `verify_strict` and pin the same forge
  `crates/gitlawb-attest/src/attestation.rs:114`
  This is still `vk.verify` (ZIP-215 / non-strict). The same small-order identity-point forge the new identity test rejects verifies successfully on that path, and `Registry::verify` treats a passing signature plus a registered payload shape as `fully_verified` without pinning the recovered key. Replace with `vk.verify_strict`, drop the unused `Verifier` import, and add a weak-key rejection test mirroring `identity::tests::verify_rejects_weak_key_signature`. `verify_strict` is an inherent method on dalek 2.2.0 `VerifyingKey`; `cargo check -p gitlawb-attest` is clean with that swap.

One process note, not a finding: rebase onto current main before the next push. Collision with #193 is mechanical (shallow file overlap); this head is ~40 commits behind.

Not an ask, recorded only: `Did::to_verifying_key` still admits weak public keys; fail-closed is at verify time. A small-order-R-only fixture would tighten the regression further, but the current weak-A fixture already reds a `verify_strict` revert.

euxaristia:
@coderabbitai full review

euxaristia:
Both requested changes are addressed in the rebased head (1fcbbbd):

- **Attestation strict verification:** `crates/gitlawb-attest/src/attestation.rs` now calls `vk.verify_strict`, the unused `Verifier` import is dropped, and a `verify_rejects_weak_key_signature` regression test mirrors the identity-side one.
- **Rebase:** the branch is rebased onto current `main` (225644a), so the shallow overlap with #193 is resolved.

Verified locally: `cargo test -p gitlawb-core -p gitlawb-attest` passes (including both weak-key rejection tests), and `cargo clippy -p gitlawb-core -p gitlawb-attest --all-targets` is clean.

beardthelion (approved):
Both round-1 asks landed. `attestation.rs:114` is now `verify_strict`, the `Verifier` import is gone, and the mirrored weak-key test is in place at `attestation.rs:563`. The branch is caught up with main.

I checked that both new tests are load-bearing rather than green by luck: reverting each `verify_strict` back to the non-strict `verify`, one at a time, reddens exactly its own test, and on the attestation side the non-strict path returns `Ok` with the identity point as the recovered signer, which is the forgery being accepted. Dropping the `Verifier` import turns out to carry weight too, since neither revert compiled until I put it back.

No compatibility regression. Every key here comes from a clamped scalar, so nothing we sign has a small-order A or R, and the accepting tests in both crates drive real signatures through the changed line and stay green. That covers the three primitive verification sites in the workspace: these two plus `icaptcha.rs:367`, which was already strict.

The two failing test jobs are not from this PR. `every_peers_table_write_is_dispositioned` fails on two peers-table writers in `api/peers.rs` and `main.rs`, neither of which this branch touches. It is red on main and a separate fix is in flight.

Two small things, neither blocking. The test plan still says `cargo test -p gitlawb-core`; with the attest change reverted that command stays green, so it is worth adding `-p gitlawb-attest` when you next touch the body. And the attestation test asserts `Error::Signature(_)`, which three sites in `verify_signature` can emit, so asserting the message starts with `ed25519:` would keep it pinned if those checks are ever reordered.

@kevincodex1 this is ready.

PR #320, opened 2026-08-11
fix(node): Escape LIKE wildcards in certificate prefix search

### LINKED ISSUE #319: Certificate prefix search treats LIKE metacharacters as wildcards
`Db::list_ref_certificates_by_prefix` (used by the CLI for short-ID resolution) builds its match pattern by concatenating the caller's prefix with `%` and passing it to a `LIKE` clause.

Because the prefix itself is not escaped, any `%`, `_`, or `\` characters it contains are interpreted as `LIKE` wildcards instead of literal characters. A prefix containing those characters can therefore match certificates outside the requested prefix, which makes short-ID resolution return over-broad or incorrect result sets.

Expected: prefix matching treats the prefix literally, so `abc%` only matches IDs that literally start with `abc%`.

Fix: escape `%`, `_`, and `\` in the prefix and add an explicit `ESCAPE '\'` clause to the query.

euxaristia:
## Summary

Short-ID resolution passes a caller-provided prefix into a `LIKE` pattern without escaping, so any `%` or `_` in the prefix is interpreted as a wildcard instead of a literal character, and prefix lookups can match certificates outside the requested prefix. This escapes the metacharacters and pins the escape character with an explicit `ESCAPE` clause, so prefix matching is always literal.

The escape character is `!` rather than the backslash default. Spelling it as `ESCAPE '\'` would put a backslash inside an ordinary SQL string literal, and under the legacy `standard_conforming_strings=off` setting that backslash escapes the closing quote, leaving the statement unterminated. That parse failure happens before the bound parameters are considered, so it would break *every* prefix lookup, including prefixes holding no metacharacter at all. Connection settings come from outside the process, so that mode can reach the query from a database- or role-level setting. Using `!` keeps the statement free of backslashes, so it parses the same under either mode without a second layer of backslash escaping. A backslash in the prefix then becomes an ordinary character, which is the literal match the search wants anyway.

Fixes #319

## Changes

- `crates/gitlawb-node/src/db/mod.rs`
  - `list_ref_certificates_by_prefix`: escape `%`, `_`, and `!` in the `prefix` argument before appending the trailing `%`, and add `ESCAPE '!'` to the query.
  - Add regression test `list_ref_certificates_by_prefix_treats_wildcards_literally` covering plain, `%`, `_`, `!`, and `\` prefixes.
  - Add regression test `list_ref_certificates_by_prefix_parses_under_legacy_string_mode`, which runs the production query through a second single-connection pool that requests `standard_conforming_strings=off` at connection startup, so a future return to a parser-mode-dependent SQL literal is caught rather than only wildcard handling under the default configuration.
  - Extract the shared seeding into `seed_prefix_certs`, over ids covering both wildcards, the escape character, and a backslash.

## Test plan

Run against Postgres 16 on Linux (the crate does not compile on Windows):

- `cargo fmt --all -- --check` (pass)
- `RUSTFLAGS="-Dwarnings" cargo clippy --locked -p gitlawb-node --all-targets` (pass)
- `cargo test --locked -p gitlawb-node ref_certificate_tests::` (pass, 8 tests)

Mutation check that the new test is load-bearing: reverting only the production line back to the escaping loop over `\` plus `ESCAPE '\'` fails `list_ref_certificates_by_prefix_parses_under_legacy_string_mode` with `unterminated quoted string at or near "'\' ORDER BY issued_at DESC LIMIT $3"`, raised on the plain `cert` prefix that contains no metacharacter. Under the same mutation `list_ref_certificates_by_prefix_treats_wildcards_literally` stays green, so the wildcard test alone does not cover this.

Behavior was also confirmed at the SQL level with `standard_conforming_strings=off`: `ESCAPE '\'` fails to parse, while `ESCAPE '!'` parses and matches literally for the `%`, `_`, `!`, and `\` cases, including with the pattern supplied as a bound parameter.

## Prior reviewer feedback addressed

- [x] [P3] Make the escape literal independent of `standard_conforming_strings` ([review](https://github.com/Gitlawb/node/pull/320#pullrequestreview-4909100150)). Switched the escape character to `!` so the SQL holds no backslash, and added coverage that executes the production query over a connection configured with `standard_conforming_strings=off`.

### DIFF
--- a/crates/gitlawb-node/src/db/mod.rs
+++ b/crates/gitlawb-node/src/db/mod.rs
@@ -2107,10 +2107,17 @@ impl Db {
     ) -> Result<Vec<RefCertificate>> {
         let limit = limit.max(1);
 
+        // `!` is the LIKE escape character rather than the backslash default: a
+        // backslash in the SQL text would be an escape inside the string literal
+        // when the session runs with the legacy `standard_conforming_strings=off`,
+        // leaving `ESCAPE '\'` unterminated and failing every prefix lookup. The
+        // pool is opened from externally supplied connection settings, so that
+        // mode can arrive from a database- or role-level setting. `!` keeps the
+        // statement free of backslashes and parses the same either way.
         let mut escaped_prefix = String::with_capacity(prefix.len() + 4);
         for c in prefix.chars() {
-            if c == '%' || c == '_' || c == '\\' {
-                escaped_prefix.push('\\');
+            if c == '%' || c == '_' || c == '!' {
+                escaped_prefix.push('!');
             }
             escaped_prefix.push(c);
         }
@@ -2118,7 +2125,7 @@ impl Db {
 
         let rows = sqlx::query(
             "SELECT id, repo_id, ref_name, old_sha, new_sha, pusher_did, node_did, signature, issued_at
-             FROM ref_certificates WHERE repo_id = $1 AND id LIKE $2 ESCAPE '\\' ORDER BY issued_at DESC LIMIT $3",
+             FROM ref_certificates WHERE repo_id = $1 AND id LIKE $2 ESCAPE '!' ORDER BY issued_at DESC LIMIT $3",
         )
         .bind(repo_id)
         .bind(&pattern)
@@ -5602,6 +5609,7 @@ mod ref_update_keyset_same_timestamp_tests {
 mod ref_certificate_tests {
     use super::{Db, RefCertificate, RepoRecord};
     use chrono::Utc;
+    use sqlx::postgres::PgPoolOptions;
     use sqlx::PgPool;
 
     async fn db(pool: PgPool) -> Db {
@@ -5821,9 +5829,13 @@ mod ref_certificate_tests {
         assert!(certs.is_empty());
     }
 
-    #[sqlx::test]
-    async fn list_ref_certificates_by_prefix_treats_wildcards_literally(pool: PgPool) {
-        let db = db(pool).await;
+    /// Certificate ids covering every character the prefix search has to treat
+    /// literally: the two LIKE wildcards, the escape character itself, and a
+    /// backslash (the default escape, which must stay an ordinary character).
+    const PREFIX_CERT_IDS: [&str; 5] = ["cert-a", "cert%b", "cert_c", "cert\\d", "cert!e"];
+
+    /// Seed one repo holding `PREFIX_CERT_IDS` and return the repo id.
+    async fn seed_prefix_certs(db: &Db) -> String {
         let repo_id = uuid::Uuid::new_v4().to_string();
 
         db.create_repo(&RepoRecord {
@@ -5842,7 +5854,7 @@ mod ref_certificate_tests {
         .await
         .unwrap();
 
-        for (i, id) in ["cert-a", "cert%b", "cert_c", "cert\\d"].iter().enumerate() {
+        for (i, id) in PREFIX_CERT_IDS.iter().enumerate() {
             db.insert_ref_certificate(&make_cert(
                 id,
                 &repo_id,
@@ -5855,15 +5867,23 @@ mod ref_certificate_tests {
             .unwrap();
         }
 
+        repo_id
+    }
+
+    #[sqlx::test]
+    async fn list_ref_certificates_by_prefix_treats_wildcards_literally(pool: PgPool) {
+        let db = db(pool).await;
+        let repo_id = seed_prefix_certs(&db).await;
+
         // A plain prefix still matches every id that starts with it.
         let all = db
             .list_ref_certificates_by_prefix(&repo_id, "cert", 10)
             .await
             .unwrap();
-        assert_eq!(all.len(), 4, "plain prefix matches all four certs");
+        assert_eq!(all.len(), 5, "plain prefix matches all five certs");
 
         // `%` in the prefix must match a literal percent sign, not act as a
-        // wildcard that would also return "cert-a", "cert_c", "cert\d".
+        // wildcard that would also return every other seeded id.
         let pct = db
             .list_ref_certificates_by_prefix(&repo_id, "cert%", 10)
             .await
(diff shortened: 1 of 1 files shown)


euxaristia:
@coderabbitai full review

beardthelion (approved):
The escaping is right and the regression test is load-bearing. I gutted it three ways in a scratch worktree: reverting `format!("{}%", escaped_prefix)` to the raw prefix turns the test RED on the percent case (4 rows returned, 1 expected), and dropping `_` or `\` from the escape set individually turns it RED on those arms, which the percent assertion otherwise masks. The pristine head passes.

I also drove it over HTTP, since the test stops at the db layer. `?prefix=cert%25` reaches the handler as a literal percent and returns only the one literal match; with the escaping reverted the same request returns every cert in the repo. That path is open to an unauthenticated caller: the route is registered in the read-routes group at `server.rs:378` under `optional_signature`, which passes an unsigned request straight through, so `caller` is `None` and a public repo still grants the read. An anonymous read of a private repo's certs with the same prefix returns 404, so the prefix branch stays behind `authorize_repo_read`.

Every `LIKE` in the crate is swept: the other ten are constant patterns, so `list_ref_certificates_by_prefix` was the only site building a pattern from caller input, and there is no `ILIKE` or `SIMILAR TO`. Removing the `ESCAPE '\'` clause leaves the test green, since backslash is already the Postgres default, so that clause pins the default rather than changing behavior. Keeping it explicit here is the right call.

Not blocking, worth doing sometime: `list_certs_prefix_resolves_deep_cert` is the natural home for a metacharacter case at the route level.

One note: CodeRabbit hit its review limit and never ran, so this rests on my pass.

euxaristia:
@coderabbitai full review

jatmn (changes requested):

## Findings

- [P3] Make the escape literal independent of `standard_conforming_strings`
  `crates/gitlawb-node/src/db/mod.rs:2121`
  The escaping loop correctly turns `%`, `_`, and `\` into literal LIKE characters, but the SQL clause expresses the escape character as an ordinary string literal: the Rust source emits SQL `ESCAPE '\'`. PostgreSQL changes the meaning of backslashes in ordinary string literals according to the session's `standard_conforming_strings` setting. With its non-default legacy value `off`, the backslash escapes the following quote, leaving this statement unterminated. The failure occurs before PostgreSQL evaluates the bound parameters, so every request using `?prefix=...` returns a database/server error—including prefixes containing no metacharacters—and the CLI can no longer resolve any short certificate ID through this path.

  This is introduced by the PR rather than inherited behavior: the base query contains no backslash SQL literal and still parses in the same session configuration. I reproduced the delta by running the new focused SQLx test normally, where it passes, and then with `PGOPTIONS='-c standard_conforming_strings=off'`, where the first prefix lookup fails with `unterminated quoted string at or near "'\' ORDER BY issued_at DESC LIMIT $3"`. The repository accepts externally configured PostgreSQL connections and does not establish `standard_conforming_strings=on` when opening the pool, so a database- or role-level setting reaches this query unchanged.

  Please address the root cause by making the SQL representation of the escape character independent of the session parser mode. One option is to emit the explicit PostgreSQL escape string `ESCAPE E'\\'` (which requires the corresponding additional escaping in the Rust literal). Another is to choose a non-backslash escape character such as `!`, escape `!`, `%`, and `_` in the bound pattern, and use `ESCAPE '!'`. The latter avoids crossing two layers of backslash escaping. Please also add regression coverage that executes the production query through a connection configured with `standard_conforming_strings=off`—for example, a dedicated single-connection test pool/session—so the test detects a future return to a parser-mode-dependent SQL literal rather than only checking LIKE matching under PostgreSQL's default configuration.

euxaristia:
Confirmed and fixed in 1d3987b.

I reproduced the parse failure before changing anything. Against Postgres 16 with `standard_conforming_strings = off`, `SELECT ... WHERE id LIKE $1 ESCAPE '\'` ends the statement at the escaped quote and the literal runs on into whatever follows, so the query errors out rather than returning wrong rows. As you said, it fails at parse time, so it takes down prefixes containing no metacharacter at all.

I took the second option you offered: `!` as the escape character, escaping `%`, `_`, and `!` in the bound pattern, with `ESCAPE '!'` in the query. That leaves no backslash anywhere in the statement, so there is no parser-mode dependence to reason about and no second layer of backslash escaping. A backslash in the prefix becomes an ordinary character under that escape, which is the literal match this search wants, so the existing backslash case still asserts the same thing.

For the regression coverage: `list_ref_certificates_by_prefix_parses_under_legacy_string_mode` opens a second single-connection pool from the test pool's own connect options with `standard_conforming_strings=off` requested as a startup option, asserts the session really is in that mode, and then runs the production method through it. It checks a plain `cert` prefix first, since that is the case a parser-mode-dependent literal breaks and a LIKE-semantics test never would.

The new test is load-bearing. Reverting only the production line back to escaping `\` with `ESCAPE '\'`, and leaving both tests untouched, fails it with exactly the error you quoted:

```
unterminated quoted string at or near "'\' ORDER BY issued_at DESC LIMIT $3"
```

raised on the plain `cert` lookup. Under that same mutation `list_ref_certificates_by_prefix_treats_wildcards_literally` stays green, which confirms the original test could not have caught this.

The wildcard test now also covers a literal `!`, since the escape character has to be escaped in caller input. Full run on Linux against Postgres 16: `cargo fmt --all -- --check`, `RUSTFLAGS="-Dwarnings" cargo clippy --locked -p gitlawb-node --all-targets`, and `cargo test --locked -p gitlawb-node ref_certificate_tests::` all pass, 8 tests.

jatmn (approved):
LGTM

beardthelion (approved):
The parser-mode fix is right and both regression tests are load-bearing. I checked that by gutting them rather than reading them: reverting the SQL to `ESCAPE '\'` turns `list_ref_certificates_by_prefix_parses_under_legacy_string_mode` red on the unterminated-quoted-string parse error, and dropping `!` from the escape loop turns `list_ref_certificates_by_prefix_treats_wildcards_literally` red on the literal-bang assertion. Both green on the pristine head, tree clean afterwards.

Choosing `!` over the backslash default is the right call, and the comment at `crates/gitlawb-node/src/db/mod.rs:2110` explains the reasoning without overstating it. Backslash correctly stops being escaped once it is no longer the escape character, and the `cert\` case pins that behavior rather than leaving it implied.

Scope is contained. `list_ref_certificates_by_prefix` has one caller at `crates/gitlawb-node/src/api/certs.rs:37`, and it is the only LIKE in the crate whose pattern is built from caller input; the rest are constant `did:key:%` literals. The prefix branch still sits behind `authorize_repo_read` and passes a constant limit, so nothing here widens a match past the authorized repo.

euxaristia:
@kevincodex1

kevincodex1 (approved):

PR #321, opened 2026-08-11
fix(node): repair source-less legacy pins by bounded additive discovery

beardthelion:
Answers the third P1 from the #173 review round: the legacy CID sweep could not repair pre-provenance pins.

**Stacked on #173, deliberately.** The base is `fix/issue-135-ipfs-cid-tree-gate`, not `main`. Everything this touches (`sweep_pass`, `repair_legacy_provider_cid`, `pin_repair_cursor`, `record_pin_source`, `pin_sources_for_oid`) is introduced by #173 and does not exist on main, so this cannot branch independently. It is split out rather than folded in because it adds a new mechanism to startup behavior, where the other two findings were clamps on existing awaits, and reviewing it properly turned up enough to deserve its own round.

## The defect

The sweep resolved each legacy row's bytes only through `pin_sources_for_oid`. A row predating provenance has `repo_id` NULL and no `pin_repo_sources` entry, so the source set is empty, the `for repo_id in sources` body never runs, and the cursor advances past it. `list_pinned_cids` also filters those rows out on `is_raw_cidv1`, so they were both unadvertised and unrepairable. The sweep exists precisely for pre-provenance rows, and that was the one shape it could not handle.

## The fix, and the design decision worth reviewing

Discovery probes a bounded set of warm local repos for the object and, on a hit, repairs the key through the existing `repair_legacy_provider_cid`.

What it records is the part to look at. It calls `record_pin_source` additively and then `mark_pin_sources_incomplete`. It does **not** call `backfill_pin_provenance`, and `repo_id` stays NULL.

The tempting version writes the discovered repo as the first-pinner. That is unsafe. Reading identical bytes proves a repo holds the object, not that it is the source, and objects collide across repos routinely: forks, a vendored file, a shared LICENSE blob, the empty tree. Because `backfill_pin_provenance` guards on `AND repo_id IS NULL`, the first claim written is permanent, so probe order would decide which repo's visibility rules gate a public read surface with no correction path. It would also switch off the resolver's own fallback, since `needs_scan` is `sources.is_empty() || at_cap || incomplete` and a non-empty below-cap unmarked set reads as complete. The additive record plus the marker keeps that fallback alive and makes nothing permanent. There is a mutation that adds the exclusive claim back and must redden the no-claim assertion, so the decision is guarded rather than just described.

## Properties worth checking in review

- **Quarantine.** `list_all_repos` has no quarantine or visibility filter, while the resolver's own legacy scan loads `list_quarantined_repos` and gates on it, so discovery loads the same set and drops those ids. Private non-quarantined repos stay: an additive source record does not widen what the resolver serves, since every source is gated independently.
- **No remote fetch.** Nothing goes through `repo_store.acquire`; a repair pass must not become a bulk restore. Asserted from both sides: the cold candidate's path still does not exist after two runs, with a control proving the sweep declined an available copy, plus a source scan that a planted `acquire` turns red.
- **A cold or quarantined candidate does not mark the row retryable.** The candidate list is every repo on the node rather than the row's holders, and the sweep rewinds its cursor whenever any retryable skip occurred. Counting a cold candidate would make every source-less row retryable on every run on a cold-storage node, so the cursor would rewind forever and the sweep would never drain.
- **Cap exhaustion is retryable, and candidates are ordered oldest-first by `(created_at, id)`.** `repo_id` derives from the owner DID, which anyone can grind, so a first-N-wins cap over an id-sorted list would let someone bury the true holder permanently behind cheap registrations.
- **The probe cap bounds expensive work, not candidates considered.** A candidate dropped cold, quarantined, or unsafe costs no probe.

## Known residual, not introduced here

`pin_sources_incomplete` is one boolean per object rather than per `(object, repo)`, so a genuine pusher whose `record_pin_source` commits strictly after discovery's mark clears it, leaving a non-empty below-cap unmarked set that tells `needs_scan` to stop scanning. No ordering inside this sweep prevents it, and `record_pin_source`'s own doc already states the same residual for its existing call sites. It is bounded in practice, since the clearing writer really holds the bytes so every source stays servable, and for these rows it cannot be a regression, because before discovery ran they were keyed on a provider CID and were neither served nor advertised. A test pins it as known. Closing it properly needs a per-`(oid, repo)` marker, which is a change to the marker's shape and belongs on its own.

## Verification

Fifteen regressions driving the sweep entry points. The repair, cheap second pass, probe cap and multi-holder cases were observed failing first. The must-not cases carry mutations instead, since a guard against something happening has no honest pre-fix red: quarantined holder never serves, a cold candidate is neither fetched nor materialized and does not rewind the cursor, an unsafe candidate name is dropped without aborting the pass while a later safe candidate still repairs, a raw-CIDv1 row never enters discovery, and a provenanced row still resolves through the source loop alone.

Every guard carries a mutation that reverts it and requires the failure to match a named message, so a red is attributed to the property rather than to removing the seam the test injects at. Three of those target the design itself: adding `backfill_pin_provenance` beside the additive record, neutralizing the incomplete marker, and swapping the record and mark order. The suite passes with fmt, clippy, and `metadata --locked` clean; CI on this head is the check to read.

Reasoned-not-run and stated rather than folded in: a fetch reached transitively through `git::store` or `db`, and git's promisor lazy-fetch on a partial clone. Neither shape exists in the tree today.

### DIFF
--- a/crates/gitlawb-node/src/ipfs_pin.rs
+++ b/crates/gitlawb-node/src/ipfs_pin.rs
@@ -11,6 +11,22 @@ use anyhow::Result;
 use gitlawb_core::cid::Cid;
 use std::time::{Duration, Instant};
 
+/// How much READ work one source-less legacy row may cost a boot-time sweep.
+///
+/// The contract this number encodes: discovery for one pre-provenance row is allowed
+/// at most this many bounded object reads from warm local repos, whatever the node's
+/// repo count. The unit counted is the expensive one, a `git cat-file` pair against a
+/// candidate repo; a candidate rejected at filter time (quarantined, cold, an unsafe
+/// path) costs nothing against it. Without the cap the sweep is the O(repos x objects)
+/// fan-out this subsystem's cost rule exists to forbid, paid at boot on the node with
+/// the most history.
+///
+/// It happens to equal the resolver's serve-time per-request source cap
+/// (`db::MAX_PIN_SOURCES`), but it is a different bound with a different owner: that
+/// one bounds how many sources ONE `/ipfs` request may gate, this one bounds how many
+/// repos ONE background row may read. If either moves, the other does not follow.
+pub(crate) const MAX_LEGACY_DISCOVERY_PROBES: usize = 16;
+
 /// Attempts (including the first) for a transient DB-record retry.
 const PIN_RECORD_ATTEMPTS: u32 = 3;
 /// Backoff between DB-record retry attempts.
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):

## Findings

- [P1] Make capped discovery eventually examine every eligible holder
  `crates/gitlawb-node/src/ipfs_pin.rs:442`
  When a node has more than 16 warm, non-quarantined repositories, this always probes the same oldest 16.  The `Retryable` result only rewinds the pinned-CID cursor, so the next run rebuilds the same ordering and repeats the same prefix.  A source-less legacy object held only by repository 17 (or later) is therefore never repaired or advertised, while every boot repeats the work indefinitely.  Persist/derive a bounded rotating candidate continuation so retries can reach later candidates without losing the per-run work cap.

- [P1] Do not let one timed-out row exhaust discovery for the whole pass
  `crates/gitlawb-node/src/ipfs_pin.rs:379`
  `DiscoveryCtx` gives every source-less row in a pass one shared deadline.  If the first row spends that budget on a slow or hung candidate, all later rows call `repair_legacy_provider_cid` with an expired deadline, become retryable without a meaningful probe, and trigger a rewind.  Stable CID ordering then lets the same first row consume the budget on every boot, permanently starving later, otherwise repairable rows.  Allocate a bounded deadline per row (or persist fair resume state) instead of sharing it across the pass.

- [P1] Keep the incomplete marker durable when discovery adds its source
  `crates/gitlawb-node/src/ipfs_pin.rs:445`
  `record_pin_source` and `mark_pin_sources_incomplete` are separate best-effort writes.  If the source insert succeeds but the marker update fails, the repaired row is raw CIDv1 with a nonempty, below-cap, unmarked source set; later sweeps skip it, and `/ipfs` treats that set as complete and suppresses its fallback scan.  Thus, after discovery selected a private holder, an unrecorded public duplicate can remain permanently unavailable once the transient DB error clears.  Commit the source and scan-required state atomically, or retain a durable retry/scan trigger when the marker cannot be written.

beardthelion:
Closing this in favour of #173, which now carries the work.

Stacking it separately was my call and it was the wrong one. The argument I made was that #173 had to land first and this immediately after with no release between them, which only holds if nothing ever ships in that gap. That is not a property I can guarantee, and jatmn's point stands: the resolver change in #173 withholds the provider CID for an unrepaired row, so shipping it without the repair leaves an upgraded node's existing pins unavailable. They belong in one change.

What landed is ahead of this branch's pushed head. This PR points at `35dac128`; the folded tip is `4c665fbc`, which adds the round-12 fixes rebased onto #173. So #173 supersedes this rather than duplicating it, and the five commits appear there unmodified.

The three findings from the review round here are fixed in #173 rather than carried over:

The capped discovery that always probed the same oldest sixteen is now a persisted keyset continuation advanced once per completed table traversal, over probes that started with live budget only, persisted from inside the pass so a shutdown cannot swallow it. The sweep also re-arms on a timer instead of running once per boot, which is what makes the reach bound wall-clock rather than a count of restarts.

The shared per-pass deadline that let one hung candidate starve every later row is now a per-row slice of the pass budget, and a row reached after the budget is spent is skipped free rather than charged for a probe it cannot make.

The source insert and the fallback-arming marker now commit in one transaction, so the nonempty-below-cap-unmarked state that suppressed the resolver's fallback is unreachable from discovery.

Each was verified by reverting the exact line it depends on and watching the named assertion go red, including present-but-wrong variants where the fix adds a condition rather than removing one.

Discussion continues on #173.

Issue #96, opened 2026-06-24
sync_queue: enqueue_sync ON CONFLICT DO NOTHING never dedupes; per-ref notify enqueues N redundant full fetches


beardthelion:
## What

`enqueue_sync` (`crates/gitlawb-node/src/db/mod.rs:1159`) inserts into `sync_queue` with a fresh `Uuid` for `id` and a bare `ON CONFLICT DO NOTHING`:

```rust
INSERT INTO sync_queue (id, repo, node_did, ref_name, new_sha, cid, status, enqueued_at)
VALUES ($1, $2, $3, $4, $5, $6, 'pending', $7)
ON CONFLICT DO NOTHING
```

`sync_queue` (`db/mod.rs:503`) has exactly one unique constraint — `id TEXT PRIMARY KEY` — and `id` is a fresh `Uuid::new_v4()` on every call. A bare `ON CONFLICT DO NOTHING` only fires on a unique violation, so the clause can never fire: it does nothing. There is no `UNIQUE(repo, node_did, ref_name, new_sha)`, so identical pending rows are never collapsed.

## Consequence

The 30s worker `process_batch` (`crates/gitlawb-node/src/sync.rs:148`) dequeues up to 10 pending rows and processes each independently, running a full refspec-less `git fetch --prune origin` (`sync.rs:528`) per row — no per-repo coalescing.

So duplicate pending rows for the same repo each trigger a full-repo fetch, even though the first fetch already pulls every ref.

This was latent before but is amplified by the per-ref sync-notify fan-out (#72): a push of N refs to P peers now sends N notifies per peer, each calling `enqueue_sync`, so a receiver gets up to N pending rows per push and runs N full fetches where one would do. With the 10-rows-per-30s batch, a large multi-ref push can also starve the queue and delay other repos' syncs. Owner-gated (only the repo owner can push), so the amplification ceiling is bounded, but a crafted high-ref-count push is the worst case.

Not a correctness bug — sync stays eventually consistent — but real redundant network/CPU and queue pressure.

## Fix options

1. Coalesce in the worker: dequeue distinct repos (or dedupe pending rows by `(repo, node_did)`) and fetch once per repo per batch, regardless of how many per-ref rows are pending. Cleanest, since even distinct refs of one repo only need one whole-repo fetch.
[...]

PR #325, opened 2026-08-12
fix(node): authenticate gossip ref-update events before writing them

### LINKED ISSUE #323: Gossip ref-update ingest accepts events from any peer without authenticating the sender
The gossipsub `Message` arm in `p2p::start` deserializes a `RefUpdateEvent` and writes it, with no check tying the sender to the `node_did` the payload claims. The HTTP route carrying the same payload does considerably more: `notify_sync` (`api/peers.rs`) rejects when the signature keyid does not match `req.node_did`, rejects unknown peer DIDs, and validates the repo slug, and `server.rs` puts a per-IP brake on that route specifically because it reaches `enqueue_sync`. The p2p path has none of the four.

`ValidationMode::Permissive` does not help: it verifies `signature`, `seqno`, and `from` only when present, so omitting them skips verification. Switching to `Strict` would not fix this either, because a gossipsub signature binds a libp2p `PeerId`, never the `node_did` inside the JSON payload.

Two consequences. Forged push events reach the public REST and GraphQL feeds with attacker-chosen `pusher_did`, ref name, SHAs, and `cert_id`. And on a node with `GITLAWB_AUTO_SYNC=true` (all three Fly nodes set it), naming any publicly known peer DID makes `resolve_origin_url` succeed, after which the sync worker clones an attacker-chosen slug from that peer and registers a mirror.

Existing defenses that limit it, and are worth crediting: `owner_did` is re-derived rather than echoed, private and quarantined local repos are gated out of the feeds, and the sync worker re-validates the slug so traversal is not reachable.

Related: #96 covers the vacuous `ON CONFLICT DO NOTHING` on `enqueue_sync`. #82, #270, and #272 fixed this class on the HTTP side.

beardthelion:
Closes #323. Ref-update events are now signed on the way out and authenticated on the way in, with the same gates the HTTP route already applies.

The signature is a flat optional field on `RefUpdateEvent`, not a nested envelope. A nested envelope would have been a flag-day break: an un-upgraded peer running `from_slice::<RefUpdateEvent>` finds no `node_did` or `repo` at top level and drops every event from an upgraded node. The flat optional field follows the `owner_did` precedent from #144, so old peers keep parsing. A golden test pins the signing input, because a future field reorder changes what every peer signs and no round-trip test can see that.

Ingest runs: a loose pre-parse bound on the forwarding peer, parse, the DID-method gate, signature policy, the known-peer check, slug validation, a tighter bound on the authenticated author, then the writes. Rejection writes nothing to either sink, and the tests assert those separately.

The two bounds are split deliberately. A forwarding peer id is free to mint and gossipsub re-shares mesh-wide, so a single budget charged to it is both evadable by rotation and usable against a victim: a flood routed through an honest neighbour would deny that neighbour on every receiver. The loose bound caps CPU before anything is parsed, which is all that identity can honestly buy; the tight bound charges the durable writes to the authenticated author. Both maps are capped, and both rejections log the budget and window that refused them.

`GITLAWB_REQUIRE_SIGNED_PEER_WRITES` governs both transports now, which widens its meaning from an HTTP route group to a cross-transport policy. With it off (the default, and what every config sets today), unsigned events from known peers are accepted with a warning, so the fleet upgrades before enforcement turns on.

## What this does not close

Authentication is not authorization. `upsert_peer` admits an unproven announce for an unseen `did:key`, so a self-registered peer passes these gates in either flag state. The signature stops impersonation of an existing peer; it does not make the peer list a membership boundary.

Replay of a captured signed event is not bound. Gossip is a public broadcast, so capturing one is free, and neither sink dedupes (#96).

A push above the author bound loses its tail, loudly rather than silently. Gossip publishes one event per ref while `sync/notify` batches a push, so the two transports still disagree about events per push.

The swarm loop's wiring is not covered by tests: every ingest test calls the function directly, so mutating the loop to hardcode the flag leaves the suite green.

## Summary by CodeRabbit

* **New Features**
  * Added cryptographic signing and verification for gossip ref-update events.
  * Added configurable enforcement for signed gossip payloads.
  * Added authentication checks for known peers and repository identifiers.
  * Added separate rate limits for incoming and authenticated gossip events.
  * Added monitoring metrics for accepted, rejected, failed, and rate-limited gossip events.

* **Bug Fixes**
  * Improved peer creation and update detection.
  * Gossip write failures are now reported more clearly.
  * Maintained compatibility with unsigned events when signature enforcement is disabled.

* **Documentation**
  * Expanded configuration and rollout guidance for signed peer writes.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -88,8 +88,12 @@ GITLAWB_P2P_PORT=7546
 # GITLAWB_BOOTSTRAP_DISABLE_SEEDS=true for isolated dev/test networks.
 GITLAWB_BOOTSTRAP_PEERS=
 GITLAWB_BOOTSTRAP_DISABLE_SEEDS=false
-# Require signatures on peer announce/sync write routes. Keep false during
-# rolling upgrades so existing live nodes can still communicate.
+# Require a peer write to prove its DID on every transport: RFC 9421 signatures
+# on the announce/sync routes, and an Ed25519 payload signature on gossip
+# ref-update events. Keep false during rolling upgrades so existing live nodes
+# can still communicate. Upgrade every gossip-publishing peer to a build that
+# signs events before setting this true, or their updates are dropped on
+# arrival with no error visible on their side.
 GITLAWB_REQUIRE_SIGNED_PEER_WRITES=false
 
 # Require the authenticated pusher to be the repo owner on git-receive-pack.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2630,7 +2630,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ab67060fc6b8ef687992d439ca0fa36e7ed17e9a0b16b25b601e8757df720de"
 dependencies = [
  "data-encoding",
- "syn 2.0.117",
+ "syn 1.0.109",
 ]
 
 [[package]]
@@ -3002,7 +3002,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -3542,9 +3542,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.4.13"
+version = "0.4.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2f44da3a8150a6703ed5d34e164b875fd14c2cdab9af1252a9a1020bde2bdc54"
+checksum = "ef8e5e5a340588f4452631496976cf8636d4a7ecf600239fdc27615d2530bc16"
 dependencies = [
  "atomic-waker",
  "bytes",
@@ -3863,7 +3863,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "tokio",
  "tower-service",
  "tracing",
@@ -5431,7 +5431,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash",
  "rustls",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
@@ -5468,7 +5468,7 @@ dependencies = [
  "cfg_aliases",
  "libc",
  "once_cell",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "tracing",
  "windows-sys 0.52.0",
 ]
@@ -5900,7 +5900,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -6794,7 +6794,7 @@ dependencies = [
  "getrandom 0.4.2",
  "once_cell",
  "rustix",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
--- a/README.md
+++ b/README.md
@@ -322,6 +322,10 @@ GITLAWB_REQUIRE_SIGNED_PEER_WRITES=true
 
 `POST /api/v1/sync/trigger` is not part of the staged rollout: it always requires a signature in both config modes and returns 401 without one, because each call drives an O(peers) outbound fan-out.
 
+The flag is not HTTP-only. It also gates inbound gossip ref-update events on the libp2p mesh, which carry an Ed25519 payload signature rather than RFC 9421 headers because there are no HTTP headers to sign. Set to `true`, an unsigned gossip event is dropped. Set to `false`, an unsigned event is accepted with a warning during the rolling-upgrade window, and the write budget it consumes is charged to the forwarding peer rather than to the DID it claims, since an unsigned event's claimed DID is asserted and not proven. A present-but-invalid signature is refused in both modes; that is forgery, not an un-upgraded peer.
+
+Because the flag now spans both transports, the rollout order matters in one direction: **upgrade every gossip-publishing peer to a build that signs events before you set this to `true`.** Enabling it while an old publisher is still live drops that publisher's ref-updates on arrival, and the publisher sees no error, because gossip has no response to carry one. Upgrading the HTTP peers alone is not sufficient.
+
 ---
 
 ## Configuration
@@ -345,7 +349,7 @@ Important node settings:
 | `GITLAWB_BOOTSTRAP_PEERS` | Comma-separated HTTP peer URLs. |
 | `GITLAWB_P2P_BOOTSTRAP` | Comma-separated libp2p multiaddrs. |
 | `GITLAWB_BOOTSTRAP_DISABLE_SEEDS` | Disable embedded seed peers for isolated dev/test networks. |
-| `GITLAWB_REQUIRE_SIGNED_PEER_WRITES` | Require signed peer announce/sync writes. |
+| `GITLAWB_REQUIRE_SIGNED_PEER_WRITES` | Require peer writes to prove their DID on every transport: RFC 9421 signatures on the announce/sync routes, and an Ed25519 payload signature on gossip ref-update events. Upgrade every gossip publisher before enabling, or their updates are dropped with no error on their side. |
 | `GITLAWB_AUTO_SYNC` | Enable automatic sync from known peers. |
 | `GITLAWB_MAX_PACK_BYTES` | Max git pack body size for smart-HTTP routes. |
 | `GITLAWB_GIT_SERVICE_TIMEOUT_SECS` | Max seconds a served git upload-pack, receive-pack, or `info/refs` advertisement may run before it is aborted (504). Default 600. Also bounds the withheld-blob classification walk (on both the upload-pack serve and receive-pack replication paths) and the push-side pin-candidate discovery (`rev-list` / `cat-file`), each reaped via process-group teardown at the deadline. On the path-scoped upload-pack path the classification walk and the pack serve share ONE deadline, so this value bounds their combined duration rather than granting each stage a full budget: a walk that consumes it leaves the serve nothing and the clone gets a 504. Serving large path-scoped repos may therefore need a higher value than they did when each stage was budgeted separately. Accepted range is 1 to 3153600000 (100 years), since the node derives deadlines from this value and a larger one cannot be represented. |
--- a/README.md
+++ b/README.md
@@ -326,6 +326,8 @@ The flag is not HTTP-only. It also gates inbound gossip ref-update events on the
 
 Because the flag now spans both transports, the rollout order matters in one direction: **upgrade every gossip-publishing peer to a build that signs events before you set this to `true`.** Enabling it while an old publisher is still live drops that publisher's ref-updates on arrival, and the publisher sees no error, because gossip has no response to carry one. Upgrading the HTTP peers alone is not sufficient.
 
+There is a second precondition on gossip ingest that the flag does not control. A publisher's `node_did` must already exist in the receiving node's peers table, or the event is dropped as an unknown peer DID. This holds in both modes, signed and unsigned alike: a valid signature proves key possession, not membership. Rows reach that table over HTTP, either from `GITLAWB_BOOTSTRAP_PEERS` when this node contacts a bootstrap peer and records the DID it reports, or from a prior `POST /api/v1/peers/announce`. So a peer that only ever joined the libp2p mesh, with no HTTP announce and no bootstrap contact in either direction, will have its ref-updates dropped even with a good signature and the flag off. If gossip is silently not landing from a peer you can see on the mesh, check that its DID is in `GET /api/v1/peers` on the receiving side first.
+
 ---
 
 ## Configuration
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -2471,6 +2471,8 @@ async fn post_receive_replication_tail(
                             timestamp: chrono::Utc::now().to_rfc3339(),
                             cert_id: None,
                             cid: cid.map(|s| s.to_string()),
+                            // Signing is U4's job; unsigned for now.
+                            sig: None,
                         })
                         .await;
                     }
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -2472,7 +2472,14 @@ async fn post_receive_replication_tail(
                             timestamp: chrono::Utc::now().to_rfc3339(),
                             cert_id: None,
                             cid: cid.map(|s| s.to_string()),
-                            // Signing is U4's job; unsigned for now.
+                            // Left unsigned here on purpose. The swarm loop
+                            // signs the event with the node keypair via
+                            // `p2p::signed_publish_bytes` immediately before it
+                            // publishes, and skips the publish outright if
+                            // signing fails, so the wire always carries a
+                            // signature even though this construction site does
+                            // not. Setting one here would be signed over a
+                            // payload the publisher has not finished building.
                             sig: None,
                         })
                         .await;
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -2461,7 +2461,11 @@ async fn post_receive_replication_tail(
                 if announce {
                     if let Some(p2p) = &p2p_handle {
                         p2p.publish_ref_update(crate::p2p::RefUpdateEvent {
-                            v: 0,
+                            // Named, not literal: the ingest gate compares `v`
+                            // against this same constant, so a bump has to move
+                            // both ends together rather than leaving the emitter
+                            // on a version the gate no longer accepts.
+                            v: crate::p2p::CURRENT_REF_UPDATE_VERSION,
                             node_did: node_did_str.clone(),
                             pusher_did: pusher_did_clone.clone(),
                             repo: repo_slug.clone(),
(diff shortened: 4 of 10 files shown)


jatmn (changes requested):

## Findings

- [P2] Version the signed event format and retain a v0 verifier
  `crates/gitlawb-node/src/p2p/mod.rs:150`
  `signing_bytes` signs the current serialization of the entire `RefUpdateEvent`, but the payload contains no format version and `verify_ref_update` has only that one representation.  On the next signed-field addition or reorder, an upgraded publisher will sign bytes that an older receiver drops while reserializing without the unknown field.  This partitions gossip during the very rolling deployment the flag is intended to support.  The repository guidance requires signature-covered formats to be versioned and to retain verification for earlier artifacts; add a defaulted version and preserve verification of the current v0 byte sequence.

- [P1] Do not debit a claimed author for unsigned legacy traffic
  `crates/gitlawb-node/src/p2p/mod.rs:353`
  In the default flag-off mode, a mesh participant can submit 500 unsigned messages claiming any known `did:key`; the known-peer lookup confirms only that the string has a row, then the author limiter consumes the victim's bucket.  The victim's subsequent valid signed updates are rejected for the rest of the minute.  Invalid-slug messages make this denial cheap because the debit happens before slug validation, but valid forged messages have the same attribution flaw.  Charge unauthenticated compatibility traffic to its forwarding source (and validate structural fields before charging), reserving the author bucket for a verified signature.

- [P2] Update the public description of the widened enforcement flag
  `README.md:343`
  This change makes `GITLAWB_REQUIRE_SIGNED_PEER_WRITES` reject unsigned gossip events too, but the README still says it controls only peer announce/sync writes.  An operator can therefore enable it after upgrading the HTTP peers while old gossip publishers remain, producing the exact silent federation drop the new config comment warns about.  Document the cross-transport behavior and rollout requirement.

- [P3] Correct the now-stale publisher comment
  `crates/gitlawb-node/src/api/repos.rs:2474`
  The caller intentionally passes an unsigned event so `p2p::signed_publish_bytes` can sign it immediately before publication, but this comment says signing is a future task and the event remains unsigned.  That is misleading at a security-sensitive boundary and leaves the earlier reviewer request unresolved; describe the actual signing handoff.

beardthelion:
All four addressed, plus a few things the round turned up on the way.

**Version the signed event format.** `RefUpdateEvent` carries `v` now, skipped when zero, so a v0 event's wire bytes and signing bytes are byte-identical to what shipped before and every signature already issued still verifies. `GOLDEN_SIGNING_BYTES` is unchanged, which I treated as the constraint rather than something to re-pin. The compatibility test verifies a frozen artifact captured before the field existed and signs nothing itself, since a signature the test just produced is self-consistent by construction. Its provenance is a SHA-256 pin rather than a comment, because the obvious check (assert no version key) passes just as happily on an artifact regenerated from current code.

I also added a guard for an unknown version ahead of the signature match. Without it a v1 event was not merely mislabelled, it was admitted and its row landed, which I only noticed because the test I wrote to prove the "loud failure" claim went red for the wrong reason.

**Do not debit a claimed author for unsigned traffic.** The author budget is charged only when the signature verified. Unsigned compatibility traffic charges a separate bucket keyed on the forwarder at `GOSSIP_UNSIGNED_SOURCE_MAX_EVENTS` (1500), and the compile-time assertion is now a three-way chain so the relation is enforced rather than remembered. I sized that bucket above the per-author cap on purpose: a forwarder aggregates many authors, so sizing it *at* the author cap would have re-imposed the mesh-edge denial `a_junk_flood_down_one_edge_does_not_deny_a_valid_author_on_that_edge` exists to prevent.

On your parenthetical about validating structural fields first: I moved `validate_repo_slug` above `peer_exists` and above the signature match rather than just above the debit. Above the debit alone buys the fairness property but still spends a database round trip on a malformed event. There is a test-only call counter asserting `peer_exists` is never reached, because an outcome assertion cannot tell those two placements apart.

**Widen the flag's description.** README and `.env.example` now cover both transports and state the rollout ordering: every gossip publisher has to sign before the flag goes on, or their updates stop federating with no error on their side. I derived the wording from the ingest arms rather than from the config docstring, since two prose sources agreeing is not corroboration.

While there I documented a precondition this branch changed and I had not called out: a gossip publisher must also be a known peer on the receiving node, in both flag modes.

**Correct the publisher comment.** Done, and it now describes the actual handoff.

**Also fixed, found while working the four.** One signature verified against multiple wire encodings, so gossipsub's duplicate cache was bypassable: the frozen 454-byte artifact and the same artifact with `"v":0` injected are 460 bytes, both verify, identical signing bytes. Ingest outcomes reached only tracing and no metric, so gossip health was invisible on `/metrics`. The three ingest limiters were built as locals in the swarm task, so the periodic sweep could not reach them. Both frozen signing constants pinned an all-`Some` shape the publish site never emits, which I proved by adding `skip_serializing_if` to `cert_id` and watching the suite stay green while the signing input changed for every real event. `WriteFailed` was the one outcome no test constructed. And the unsigned-acceptance warning fired before two gates that could still drop the event.

**Two things I have not fixed here, deliberately.**

A validly signed event still replays. There is no freshness or uniqueness binding, and each replay writes a row and debits the claimed author, so a captured signature partially reopens the victim-selection problem your P1 is about. A follow-up branch implements a freshness window plus a seen-set keyed on canonical signing bytes; it is not in this PR, because with this merged and that not, an attacker is reduced to replaying events a legitimate node actually published, which is strictly less than forging arbitrary ones beforehand. Happy to fold it in instead if you would rather not have the gap.

Separately, and pre-existing: a registered peer can claim any repo slug. The sync worker resolves the origin URL from the claiming DID's peer row, so proving who sent an event does not bind them to authority over the repo it names. Not introduced here and not closed here.

Full suite, clippy and fmt all clean locally; CI has the authoritative run for this head.

jatmn (changes requested):

## Overall guidance

This PR is attempting a security-sensitive protocol change across the gossip wire format, cryptographic verification, rate limiting, persistence, metrics, configuration, documentation, and compatibility behavior. The remaining issues have a common cause: the implementation is being assembled and corrected incrementally on top of an out-of-date branch, while several security-relevant properties are validated at only one layer at a time. That can make local tests and individual fixes look convincing while the complete deployed contract is still inconsistent.

This guidance is intentionally limited to the PR's stated goal: authenticated gossip ref-update handling with a rolling-upgrade compatibility mode. It does **not** ask this PR to turn the open peer list into an authorization boundary, to bind a peer to authority over every repository it names, or to solve generic captured-message replay with freshness/seen-state. Those are explicitly documented follow-up limitations. The author also reports the alternate-encoding duplicate-cache bypass as fixed; this review treats that work as resolved rather than reopening it. The requested work is to preserve current-main protections and make the authentication and rollout-observability behavior that this PR *does* introduce internally consistent.

Before another review round, please treat this as one integration pass rather than a sequence of narrowly targeted responses:

1. Rebase the branch onto current `main` first, resolve every conflict by preserving current-main security/correctness fixes, and review the resulting *entire* base-to-head diff. Do not resolve a conflict by copying an older implementation or its test into the new tree merely because it makes the local branch pass. In this case, the stale resolution already reintroduced the fixed advisory-lock and certificate-prefix defects. A rebase is not administrative work here: it is part of the safety proof for a change that touches shared-Postgres writes and public query behavior.

2. Keep the event contract written as one end-to-end set of invariants, then make each invariant hold at every boundary. At minimum, record the signed wire version, the accepted legacy compatibility behavior, which identity each rate-limit bucket may charge, what is persisted/enqueued, and which metric outcome represents each admission state. The signer, verifier, rate limiter, database writer, and metric must use compatible meanings for the sender and the admission decision.

3. Separate three distinct concepts that are currently easy to conflate: authentication, authorization/membership, and delivery/persistence. A valid signature proves possession of the claimed key; a known-peer lookup establishes only the project’s current membership rule; and a successful database write says the node retained the event. Model those separately in names, outcomes, logs, metrics, and tests. In particular, an unsigned compatibility admission must not be reported as an authenticated acceptance merely because its writes succeeded.

4. Keep the new protocol tests load-bearing at the externally visible boundaries, not only as struct round trips. Cover signed and unsigned legacy messages, malformed signatures, unknown versions, known-peer rejection, invalid repository slugs, each rate-limit bucket, persistence failure, queue failure, and the metric/log outcome for each admission state. Preserve the author’s documented generic-replay limitation for its dedicated follow-up rather than expanding this PR to solve it.

5. Exercise the real production seams in addition to direct helper tests where practical. The production path is a swarm loop receiving configuration from `main` and publishing through a command channel, while most ingestion tests call the helper directly. Add focused wiring coverage or a small testable adapter for configuration propagation and sign-before-publish behavior, so a later refactor cannot silently diverge from the helper-level guarantees.

6. Keep compatibility policy explicit and bounded. Supporting legacy unsigned gossip during rollout is a product choice, but it needs a distinct observability state and a documented exit condition. Define what upgraded and un-upgraded peers send and accept, which configuration enables enforcement, how operators detect remaining legacy traffic, and when the compatibility path can be removed. Compatibility should not silently weaken the meaning of security metrics or make a peer’s update disappear without an operational signal.

7. For every finding fixed in this PR, search adjacent contracts before declaring the class closed: the HTTP peer-write twin, outbound publishing, database uniqueness/idempotency, sync queue behavior, rate-limit keying, dashboards, docs, and current-main conflict resolutions. This is especially important here because the author has already correctly identified generic replay and peer-to-repository authority as separate work. Those accepted limitations should remain explicit and tracked, rather than being silently folded into this change.

The practical completion criterion should be: a freshly rebased branch, preserved current-main protections, load-bearing tests for the documented authentication and rollout states, and a review of the new full diff. That approach should surface the remaining interaction bugs in one author pass without expanding this PR into the separately scoped replay- or authorization-design work.

## Findings

- [P1] Rebase onto current main before merging
  `crates/gitlawb-node/src/git/repo_store.rs:733`
  This head is based on `0e2328b`, while current `main` is `96d8123`; resolving the stale branch by taking its versions silently drops protections that have already merged. In particular, this restores `DefaultHasher` for the PostgreSQL advisory-lock key, whereas current main uses a fixed SHA-256 derivation. A current-main node and a node built from this head can therefore derive different keys for the same repository, acquire independent locks, and concurrently perform receive-pack or other repository writes against shared Postgres. The same stale resolution removes `db::list_ref_certificates_by_prefix`'s literal `LIKE` escaping, so a public `?prefix=%` or `_` is interpreted as a SQL wildcard rather than the requested certificate-ID prefix.

  The root cause is carrying an old implementation across a branch that predates both fixes, not either feature in isolation. Please rebase onto current main and resolve conflicts by retaining the stable SHA-256 key derivation, its golden/stability coverage, the `ESCAPE`-based literal-prefix query, and its wildcard/legacy-parser tests. Then request re-review of the resulting base-to-head diff; do not preserve the copied `DefaultHasher` test as a substitute for the production contract.

- [P3] Do not count unsigned compatibility writes as authenticated accepts
  `crates/gitlawb-node/src/p2p/mod.rs:427`
  With enforcement disabled, the `None` signature arm intentionally admits a known unsigned event, but the function subsequently returns `IngestOutcome::Accepted`. That variant is documented as "authenticated AND every write it implies landed" and maps to the `accepted` metric label, so dashboards combine unproven rolling-upgrade traffic with signature-verified admissions. An operator cannot tell whether the fleet is still relying on the compatibility allowance or whether authenticated gossip is healthy.

  The root cause is modeling persistence success and authentication state with one outcome. Add an explicit unsigned-admitted outcome and metric label, preserving `Accepted` for verified events only; propagate it through the swarm-loop logging and tests. If the intended contract is instead that `accepted` is transport-neutral, rename and document the outcome/metric consistently, but retain a separate authentication-state signal so the rollout can be monitored safely.

beardthelion:
Both blocking asks are done, and re-checking the signature path on the way turned up something I think matters more than either of them.

**Stale base.** Rebased onto current main, no conflicts. That brings back the three protections the old base was missing: the SHA-256 advisory lock key with its domain separator, the `ESCAPE '!'` prefix escaping in `list_ref_certificates_by_prefix`, and the distinct-signer certificate threshold fix from #332. I ran main's own golden and wildcard tests against the rebased tree rather than assuming the rebase preserved them.

**Unsigned admissions counted as authenticated.** Fixed, and you and CodeRabbit were both right that the label was the problem rather than the behavior. There is now a separate `IngestOutcome::UnsignedAdmitted` with its own `unsigned_admitted` metric label, so `accepted` means signature-verified and an operator can see whether the fleet still leans on the compatibility window. `Accepted` no longer covers unauthenticated traffic.

**A replay hole in the thing this PR exists to harden.** A signature proves the bytes were signed by `node_did` once. It does not prove they are fresh, and nothing on the ingest path consults `timestamp`. There is no nonce and no sequence number, and the stored row id is a fresh uuid, so the `ON CONFLICT(id) DO NOTHING` on `insert_ref_update` cannot collapse duplicates either.

That made one captured signed event a reusable proof of authorship. I reproduced it with enforcement on: replaying a single captured event from rotating peer ids up to the author cap, every copy verified, every copy landed its own row, every copy charged the author's budget, and then the real publisher's next genuine push came back `AuthorRateLimited`. Any node on the mesh can do this to any publisher whose event it has seen, and it defeats exactly the property the author brake claims a few lines below itself, that a named victim's budget is not spendable by anyone else. The suite already covered the unsigned version of this attack; the signed twin walked in through the path the signature was supposed to protect.

Fixed with a fourth ingest limiter keyed on the signature, budget of one per window, checked above the peer lookup and above the author budget so a rebroadcast costs neither a database round trip nor a debit against the captured author. The signature is the key rather than a hash of the payload because Ed25519 signatures are deterministic and every other field sits inside the signed bytes, so mutating anything to dodge the guard invalidates the signature that made the event worth replaying. Rebroadcasts return a distinct `ReplayDropped` outcome with a `replay_dropped` label and log at debug, since a duplicate is ordinary mesh traffic rather than something to warn on.

Two things worth flagging in what that change touched. The author-budget test used to exhaust the budget by resending one event's bytes, which is now the attack, so it publishes distinct signed events instead. And the sweep test caught the new limiter missing from the sweep after adding the field to the destructure had already satisfied the compiler, which is the sort of thing that would otherwise have shipped as an unreclaimed map.

**On the swarm-loop database await, I would rather split it than fold it in here.** It is real and I am not disputing it, but it is not this PR's doing: at the merge base the same `select!` arm already awaited `insert_ref_update` and `enqueue_sync` inline, so this PR relocated that code into `ingest_ref_update` rather than introducing the await-on-the-loop shape. What it does add to that region is one round trip, the `peer_exists` lookup. A bounded ingestion worker with explicit database deadlines is the right fix and it is an architecture change to code that predates this branch, on a change already carrying a live-exploit fix.

Related and worth doing in that same follow-up: the author budget is currently checked below `peer_exists`, so a signed author past its budget still pays a lookup before being shed. The brake belongs in front of the work it bounds. Happy to file both together, or to pull the ordering change forward into this PR if you would rather see it now.

The full suite passes in CI, fmt and clippy are clean, and the lockfile is in sync with the rebased base.

beardthelion:
Correction to my previous comment: the replay section does not belong on this PR, and I have reverted it.

Replaying a captured signed event was already known and is already fixed in #334, which is stacked on this branch and predates that comment. It carries the freshness window with a separate future-skew allowance, a replay key over the canonical signing bytes so two wire encodings of one signature collapse to a single slot, and a seen-set that releases a reserved slot when a write fails and fails open on saturation rather than reporting saturation as a replay. That is the better design and it is where the work belongs.

I pushed a duplicate guard onto this branch and have removed it. The head is back to c003f652, and #334 is rebased onto it, with one commit reconciling it against the unsigned-admission outcome split below.

What stands from that comment: the rebase and the three protections it restores, the separation of unsigned admissions from authenticated ones, and the note that the swarm-loop database await predates this branch rather than being introduced by it.

The brake-ordering point I raised at the end also needs no separate follow-up. #334 already places both of its gates above the peer lookup and above the author debit, for the same reason I gave.

beardthelion:
One more fix on this head, from re-checking the ingest path after the last push.

**The author brake sat behind the work it bounds.** It was checked after `db.peer_exists`, so a signed author already past its budget still paid a Postgres round trip before being refused. The lookup was being done for a request that was never going to be admitted, and a signed author can drive that at the pre-parse source rate.

It now answers first. What is worth saying is that the obvious version of this fix is worse than the problem: the charging call inserts a window for a key it has never seen, so probing with it there would let a flood of self-minted signed DIDs occupy the bounded author map before the peer lookup could refuse them, and legitimate new authors would be shed once it filled. The early answer is therefore a read that allocates nothing and reports an unseen key as within budget. The charge itself stays below the lookup, unmoved, so the only behaviour that changes is when an already-over-budget author is refused.

Two guards cover it, and both were proven by reverting them. The first observes the peer lookup not happening through a call counter rather than restating the ordering, and went red at one call versus zero before the change. The second asserts the author map is unchanged when an unregistered DID is refused, which is what breaks if the read is ever simplified back into the charge.

**The swarm-loop database await stays deferred, on the same basis as before.** The defect is visible at the merge base, where that `select!` arm already awaited `insert_ref_update` and `enqueue_sync` inline, so a bounded ingestion worker is an architecture change to code this branch relocated rather than introduced. Happy to file it, and happy to be argued out of the split.

Suite, fmt, clippy and the lockfile are all clean on this head.

jatmn (changes requested):

## Findings

- [P2] Rerun the failed CodeQL matrix before merging
  The `Analyze (swift)` CodeQL job is red on this head. Its setup failed after
  GitHub returned HTTP 429 while downloading `github/codeql-action`, so this is
  not evidence of a source-level Swift finding, but it leaves the required code
  scan incomplete. Please rerun the workflow and verify that the Swift matrix
  completes successfully before merge.

beardthelion:
Your round-3 ask was the CodeQL Swift matrix. That job failed in setup on `2eaabff3` (HTTP 429 fetching the action). I pushed an empty retrigger (`b7416c68`); run 32191230343 completed Initialize CodeQL, Autobuild, and Perform CodeQL Analysis successfully, and the CodeQL check is green on that head.

CI was still red on `cargo audit` because this branch's lockfile pinned h2 0.4.13 while main had already moved on. The new head bumps h2 to 0.4.19; audit is green locally and that commit is lockfile-only.

The round-2 rebase onto current main is still outstanding. The specific defects you named there (advisory-lock hashing, cert-prefix ESCAPE) are fixed on this branch's code; I have not rebased yet because #334 is stacked on this branch and a rebase would detach it until re-stacked. Happy to do that as a follow-up if you want it before merge.

jatmn (approved):
lgtm

beardthelion:
@kevincodex1 LGTM

PR #330, opened 2026-08-14
fix(node)!: enforce owner-only push by default

Vasanthdev2004:
## Summary

A `did:key` signature is authentication, not authorization. The method is self-certifying, so any party can generate a keypair, derive its DID and sign. With `GITLAWB_ENFORCE_OWNER_PUSH` defaulting to `false`, `owner_push_rejection` short-circuited (`api/repos.rs:1600`, `if !enforce { return None; }`) and every `git-receive-pack` carrying any valid signature was accepted — **including pushes to a repository the signer does not own, and including private ones**, on every branch not explicitly protected.

The gate itself already worked when switched on, and is well covered by unit tests. What was missing is that no reachable default configuration switched it on. A node started with nothing configured accepted a push from anyone who could run `gl identity new`. That default is what this PR changes.

Two commits, reviewable independently:

| Commit | Scope |
|---|---|
| `fix(node): gate two Unix-only tests…` | Prerequisite. Unblocks Windows development. No behaviour change on Linux. |
| `fix(node)!: enforce owner-only push by default` | The default flip, three test fixtures, three docs. |

Happy to split the first commit into its own PR if that reads better — it stands alone.

## The argument had to become value-taking

Flipping `default_value_t` alone would have been a trap. The field is a bare `bool`, so clap derives `ArgAction::SetTrue` — presence-only. A test asserting the escape hatch failed with:

```
error: unexpected argument 'false' found
```

With a `true` default and a presence-only flag, `--enforce-owner-push false` is a parse error and the gate can never be turned off. The argument now takes a value, so `--enforce-owner-push false` and `GITLAWB_ENFORCE_OWNER_PUSH=false` both disable it, while the bare `--enforce-owner-push` form still means `true`. Env and CLI share one value parser, so both resolve consistently.

## Three tests updated, none weakened

`test_support.rs:67` builds every fixture with `Config::parse_from(["gitlawb-node"])`, so the flip reaches every behavioural test. Three broke:

- `f1_write_cap_key_is_per_source_not_per_repo`
- `receive_pack_per_source_write_cap_sheds_capped_source_not_others`
- `receive_pack_acquire_deadline_sheds_and_releases_permit`

All three push as a stranger in order to reach the concurrency behaviour they actually assert on, and now stop at the gate instead. Each fixture now pushes as the repo owner, so they keep exercising the **shipped default** rather than setting `enforce_owner_push = false` and testing a configuration nobody runs. `did_matches` collapses the `did:key:` prefix against the bare owner the rows store, so this is the same identity.

## Verification

Run on Windows against a local PostgreSQL 17.

| Check | Result |
|---|---|
| `cargo test -p gitlawb-node --bin gitlawb-node` | 706 passed / 11 failed |
| Same command, flip reverted (baseline) | 704 passed / 11 failed |
| Regressions introduced | **0** — failure sets identical |
| `cargo fmt --all -- --check` | exit 0 |
| `cargo clippy --workspace --all-targets` | exit 0 |

The +2 is the two new config tests. The 11 failures are pre-existing on a clean tree and unrelated to this change — `sync::tests::*promisor*` die on `fatal: invalid filter-spec 'blob:limit=10g'` from the Windows git build, and the `ipfs_cid_*` walks return 503 where 200 is expected. Both are Windows environment issues that predate this branch; Linux CI should be unaffected. Worth a separate issue.

Both new tests were watched failing before the implementation existed, and each failed for its own distinct reason.

## Breaking change — please read before approving

`git-receive-pack` now rejects a push whose authenticated DID is not the repo owner, returning 403 before any ref update is applied.

**Delegated and CI keys count as non-owners.** A UCAN `git/push` capability is verified but not yet honored for authorization, so an agent pushing under its own DID cannot push while this is on, even holding a valid capability for the repo. This is a real trade-off, not an oversight: the previous default meant anyone could push anywhere, which is worse. But it does mean scoped delegation is a claim the forge cannot currently honor, and completing UCAN authorization is the work that removes the trade-off.

**Merge order.** That work is #331, which honors an owner-rooted `git/push` delegation on this same path. Land #331 first or together with this one; landing this alone locks out every delegated and CI key until #331 follows.

Operators who need the old behaviour during a rolling upgrade set `GITLAWB_ENFORCE_OWNER_PUSH=false`. Documented in `README.md`, `.env.example`, and `docs/RUN-A-NODE.md`.

## Two things to confirm

1. **No cutover date for `GITLAWB_REQUIRE_SIGNED_PEER_WRITES`** — it stays `false` in this PR, and `README.md` documents the flip as conditional ("once all live peers upgrade") with no date attached. An earlier revision of this description asserted a 15 September 2026 commitment; the branch never made one, and this corrects that. If the rollout wants a fixed date, it should be set deliberately in a follow-up rather than inferred from here.
2. **`CHANGELOG.md` is untouched** — it is release-please generated, so the breaking change is carried in the commit footer rather than hand-edited.

## Summary by CodeRabbit

- **Security**
  - Owner-only push authorization is now enabled by default on every branch.
  - Signed pushes from non-owner, delegated, and CI keys return HTTP 403 unless enforcement is disabled.

- **Configuration**
  - Added documented environment and CLI options to temporarily disable enforcement during rolling upgrades.
  - Docker Compose and macOS app deployments now pass through this setting, defaulting to enabled.
  - Signed peer writes remain disabled by default during staged rollout.

- **Documentation**
  - Upgrade guidance now explains configuration refresh behavior and Terraform application requirements.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -196,8 +196,9 @@ GITLAWB_MAX_CONCURRENT_READS_PER_CALLER=16
 #     many WRITE-pool slots. This one is load-bearing for push availability, since
 #     it is acquired before the global write permit: without it, one host minting
 #     disposable did:key identities could open enough slow pushes to monopolize the
-#     write pool and 503 every other source (owner enforcement defaults off, and
-#     the push rate limiter caps arrival rate, not in-flight concurrency).
+#     write pool and 503 every other source (the push rate limiter caps arrival
+#     rate, not in-flight concurrency, and owner enforcement can be turned off for
+#     a rolling upgrade).
 # Keyed on the resolved source IP, never the DID, so a DID farm does not defeat
 # them; keying granularity follows GITLAWB_TRUSTED_PROXY like the read cap above.
 
--- a/.env.example
+++ b/.env.example
@@ -94,9 +94,14 @@ GITLAWB_REQUIRE_SIGNED_PEER_WRITES=false
 
 # Require the authenticated pusher to be the repo owner on git-receive-pack.
 # A valid did:key signature is authentication, not authorization: anyone can
-# sign as their own DID. When true, pushes from a non-owner DID are rejected.
-# Keep false until the repo owner is ready for owner-only writes.
-GITLAWB_ENFORCE_OWNER_PUSH=false
+# mint a key, derive its DID and sign, so with this off every signed caller may
+# push to every repository, private ones included. On by default.
+#
+# Set to false only for a rolling upgrade whose pushers are not yet the repo
+# owner. Note that delegated and CI keys count as non-owners: UCAN git/push is
+# verified but not yet honored for authorization, so they cannot push while this
+# is on. See docs/RUN-A-NODE.md.
+GITLAWB_ENFORCE_OWNER_PUSH=true
 
 # Comma-separated libp2p multiaddrs.
 # Example: /ip4/1.2.3.4/udp/7546/quic-v1/p2p/12D3KooW...
--- a/README.md
+++ b/README.md
@@ -315,13 +315,6 @@ When `GITLAWB_REQUIRE_SIGNED_PEER_WRITES=false`, unsigned legacy peers are accep
 GITLAWB_REQUIRE_SIGNED_PEER_WRITES=true
 ```
 
-**Planned cutover: this default flips to `true` on 15 September 2026.** Until then a
-node started with no configuration still accepts unsigned announces, which is an open
-write surface kept open only so live peers can finish upgrading. Operators should set it
-to `true` now; after the cutover, setting it to `false` becomes the explicit opt-out.
-This mirrors what `GITLAWB_ENFORCE_OWNER_PUSH` has already done — that one defaults to
-`true` today.
-
 `POST /api/v1/sync/trigger` is not part of the staged rollout: it always requires a signature in both config modes and returns 401 without one, because each call drives an O(peers) outbound fan-out.
 
 ---
--- a/README.md
+++ b/README.md
@@ -315,6 +315,13 @@ When `GITLAWB_REQUIRE_SIGNED_PEER_WRITES=false`, unsigned legacy peers are accep
 GITLAWB_REQUIRE_SIGNED_PEER_WRITES=true
 ```
 
+**Planned cutover: this default flips to `true` on 15 September 2026.** Until then a
+node started with no configuration still accepts unsigned announces, which is an open
+write surface kept open only so live peers can finish upgrading. Operators should set it
+to `true` now; after the cutover, setting it to `false` becomes the explicit opt-out.
+This mirrors what `GITLAWB_ENFORCE_OWNER_PUSH` has already done — that one defaults to
+`true` today.
+
 `POST /api/v1/sync/trigger` is not part of the staged rollout: it always requires a signature in both config modes and returns 401 without one, because each call drives an O(peers) outbound fan-out.
 
 ---
@@ -340,7 +347,8 @@ Important node settings:
 | `GITLAWB_BOOTSTRAP_PEERS` | Comma-separated HTTP peer URLs. |
 | `GITLAWB_P2P_BOOTSTRAP` | Comma-separated libp2p multiaddrs. |
 | `GITLAWB_BOOTSTRAP_DISABLE_SEEDS` | Disable embedded seed peers for isolated dev/test networks. |
-| `GITLAWB_REQUIRE_SIGNED_PEER_WRITES` | Require signed peer announce/sync writes. |
+| `GITLAWB_REQUIRE_SIGNED_PEER_WRITES` | Require signed peer announce/sync writes. Defaults to `false` during the staged rollout below. |
+| `GITLAWB_ENFORCE_OWNER_PUSH` | Require the authenticated pusher to be the repo owner on `git-receive-pack`. **Defaults to `true`.** A `did:key` signature is authentication, not authorization — anyone can mint a key and sign — so with this off every signed caller may push to every repository, private ones included. Delegated and CI keys count as non-owners: a UCAN `git/push` capability is verified but not yet honored for authorization, so they cannot push while this is on. Set `false` only for a rolling upgrade; see [`docs/RUN-A-NODE.md`](docs/RUN-A-NODE.md). |
 | `GITLAWB_AUTO_SYNC` | Enable automatic sync from known peers. |
 | `GITLAWB_MAX_PACK_BYTES` | Max git pack body size for smart-HTTP routes. |
 | `GITLAWB_GIT_SERVICE_TIMEOUT_SECS` | Max seconds a served git upload-pack, receive-pack, or `info/refs` advertisement may run before it is aborted (504). Default 600. Also bounds the withheld-blob classification walk (on both the upload-pack serve and receive-pack replication paths) and the push-side pin-candidate discovery (`rev-list` / `cat-file`), each reaped via process-group teardown at the deadline. On the path-scoped upload-pack path the classification walk and the pack serve share ONE deadline, so this value bounds their combined duration rather than granting each stage a full budget: a walk that consumes it leaves the serve nothing and the clone gets a 504. Serving large path-scoped repos may therefore need a higher value than they did when each stage was budgeted separately. Accepted range is 1 to 3153600000 (100 years), since the node derives deadlines from this value and a larger one cannot be represented. |
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -4993,6 +4993,13 @@ mod tests {
     /// task the global slot stays occupied until the walk finishes; on the pre-fix code
     /// the handler-local permits drop on future-drop and the slot frees instantly (RED),
     /// letting disconnect-spam exceed the cap while real git work keeps running.
+    ///
+    /// Unix-only: the fake git is a `/bin/sh` script made executable through
+    /// `PermissionsExt::set_mode`, and the hung `rev-list` is reaped with
+    /// `libc::kill(SIGKILL)`. Neither exists on Windows, so without this gate the
+    /// whole `gitlawb-node` test target fails to compile there (#228) and no test
+    /// in the crate can run on a Windows checkout.
+    #[cfg(unix)]
     #[sqlx::test]
     async fn upload_pack_permit_held_through_walk_after_disconnect(pool: sqlx::PgPool) {
         use axum::body::Body;
@@ -5860,6 +5867,11 @@ mod tests {
     /// sheds — releasing the permit lets the SAME walk run and pin (durability stays
     /// fail-closed). Exercises the gating seam directly; the detached push task calls
     /// this exact helper.
+    ///
+    /// Unix-only for the same reason as
+    /// `upload_pack_permit_held_through_walk_after_disconnect`: the fake git is a
+    /// `/bin/sh` script made executable through `PermissionsExt::set_mode` (#228).
+    #[cfg(unix)]
     #[tokio::test]
     async fn encrypt_walk_defers_when_pool_exhausted() {
         use std::sync::Arc;
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -9195,7 +9195,9 @@ mod tests {
         (state, log)
     }
 
-    const P2_PUSHER: &str = "did:key:z6p2tail";
+    // No shared pusher constant: the two P2 tests own different repos (`z6p2tail`
+    // and `z6p2fail`), and owner-only push is on by default, so the identity has to
+    // follow the repo each push targets rather than being fixed for both.
 
     fn p2_push(
         state: &AppState,
@@ -9208,7 +9210,7 @@ mod tests {
         git_receive_pack(
             State(state.clone()),
             Path((owner.to_string(), name.to_string())),
-            Extension(crate::auth::AuthenticatedDid(P2_PUSHER.to_string())),
+            Extension(crate::auth::AuthenticatedDid(format!("did:key:{owner}"))),
             crate::rate_limit::PeerAddr(Some("203.0.113.90:5000".parse::<SocketAddr>().unwrap())),
             axum::http::HeaderMap::new(),
             axum::body::Bytes::from_static(b"0000"),
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -5622,7 +5622,10 @@ mod tests {
             .await
             .unwrap();
 
-        let did = "did:key:z6MkReceivePackWriteCapProofDidAAAAAAAAAA";
+        // Owner of the `rp4` row, so the push reaches the per-source write cap instead
+        // of stopping at the owner-push gate (on by default). `did_matches` collapses
+        // the `did:key:` prefix against the bare stored owner.
+        let did = "did:key:z6rp4wr";
         let capped: SocketAddr = "203.0.113.44:5000".parse().unwrap();
         let other: SocketAddr = "203.0.113.45:5000".parse().unwrap();
 
@@ -5741,7 +5744,10 @@ mod tests {
             .await
             .expect("hold the advisory lock on the second connection");
 
-        let did = "did:key:z6MkAcquireDeadlineProofDidAAAAAAAAAAAAAAAA";
+        // Owner of the seeded rows, so the push reaches `acquire_write` instead of
+        // stopping at the owner-push gate (on by default). `did_matches` collapses the
+        // `did:key:` prefix against the bare `owner` above.
+        let did = "did:key:z6acqdead";
         let peer: SocketAddr = "203.0.113.61:5000".parse().unwrap();
 
         let sem = state.git_write_semaphore.clone();
@@ -8584,7 +8590,11 @@ mod tests {
                 .unwrap();
         }
 
-        let did = "did:key:z6MkF1KeyPusherAAAAAAAAAAAAAAAAAAAAAAAAAA";
+        // The pusher must be the repos' owner: this test is about the write-cap key's
+        // shape, so the push has to reach the cap rather than stop at the owner-push
+        // gate, which is on by default. `did_matches` collapses the `did:key:` prefix,
+        // so this is the same identity as the bare `z6f1key` the rows are owned by.
+        let did = "did:key:z6f1key";
         let capped: SocketAddr = "203.0.113.65:5000".parse().unwrap();
         let other: SocketAddr = "203.0.113.66:5000".parse().unwrap();
         let _slot = state
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -8483,8 +8483,11 @@ mod tests {
     /// the per-source permit above the park, four parked pushes deny every push on the
     /// node for up to 1260s.
     ///
-    /// Same one source IP for all three pushes (the collapsed-key shape), distinct pusher
-    /// DIDs, per-source cap 2 so the holder plus a parked push would exhaust it.
+    /// Same one source IP for all three pushes (the collapsed-key shape), per-source
+    /// cap 2 so the holder plus a parked push would exhaust it. A and B are the owner
+    /// of `n1`; C is a second tenant, `z6u1nat2`, pushing to its own `n2` — owner-only
+    /// push means a tenant is identified by the repo it may write, not by an arbitrary
+    /// DID, and the cap keys on the resolved peer address rather than either.
     #[cfg(unix)]
     #[sqlx::test]
     async fn u1_parked_push_does_not_shed_another_pusher_behind_the_same_ip(pool: sqlx::PgPool) {
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -1780,7 +1780,7 @@ pub async fn git_receive_pack(
         "parsed ref updates from pack"
     );
 
-    // ── Owner-only push enforcement (opt-in: GITLAWB_ENFORCE_OWNER_PUSH) ──
+    // ── Owner-only push enforcement (on by default; GITLAWB_ENFORCE_OWNER_PUSH) ──
     // Runs before branch protection on purpose: when enabled, a non-owner is
     // rejected here regardless of whether the target branch is protected, so a
     // single rejection never yields two different error bodies. The identity is
@@ -1903,9 +1903,12 @@ pub async fn git_receive_pack(
     // (INV-10) and a saturated pool sheds 503 before spawning git.
     //
     // Per-source sub-cap first (#174 P1-d): one source IP cannot occupy the whole write
-    // pool via many slow pushes. Owner enforcement defaults off, so any valid did:key is
-    // accepted (auth != authz) and the push rate limiter bounds arrival RATE, not in-flight
-    // concurrency. Keyed on the resolved source IP, NEVER the signed DID (a DID farm defeats
+    // pool via many slow pushes. The push rate limiter bounds arrival RATE, not in-flight
+    // concurrency, so this cap is what bounds occupancy. Owner enforcement is on by
+    // default now, but it is not what makes this cap load-bearing: an operator may turn it
+    // off for a rolling upgrade, and even with it on a single owner can open many
+    // concurrent slow pushes to their own repo.
+    // Keyed on the resolved source IP, NEVER the signed DID (a DID farm defeats
     // a DID key); no resolvable key -> global write pool only. Then the global write permit:
     // pushes draw from the dedicated WRITE pool, separate from reads, and it is held for the
     // whole op (moved into the AdmissionGuard below).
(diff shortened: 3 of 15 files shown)


beardthelion (dismissed):
The default should have been `true` from the start, and the reasoning in the PR body is right: a
self-certifying `did:key` is authentication, and a gate that no reachable configuration switched on
was decoration. The fixture sweep in `586896d8` and `7de6f7cf` is the correct fix and the suite is
green at `7de6f7c` (825 passed, 0 failed). Deriving the pusher from the repo it targets rather than
hard-coding a matching literal is the better of the two shapes, and it is what makes the remaining
item below visible.

I checked the flag's blast radius first. `owner_push_rejection` is the only production reader,
called once from `git_receive_pack`, and it runs before the per-source write cap and before
`acquire_write`, so a rejected push consumes no permit or lease. Nothing in `sync.rs`, `p2p/` or
`api/peers.rs` reaches receive-pack, so replication is unaffected, and `git-remote-gitlawb` surfaces
the 403 body rather than rendering the denial as an empty success. The mechanism is sound. What is
left is one test the sweep missed and some prose.

## Findings

- [P1] Push as the owner in `f1_write_cap_is_inert_without_a_resolvable_source_key` too
  `crates/gitlawb-node/src/api/repos.rs:8663`
  This is the one fixture the sweep did not reach, and it is invisible precisely because it never
  went red. The repo is seeded to `z6f1none` while the push authenticates as
  `did:key:z6MkF1NoKeyPusherAAAAAAAAAAAAAAAAAAAAAAA`, so the request is now rejected at the owner
  gate and never reaches the per-source cap the test exists to measure. Its assertion is
  `!matches!(r, Err(AppError::Overloaded(_)))`, and a `Forbidden` satisfies that, so it reports `ok`
  inside the green run while proving nothing. Confirmed by running it: swapping only that
  assertion to `matches!(r, Err(AppError::Forbidden(_)))` makes the test pass, so the request
  really is stopping at the gate. Give it `did:key:z6f1none` like the others, and
  tighten the assertion so a denial cannot pass it: the current negative shape at `:8671` accepts any
  error at all, which is what let this hide. One sibling carries the same shape,
  `f1_write_cap_key_is_per_source_not_per_repo` at `:8626`. That one passes for the right reason now
  that you gave it the owner DID, but the shape would hide an identical regression, so it is worth
  tightening while you are in here.

- [P2] Restore the cross-tenant shape in `u1_parked_push_does_not_shed_another_pusher_behind_the_same_ip`
  `crates/gitlawb-node/src/api/repos.rs:8482`
  The sweep collapsed this test's three pusher identities into one `did:key:z6u1nat`, but the
  docstring still reads "distinct pusher DIDs" and the line above it still frames the case as "a
  DIFFERENT pusher's push to a DIFFERENT repo". The mechanism is fine and I checked it:
  `read_caller_key` delegates to `rate_limit::client_key(headers, peer, trust)`, which never sees the
  DID, so the three pushes contend on the resolved edge IP exactly as before. But this is the test
  named THE CROSS-TENANT GUARD, and after the collapse nothing drives two distinct tenants sharing
  one source key. That shape is still expressible under owner-only push and cheaply so: `f1_add_repo`
  already takes an owner, so seeding `n2` to a second owner and signing push C as that owner keeps
  the guard matching its name. Failing that, fix the two docstring lines, since they currently
  describe a scenario the code no longer runs.

- [P1] Test the behavior the PR exists to change
  `crates/gitlawb-node/src/config.rs:998`
  Following the flag through every consumer: the gate's own tests at `repos.rs:3327-3366` pass
  `enforce` as a literal and are insensitive to the default, no fixture sets the field, and the
  string `only the repo owner may push` appears exactly once in the file, at its own definition. So
  reverting `default_value_t` to `false` turns exactly one test red, and that test parses the
  process environment, which means on a host with `GITLAWB_ENFORCE_OWNER_PUSH=false` exported the
  revert turns nothing red at all. Add a handler-level test: default `Config`, repo owned by A, push
  as B, assert 403. For the other half, asserting the declared default is the env-independent form:

  ```rust
  let cmd = Config::command();  // clap::CommandFactory
  let arg = cmd.get_arguments().find(|a| a.get_id() == "enforce_owner_push").unwrap();
  assert_eq!(arg.get_default_values(), ["true"]);
  ```

  I compiled that against the pinned clap and ran it with `GITLAWB_ENFORCE_OWNER_PUSH=false`
  exported; it still reads `["true"]`, which is the property the current test lacks.

- [P2] Forward the knob in the shipped compose files
  `docker-compose.yml:40`
  The node service uses an explicit `environment:` block with no `env_file:`, and
  `GITLAWB_ENFORCE_OWNER_PUSH` is absent from it while the sibling security knob
  `GITLAWB_REQUIRE_SIGNED_PEER_WRITES` sits at line 46. Same gap in
  `macos-app/Sources/GitlawbNode/Resources/docker-compose.yml` and `infra/aws/compose.yaml.tftpl`.
  An operator who follows the rolling-upgrade instructions this PR adds to three docs gets nothing
  on the shipped container path. `GITLAWB_ENFORCE_OWNER_PUSH: ${GITLAWB_ENFORCE_OWNER_PUSH:-true}`
  keeps the new default while making the opt-out real.

- [P2] Update the four comments that still say the gate defaults off
  `crates/gitlawb-node/src/api/repos.rs:1783`
  Also `repos.rs:1906`, `repos.rs:5598`, and `.env.example:199`, the last one in a file this PR
  otherwise rewrites. At `:1906` the IP-keying rationale is still correct (a DID farm defeats a DID
  key); it is only the "enforcement defaults off" premise underneath it that died. `repos.rs:1594`
  and `:3355` describe the flag-off branch and are accurate as they stand.

- [P2] Drop the 15 September 2026 cutover from the README
  `README.md:318`
  That paragraph is about `GITLAWB_REQUIRE_SIGNED_PEER_WRITES`, which this PR does not change, and
  the date appears nowhere on `main`. Answering the question you raised in the PR body: it is not a
  promise the project has made, and a PR about a different flag is not where we would make one. Cut
  the paragraph. The table row you added for `GITLAWB_ENFORCE_OWNER_PUSH` carries the part that belongs.

Two notes, not asks. The docstring at `config.rs:1011-1013` credits the value-taking action with
making `GITLAWB_ENFORCE_OWNER_PUSH=false` resolve to `false`. I built both the old and the new
declarations against the pinned clap 4.5.60 and ran them: the env form already resolved under
`SetTrue`, so what the change actually buys is `--enforce-owner-push false` parsing at all. That
same run says the new shape is fail-closed, which is the answer to the parser question raised in
review: bare `--enforce-owner-push` is `true`, it does not swallow a following `-` token, and `=1`,
`=yes`, `=TRUE` and empty are startup errors rather than a silent `false`, before and after.
Separately, I checked the Windows premise behind the `#[cfg(unix)]` gates rather than take it on
trust, and it holds. Cross-compiling the crate for `x86_64-pc-windows-gnu --all-targets`
succeeds with both gates in place; removing just those two attributes fails with six errors
naming exactly the constructs your comments cite (`os::unix`, `libc::kill`, `SIGKILL`,
`Permissions::set_mode`). Both gates are load-bearing. Nothing in CI covers that, since
`pr-checks.yml:269` builds only `gl` and `git-remote-gitlawb` on Windows, and a
`cargo check --target x86_64-pc-windows-gnu -p gitlawb-node --all-targets` step needs neither
Postgres nor a Windows runner, so that lane could carry it. Its own change, not yours.

I am not asking for a split. The `#[cfg(unix)]` commit is a genuine prerequisite and small enough
that separating it would cost more than it saves.

jatmn (changes requested):

## Findings

- [P1] Preserve the opt-out for existing AWS deployments
  `infra/aws/compose.yaml.tftpl:77`
  The new interpolation is rendered only by first-boot user data. Existing instances retain the pre-PR `/opt/gitlawb/compose.yaml`: `aws_instance` deliberately ignores `user_data` drift, and the advertised SSM upgrade command at `infra/aws/main.tf:446` only pulls the image and runs `docker compose up -d`. Those old Compose files have an explicit `environment:` block without this key, so Compose does not pass an operator's `/opt/gitlawb/.env` value into the node container.

  Reproduction is the documented rolling-upgrade sequence on an existing AWS node: write `GITLAWB_ENFORCE_OWNER_PUSH=false` to `/opt/gitlawb/.env`, then invoke `upgrade_command`. The node process receives no such environment variable, takes the new binary default of `true`, and immediately rejects delegated/CI pushes that the docs say can remain enabled during migration. This is a deployment-state migration problem, not a template-default problem. Make the upgrade action install the newly rendered Compose file before restarting (or provide a versioned, idempotent Compose migration), and document the required one-time migration for existing instances. Please add a regression check that starts from the old rendered Compose file and proves the opt-out reaches the upgraded container.

- [P1] Wire the macOS app's actual generated Compose file to the opt-out
  `macos-app/Sources/GitlawbNode/DockerCompose.swift:292`
  The modified resource is not the Compose file a normal packaged app runs. When no source checkout is selected, `composeFilePath` writes `Application Support/GitlawbNode/docker-compose.yml` from `generateComposeFile()`. Its node `environment:` block at this location does not include `GITLAWB_ENFORCE_OWNER_PUSH`; moreover, each start calls `Config.writeEnvFile()`, which recreates that app's `.env`. The bundled resource is not read by this fallback path.

  Consequently, putting `GITLAWB_ENFORCE_OWNER_PUSH=false` in the app's environment file cannot disable the gate: the generated Compose file never forwards it, and a manual edit is discarded on the next start. Packaged-app users with delegated or CI pushers are therefore forced into the breaking default with no usable migration control. Address the root cause by making the app have one source of truth for its runtime Compose configuration: add the interpolation to `generateComposeFile()`, persist an explicit owner-push setting in the app configuration (or deliberately preserve an operator-provided value), and make the bundled and generated templates share that definition. A packaged-app smoke test should set the opt-out and assert it appears in `docker compose config` / the container environment.

- [P2] Make the default-behavior test independent of the process environment
  `crates/gitlawb-node/src/test_support.rs:2182`
  The test is named and documented as a no-configuration/default-behavior assertion, but `test_state()` builds `AppState` through `Config::parse_from(["gitlawb-node"])`. Clap reads `GITLAWB_ENFORCE_OWNER_PUSH` from the surrounding process, so a test run launched with the documented rolling-upgrade override (`false`) constructs a state with enforcement disabled. The signed non-owner request then passes the owner gate and reaches the Git backend instead of returning the asserted 403.

  The parser-introspection test correctly proves the declared default, but it does not make this handler-level test hermetic. Do not mutate the process environment around an async test suite: construct the test state with an explicit, environment-independent `Config` (or add a narrowly scoped state-builder override) and set `enforce_owner_push = true` for the handler wiring assertion. Keep the declared-default test as the proof of the shipped default, and add an explicit override test if the behavior under `false` needs coverage. That separates parser precedence from authorization behavior and prevents a developer or CI environment setting from turning this into a spurious failure.

## Guidance for addressing the whole change

The authorization change itself is the right direction: a self-certifying `did:key` proves who signed, not that the signer owns the target repository. Nothing below asks to restore the insecure default, to honor delegated UCAN pushes in this PR, or to change the separate signed-peer-write rollout. The PR expressly chooses owner-only authorization by default and a temporary `false` override for rolling upgrades; the guidance is limited to making that stated transition work consistently.

The remaining problems are consequences of treating that default flip as only a Rust/Clap change. For the launch paths this PR documents or modifies, the complete contract is broader than `Config::default()`:

1. A fresh source build must default to owner-only push.
2. Each documented or changed Compose/runtime path must be able to pass the temporary `false` override to the node process.
3. The documented AWS rolling-upgrade path must migrate existing installations, not merely render future installations correctly.
4. Tests must distinguish the declared default, explicit configuration precedence, and the handler's authorization behavior without inheriting ambient process state.

The current diff satisfies the first item and the new parser configuration supports the second in a direct CLI/source invocation. It does not yet satisfy the deployment lifecycle cases. Root Compose, the first-boot AWS template, and the macOS resource were updated, but existing AWS instances retain a previous generated file and the packaged macOS application runs a separately generated template. Those are independent configuration sources with different upgrade lifecycles. Updating a checked-in template is not enough when a deployed copy or a runtime generator remains authoritative.

Please address this as one compatibility-migration change rather than adding isolated forwarding lines:

- Define the owner-push setting in the actual configuration source used by each changed/documented runtime, with the same default and interpolation semantics.
- Inventory the launch paths touched here: source/CLI, root Compose, newly bootstrapped AWS, already-running AWS, packaged macOS fallback, and source-checkout macOS. For each, verify both the omitted-value default and an explicit `false` value reach the node process where that path exposes the documented override.
- Add an idempotent AWS upgrade migration. It should safely update or regenerate the persisted `/opt/gitlawb/compose.yaml` before the new image starts; a future instance replacement alone is not an adequate migration for a documented rolling upgrade.
- Avoid maintaining two macOS Compose definitions that silently diverge. Generate the fallback from the same canonical template or test that both rendered outputs contain the same node environment contract; ensure the documented override is forwarded rather than overwritten on app restart.
- Split the test responsibilities: parser metadata/default tests, CLI/env override tests, and handler authorization tests should each build only the state they are proving. Do not rely on whichever environment happens to run the test process.

Before requesting another review, please run a small deployment-focused matrix in addition to the normal Rust suite: fresh source default; source CLI `false`; fresh root Compose `false`; an old AWS Compose file upgraded through the published SSM command; a newly bootstrapped AWS instance; and the packaged macOS fallback with an explicit `false` value. In each case, inspect the resolved Compose configuration or container environment and perform a signed non-owner push. The expected outcomes should be explicit: default is 403, temporary override reaches the process and preserves the legacy path. This will make the breaking change reviewable as one end-to-end contract and avoid finding the next un-migrated launcher in a later round.

beardthelion (dismissed):
Every ask from the last two rounds landed, and the node half of this is done. Both f1 fixtures push
as the owner now and assert the positive `AppError::Git(_)` instead of the negative shape that hid
the first regression; the cross-tenant test seeds `n2` under `z6u1nat2` so it drives two real tenants
again; the declared default is read off the parser rather than out of a parse, which is the
env-independent form. Suite is green at `1f0d9ff3`: 826 passed, 0 failed. `Config::parse()` in
`main.rs:72` is the only production construction site and there is no `Default` impl, so the flip
cannot be bypassed by a second builder. Fly needs nothing: its `[env]` block is direct process
environment with no allowlist, so the opt-out already reaches the process there.

What is left is all on the deployment half added in `1f0d9ff3`, and none of it is in the Rust.

## Findings

- [P2] Split the AWS "Changing configuration" section so it stops contradicting the upgrade section
  `infra/aws/README.md:124`
  That section still says editing `image_tag` "does **not** affect a running instance" and routes the
  operator to `terraform apply -replace=aws_instance.node`. Thirty lines above, the section this PR
  added says the opposite and is now the correct one: the upgrade reinstalls the rendered compose,
  which carries `image_tag`. The same paragraph also tells operators to edit `/opt/gitlawb/.env` and
  restart, which is exactly the advice your own "Why the upgrade rewrites the compose file" section
  explains is silently ineffective for any variable the installed compose does not already name.
  Keep the invariant that both halves have to preserve: every rollout instruction in this file must
  route through `apply` plus `upgrade_command`, and `.env` must be described as reaching the
  container only for keys the installed compose already interpolates. `bootstrap_peers` and
  `public_url` do stay replacement-only, so the section needs splitting rather than deleting.

- [P2] Say what `--remove-orphans` now does, or drop it from the upgrade
  `infra/aws/main.tf:463`
  The flag is not new, but until this commit it was inert: the upgrade never rewrote
  `/opt/gitlawb/compose.yaml`, so the file it reconciled against was the instance's own and the only
  orphans were containers an operator had started by hand. Now it reconciles against a fresh
  rendering whose service set is conditional (`compose.yaml.tftpl:22` renders `postgres` only when
  `db_host == ""`, `:7` renders `caddy` only when `domain_name != ""`). An operator who sets
  `use_rds` or clears `domain_name`, applies, then runs the command the README describes as pull and
  restart, has their local postgres or their TLS terminator stopped and deleted. The data survives,
  since both are bind mounts under `/mnt/data`, but the topology change is unannounced. Either state
  it in the README caution next to "Terraform-owned and is overwritten on every upgrade", or drop
  `--remove-orphans` and let a service-set change require the explicit replace.

- [P2] Give the macOS setting a writer
  `macos-app/Sources/GitlawbNode/Config.swift:26`
  `enforceOwnerPush` is persisted, loaded, and emitted into `.env`, and the generated compose
  forwards it, so the plumbing is right. Nothing sets it. `SettingsWindow.swift` has no `@State` and
  no assignment for it, and `save()` at `:89-98` writes only the fields it tracks, so the only way to
  reach the documented opt-out on this launch path is hand-editing `config.json` in Application
  Support, which no doc mentions. Editing `.env` does not work here: `writeEnvFile()` rebuilds that
  file from `Config` on every `start()` (`DockerCompose.swift:77`). I confirmed a hand-edit does
  survive a Settings save, since `persist()` re-emits the value `load()` read, so this is
  reachability rather than data loss. The invariant is that the opt-out this PR documents must have
  an operator-reachable setter on every launch path it ships. A `Toggle` beside the existing
  `Sync repos from peers` one at `SettingsWindow.swift:51`, with the `@State` at `:30` and the
  assignment in `save()`, matches the pattern already there; documenting the `config.json` key would
  also close it.

- [P3] Fix three prose lines the change made untrue
  `crates/gitlawb-node/src/test_support.rs:2259`
  The assertion reads "with no configuration at all" while the test sets `enforce_owner_push = true`
  explicitly two dozen lines up, which is the right thing for the test to do and the wrong thing for
  the message to claim. Also `api/repos.rs:8486`, whose docstring still says all three pushes carry
  distinct pusher DIDs when A and B now share `n1`'s owner, and `docs/OSS-READINESS-AUDIT.md:91`,
  which still lists "unprotected repo branches do not yet enforce owner/UCAN capability checks" as a
  live-network blocker. Owner checks are enforced on every branch as of this PR; the UCAN half of
  that sentence is the part that stays true.

Three notes, not asks. The `#[allow(dead_code)]` at `test_support.rs:54` is redundant now that the same
commit adds the call site; I removed it locally and nothing warns. `repos.rs:5673` and `:5866` still carry the lone-negative
`!matches!(.., Overloaded)` shape in two of the fixtures you swept. Both push as the repo owner now,
so they pass for the right reason and I am not asking you to touch them; I mention it only because
the same shape is what hid the original defect and you tightened its two siblings. Separately, a
non-owner push to a private repo returns 403 with a body naming the reason, which is an existence
oracle the read path does not give away. It is strictly smaller than what it replaces, so it is not
a blocker on this PR, but it is worth its own issue.

I checked the aggregate rather than just this round, since this is the third. The remedy has grown a
long way from a one-line default flip and every finding above traces to the fix for the previous
round rather than to the original defect, which is usually the point to unwind the scope. I am not
asking for a split, and the reason is that the deployment migration cannot ship after the flip: doing
that lands the breaking default on existing AWS instances with the opt-out unreachable, which is the
problem the migration exists to solve. The bundling is load-bearing, so this stays one PR.

beardthelion (dismissed):
The node half is done and I have nothing further on it. Both f1 assertions hold on Linux, not just on the Windows box the PR body reports from: I ran `f1_write_cap_is_inert_without_a_resolvable_source_key` and `f1_write_cap_key_is_per_source_not_per_repo` at `579eee07` against Postgres and both pass. The gate still sits ahead of branch protection with one production call site at `repos.rs:1789`, the macOS toggle is a real setter end to end, and the AWS README now splits correctly by which file carries a change. CI is green, seventeen checks.

One thing in the new upgrade path needs fixing before this lands.

## Findings

- [P2] Write the compose file through a unique temp path and serialize the upgrade
  `infra/aws/main.tf:467`
  Two executions of the upgrade document race on the fixed `/opt/gitlawb/compose.yaml.new`. The second execution's redirect truncates the file the first is about to install, and because the first has already renamed the inode, the second keeps writing through into the live `compose.yaml`. I reproduced both outcomes with the same write-then-rename shape: two different renderings produce an installed file blending both service sets, and a truncating second execution produced a short file in 12 of 12 trials across the timing window, between 66 and 141 of 150 services surviving. Either result is valid YAML, so `docker compose up -d --remove-orphans` on the next line accepts it and deletes whatever fell off, which on a live node is the local postgres or the TLS terminator. `main` has no compose write in this document, so the race arrives with this commit. `mktemp` for the temp path plus an `flock` around the whole step closes it.

- [P3] Drop the README cutover item from the PR description
  `README.md:343`
  The description's "Two things to confirm" still says a 15 September 2026 cutover for `GITLAWB_REQUIRE_SIGNED_PEER_WRITES` is asserted in `README.md`. You removed it, correctly, and the row at `:343` now says only that it defaults to `false` during the staged rollout. The description is the last place still claiming a commitment the branch does not make.

jatmn (changes requested):

## Findings

- [P2] Serialize the AWS upgrade compose write with a unique temp path
  `infra/aws/main.tf:467`

  The upgrade SSM document now reinstalls `/opt/gitlawb/compose.yaml` before `docker compose pull && up -d --remove-orphans`. That migration is the right fix for the opt-out problem on existing instances, but the install step is not safe under overlap.

  **What happens today**

  The document uses a fixed intermediate path and a plain redirect:

  ```text
  cat > /opt/gitlawb/compose.yaml.new <<'COMPOSE_EOF'
  …
  COMPOSE_EOF
  mv /opt/gitlawb/compose.yaml.new /opt/gitlawb/compose.yaml
  ```

  `main` never wrote compose from this document, so this race is new in this PR. If two `upgrade_command` invocations overlap on the same instance — a double-run, a retry while the first is still pulling, or two operators — both executions share `/opt/gitlawb/compose.yaml.new`.

  Two failure modes follow from the usual write-then-rename shape:

  1. **Blended file.** Execution A is still writing service definitions while execution B truncates `.new` and writes a different rendering. A then `mv`s its (possibly partial) file into place; B keeps writing through the open fd into the live `compose.yaml` inode after A's rename. The installed file can contain a mix of both renderings.
  2. **Truncated file.** A second execution truncates `.new` while the first is mid-write. The result can be a short, syntactically valid YAML fragment.

  Either outcome is still valid YAML enough for `docker compose up -d --remove-orphans` to accept. The next line in the script then reconciles against that file. On a live node that can drop conditional services — local `postgres` when `db_host` is set, or `caddy` when `domain_name` is cleared — which is exactly what the README now warns about for `--remove-orphans`.

  **Why this is merge-blocking**

  The compose rewrite is load-bearing for the rolling-upgrade contract this PR adds: without it, `GITLAWB_ENFORCE_OWNER_PUSH` in `/opt/gitlawb/.env` never reaches the container on pre-template instances. The migration needs to be safe to run in production, not only correct on a single serial invocation.

  ## Guidance

  The root cause is treating a **shared temp path** as an atomic install and assuming **at-most-one upgrade** without enforcing it. Address both halves:

  1. **Unique temp path per run.** Do not write through a fixed `compose.yaml.new`. Use `mktemp` under `/opt/gitlawb` (or `/tmp` with a stable prefix) so concurrent writers never truncate each other's file.
  2. **Serialize the whole compose-install + restart step.** Wrap the write, rename, `pull`, and `up` in an `flock` on a dedicated lock file (for example `/opt/gitlawb/.upgrade.lock`). If the lock is held, exit non-zero with a clear message rather than proceeding. That closes the inode-rename race where the second writer follows the first `mv` into the live path.

  A shape that matches the intent of the current script:

  ```bash
  set -euo pipefail
  install -d -m 0755 /opt/gitlawb
  exec 9>/opt/gitlawb/.upgrade.lock
  flock -n 9 || { echo "upgrade already in progress" >&2; exit 1; }

  tmp="$(mktemp /opt/gitlawb/compose.yaml.XXXXXX)"
  trap 'rm -f "$tmp"' EXIT
  cat >"$tmp" <<'COMPOSE_EOF'
  ${local.compose_yaml}
  COMPOSE_EOF
  mv "$tmp" /opt/gitlawb/compose.yaml
  trap - EXIT

  cd /opt/gitlawb && docker compose pull && docker compose up -d --remove-orphans && docker image prune -f
  ```

  Keep the heredoc delimiter quoted so shell does not expand compose interpolation markers. If the rendered YAML can ever contain a line that exactly matches the delimiter, pick a delimiter that cannot appear in the template or escape accordingly — the current `COMPOSE_EOF` choice is probably fine given the tftpl content, but it is worth a quick grep when you touch this.

  **Verification to add before re-requesting review**

  - Manually or in a small shell test, run two overlapping installs that use the same write-then-rename pattern and confirm the fixed version always leaves a byte-identical `compose.yaml` matching one complete rendering.
  - Confirm a second concurrent invocation exits immediately with the lock message and does not mutate `compose.yaml` or run `docker compose up`.
  - After a serial successful upgrade on a fixture that mimics a pre-PR compose file (no `GITLAWB_ENFORCE_OWNER_PUSH` in the service `environment:` block), confirm `GITLAWB_ENFORCE_OWNER_PUSH=false` in `/opt/gitlawb/.env` reaches the running container — that is the migration this step exists to make safe, and it should keep working once the install is serialized.

jatmn (approved):
LGTM

beardthelion (approved):
Approving. The round-5 ask landed, and I re-verified both it and the premise underneath it.

On the upgrade rewrite I extracted the rendered shell body and ran it rather than reading it. Two overlapping executions serialize correctly: the second is refused with the documented message and exits 1, and compose.yaml survives intact and unblended at mode 0644 with no stray temp files. I also forced a write failure mid-decode, and the script aborts, the EXIT trap removes the temp file, and the pre-existing compose.yaml is untouched. Worth noting for anyone reading the transport later: a deliberately truncated base64 payload does install partially, because `base64 -d` exits 0 on short input and pipefail never fires. `base64encode` cannot emit one, so there is no reachable path today, but the safety there comes from the renderer rather than from the guard.

The world-readable compose.yaml carries no secrets. POSTGRES_PASSWORD, GITLAWB_OPERATOR_PRIVATE_KEY, and both S3 keys are `$${VAR}` runtime references resolved from the 0600 .env.

The premise is still load-bearing: `default_value_t = true` in `crates/gitlawb-node/src/config.rs`, pinned by the test at `config.rs:1001`. Gutting the default turns it red and restoring it turns it green.

One non-blocking item.

## Findings

- [P3] Restrict the upgrade lock file so a local user cannot wedge it
  `infra/aws/main.tf:479`
  `exec 9>/opt/gitlawb/.upgrade.lock` creates a root-owned 0644 file, and flock(2) grants an exclusive lock through a read-only descriptor. I confirmed both by execution: the file comes out 644 under umask 022, a read-only fd acquires the free lock, and the upgrade's own writer is then refused. Any local user who can open that path read-only can make every subsequent upgrade fail fast with "another upgrade is already running". Not reachable on the default deployment, which has no SSH ingress, so a `chmod 0600` after the open or a follow-up issue both close it.

Recorded without asking for changes: first boot writes compose.yaml at `infra/aws/user-data.sh.tftpl:108` with a plain `cat >`, outside the lock, so an upgrade racing first boot is not serialized. A failed `docker compose pull` leaves the new rendering on disk over the still-running old stack. Both predate this round. The PR body still describes only the node commits and not the infra work that landed here, worth a one-line edit before merge.

Live SSM execution is the one thing I could not run locally, so that integration point is reasoned rather than executed.

beardthelion:
@kevincodex1 ready for merge. Both jatmn and I approved, 19/19 green on 888fab6, no unresolved threads. Note it is a breaking change, so it wants the release notes that go with it.

kevincodex1 (approved):

PR #332, opened 2026-08-14
docs: correct security and workspace status

anandh8x:
## Summary

- Correct the workspace inventory to include all crates.
- Align the README and security policy with the existing read-visibility controls and current authorization and review limitations.
- Update near-term documentation priorities to match the implementation.

## Why

The public documentation described workspace contents and security behavior that no longer match the current code, which could lead operators and contributors to make incorrect assumptions.

## Validation

- `git diff --check`
- Source review against the current authorization and visibility implementation
- `cargo test --locked --workspace` — 821 tests passed; two unchanged node tests failed in this environment, including when run in isolation after their 30-second test budgets.

## Summary by CodeRabbit

* **Documentation**
  * Updated the README and contribution guide with the current workspace layout and repository/path-scoped read visibility.
  * Clarified security limitations and safeguards covering write authorization, UCAN trust and revocation, private reads, GraphQL mutations, and pull-request approvals.
  * Updated the maintainer roadmap with priorities for authorization, trusted delegation, metadata visibility, and irreversible publication.
  * Added a historical-snapshot notice to the open-source readiness audit and linked current security guidance.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,9 @@ crates/
 ├── gitlawb-core/       crypto primitives (DID, CID, HTTP sigs, UCAN, ref certs)
 ├── gitlawb-node/       axum HTTP server, git smart HTTP, P2P, GraphQL
 ├── gl/                 CLI — identity, repos, MCP server, Base L2 names
-└── git-remote-gitlawb/ git remote helper for gitlawb:// URLs
+├── git-remote-gitlawb/ git remote helper for gitlawb:// URLs
+├── gitlawb-attest/     provenance attestations for ref-update certificates
+└── icaptcha-client/    client for the iCaptcha proof-of-intelligence service
 docs/                   Operator guides
 scripts/                Build helpers
 ```
@@ -106,7 +108,7 @@ cargo test -p gitlawb-node
 
 - **TypeScript SDK** (`@gitlawb/sdk`) — client library for the HTTP API
 - **Python SDK** (`gitlawb`) — for ML/agent pipeline integration
-- **UCAN chain validation** — complete the auth middleware
+- **UCAN authorization** — add trusted issuer anchoring, capability checks, and authorization-aware revocation
 - **Filecoin storage tier** — wire up cold storage deals
 - **Documentation** — guides, tutorials, API examples
 - **Node operators** — run a public node and report issues
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Good today:
 - Bare git repository storage.
 - Git smart-HTTP clone/fetch/push.
 - RFC 9421-signed writes.
-- Repository and path-scoped read visibility enforcement, including 404-shaped denials.
+- Repository and path-scoped visibility enforcement for repository and Git content reads, with 404-shaped repository denials.
 - DID identities.
 - `gl` CLI workflows.
 - libp2p peer discovery/gossip foundation.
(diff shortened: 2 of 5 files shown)


beardthelion (dismissed):
The corrections here are real and I verified each one against main: six workspace crates, the signed JSON UCAN wire format, `GITLAWB_ENFORCE_OWNER_PUSH` defaulting false, capability grants not consulted by write authorization, revocation not checked at signature time, merges not enforcing approvals. The private-read section had genuinely gone stale and this is the right direction.

What holds the round is one deletion and three claims that are stronger than the code supports. All four are in the paragraph describing read enforcement, which is the part operators use to decide what is safe to host.

## Findings

- [P1] Keep the warning against storing secrets on a public node
  `SECURITY.md:75`
  The read gate governs what this node serves, not what has already left it. A push to an announceable repo pins objects to IPFS/Pinata and anchors each ref update to Arweave, and there is no unpin or retraction path in the node source. `announce` is evaluated at push time in `api/repos.rs:2489` and never re-evaluated, so tightening a repository's visibility afterwards does not retract what was already published. Replacing that warning with "Callers can read only repositories and paths their visibility rules permit" removes the only guidance covering the irreversible case.

- [P2] Scope the 404 sentence to repository existence, since withheld path globs are published
  `SECURITY.md:74`
  `withheld_paths` in `api/visibility.rs:209` runs the root read gate and then returns the denied globs themselves, and it is mounted with an optional auth extractor, so an anonymous caller on a public repo with a private subtree learns that subtree's name. That is deliberate, since sparse-clone clients need it. Repository existence is hidden; withheld-path existence is not, and an operator could reasonably read the current wording as license to put sensitive information in a directory name.

- [P2] Qualify the read-scope claim: three read surfaces take no caller and no visibility check
  `SECURITY.md:75`
  `GET /api/v1/tasks` and `GET /api/v1/tasks/{id}` are mounted at `server.rs:81-82` with no auth layer and no visibility check, and `task_to_json` at `api/tasks.rs:83` serializes `ucan_token`, so an unauthenticated caller reads a credential. `/api/v1/ipfs/pins` and `/api/v1/arweave/anchors` are likewise ungated, and the anchors response carries `repo`, `owner_did`, `ref_name` and both SHAs. The comment at `server.rs:216` already concedes the pin index gap. Worth naming these as a remaining boundary rather than stating the rule without exception.

- [P2] Reconcile the threat model with the new write-authorization text
  `SECURITY.md:118`
  "Unauthorized writes" still sits under "designed to be secure against" while the paragraph added at line 56 says the default does not require the pusher to be the repository owner. Scoping that bullet to authentication and pointing at Known Limitations would settle it. This one resolves itself if #330 lands first, but it contradicts the file as written today.

- [P3] Fix the UCAN row in the Cryptographic Primitives table
  `SECURITY.md:111`
  The table still reads `JWT (Ed25519 signatures)` while line 34 now says signed JSON. `Ucan::encode` is `serde_json::to_string` and the signature is a base64url field inside the JSON object, so the new prose is right and the table is the half to change. CodeRabbit flagged the conflict without picking a direction; the code picks it.

- [P3] Carry the correction into the two other docs that still assert the old state
  `CONTRIBUTING.md:29-35`, `docs/MAINTAINER-ROADMAP.md:42`
  CONTRIBUTING still draws a four-crate tree with no `gitlawb-attest` or `icaptcha-client`, and line 109 still asks for help completing UCAN chain validation. The roadmap carries verbatim the sentence this PR deletes from the README, under its own principle that every public claim in docs should match current behavior. `docs/OSS-READINESS-AUDIT.md:93` says the same thing but reads as a dated snapshot, so a superseded note is enough there.

Separately, `config.rs:58` and the startup warning at `main.rs:87` both still say per-repo read enforcement is not wired, which now contradicts this PR directly. An operator who sets `GITLAWB_PUBLIC_READ=false` after reading the new text gets told by the node that the feature does not exist. That is a code change rather than a docs one, so please open a follow-up for it rather than pulling it in here.

On ordering: #330 and #331 both touch README.md and each one invalidates a paragraph this PR adds, so landing this first and rebasing their README hunks onto it is the cheaper direction.

beardthelion (approved):
All six findings from the last round are addressed at `0b47e1fc`, and I checked each corrected claim against the code rather than against the prose. The irreversibility warning is back and scoped correctly, the 404 sentence is now limited to repository existence, the withheld-path caveat matches what `withheld_paths` actually returns to a caller who can read the root, and the three ungated metadata routes are named with the UCAN exposure called out. The threat-model bullet reads "unauthenticated" now, the primitives table says signed JSON, and the crate-count and UCAN corrections carried into CONTRIBUTING and the roadmap with a dated-snapshot note on the OSS audit.

One follow-up from last round is still open and stays out of this PR: `config.rs:58` and the startup warning in `main.rs:87` both still say per-repository read enforcement is not wired, which this PR now contradicts. Please open an issue for it so it does not get lost.

Worth noting the full CI suite has not executed on this head (fork approval gate), so the checks showing green are triage only. I will approve the run before merging.

kevincodex1 (approved):

PR #334, opened 2026-08-14
fix(node): bound replay of authenticated gossip ref-update events

beardthelion:
Bounds replay of authenticated gossip ref-update events. Stacked on #325, which is what makes these events authenticated in the first place.

## Why this is separate from #325 rather than folded into it

With #325 merged and this not yet merged, an attacker's capability is strictly lower than it was before #325: forging arbitrary ref-updates for any peer required no key at all, and afterwards they are reduced to replaying events a legitimate node actually published, with the same repo, refs, and shas. The intermediate state is better than the pre-state, not worse, so the split does not leave a window where the system is more exposed than it was.

That is the test I applied, and it is the one the #173/#321 split failed: that resolver hardening withheld the provider CID for unrepaired rows, so deploying it alone actively broke existing pins, and the ordering argument only held as long as nothing shipped in the gap. Writing this out here rather than leaving it in a comment is the other half of that lesson.

## What it does

Two layers, following the shape production gossip systems use, where a dedup-layer fix and an authentication-layer fix are paired rather than either being treated as sufficient.

**A freshness window** on the already-signature-covered `timestamp`: 600 seconds into the past, 60 into the future. Two comparisons, never a distance. An `abs()` window accepts a future-dated event as readily as a late one, and a future-dated event also pins a seen-set slot while sitting outside the past-window check until the clock catches up.

**A bounded seen-set** keyed on SHA-256 of the canonical signing bytes, not the raw wire bytes. That distinction is the whole point: one signature verifies against a family of wire encodings, so a raw-bytes key deduplicates nothing. The frozen pre-version artifact and the same artifact with `"v":0` injected are different lengths, both verify against one signature, and produce identical signing bytes.

The guard runs immediately after signature verification and above the per-author debit. Below the debit it would still stop the duplicate row and the duplicate sync while letting every replay drain the victim author's budget, which is the harm this is about. It runs on the verified path only: unsigned bytes are predictable, so applying dedup there would let an attacker pre-send a victim's expected event and have the genuine one dropped as a replay.

A key is recorded only on acceptance, through a reservation whose drop releases it, so a transient write failure does not permanently burn an event's slot. At capacity the guard fails open and counts the degradation, because a saturated set that dropped all fresh gossip would convert a loud resource attack into quiet mesh-wide censorship.

Also included: `check_created` in gitlawb-core used a symmetric `abs()` window, so it tolerated 300 seconds of future-dating and a signer could roughly double a signature's effective validity by stamping forward. It is now two comparisons like the gossip path. Adding a freshness window obliges auditing the siblings, which is how that surfaced.

## What is proven rather than argued

Every guard here is backed by a mutation that turns a named test red, and they are present-but-wrong re-implementations rather than deletions, because each of these compiles and reads sensibly: keying on raw wire bytes, keying fresh per ingest, keying on `(repo, ref, sha)`, an `abs()` freshness window, recording before accept, failing closed at saturation, placing the guard below the author debit, a sweep that stops evicting, confirming on a write failure, and treating an expired entry as absent at capacity.

Two are worth calling out because the obvious version of each test cannot catch what it names. The revert case needs three events, since two never collide under a `(repo, ref, sha)` key. The layer-composition case has to run at the future-skew edge, since a present-stamped probe is rejected by freshness anyway and stays green under a shortened retention.

An attacker-supplied `created` could overflow the window subtraction and panic a debug build; that is fixed with saturating arithmetic and a test driving all four extremes of the type.

## Known open, not closed here

A compromised or malicious registered signer can still mint fresh signed events with fresh timestamps, each of which gets a distinct key and passes freshness. This bounds third-party replay of a captured event; it is not an aggregate write-volume defense and should not be described as one.

The seen-set is in-process, so a restart readmits any event still inside its freshness window, once per restart. Bounded by the window rather than unbounded, and documented on the guard.

`POST /api/v1/sync/notify` reaches the same two sinks with no dedup. Bounded differently there, since that handler still accepts unsigned notifications naming any known peer DID, so forgery already beats replay and the per-IP limiter is the real bound. Worth closing separately.

The freshness window is an availability bound with no resume: a peer returning from a partition longer than the window drains a backlog stamped at push time and every one of those events is dropped, with no republish and no repair path. That is a product decision about what federation should do with stale-but-genuine updates, not something to patch quietly.

Full suite, clippy and fmt clean locally; CI has the authoritative run for this head.

### DIFF
--- a/crates/gitlawb-core/src/http_sig.rs
+++ b/crates/gitlawb-core/src/http_sig.rs
@@ -134,14 +134,21 @@ impl HttpSignature {
 
         let now = Utc::now().timestamp();
 
-        let age = now - self.created;
+        // Saturating, not plain subtraction. `created` is parsed as an
+        // unrestricted i64 straight from the Signature-Input header, so a
+        // sender picks it: at i64::MIN the past-side subtraction overflows,
+        // which panics a debug build and wraps a release one into an absurd
+        // value that can read as acceptable. Saturating gives the answer the
+        // check wants at both extremes, since a timestamp that far away is
+        // refused by whichever side it saturates toward.
+        let age = now.saturating_sub(self.created);
         if age > MAX_AGE_SECS {
             return Err(Error::HttpSignature(format!(
                 "clock skew too large: created {age}s in the past (max {MAX_AGE_SECS}s)"
             )));
         }
 
-        let ahead = self.created - now;
+        let ahead = self.created.saturating_sub(now);
         if ahead > MAX_FUTURE_SECS {
             return Err(Error::HttpSignature(format!(
                 "clock skew too large: created {ahead}s in the future (max {MAX_FUTURE_SECS}s)"
@@ -371,6 +378,36 @@ mod tests {
             r#"sig1=("@method" "@path" "content-digest");keyid="{did}";alg="ed25519";created={created}"#
         );
         HttpSignature::parse(&sig_input, "sig1=:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA:").unwrap()
+    }
+
+    /// Build a signature at an ABSOLUTE `created`, so a test can drive the
+    /// extremes an offset from now cannot reach.
+    fn sig_created_absolute(created: i64) -> HttpSignature {
+        let kp = Keypair::generate();
+        let did = kp.did();
+        let sig_input = format!(
+            r#"sig1=("@method" "@path" "content-digest");keyid="{did}";alg="ed25519";created={created}"#
+        );
+        HttpSignature::parse(&sig_input, "sig1=:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA:").unwrap()
+    }
+
+    /// `created` is parsed as an unrestricted i64 from a header the sender
+    /// writes, so the window check has to survive both ends of the type. Before
+    /// the saturating form this panicked with "attempt to subtract with
+    /// overflow" on the past side, which is a debug-build crash any client
+    /// could trigger by sending one header value.
+    #[test]
+    fn extreme_created_values_are_refused_rather_than_overflowing() {
+        for created in [i64::MIN, i64::MIN + 1, i64::MAX - 1, i64::MAX] {
+            let err = sig_created_absolute(created)
+                .check_created()
+                .expect_err("an extreme created must be refused, not accepted");
+            let msg = err.to_string();
+            assert!(
+                msg.contains("past") || msg.contains("future"),
+                "created={created} must be refused by a named direction, got: {msg}"
+            );
+        }
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):

## Findings

- [P1] Sanitize every rejected timestamp before logging it
  `crates/gitlawb-node/src/p2p/mod.rs:1250`
  `DateTime::parse_from_rfc3339` accepts RFC 3339 fractional seconds with arbitrarily many digits (discarding precision after nanoseconds). A signed timestamp such as an old instant followed by megabytes of fractional digits therefore parses successfully, reaches the `TooOld` or `TooFarFuture` arm, and is copied into `StaleTimestamp`; the swarm loop then logs that reason verbatim. Only the parser-error branch calls `sanitize_for_log`.

  The root cause is treating successful parsing as proof that the original wire representation is safe to expose. The parsed `DateTime` is bounded, but the source string is not. A registered peer can consequently turn bounded gossip deliveries into unbounded log, disk, and log-pipeline traffic despite the per-source event limiter. Apply the same bounded/control-character-safe rendering to all three freshness failure branches, preferably by logging a bounded canonical rendering of the parsed timestamp for `TooOld` and `TooFarFuture` rather than the original wire string. Please add a regression test using an overlong but parseable fractional timestamp that exercises both directions and asserts that the emitted detail is bounded and contains no control characters.

jatmn (changes requested):

## Merge readiness

- [P1] Rebase onto the current target branch before merge
  The PR is currently `CONFLICTING` / `DIRTY`. Its target (`fix/p2p-gossip-ingest-auth`) and head each merged `main` independently and have diverged across `p2p/mod.rs`, `metrics.rs`, and `db/mod.rs`, so this cannot merge as reviewed. Rebase or reconstruct the head on the current target, resolve the conflicts, and request review of the resulting diff.

beardthelion:
Rebased onto the current target. The branch now carries the seven commits replayed on top of `fix/p2p-gossip-ingest-auth` plus four fixups; the old independent merge of main is gone, and the PR diff is back to the three files it owns (`http_sig.rs`, `metrics.rs`, `p2p/mod.rs`).

Verified locally:

- `git range-diff` of the old series against the replayed one shows all six pre-merge commits patch-identical, and the head commit's `p2p/mod.rs` diff is byte-identical to the reviewed head.
- The divergence lived in the dropped merge resolution: `db/mod.rs` is now identical to the target, `metrics.rs` and `http_sig.rs` are unchanged, and `p2p/mod.rs` gained only the target's early-shed block plus its two tests. Those two tests still called `ingest_ref_update` with the pre-guard signature, so a fixup threads a fresh `ReplayGuard` through each call.
- `cargo test -p gitlawb-node --bin gitlawb-node p2p::` 90/90, `cargo test -p gitlawb-core` 105/105, fmt and clippy clean.

Three further commits on the same surface, each with a regression test:

- A live-mesh test drives the publish and receive path end to end: two real swarms on QUIC loopback, a signed ref-update is admitted and stored, an encoding-malleated re-delivery of the same signed content is refused by the replay guard, and an unsigned event is refused under `require_signed`.
- Every `Rejected` and `WriteFailed` reason now renders through the log sanitizer at the warn site, since resolve failures carry wire fields like the claimed `node_did`. And an unsigned event under `require_signed` is refused before its `node_did` is resolved, so it buys no pre-auth work.
- A republish after a partial write failure no longer stores a duplicate `received_ref_updates` row: the row id derives from the replay key, so the retry repairs the missing queue entry without repeating the insert.

jatmn (approved):
LGTM

PR #324, opened 2026-08-12
fix(node): persist the libp2p identity instead of deriving it from the node DID

beardthelion:
Closes #322. The node now generates its libp2p keypair once from the OS RNG and keeps it on disk, rather than recomputing it from a public value on every start.

The key path is configurable (`--p2p-key-path` / `GITLAWB_P2P_KEY`, default `~/.gitlawb/p2p.key`), following the existing `key_path` idiom, and is pinned onto the mounted volume in the Docker and fly configs. That pinning matters: without it the file lands on persistent storage only through home-directory resolution, and a change there would rotate the PeerId on every deploy.

Publishing is atomic. The key is written to a scratch file in the same directory and hard-linked onto the final path, so a crash cannot leave a partial key that fails to load and takes the node off the network, and a concurrent reader cannot observe a half-written file. `hard_link` rather than `rename` because `rename` replaces its destination silently, so refusing to clobber would need a separate check with a gap a concurrent start can land in. On `AlreadyExists` the key that landed is read and used, so two concurrent starts agree.

The file is created `0600` at open rather than chmod'd afterwards, so the secret is never on disk under a wider mode. An existing key whose mode grants group or other access is refused with the observed mode named. The directory is created `0700` and tightened if it is looser; #231 still owns the sibling identity key's own creation path.

## Migration

Every node's PeerId rotates once, on first start after upgrade. The `peers` table keys on `did`, bootstrap uses https URLs, and `from_peer` is provenance only, so nothing is orphaned. Pre-rotation `from_peer` values refer to pre-rotation identities.

## Not in scope

The gossipsub `message_id_fn` still uses `DefaultHasher`. Different severity, different fix, deliberately untouched.

A failure to load the key still logs a warning and continues with p2p disabled, which means a tampered or unreadable key file is a silent network outage with a healthy `/health`. A comment names that at the call site; changing it is a separate call.

## Summary by CodeRabbit

- **New Features**
  - Added persistent P2P identity keys, preserving node identity across restarts.
  - Added configurable key locations and deployment defaults.
  - Added configurable limits for legacy IPFS probing, scan rows, and CID resolution.
  - Enabled owner-only push enforcement by default, with an explicit opt-out.
  - Added automatic background repair handling for legacy CID records.

- **Documentation**
  - Documented identity persistence, permissions, volume requirements, upgrade guidance, IPFS limits, and legacy-pin handling.

- **Bug Fixes**
  - Improved key validation, secure permissions, atomic writes, and protection against invalid or unsafe key files.

PR #335, opened 2026-08-14
fix(node): refuse a p2p identity key owned by another user

beardthelion:
**Stacked on #324, deliberately.** The base is `fix/p2p-keypair-derivation`, not `main`. Everything this touches (`read_p2p_keypair`, `ensure_key_dir`, `load_or_create_p2p_keypair`) is introduced by #324 and does not exist on main, so it cannot branch independently.

Deployable on its own. Merging #324 without this leaves the key exactly as #324 ships it, persisted with owner-only modes and no ownership check, which is already better than deriving it from public data. This adds a check on top rather than restoring something #324 removed.

## The defect

#324 pins the key file to `0600` and its directory to `0700`, and verifies both on load. Neither check asks who owns them. A `0600` file owned by a different user passes every check on the read path, and the node adopts it, so that user chooses the node's libp2p identity. Mode answers who may read the file; it says nothing about who may replace it.

## The fix, and the part worth reviewing

The leaf check is the easy half: refuse when the key file or its directory is owned by someone other than the running user.

The ancestor walk is the half I would look at. Checking only the leaf is not enough, and the way it fails looks safe. A user who owns `/home/them/base` can point the node at `/home/them/base/keys/p2p.key`. On first start the node creates `keys` and the key itself, so both are node-owned, `0700` and `0600`, and pass every leaf check. That owner never needs to own either one. They can rename `keys` aside, let the node generate a fresh identity in a new `keys`, and move the old directory back before a later restart. Both directories pass at every point, and they decide which identity the node presents and when it rolls back.

So the trust boundary is the whole existing chain, not the leaf. Walking up from the deepest component that exists, every ancestor must be owned by this user or by root, and must not be world-writable unless it is sticky.

Root counts as trusted on purpose. Requiring every ancestor to be node-owned would refuse `/data/keys` under a root-owned `/data`, and `/` itself, which is most real deployments. Root can already replace the binary, so treating it as an attacker here buys nothing.

The read also became a single `O_NOFOLLOW` descriptor feeding the ownership check, the mode check, and the read. Before this, the mode came from a separate `metadata` call and the bytes from a later `fs::read`, which is a stat-then-read window and follows a symlink at the final component.

## Testing it

A test cannot chown to another user without root, so both directions run through a `cfg(test)` override of the effective uid, and the override is thread-local so an armed test cannot disturb the ones beside it. The seam itself carries a mutation, because a seam that stopped consulting `geteuid` would leave every other ownership test passing against nothing.

Eight mutations, each reverting one guard and requiring the failure to match a named message, so a red is attributable to the property rather than to removing the seam the test injects at. Two of them exist because earlier versions were misattributed: gating the ordering check on the directory mode made the `0777` case skip the check entirely, so the observed failure was "no refusal happened" rather than "the repair ran first", and the uid-versus-gid mutation was invisible on a machine where the two are equal until the test named the expected owner explicitly.

## Known limitation, and why it is not in here

The refusal is still non-fatal. It reaches `main.rs`, which logs it and continues without p2p, so a foreign-owned key takes the node off the network for the run rather than stopping it. Moving the check to `Config::validate`, where it would stop the process, is its own change. It needs a look at what actually owns the data volume on a deployed node first: a boot-fatal version against an unexpected volume ownership refuses to start, and that is a worse outcome than the one it prevents.


jatmn (changes requested):

## Merge readiness

  The PR head `66fb3ffc954fc094733465b6e1bd931eb5609a42` is not contained in live `main` (`e4c7458fb3da87c8bca81e43e2db18a3e9a5a60e`); its merge base is `96d8123e0f84e85a7b0234ad84f5e603a5a303aa`. The current target does not alter this module, but the stale head still needs a rebase and a rerun of the resolved diff's checks before merge.

## Findings

- [P1] Reject writable group ancestors as well
  `crates/gitlawb-node/src/p2p/mod.rs:343`
  The new check rejects only other-writable ancestors, so a `0770` service directory is accepted even when another account belongs to its group. That account can rename `base/keys` while the node is stopped, retain the directory containing an earlier legitimate key, and put a different node-created `keys` directory in its place before the next start. Each replacement leaf is still owned by the node and is `0700`; its `p2p.key` is likewise node-owned and `0600`. Consequently, the leaf ownership and mode checks all pass, but the group member can select or roll back the PeerId on every restart.

  Please address the root cause—untrusted write authority over any ancestor—not just the leaf mode. Group write is an identity-control capability in this path just as other-write is. The conservative fix is to reject both group- and other-writable ancestors (with the existing sticky-directory exception only where its semantics are deliberately supported). If shared group directories are a required deployment model, introduce an explicit, documented trusted-group policy and test both its allowed and denied members; accepting every group-writable ancestor cannot establish the claimed trust boundary.

- [P1] Bind the ancestor validation and key operations to trusted directory handles
  `crates/gitlawb-node/src/p2p/mod.rs:317`
  The ancestor walk skips missing components and validates only pathnames, then `DirBuilder`, metadata, scratch creation, and reads reopen those pathnames. For a configured path such as `/tmp/created-later/keys/p2p.key`, `/tmp` passes because it is sticky and `created-later` is skipped because it does not yet exist. An attacker can create `created-later` after the walk but before recursive creation, so the node creates a correctly owned `keys` leaf under the attacker's parent. The attacker can then swap that leaf between restarts and choose the identity. Directory symlinks have the same root problem: `metadata` follows them, while `O_NOFOLLOW` protects only the final `p2p.key` component; an attacker-owned intermediate symlink can be retargeted after validation.

  Please fix this as a pathname-resolution problem, rather than adding another check after `DirBuilder::create`. Each check on a pathname creates another check-then-use window. Resolve and create the directory chain through trusted directory file descriptors without following untrusted components (for example, `openat`/`mkdirat`-style operations with no-follow semantics), verify ownership and permissions on those descriptors, and perform scratch-file creation, publication, and existing-key opens relative to the verified directory descriptor. Add race-oriented coverage for a missing intermediate component created after validation and for an intermediate directory symlink that changes target; both tests should prove that no attacker-selected key can be loaded or created.

beardthelion:
Closing this: everything it adds is already on #324, which has since carried the same work through four more review rounds.

Checked rather than assumed. All four of this branch's own commits have their distinctive markers present in #324's current head (`01e04637`):

| commit | landed in #324 as |
|---|---|
| `3a296485` scrub the serialized key from memory | `Zeroizing` in the read path, `zeroize` in the manifest |
| `6dde36b0` refuse a key or key directory owned by another user | `foreign_ownership_error`, 5 call sites |
| `5a93f243` close the ancestor path and the stat/read window | the descriptor-anchored walk, plus a single-handle read |
| `66fb3ffc` make the ownership guards observable | all four ownership tests, by name |

I went through `5a93f243` line by line, since it and #324 solve the same problem by different means. Its five stated behaviors are all present: ancestors are walked before anything is created (`verify_and_create_ancestor_chain`, called before leaf creation), root counts as trusted (`owner != euid && owner != 0`), the key open refuses a symlink at the final component (`O_NOFOLLOW` is added unconditionally in `open_child`), the stat and the read are the same handle (`read_p2p_keypair_from` takes the `File`, calls `file.metadata()`, and reads from that same handle), and the ordering test asserts the mode is untouched rather than matching on a message that appeared either way.

The one place the two differ, #324 is stricter. This branch deliberately allowed group-write on ancestors (`mode & 0o002`), reasoning that refusing it would reject an ordinary umask-002 directory, and it named the residual honestly: someone in an ancestor's group can still rename the key directory. #324 uses `mode & 0o022`, so that residual is closed rather than carried, and the ordinary shapes still boot (the ancestor test covers `0755` and sticky `1777`). Nothing here would restore something missing; on that axis it would argue for loosening.

The 14 commits this branch shows ahead of #324 are #324's own older history under different SHAs, from before it was rebased. A trial rebase onto the current head conflicts on the first of them, which adds the persistent-key feature #324 already has. The branch is also 121 commits behind main.

One thing worth not losing: `5a93f243` added a test named for a root-owned ancestor. #324 enforces root-trust in the predicate but has no test named for that case, only the `0755` and sticky shapes. Behavior is there, dedicated coverage is not.

Issue #340, opened 2026-08-15
Outbound clients never validate the address a hostname resolves to


beardthelion:
`is_public_http_url` (`crates/gitlawb-node/src/api/peers.rs:109`) classifies a URL by its host *string*. When the host is an IP literal that works. When it is a name, nothing resolves it, so a hostname whose A or AAAA record points at `127.0.0.1`, `169.254.169.254`, or RFC1918 space passes every check we have.

`POST /api/v1/peers/announce` is unauthenticated, and the resulting `http_url` is used for outbound sync-notify fan-out, so the caller who chooses the name is not required to be anyone in particular.

There is a second half. Validation and connection resolve independently, so even a name that resolves to public space at announce time can resolve somewhere else when the request is actually made. A syntactic gate cannot see that by construction, which is why OWASP's SSRF guidance treats hostname-only validation as insufficient rather than as a partial win.

### Why this is the unfinished third step

We have closed SSRF here twice, both times at the syntactic layer. #78 added the host-class gate and the peer prune; #93 and #140 stopped the gossip path following redirects into private space. Neither addressed what a name resolves to. Nothing in the tree records a decision that syntax-only is sufficient, and #333's description states the same conclusion from the other direction, deferring resolved-address validation to the fetch worker that will consume `upstream_url`.

### Fix direction

`reqwest::ClientBuilder::dns_resolver` takes an `Arc<R: Resolve>`, and `reqwest::dns` carries no feature gate, so it is available in our `default-features = false` build of 0.12.28. A resolver that applies the address policy and drops non-public results closes the name case and the rebinding case together, because the check then happens on the addresses actually used to connect.

Two things to get right:

1. **Reuse the address checks we already have rather than adding a dependency.** Extract the IP-classification half of `is_public_http_url` into a function over `IpAddr` and call it from both the existing literal path and the resolver. `IpAddr::is_global` is still nightly-only, so std is not an option at our MSRV, and our hand-rolled ranges are already covered by tests.
2. **Both clients, not one.** `build_http_client` (`crates/gitlawb-node/src/main.rs:1173`) is the named shared builder, and `sync.rs:149` builds a second client of its own. Both set `redirect::Policy::none()`. A resolver policy attached only to the first leaves the sync worker on the old behavior, which is the same shape as the bug #140 fixed.

Keep the string checks. They stay useful as a cheap reject before any lookup happens, and defense in depth is the point; they are just not the boundary.

The narrower predicate gaps in that same function are tracked separately.

PR #346, opened 2026-08-15
fix(node): close three host-class gaps in the public-URL gate

### LINKED ISSUE #339: Host gate accepts bare single-label names, fec0::/10, and a double trailing dot
`is_public_http_url` (`crates/gitlawb-node/src/api/peers.rs:109`) is the shared host-class gate for every operator-supplied outbound URL: peer announce, webhook creation, the boot-time peer prune. Three host forms get past it that the checklist around them clearly means to exclude.

**Bare single-label names.** The gate tests `.local` and `.internal` as suffixes, so a name with no dot never matches:

```
http://local/      ACCEPTED     http://foo.local/      rejected
http://internal/   ACCEPTED     http://foo.internal/   rejected
http://intranet/   ACCEPTED
http://wpad/       ACCEPTED
http://metadata/   ACCEPTED
```

A dotless name is resolved through whatever search domain the host is configured with, so on a node with `search corp.example.com` a peer row pointing at `http://internal/` reaches `internal.corp.example.com`. `POST /api/v1/peers/announce` is unauthenticated, and a peer row's `http_url` is used for outbound sync-notify fan-out. ICANN's SAC053 and the IAB statement on dotless domains both treat these as names that should not be trusted to mean what they look like.

**`fec0::/10`.** The IPv6 arm rejects `fc00::/7` and `fe80::/10` with `(s[0] & 0xfe00) == 0xfc00 || (s[0] & 0xffc0) == 0xfe80`. Site-local is `0xfec0`, and `0xfec0 & 0xffc0` is `0xfec0`, so it falls through. RFC 3879 deprecated the range in 2004, which is an argument for rejecting it rather than for ignoring it.

**A double trailing dot.** The normalizer strips exactly one root dot, so `localhost..` becomes `localhost.`, which matches neither `== "localhost"` nor `.ends_with(".localhost")` and is accepted. I could not resolve that name here (`getaddrinfo` rejects the empty label), so this is a gap in the predicate rather than a demonstrated reachable target, and it is listed for completeness.

### Fix direction

Reject a host with no dot that is not an IP literal, widen the site-local mask, and strip trailing dots to fixation instead of once. Table-driven cases for the accepted forms above, with the dotted forms kept as negative controls so the suffix checks stay proven.

### Related, separate work

The peer surface has `prune_non_public_peers` (`crates/gitlawb-node/src/db/mod.rs:2457`) to sweep rows written before announce-time validation existed. Webhooks have no equivalent, so a webhook row registered before any tightening is still delivered to. Owner-gated and lower reach than the announce path, but it is the reason a widened predicate does not fully take effect on existing data. Worth doing in the same area, not in the same patch.

This issue covers the predicate only. The larger gap, that nothing validates what a hostname actually resolves to, is tracked separately.

beardthelion:
Closes #339.

`is_public_http_url` is the shared host-class gate for every operator-supplied outbound URL: peer announce, webhook creation, and the boot-time peer prune. Three host forms got past it that the checks around them plainly meant to exclude.

**Dotless names.** `.local` and `.internal` are suffix tests, so a single label matches neither, and `http://internal/`, `http://wpad/`, `http://metadata/` were accepted and left to the resolver to complete from its search domain. `POST /api/v1/peers/announce` is unauthenticated and a peer row's `http_url` drives outbound sync-notify fan-out, so the caller choosing that name need not be anyone in particular.

The new rule sits on the branch where the host did **not** parse as an IP literal. That placement is the whole trick: a bracketed IPv6 literal is also dotless, and the existing 6to4 and NAT64 accept cases pin that it stays accepted.

**`fec0::/10`.** The IPv6 arm missed it because the link-local mask does not cover it: `0xfec0 & 0xffc0` is `0xfec0`, not `0xfe80`. RFC 3879 deprecated the range in 2004, which argues for rejecting it rather than ignoring it.

**Repeated root dots.** Trailing dots were stripped once, so `localhost..` reduced to `localhost.` and matched neither the equality nor the suffix check. Stripping to fixation closes it. I could not resolve that name locally, so it is a predicate gap rather than a demonstrated reachable target.

## Verification

Each of the three is covered by a test that goes red when its own production line is reverted, checked one at a time rather than as a batch:

| reverted | test that reddens |
|---|---|
| the dotless branch | `rejects_dotless_single_label_hosts` |
| the `fec0::/10` clause | `rejects_deprecated_site_local_v6` |
| strip-to-fixation | `rejects_repeated_trailing_root_dots` |

The dotted forms stay in the tables as negative controls so the suffix checks remain proven, and the IPv6 accept cases guard against the dotless rule over-rejecting.

`cargo test -p gitlawb-node --bin gitlawb-node --locked` is 826 passed / 0 failed, with `fmt` and `clippy --workspace --all-targets -- -D warnings` clean.

## Scope

Predicate only. The larger gap, that nothing validates the address a hostname actually resolves to, is #340 and needs a resolver policy on both outbound clients rather than a change here.

This touches the same span of `peers.rs` as #333, so whichever lands second wants a rebase. The conflict is adjacent lines, not competing logic: #333 adds `.localhost` to the suffix list, this adds a dotless branch further down. The `node.localhost..` case is deliberately not asserted here, since `.localhost` is not a rejected suffix until #333 lands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL validation by handling hostnames with repeated trailing dots.
  * Rejected deprecated site-local IPv6 addresses.
  * Rejected single-label hostnames that are not valid IP addresses.
  * Added coverage for these validation scenarios.

### DIFF
--- a/crates/gitlawb-node/src/api/peers.rs
+++ b/crates/gitlawb-node/src/api/peers.rs
@@ -118,11 +118,11 @@ pub fn is_public_http_url(raw: &str) -> bool {
         Some(h) => h.to_ascii_lowercase(),
         None => return false,
     };
-    // Drop a single trailing dot (FQDN root): `localhost.` resolves the same as
+    // Drop trailing dots (FQDN root): `localhost.` resolves the same as
     // `localhost`, so normalize before the suffix/equality checks below.
-    if let Some(stripped) = host.strip_suffix('.') {
-        host = stripped.to_string();
-    }
+    // Strip to fixation rather than once, or `localhost..` reduces to
+    // `localhost.` and matches neither the equality nor the suffix check.
+    host.truncate(host.trim_end_matches('.').len());
     if host.is_empty()
         || host == "localhost"
         || host.ends_with(".local")
@@ -167,8 +167,13 @@ pub fn is_public_http_url(raw: &str) -> bool {
             }
             std::net::IpAddr::V6(v6) => {
                 let s = v6.segments();
-                // fc00::/7 (unique-local) or fe80::/10 (link-local)
-                if (s[0] & 0xfe00) == 0xfc00 || (s[0] & 0xffc0) == 0xfe80 {
+                // fc00::/7 (unique-local), fe80::/10 (link-local), or fec0::/10
+                // (site-local, deprecated by RFC 3879 and so never a legitimate
+                // peer address, but still routable on networks that kept it).
+                if (s[0] & 0xfe00) == 0xfc00
+                    || (s[0] & 0xffc0) == 0xfe80
+                    || (s[0] & 0xffc0) == 0xfec0
+                {
                     return false;
                 }
                 // Any NAT64 address (64:ff9b::/32) that is not the cleanly
@@ -183,6 +188,13 @@ pub fn is_public_http_url(raw: &str) -> bool {
                 }
             }
         }
+    } else if !host.contains('.') {
+        // A name with no dot at all. The `.local` / `.internal` rules above are
+        // suffix checks, so a single label slips both, and the resolver decides
+        // what it means by appending a search domain we cannot see from here.
+        // Only reachable when the host did not parse as an IP literal, which is
+        // what keeps bracketed IPv6 (also dotless) on the accepting path.
+        return false;
     }
     true
 }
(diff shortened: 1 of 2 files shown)


jatmn (approved):
LGTM

beardthelion:
@kevincodex1 this one is ready when you are. jatmn approved, 19/19 checks green on ceadf8b, no open threads.

jatmn (approved):
LGTM

beardthelion:
@kevincodex1 LGTM

PR #363, opened 2026-08-17
docs: document the two-tier storage and pinning model

andreolf:
## Summary

Adds `docs/STORAGE-AND-PINNING.md` documenting how a node stores git objects and keeps them available. Docs only — no behavior change.

## Motivation & context

`RUN-A-NODE.md` covers provisioning, staking and rewards, but there's no doc on how a node actually stores git objects and keeps them available — a question that comes up often (e.g. "where is the data stored?"). No linked issue; docs-only addition.

## Kind of change

- [x] Docs

## What changed

Crate touched: none (docs only, `gitlawb-node` behavior described).

- New `docs/STORAGE-AND-PINNING.md` covering:
  - The two opt-in sinks: hot = local Kubo/IPFS (`ipfs_pin.rs`, `GITLAWB_IPFS_API`), warm = Pinata/Filecoin (`pinata.rs`, `GITLAWB_PINATA_JWT`).
  - Env vars and defaults.
  - Post-push execution off the critical path, the shared `max_concurrent_pin_tasks` admission semaphore (defers rather than sheds), `PIN_BATCH_BUDGET`, Pinata post-acquire re-derivation for O(refs) memory bounding, and per-sink best-effort de-duplication.

## How a reviewer can verify

Docs only; verify by reading against the referenced modules:

```sh
sed -n '1,80p' docs/STORAGE-AND-PINNING.md
grep -n "GITLAWB_IPFS_API\|GITLAWB_PINATA_JWT\|max_concurrent_pin_tasks" crates/gitlawb-node/src/config.rs
```

## Before you request review

- [x] `cargo test --workspace` passes locally (N/A — docs only, no code changed)
- [x] New behavior is covered by tests (N/A — docs only)
- [x] `cargo fmt --all` and `cargo clippy` clean (N/A — docs only)
- [x] Commit titles use Conventional Commits (`docs(...)`)
- [x] Docs updated (this PR is the doc)

## Notes for reviewers

- Sourced by reading `ipfs_pin.rs`, `pinata.rs`, `api/repos.rs`, and `config.rs`; happy to correct anything mischaracterized.
- Addressed CodeRabbit's note: de-dup wording now describes it as best-effort (check-upload-record is not atomic), not a zero-redundant-upload guarantee.
- Separate follow-up (not in this PR): the two sinks are invoked as hardcoded call paths with no shared backend interface — opened as a discussion issue.

## Summary by CodeRabbit

* **Documentation**
  * Added documentation describing optional local IPFS and Pinata storage tiers for post-push Git object pinning.
  * Documented configuration, asynchronous processing, timeouts, deferred work, memory limits, deduplication, and durability characteristics.
  * Added a link to the new storage and pinning guide in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,6 +392,7 @@ Read:
 
 - [`docs/RUN-A-NODE.md`](docs/RUN-A-NODE.md)
 - [`docs/ECONOMICS.md`](docs/ECONOMICS.md)
+- [`docs/STORAGE-AND-PINNING.md`](docs/STORAGE-AND-PINNING.md)
 
 Use a dedicated low-balance operator wallet. Do not use a treasury wallet as the heartbeat key.
 
--- a/docs/STORAGE-AND-PINNING.md
+++ b/docs/STORAGE-AND-PINNING.md
@@ -43,9 +43,11 @@ slot indefinitely. The Pinata tail re-derives its object list only *after*
 acquiring a slot, which bounds outstanding memory to O(refs) rather than
 O(pushes × objects).
 
-De-duplication is per sink: the `pinned_cids` and `pinata_cids` tables record
-what each sink already holds, so re-pushing unchanged objects does no redundant
-upload.
+De-duplication is per sink and best-effort: the `pinned_cids` and `pinata_cids`
+tables record what each sink already holds, so later pushes normally skip objects
+whose successful pin is already recorded. The check-upload-record sequence is not
+atomic, so concurrent post-push tasks for the same object, or a failure to record
+after a successful upload, can still cause a repeat upload attempt.
 
 ## Durability notes
 
--- a/docs/STORAGE-AND-PINNING.md
+++ b/docs/STORAGE-AND-PINNING.md
@@ -16,9 +16,12 @@ both configured.
 | Warm | Pinata (Filecoin-backed) | `pinata.rs` | `GITLAWB_PINATA_JWT` set | Off-node durability + public IPFS gateway reachability |
 
 If a sink's config value is empty, its **pin** paths are no-ops — so leaving
-`GITLAWB_PINATA_JWT` unset simply disables warm-tier pinning. Note this applies to
-pinning only, not reads: the hot-tier read path (`ipfs_pin::cat`) returns an error
-rather than a no-op when `GITLAWB_IPFS_API` is unset, so a node that serves the
+`GITLAWB_PINATA_JWT` unset disables warm-tier *uploads*, though not the tail's
+cost: the post-push Pinata tail still takes a global pin permit and runs the full
+object-list re-derivation walk before the upload step returns empty, since its
+gate is the announce/walk state, not the JWT. Note the no-op applies to pinning
+only, not reads: the hot-tier read path (`ipfs_pin::cat`) returns an error rather
+than a no-op when `GITLAWB_IPFS_API` is unset, so a node that serves the
 encrypted-blob read endpoint needs Kubo configured.
 
 ## Configuration
@@ -43,7 +46,13 @@ Both tiers share a single global **pin admission semaphore**
 saturated, a pin loop waits for a slot instead of dropping the pin. Each batch is
 bounded by `PIN_BATCH_BUDGET` (120s) so the pin batch itself cannot hold a slot
 indefinitely. The Pinata tail re-derives its object list only *after* acquiring a
-slot, which bounds outstanding memory to O(refs) rather than O(pushes × objects).
+slot, which bounds *its* outstanding memory to O(refs) rather than
+O(pushes × objects) — but that bound is the warm tail's alone. A **hot-tier** pin
+loop takes an owned object list materialized by its caller *before* it waits for
+a slot, so every hot loop parked on a saturated pool is holding its full list:
+the hot side of the parked-list memory is O(pushes × objects), capped per repo by
+`EncryptInflight` but not across repos. Do not size a node's memory from the
+semaphore knob (README says the same).
 Note the budget covers the pin batch, not the preceding object-list re-derivation
 walk: that walk holds the slot too, and bounds only each child git process
 individually (no aggregate deadline).
@@ -51,7 +60,11 @@ individually (no aggregate deadline).
 De-duplication is per sink and best-effort, backed by the single `pinned_cids`
 table: the hot tier keys on its `cid`/`sha256_hex` rows and the warm tier on the
 nullable `pinata_cid` column, so later pushes normally skip objects whose
-successful pin is already recorded. The check-upload-record sequence is not atomic,
+successful pin is already recorded. The isolation is one-directional: a hot-only
+pin does not suppress the warm tier (`pinata_cid` stays null), but the hot tier's
+skip check is row *presence*, and a warm-tier pin inserts the row — so an object
+warm-pinned while `GITLAWB_IPFS_API` was unset is skipped by the hot tier after
+Kubo is configured, until something re-pins it. The check-upload-record sequence is not atomic,
 so concurrent post-push tasks for the same object, or a failure to record after a
 successful upload, can still cause a repeat upload attempt.
 
--- a/docs/STORAGE-AND-PINNING.md
+++ b/docs/STORAGE-AND-PINNING.md
@@ -1,0 +1,68 @@
+# Storage and pinning
+
+How a gitlawb node stores git objects and keeps them available. This documents
+the behavior implemented in `crates/gitlawb-node/src/ipfs_pin.rs` and
+`crates/gitlawb-node/src/pinata.rs`; it does not change any behavior.
+
+## Two-tier model
+
+After a push lands, new git objects are pinned to up to two independent sinks.
+Both are **opt-in and independent** — a node runs fine with neither, either, or
+both configured.
+
+| Tier | Sink | Module | Enabled by | Purpose |
+|------|------|--------|-----------|---------|
+| Hot  | Local Kubo (IPFS) | `ipfs_pin.rs` | `GITLAWB_IPFS_API` set | Node-local availability; the node is itself an IPFS peer |
+| Warm | Pinata (Filecoin-backed) | `pinata.rs` | `GITLAWB_PINATA_JWT` set | Off-node durability + public IPFS gateway reachability |
+
+If a sink's config value is empty, every call into that sink is a no-op — so
+leaving `GITLAWB_PINATA_JWT` unset simply disables the warm tier.
+
+## Configuration
+
+| Env var | Default | Meaning |
+|---------|---------|---------|
+| `GITLAWB_IPFS_API` | `""` (disabled) | Base URL of the local Kubo HTTP API, e.g. `http://127.0.0.1:5001` |
+| `GITLAWB_PINATA_JWT` | `""` (disabled) | Pinata bearer JWT enabling the warm tier |
+| `GITLAWB_PINATA_UPLOAD_URL` | `https://uploads.pinata.cloud/v3/files` | Pinata v3 upload endpoint |
+| `GITLAWB_MAX_CONCURRENT_PIN_TASKS` | `8` | Cap on concurrent post-push pin loops across all repos |
+
+## How pinning runs
+
+Pinning happens **after** a push is accepted, not on the push's critical path:
+
+- The **hot** tier pins inline in the post-push encrypt/pin task.
+- The **warm** (Pinata) tier runs in a spawned replication tail, so a slow or
+  unreachable Pinata never blocks the pusher.
+
+Both tiers share a single global **pin admission semaphore**
+(`max_concurrent_pin_tasks`). The pool **defers rather than sheds**: when it is
+saturated, a pin loop waits for a slot instead of dropping the pin. Each batch is
+bounded by `PIN_BATCH_BUDGET` (120s) so a single large or slow push cannot hold a
+slot indefinitely. The Pinata tail re-derives its object list only *after*
+acquiring a slot, which bounds outstanding memory to O(refs) rather than
+O(pushes × objects).
+
+De-duplication is per sink: the `pinned_cids` and `pinata_cids` tables record
+what each sink already holds, so re-pushing unchanged objects does no redundant
+upload.
+
+## Durability notes
+
+- With only the **hot** tier, availability depends on the node (and any IPFS
+  peers that have fetched the CIDs). If the node is down and no peer holds the
+  objects, they are unreachable until it returns.
+- The **warm** tier adds off-node durability via Pinata's Filecoin-backed
+  storage and makes objects reachable through the public IPFS gateway.
+- Running **both** gives a node-local hot copy plus an off-node warm copy.
+
+> The two sinks are currently invoked as separate call paths. Unifying them
+> behind a single pluggable backend interface (to add providers such as direct
+> Filecoin deals or self-hosted clusters without touching push logic) is tracked
+> separately.
+
+## See also
+
+- [RUN-A-NODE.md](RUN-A-NODE.md) — provisioning and running a node
+- `crates/gitlawb-node/src/ipfs_pin.rs` — hot-tier implementation
+- `crates/gitlawb-node/src/pinata.rs` — warm-tier implementation
--- a/docs/STORAGE-AND-PINNING.md
+++ b/docs/STORAGE-AND-PINNING.md
@@ -12,11 +12,14 @@ both configured.
 
 | Tier | Sink | Module | Enabled by | Purpose |
 |------|------|--------|-----------|---------|
-| Hot  | Local Kubo (IPFS) | `ipfs_pin.rs` | `GITLAWB_IPFS_API` set | Node-local availability; the node is itself an IPFS peer |
+| Hot  | Local Kubo (IPFS) | `ipfs_pin.rs` | `GITLAWB_IPFS_API` set | Node-local availability; the node is an HTTP client of a co-located Kubo daemon |
 | Warm | Pinata (Filecoin-backed) | `pinata.rs` | `GITLAWB_PINATA_JWT` set | Off-node durability + public IPFS gateway reachability |
 
-If a sink's config value is empty, every call into that sink is a no-op — so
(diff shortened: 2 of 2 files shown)


andreolf:
Thanks for the review. Fixed in 706db57 — the de-duplication wording now describes it as best-effort (the check-upload-record path isn't atomic, so concurrent tasks or a record failure can still re-upload), rather than promising zero redundant uploads. Also filled in the PR template sections.

beardthelion (dismissed):
Read the doc against the code at 706db57c. The structure is right and most of it checks out: the config table matches `config.rs` row for row, the hot-inline versus warm-spawned-tail split is accurate, and the shared defer-rather-than-shed semaphore is described correctly. The de-duplication rewrite in this round is correct on the substance too, both sinks really do check, upload, then record as three separate steps with a warn-only failure on the record. Four accuracy fixes before this lands, two of which would mislead an operator.

## Findings

- [P2] Drop the reference to a `pinata_cids` table, which does not exist
  `docs/STORAGE-AND-PINNING.md:46`
  `pinata_cids` appears exactly once in the repo, on this line. There is one table, `pinned_cids`, created at `db/mod.rs:530`, and warm-tier state lives in its nullable `pinata_cid` column added at `:537` and read by `has_pinata_cid`. As written it sends a reader looking for a table that was never created. Suggested wording: "the `pinned_cids` table, whose `pinata_cid` column records warm-tier state".

- [P2] Narrow the blanket no-op claim, which is wrong in the case an operator will hit
  `docs/STORAGE-AND-PINNING.md:18`
  True of the pin paths, which return `Ok("")` at `ipfs_pin.rs:76` and an empty vec at `:259`. Not true of `cat`, which returns `Err("IPFS not configured")` at `:145-147`. Neither caller guards on the config first: `api/encrypted.rs:64` calls it straight after the visibility check, and `AppError::Git` maps to 500 at `error.rs:164`, so an operator who leaves `GITLAWB_IPFS_API` unset on the strength of this sentence gets 500s on the encrypted-blob endpoint rather than a disabled feature. Replication takes it more gracefully and warns at `sync.rs:601`. Scope the sentence to the pin paths.

- [P3] Correct "the node is itself an IPFS peer"
  `docs/STORAGE-AND-PINNING.md:15`
  The node embeds no IPFS implementation, has no IPFS crate in any manifest, and speaks only HTTP to a separately provisioned Kubo daemon (`/api/v0/add` at `ipfs_pin.rs:84`, `/api/v0/cat` at `:148`); its own networking is libp2p. The config row nine lines down already calls it "Base URL of the local Kubo HTTP API", so the doc disagrees with itself. "An HTTP client of a co-located Kubo daemon" would be accurate.

- [P3] Scope the "cannot hold a slot indefinitely" claim to the pin batch
  `docs/STORAGE-AND-PINNING.md:41`
  `PIN_BATCH_BUDGET` does bound the pin loop, but the Pinata tail takes the permit at `api/repos.rs:2404` and re-derives its object list at `:2413`, before the budget is applied at `:2436`. Each git child in that walk carries its own timeout; the walk has no aggregate deadline, and `pinata.rs:101-106` already says so. Either scope the sentence to the pin batch or state the residual.

Minor, take it or leave it: nothing in the repo links to this file, while the README links the other four `docs/` pages. Worth an entry so it is reachable.

andreolf:
Thanks for the thorough read against the code — all four were fair. Fixed in 8a577d3:

- **[P2] `pinata_cids` table** — corrected; there's one `pinned_cids` table and the warm tier keys on its nullable `pinata_cid` column. Reworded accordingly.
- **[P2] blanket no-op** — scoped to the pin paths, and added a note that `ipfs_pin::cat` returns an error (not a no-op) when `GITLAWB_IPFS_API` is unset, so serving the encrypted-blob read endpoint needs Kubo configured.
- **[P3] 'IPFS peer'** — changed to 'an HTTP client of a co-located Kubo daemon' (no IPFS impl in any manifest; it speaks `/api/v0/*` over HTTP).
- **[P3] budget scope** — scoped the 'cannot hold a slot indefinitely' claim to the pin batch, and noted the preceding re-derivation walk holds the slot too with only per-child git timeouts, no aggregate deadline.

Also added the README link so the page is reachable like the other docs/ pages. Ready for another look.

beardthelion (dismissed):
Re-read the doc against the code at 8a577d3a. All four fixes from the last round landed and each one checks out: `pinned_cids` is the single table with the nullable `pinata_cid` column, the no-op sentence is scoped to the pin paths with the `cat` error called out, the Kubo relationship is described as an HTTP client, and the slot-hold claim is scoped to the pin batch with the walk's missing aggregate deadline stated. README links it. One thing left, in the paragraph that matters most to an operator sizing a node.

## Findings

- [P2] Name the hot tier as the unbounded side of the parked-list memory, not just the Pinata tail's bound
  `docs/STORAGE-AND-PINNING.md:45`
  Two lines after "Both tiers share a single global pin admission semaphore", the only memory property given is the warm tier's, so O(refs) reads as the story for the pool. It is the Pinata tail's alone. `pin_new_objects_gated` takes an owned object list, so the caller has materialized it before the function acquires (`crates/gitlawb-node/src/api/repos.rs:1131-1144`): a hot-tier loop parked on a full pool is holding its full list, and the parked count is capped per repo by `EncryptInflight` rather than across repos. `README.md:356` already tells operators not to size memory from this knob for exactly this reason, and a reader who takes the new page as the authority gets the opposite instruction. One clause naming the hot tier as the O(pushes x objects) side closes it.

Two smaller wording residuals while you are in there, neither blocking. De-duplication is per sink in one direction only: `has_pinata_cid` is per-column so a hot-only pin does not suppress the warm tier, but the hot tier's only skip check is `is_pinned`, which is row presence on `sha256_hex` (`db/mod.rs:2479`), and `record_pinata_cid` inserts a row when none exists (`:2603`), so an object warm-pinned while `GITLAWB_IPFS_API` was unset is skipped by the hot tier once Kubo is configured. And "leaving `GITLAWB_PINATA_JWT` unset simply disables warm-tier pinning" (`:19`) is true of the uploads and not of the cost: `do_pinata_replication` is `announce_at_root && !own_walk_failed` (`api/repos.rs:2380`) with no JWT term, so the tail still takes a global pin permit and runs the full re-derivation walk before `pinata::pin_new_objects` returns empty.

Separately, the PR Checks suite has not run on this head (approval-gated), so only the triage job is green.

beardthelion (changes requested):
Re-read the doc at bc69549f. All three items from the last round landed and each checks out: the hot-tier paragraph now matches `pin_new_objects_gated`'s own docstring on the unbounded-across-repos list memory, the one-directional dedup wording matches `is_pinned` being bare row presence against `has_pinata_cid`'s `IS NOT NULL`, and the JWT sentence matches `do_pinata_replication` having no JWT operand ahead of the permit and the walk. Good fixes. Two things the new dedup paragraph exposed, though, and one of them contradicts a bullet further down the same page.

## Findings

- [P2] Qualify the "running both gives a hot copy plus a warm copy" bullet
  `docs/STORAGE-AND-PINNING.md:78`
  Fifteen lines above it the page now says the hot tier's skip check is row presence and a warm pin inserts the row, so an object warm-pinned while `GITLAWB_IPFS_API` was unset is skipped by the hot tier once Kubo is configured. That is the warm-first migration path, and on it enabling both sinks does not produce a local hot copy for the objects already recorded. The bullet is true for objects pushed while both are on and false for everything pinned before. One clause scoping it to objects pushed with both sinks enabled closes the contradiction.

- [P2] Say where the objects actually live, or narrow the opening claim
  `docs/STORAGE-AND-PINNING.md:3`
  The page opens with "how a gitlawb node stores git objects and keeps them available" and the PR body names "where is the data stored?" as the question it answers, but the body only covers the two optional pin sinks. The source of truth that serves clone and fetch is the bare repo directory under `GITLAWB_REPOS_DIR`, with `GITLAWB_TIGRIS_BUCKET` as the optional shared backing; neither appears anywhere on the page. An operator following this link to find their data can come away thinking it lives in Kubo or Pinata. A sentence in the intro naming the primary store and pointing at the README rows is enough. Retitling to scope the page to pinning would also work, but then the opening sentence has to go with it.

- [P3] Move the README link out of the staking section
  `README.md:448`
  The only entry point is in the Optional node staking "Read:" list, between `ECONOMICS.md` and the operator-wallet warning. Storage and pinning is not a staking concern, and nobody arriving from Architecture or the Configuration table will find it. That placement is partly on me since my round-one note just said nothing links to the file. Put it near the storage discussion or the pin-related config rows.

Not an ask, recorded only: `GITLAWB_IPFS_API` is in neither the README config table nor `.env.example`, while `GITLAWB_PINATA_JWT` is in both. That predates this PR and your page documents the var, so it is not yours to fix here.

jatmn (changes requested):
I found issues that need to be addressed before this is ready. This is intended to be a **final, bounded review** — not another drip round.

## Why this PR has been through so many review cycles

The technical content on head `bc69549f` is largely accurate. The repeated rounds are not because the pinning model is wrong; they are because three structural problems keep reopening the review:

1. **Scope mismatch between marketing and charter.** The PR body and opening sentence sell a broad answer to “where is the data stored?” but the page’s own charter (lines 3–4) limits itself to `ipfs_pin.rs` and `pinata.rs` post-push pinning. Every round, reviewers (human and automated) read the broad framing and hunt for everything storage-related — bare repos, Tigris, visibility gates, encrypt-then-pin, batch truncation recovery, branch CIDs, etc. That is how a docs-only PR accumulates an endless finding list even when the pinning sections are correct.

2. **Precision fixes expose new contradictions.** Each round correctly adds detail (dedup is best-effort, warm-first skip, JWT tail cost, hot-tier memory). That is good. But detailed paragraphs in one section can contradict summary bullets elsewhere on the same page if those bullets are not updated in the same pass. Round 4’s dedup paragraph (lines 65–67) is accurate; the Durability bullet at line 78 was not revised to match. This pattern — fix the hard paragraph, leave the easy summary — is the direct cause of the current open finding.

3. **Reviewer scope creep (including mine).** Several candidate issues are real facts about the codebase but **outside this page’s charter**: `announce_at_root` gating, encrypt-then-pin writes, `PIN_BATCH_BUDGET` truncation recovery, `GITLAWB_IPFS_API` in `.env.example`, gateway-vs-`/ipfs` serving, rules snapshot staleness. Chasing them here turns a pinning overview into a full operator manual and guarantees another round. They should be follow-ups or README work, not blockers on this PR.

**Root-cause fix for the author:** Pick one charter and enforce it everywhere — title, intro, PR body, Durability bullets, and README link context. This PR’s charter should be: *optional post-push hot/warm pinning as implemented in `ipfs_pin.rs` and `pinata.rs`*. Not “all storage,” not “all IPFS touchpoints,” not “every edge case in `api/repos.rs`.” Once the charter is explicit, summary bullets must be re-read against detailed paragraphs before each push.

**Done bar for this PR (merge without another content round):**
- [ ] Intro matches the `ipfs_pin.rs` / `pinata.rs` charter (no “where is all my data?” framing unless you expand scope deliberately).
- [ ] No internal contradictions between detailed sections and Durability summary bullets.
- [ ] README link sits where an operator configuring pins will find it.
- [ ] No new sections added for out-of-scope topics listed below.

---

## Explicitly out of scope for this PR

Do **not** expand this PR to address the following — they are real topics but belong elsewhere. Adding them here will not close review; it will start another round:

| Topic | Why it is out of scope | Where it belongs |
|-------|------------------------|------------------|
| Primary bare-repo / Tigris storage | Clone/fetch source of truth; already in README config table (`GITLAWB_REPOS_DIR`, `GITLAWB_TIGRIS_BUCKET`) | README Architecture / Configuration — or a separate “Storage” doc if maintainers want one |
| `GITLAWB_IPFS_API` missing from `.env.example` / README table | Pre-existing; `beardthelion` recorded it is not this PR’s fix | Follow-up PR |
| `announce_at_root` / private-repo pinning gate | Real precondition in `api/repos.rs:2464–2500`, but visibility policy doc — not pinning-sink mechanics | `RUN-A-NODE.md` or visibility docs |
| encrypt-then-pin write path | Separate path in `encrypted_pin.rs`; read-side Kubo requirement is already noted at lines 23–25 | Encrypted-blob doc or README if needed |
| `PIN_BATCH_BUDGET` truncation / no reconciliation sweep | Already in code comments and README semaphore section | Only add here if maintainers explicitly widen charter |
| Branch→CID / gossip / provider-vs-raw CID split | Warm-tail announcement mechanics; not covered by current page structure | Separate pinning-internals or API doc |

---

## Merge readiness

- [P2] Open `CHANGES_REQUESTED` from `beardthelion` on `bc69549f` — two substantive items remain (bullet contradiction, intro scope); one discoverability item (README link).
- [P3] Full PR Checks suite has not run on this head (approval-gated). Low risk for docs-only; maintainer rerun before merge is sufficient.

---

## Findings

### [P2] Qualify the Durability bullet so it matches the dedup paragraph you just added

`docs/STORAGE-AND-PINNING.md:78` (contradicts `:65-67`)

**What is wrong.** Line 78 says unconditionally: “Running **both** gives a node-local hot copy plus an off-node warm copy.” Lines 65–67 correctly document the opposite case: if an object was warm-pinned while `GITLAWB_IPFS_API` was unset, the hot tier’s `is_pinned` skip (row *presence* on `sha256_hex`) prevents a Kubo upload after Kubo is later enabled. `record_pinata_cid` inserts that row (`db/mod.rs:3424-3443`); the hot skip branch never calls Kubo add (`ipfs_pin.rs:1732+`). Git push deltas omit objects the node already has, so those objects may never get a local hot copy.

**Why it matters.** An operator on the warm-first migration path reads the Durability section, enables both tiers, and reasonably expects local Kubo copies. They will not get them for objects already warm-pinned. The detailed dedup paragraph already tells them this; the summary bullet tells them the opposite.

**Root cause.** Summary bullets were not re-checked after the dedup paragraph was added in `bc69549f`. This is the same failure mode as earlier rounds (fix the precise paragraph, leave the summary).

**Fix (one clause, no new sections).** Change line 78 to something like:

> Running **both** gives a node-local hot copy plus an off-node warm copy **for objects pinned while both sinks are enabled**; objects already warm-pinned before Kubo was configured keep only the warm copy unless manually re-pinned.

**Do not:** Add migration runbooks, SQL repair steps, or new subsections on dedup — the paragraph at 65–67 already covers the mechanism.

---

### [P2] Narrow the opening sentence to match the page charter — do not expand into a full storage guide

`docs/STORAGE-AND-PINNING.md:3-4`

**What is wrong.** Line 3 opens with “How a gitlawb node stores git objects and keeps them available,” which reads like the answer to “where is my data?” Line 4 immediately scopes the page to `ipfs_pin.rs` and `pinata.rs` only. Those two statements fight each other. The PR body repeats the broad framing (“where is the data stored?”) while the actual content is the two optional pin sinks.

**Why it matters.** Broad framing is what triggers every reviewer (and every automated pass) to search for bare repos, Tigris, visibility gates, encrypt paths, and truncation semantics. That is the main reason this PR keeps getting new findings despite accurate pinning content.

**Root cause.** The motivation (“operators ask where data lives”) was copied into the doc opening without deciding whether this page is (a) a pinning supplement or (b) a full storage map. The page is (a); the opening says (b).

**Fix (preferred — smallest, no drift).** Replace lines 3–4 with something that matches the charter, e.g.:

> How a gitlawb node’s **optional post-push pinning tiers** work. This documents behavior in `crates/gitlawb-node/src/ipfs_pin.rs` and `crates/gitlawb-node/src/pinata.rs`; it does not change any behavior. Primary git storage for clone/fetch lives in bare repos under `GITLAWB_REPOS_DIR` (optional `GITLAWB_TIGRIS_BUCKET` backing) — see the README Configuration table.

That is **two sentences**: charter + one pointer so operators are not misled into thinking data lives only in Kubo/Pinata. You are not documenting repos/Tigris in depth; you are preventing the false impression the current opening creates.

**Alternative (also acceptable):** Retitle to “Pinning tiers” and drop any “stores git objects” language entirely; add the README pointer in “See also.”

**Do not:** Turn this PR into a full storage architecture doc, duplicate README config rows, or document Tigris acquisition/sync — that is scope expansion and will not end review.

**Also update the PR description** to match: e.g. “documents optional post-push Kubo and Pinata pinning” rather than “where is the data stored?” unless you deliberately widen scope.

---

### [P3] Move the README link to where pin operators look

`README.md:448`

**What is wrong.** The only link to this page is under Optional node staking → “Read:” list. Pin-related config (`GITLAWB_MAX_CONCURRENT_PIN_TASKS`, `GITLAWB_PINATA_JWT`) is in the Configuration table around lines 404–416; Architecture mentions bare git storage at line 56. Operators configuring pins will not look in the staking section.

**Why it matters.** Discoverability only — the link works; the placement is wrong. `beardthelion` flagged this and noted his round-one “add any link” request caused the placement.

**Root cause.** The first review asked for reachability; the link was added to the nearest existing docs list (staking) rather than the config context.

**Fix.** Add a link on the `GITLAWB_MAX_CONCURRENT_PIN_TASKS` row or immediately after the Configuration table intro (e.g. “For pinning tier details see [`docs/STORAGE-AND-PINNING.md`](docs/STORAGE-AND-PINNING.md).”). You can keep the staking-list link too or remove it to avoid duplication — either is fine.

**Do not:** Restructure README sections or add new doc index pages in this PR.

---

## Guidance before you push again

1. **Read the Durability section after every edit to “How pinning runs.”** Any new precision in the body (dedup, JWT cost, memory) must have matching summary bullets. That single habit would have prevented rounds 3 and 4.

2. **Align PR body, title, intro, and first paragraph to the same charter.** If the charter is “post-push pinning tiers,” say that everywhere. Mixed messages are what keep automated and human reviewers searching the whole codebase.

3. **When in doubt, pointer not paragraph.** For topics outside `ipfs_pin.rs` / `pinata.rs`, one README cross-link beats a new subsection. Subsections invite the next review round.

4. **Reply on the PR listing what you did not change** and pointing to the out-of-scope table above, so the next reviewer does not re-raise `announce_at_root`, encrypt-then-pin, truncation, or `.env.example` gaps as blockers on this PR.

After the three items above, the pinning documentation is accurate and complete for its stated scope. Further findings on this PR should be treated as follow-up work unless maintainers explicitly widen the charter.

Issue #349, opened 2026-08-15
One appended junk signature makes verify_all reject a threshold-satisfying certificate


beardthelion:
`RefUpdateCert::verify_all` promises a filtered list and delivers an all-or-nothing result
(`crates/gitlawb-core/src/cert.rs:113-135`):

```rust
/// Returns the list of DIDs whose signatures are valid.
pub fn verify_all(&self) -> Result<Vec<Did>> {
    ...
    for cert_sig in &self.signatures {
        let vk = cert_sig.signer.to_verifying_key()?;
        let sig_bytes_vec = URL_SAFE_NO_PAD.decode(&cert_sig.sig)
            .map_err(|e| Error::RefCert(format!("invalid base64 sig: {e}")))?;
        let sig_bytes: [u8; 64] = sig_bytes_vec.try_into()
            .map_err(|_| Error::RefCert("signature must be 64 bytes".to_string()))?;
        verify(&vk, &signing_bytes, &sig_bytes)?;
        valid.push(cert_sig.signer.clone());
    }
```

Four `?` on per-entry attacker-supplied data, under a docstring that says the opposite. The threshold
check consumes it unfiltered (`cert.rs:143-148`) and is itself written to tolerate irrelevant entries,
filtering non-maintainers and deduping, but never gets the chance.

The signature list is deliberately outside the signed bytes: `RefUpdateBody::to_signing_bytes`
(`cert.rs:53-55`) covers the body only, and `gitlawb-attest` strips `signatures` and `attestations`
before JCS encoding so countersignatures stay stable. Appending is the intended workflow
(`countersign`, `cert.rs:98-109`, a plain `push`). So an append is undetectable by construction and
requires no key material.

## Measured

Workspace copied to scratch, integration test added there, `wt-clawpatch` untouched:

```
baseline  threshold(2)       = Ok(true)
junk-b64  threshold(2)       = Err(RefCert("invalid base64 sig: Invalid symbol 33, offset 0."))
bogus-sig threshold(2)       = Err(SignatureInvalid)
junk-b64  validate_structure = Ok(())
```

Both entry classes reproduce, and structure validation does not catch either.

## Why this is low, and why it is still worth fixing now

`RefUpdateCert`, `verify_all`, and `satisfies_threshold` have **zero production callers**: grep across
`gitlawb-node`, `gl`, and `git-remote-gitlawb` returns nothing, and the node's own
[...]

PR #365, opened 2026-08-18
fix(attest): make verify_all lenient under RequireAll (#359)

### LINKED ISSUE #359: Registry::verify_all aborts the whole batch on one malformed attestation, and the test that names the DoS rule cannot see it
`gitlawb-attest` states the rule and does not implement it. `verifier.rs:7-12`:

```rust
//! Under [`Policy::RequireAll`] the registry never short-circuits on unknown
//! types — any attestation can be attached by any third party, so blocking
//! the cert because an attacker added an unrelated `attacker/spam/v1` would
//! be a denial-of-service vector.
```

Note the premise is "any attestation can be attached by any third party", which is broader than "any
unknown type". `Policy::RequireAll`'s own doc at `:38-40` says the same thing ("so a malicious extra
attestation cannot DoS the cert").

The loop propagates (`verifier.rs:153-156`):

```rust
let mut verified = Vec::with_capacity(attestations.len());
for a in attestations {
    verified.push(self.verify(a, expected_cert_hash)?);
}
```

That `?` inherits every error from `verify_signature` (`:124-128`): cert-hash mismatch, type,
signature, JCS, DID. The leniency covers exactly one arm, the `None` branch at `:131-136` where the
type is unregistered.

## Measured

Seven input classes against a `RequireAll` registry requiring `demo/v1`, with the required
attestation present and satisfying it:

```
baseline (real only)                  => Ok(1 verified)
unknown-type VALID-sig spam first     => Ok(2 verified)     <- the only class the test drives
unknown-type junk-base64 sig          => Err(base64url: Invalid symbol 33)
unknown-type bogus 64-byte sig        => Err(ed25519: Verification equation was not satisfied)
unknown-type wrong cert_hash          => Err(cert hash mismatch)
malformed type discriminator          => Err(discriminator has an empty segment)
registered type, failing payload      => Err(payload: missing or empty 'a')
junk appended last                    => Err(base64url: Invalid symbol 33)
```

Six of seven abort. Order is irrelevant.

## Why the existing test does not catch it

`verifier.rs:302-320` builds its spam entry with a genuine `Attestation::sign` and the correct cert
hash, so every field is well-formed. It reaches only the `by_type.get` miss branch and never touches
`verify_signature`'s error paths. It is load-bearing for "unknown types are lenient" and vacuous for
the DoS property the docstring claims.

That was measured rather than argued. Replacing `:155` with a skip-and-continue flips all six failing
classes to `Ok`, and the whole suite stays green under both semantics:

```
41 passed (unit, incl. require_all_is_lenient_on_unknown_types_in_the_batch)
 4 passed (canonical_vectors)
 4 passed (with_gitlawb_core)
```

No test pins fail-closed-on-malformed. `reject_unknown_blocks_unregistered_types` survives because it
calls `verify`, not `verify_all`, and the `RejectUnknown` batch test in
`tests/with_gitlawb_core.rs:109-115` asserts only the happy path's `verified.len() == 1`.

## Reachability

Anyone who can relay or store an envelope. `attestations` sits outside the signed bytes by design
(`cert.rs:104-112` strips `signatures` and `attestations` before hashing) and `attach` (`cert.rs:88-90`)
is a bare `push`, so an append is undetectable by construction. Five of the six failing classes need
no key material at all: replacing `sig` with `"!!!not-base64!!!"` on any existing attestation is enough.

## Severity

Low, matching #349 rather than inflating. It fails closed, and `gitlawb-attest` currently has no
in-tree consumer: it appears once in `Cargo.lock` as a package and in no other crate's dependency
list, so every `verify_all` call site today is a test. The argument for fixing it now is that the
module states the rule and ships a test that reads like proof of it, so the next person to wire this
crate up will trust a property that holds for one input class in six.

## Fix direction, with a trap

Partition valid from invalid entries rather than blanket-swallowing the error. Naive
skip-and-continue is not the remedy: under exactly that mutation, `RejectUnknown` silently degraded
from rejecting an unregistered type to dropping it, which trades a denial for a permission. The fix
has to keep `RejectUnknown` fail-closed while making `RequireAll` tolerate junk.

A regression test needs to drive a malformed entry, not a validly-signed unknown-type one, or it will
be as vacuous as the current one.

## Relation to #349

Sibling, not duplicate. #349 is `RefUpdateCert::verify_all` in `crates/gitlawb-core/src/cert.rs:113-135`;
this is `Registry::verify_all` in `crates/gitlawb-attest/src/verifier.rs:148-168`. Different crates,
no shared code, so fixing one does not touch the other. Both are the same shape: a `?` on per-entry
attacker-supplied data inside an append-only list that sits outside the signature.

Worth fixing together, since they are one class and the same trap applies to both.

andreolf:
## Summary

`Registry::verify_all` aborted the whole batch on any per-attestation error, violating its documented `RequireAll` leniency. Under `Policy::RequireAll`, a single malformed or unknown attestation now drops out of the result instead of failing the batch. Fixes #359.

## Motivation & context

Closes #359

The module docstring promises that under `RequireAll` the registry "never short-circuits on unknown types — any attestation can be attached by any third party, so blocking the cert because an attacker added an unrelated `attacker/spam/v1` would be a denial-of-service vector." But `verify_all` propagated every error via `?`:

```rust
for a in attestations {
    verified.push(self.verify(a, expected_cert_hash)?);
}
```

So a bad signature, cert-hash mismatch, or malformed payload on *any* attached attestation aborted the entire batch — exactly the DoS the docstring warns about.

## Kind of change

- [x] Bug fix

## What changed

Crate touched: `gitlawb-attest` (`src/verifier.rs`).

- Under `Policy::RequireAll`, `verify_all` now drops entries that fail to verify (`filter_map(.ok())`) instead of aborting. The `required_types` presence check remains the gate, and it only counts `fully_verified` entries — so a required type present **only** as a malformed attestation still fails with `RequiredMissing`. Leniency does not open a hole.
- `AcceptKnown` and `RejectUnknown` are unchanged: they stay strict and surface the first error.
- Expanded the `verify_all` docstring to state the per-policy contract.

## How a reviewer can verify

```sh
cargo test -p gitlawb-attest --lib verifier
cargo clippy -p gitlawb-attest --all-targets -- -D warnings
```

Two new tests:
- `require_all_drops_a_malformed_extra_and_still_accepts` — a malformed extra (signed against the wrong cert hash) alongside a valid required attestation no longer aborts; the junk is dropped and the cert verifies. Fails on the pre-fix code.
- `require_all_rejects_when_only_a_malformed_required_type_is_present` — when the sole copy of a required type is malformed, `verify_all` still returns `RequiredMissing`.

## Before you request review

- [x] `cargo test --workspace` passes locally (ran `gitlawb-attest`: 52 tests pass)
- [x] Commit titles use Conventional Commits (`fix(attest): ...`)

## Protocol & signing impact

- [x] Touches DID / `did:key`, Ed25519 / RFC 9421 signatures, UCAN, ref certs, or P2P wire formats — attestation verification for ref-update certs. **No wire-format or signature-scheme change**; only the batch error-handling policy under `RequireAll`. Behavior for `AcceptKnown`/`RejectUnknown` is unchanged.
- [x] Backward-compatible with existing nodes and previously signed history (strictly more permissive under `RequireAll`, and only for attestations that would otherwise have hard-failed the batch).

## Notes for reviewers

- #349 tracks the same short-circuit pattern in `gitlawb-core`'s `cert.rs`; deliberately out of scope here to keep this PR to one crate/one change. Happy to follow up on that separately.

### DIFF
--- a/crates/gitlawb-attest/src/verifier.rs
+++ b/crates/gitlawb-attest/src/verifier.rs
@@ -147,14 +147,19 @@ impl Registry {
     /// Verify a batch, then enforce `RequireAll`.
     ///
     /// Under [`Policy::RequireAll`] this never short-circuits: an attestation
-    /// that fails to verify — bad signature, cert-hash mismatch, malformed
-    /// payload, or a type with no verifier — is dropped from the result rather
-    /// than aborting the batch. Any third party can attach an attestation, so a
-    /// hard error here would let an unrelated `attacker/spam/v1` deny-of-service
-    /// the whole cert (the DoS vector the module docstring warns about). The
-    /// `required_types` check below is the real gate, and it only counts
-    /// `fully_verified` entries — so a required type represented solely by a
-    /// malformed attestation still fails with [`Error::RequiredMissing`].
+    /// that fails to verify — bad signature, cert-hash mismatch, or malformed
+    /// payload, the actual `Err` paths of [`Registry::verify`] — is dropped
+    /// from the result rather than aborting the batch. A valid attestation of
+    /// a type with NO registered verifier is not one of those: under
+    /// `RequireAll` it verifies to `Ok` with `fully_verified: false` and stays
+    /// in the result, preserving the distinction between *untrusted* and
+    /// *invalid* (see `require_all_is_lenient_on_unknown_types_in_the_batch`).
+    /// Any third party can attach an attestation, so a hard error here would
+    /// let an unrelated `attacker/spam/v1` deny-of-service the whole cert (the
+    /// DoS vector the module docstring warns about). The `required_types`
+    /// check below is the real gate, and it only counts `fully_verified`
+    /// entries — so a required type represented solely by a malformed
+    /// attestation still fails with [`Error::RequiredMissing`].
     ///
     /// Under `AcceptKnown` and `RejectUnknown` the batch stays strict: the
     /// first verification error is surfaced to the caller.
@@ -337,6 +342,38 @@ mod tests {
         assert!(verified
             .iter()
             .any(|v| v.type_ == "other/v1" && !v.fully_verified));
+    }
+
+    /// The strict policies must surface a batch member's verification error
+    /// through `verify_all`, not launder it into a partial result. This binds
+    /// the POLICY BOUNDARY itself: only the `RequireAll` arm may drop failing
+    /// entries, and an apparently harmless cleanup that applies its
+    /// `filter_map(Result::ok)` shape to the strict arms would silently turn
+    /// the fail-closed contracts into fail-open — with every per-attestation
+    /// `verify` test still green, because none of them go through the batch
+    /// path. The failing entry is a wrong-cert-hash attestation (signed over a
+    /// different hash than the batch expects), a genuine `Err` under every
+    /// policy.
+    #[test]
+    fn strict_policies_propagate_a_batch_members_error() {
+        let sk = fresh();
+        let cert_hash = sample_hash();
+        let mut wrong_hash = sample_hash();
+        wrong_hash[0] ^= 0xff;
+
+        for policy in [Policy::AcceptKnown, Policy::RejectUnknown] {
+            let mut reg = Registry::new().with_policy(policy);
+            reg.register(DemoVerifier);
+            let good = signed_demo(&sk, cert_hash, "ok");
+            let bad = signed_demo(&sk, wrong_hash, "ok");
+            let err = reg
+                .verify_all(&[good, bad], cert_hash)
+                .expect_err("a strict policy must fail the whole batch");
+            assert!(
+                matches!(err, Error::Signature(_) | Error::CertHashMismatch),
+                "expected the underlying verification error, got {err:?}"
+            );
+        }
     }
 
     /// A required type that is present but unverified (no verifier registered
--- a/crates/gitlawb-attest/src/verifier.rs
+++ b/crates/gitlawb-attest/src/verifier.rs
@@ -145,23 +145,43 @@ impl Registry {
     }
 
     /// Verify a batch, then enforce `RequireAll`.
+    ///
+    /// Under [`Policy::RequireAll`] this never short-circuits: an attestation
+    /// that fails to verify — bad signature, cert-hash mismatch, malformed
+    /// payload, or a type with no verifier — is dropped from the result rather
+    /// than aborting the batch. Any third party can attach an attestation, so a
+    /// hard error here would let an unrelated `attacker/spam/v1` deny-of-service
+    /// the whole cert (the DoS vector the module docstring warns about). The
+    /// `required_types` check below is the real gate, and it only counts
+    /// `fully_verified` entries — so a required type represented solely by a
+    /// malformed attestation still fails with [`Error::RequiredMissing`].
+    ///
+    /// Under `AcceptKnown` and `RejectUnknown` the batch stays strict: the
+    /// first verification error is surfaced to the caller.
     pub fn verify_all(
         &self,
         attestations: &[Attestation],
         expected_cert_hash: [u8; 32],
     ) -> Result<Vec<VerifiedAttestation>> {
-        let mut verified = Vec::with_capacity(attestations.len());
-        for a in attestations {
-            verified.push(self.verify(a, expected_cert_hash)?);
+        if self.policy != Policy::RequireAll {
+            let mut verified = Vec::with_capacity(attestations.len());
+            for a in attestations {
+                verified.push(self.verify(a, expected_cert_hash)?);
+            }
+            return Ok(verified);
         }
-        if self.policy == Policy::RequireAll {
-            for required in &self.required_types {
-                let present = verified
-                    .iter()
-                    .any(|v| &v.type_ == required && v.fully_verified);
-                if !present {
-                    return Err(Error::RequiredMissing(required.clone()));
-                }
+
+        // RequireAll: drop entries that fail to verify instead of aborting.
+        let verified: Vec<VerifiedAttestation> = attestations
+            .iter()
+            .filter_map(|a| self.verify(a, expected_cert_hash).ok())
+            .collect();
+        for required in &self.required_types {
+            let present = verified
+                .iter()
+                .any(|v| &v.type_ == required && v.fully_verified);
+            if !present {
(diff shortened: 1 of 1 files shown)


beardthelion (dismissed):
Verified the fix does what it claims and does not weaken the gate. Reverting the policy branch to the pre-fix strict loop turns `require_all_drops_a_malformed_extra_and_still_accepts` red, and disabling the `required_types` check turns `require_all_rejects_when_only_a_malformed_required_type_is_present` red, so both new tests bind real behavior rather than passing vacuously. The vector is reachable as described: `cert_hash()` strips `attestations` before hashing and `attach()` is public, so any relaying party can append junk without invalidating a signature. Direction is right, the gate holds, and the blast radius is one caller. Two asks, both in lines this PR adds.

Unrelated to this PR: `cargo audit` is red on `main` too, RUSTSEC-2026-0258 in `h2 0.4.13`. Not yours, and I am not holding this on it.

## Findings

- [P2] Lock the non-`RequireAll` paths with an error-propagation test
  `crates/gitlawb-attest/src/verifier.rs:166`
  The new docstring promises `AcceptKnown` and `RejectUnknown` still surface the first error, but nothing asserts it. Replacing the strict branch's `?` with a lenient `if let Ok(v)` leaves the whole crate green; gutting its return does go red, so the branch executes and it is specifically the error propagation that is unbound. A later edit applying the lenient loop to every policy would pass every test here while silently turning default `AcceptKnown` batch verification into drop-on-error. This closes it, and I confirmed it reddens under exactly that mutation:

  ```rust
  #[test]
  fn strict_policies_surface_the_first_batch_error() {
      let sk = fresh();
      let cert_hash = sample_hash();
      let bad = signed_demo(&sk, other_hash(), "ok");

      let mut accept = Registry::new().with_policy(Policy::AcceptKnown);
      accept.register(DemoVerifier);
      let err = accept.verify_all(&[bad.clone()], cert_hash).expect_err("must not drop");
      assert!(matches!(err, Error::CertHashMismatch { .. }), "got {err:?}");

      let mut reject = Registry::new().with_policy(Policy::RejectUnknown);
      reject.register(DemoVerifier);
      let err = reject.verify_all(&[bad], cert_hash).expect_err("must not drop");
      assert!(matches!(err, Error::CertHashMismatch { .. }), "got {err:?}");
  }
  ```

- [P3] Correct the docstring: unknown types are not dropped
  `crates/gitlawb-attest/src/verifier.rs:149`
  The added text lists "a type with no verifier" among entries dropped from the result. Under `RequireAll`, `verify` returns `Ok` with `fully_verified = false` for an unregistered type, so `filter_map(.ok())` keeps it, and `require_all_is_lenient_on_unknown_types_in_the_batch` already asserts the entry survives with `len == 2`. Only the genuine `Err` paths are dropped. The PR summary carries the same slip, though the Motivation section states it correctly.

jatmn (changes requested):

## Findings

- [P2] Lock the strict batch policies with an error-propagation test
  `crates/gitlawb-attest/src/verifier.rs:166`
  The root cause is that the refactor splits `verify_all` by policy but the new test coverage only exercises the lenient `RequireAll` arm. Neither `AcceptKnown` nor `RejectUnknown` is invoked through `verify_all` with an attestation that makes `self.verify` fail. Consequently, an apparently harmless future cleanup that applies the `filter_map(Result::ok)` pattern to all policies would make malformed, bad-signature, or wrong-cert-hash entries disappear instead of returning an error; the current added tests would still pass, silently changing the default and fail-closed contracts despite the promise at lines 159–160.

  Preserve the explicit strict loop, and add a batch-level regression test for both strict policies using a genuinely malformed entry (for example, the existing wrong-cert-hash helper). Assert that `verify_all` returns the underlying verification error rather than a partial result. That binds the policy boundary itself, rather than only the single-attestation `verify` behavior.

- [P3] Do not say unknown attestations are dropped
  `crates/gitlawb-attest/src/verifier.rs:149`
  The root cause is that the documentation treats “no verifier” as a verification failure, but it is an intentional successful result under `RequireAll`: `verify` returns `Ok(VerifiedAttestation { fully_verified: false })`, so `filter_map(Result::ok)` retains the entry. The existing `require_all_is_lenient_on_unknown_types_in_the_batch` test already proves this with a two-entry result containing the unverified `other/v1` attachment. The new wording therefore tells callers that the returned vector excludes unknown valid attestations when it does not.

  Correct the contract at its source by limiting the dropped-entry list to actual `Err` paths—bad signatures, cert-hash mismatches, and malformed payloads—while stating that valid unknown types remain in the result as unverified. Do not change the implementation to drop unknown entries: retaining them is the established `RequireAll` behavior and preserves the distinction between untrusted and invalid attestations.

jatmn (approved):
lgtm

beardthelion (approved):
Approving. The strict arms now have a test that fails when they are made lenient, which is what the last round asked for. I checked it by giving `AcceptKnown`/`RejectUnknown` the `filter_map(Result::ok)` shape and watching `strict_policies_propagate_a_batch_members_error` go red, and by reverting the `RequireAll` arm to `?` propagation, which reds both `require_all_*` tests. Blast radius is inside the crate: nothing in the workspace depends on `gitlawb-attest`, and `gitlawb-core::cert::verify_all` is a separate method.

Two follow-ups, neither blocking.

## Findings

- [P3] Assert `CertHashMismatch` exactly instead of the two-variant union
  `crates/gitlawb-attest/src/verifier.rs:373`
  `verify_signature` compares `cert_hash` before it touches the signature, so a member signed against a different hash always fails as `CertHashMismatch` and the `Error::Signature(_)` arm is dead. I narrowed the `matches!` to `CertHashMismatch` alone and the suite stayed at 44 passed; narrowing it to `Signature(_)` alone went red. Pinning the exact variant makes the test say which contract it is holding.

- [P3] Drive the other strict error classes through `verify_all` too
  `crates/gitlawb-attest/src/verifier.rs:171-177`
  Only the cert-hash class crosses the batch path today. I mutated the strict arm to swallow just `Err(Error::UnknownType(_))`, which makes a `RejectUnknown` registry silently accept an attacker-attached unregistered attestation, and the whole suite stayed green at 44 passed; swallowing just `Err(Error::Payload(_))` was also green. A `RejectUnknown` batch with an unregistered member asserting `Err(UnknownType)`, plus a payload-failing member under a strict policy, would close both.

beardthelion:
@kevincodex1 LGTM

PR #285, opened 2026-07-30
fix(node): fence a repo publish on the write attempt that owns it

beardthelion:
Closes #283.

The advisory-lock fix this PR opened with is already on `main`. It landed in 752fd19b and reached `main` through the PR #173 merge, and #279 is closed, so the framing here has moved on. What is left is the work that grew around that pin, and it addresses a different failure: a write attempt whose future is gone can still land its bytes, and nothing downstream could tell that apart from a committed write.

## What changes

**A publish attempt has an identity.** `crates/gitlawb-node/src/git/publish.rs` adds `PublishAttemptId`, `PublishStage` and `PublishStageCell`. The id travels in object user metadata and is stamped on the on-disk clone and the database row, so a reader can ask whether what it is looking at is this attempt's work, and how far a cancelled attempt actually got is observable from outside the future performing it.

**Object-store writes are fenced on a precondition.** Uploads and deletes now carry `If-Match` against the generation the attempt read, so a late PUT from an abandoned attempt is refused instead of overwriting a newer archive. Neither path carried any precondition before, which is what made "the lock is held, therefore nothing else can be writing" false in the first place.

**The local directory swap is fenced on the same authority, which is #283.** `decompress_repo` no longer removes and renames the live tree itself. It goes through `swap_extracted_into_validated_repo`, which takes the publish lock and then claims a commit token with a `true -> false` compare-and-exchange. A writer that loses lock ownership revokes that token, both on the explicit refusal path and in `RepoWriteGuard::drop`, so an extraction already running in an uncancellable `spawn_blocking` bails and removes its temp directory rather than deleting a tree a later writer is using.

**An unrefreshable write sheds instead of writing.** A failed archive HEAD on the write paths raises a typed `RepoUnavailable` that maps to a retryable 503 with a fixed body, rather than reading as "no archive exists" and proceeding.

**`close_issue` authorizes before it takes the lock.** It took the write lock and then ran the owner-or-author check, returning 403 with the lock held. Once exclusion actually works that is a wedge primitive for anyone with read access. The author fallback reads the issue's git-JSON blob without the lock as a pre-check, and the authoritative check runs again under the guard, because the tree that gets mutated is frequently not the one the pre-check read.

## The pin the rest of this follows from

Pinning a connection per in-flight write means writes can no longer share the 20-connection application pool, so they get a dedicated one (`GITLAWB_DB_LOCK_POOL_MAX_CONNECTIONS`, default 32, `connect_lazy`). Without it a push burst starves ordinary reads.

[...]

PR #368, opened 2026-08-18
fix(deps): bump h2 to 0.4.16 for RUSTSEC-2026-0258

Vasanthdev2004:
## What breaks

`cargo audit` fails on every branch as of today:

```
Crate:     h2
Title:     h2 unbounded empty DATA frames
ID:        RUSTSEC-2026-0258
Solution:  Upgrade to >=0.4.16
error: 1 vulnerability found!
```

The advisory was published after `main`'s last green run, so the `SUCCESS` ticks currently showing on open PRs are stale cached results, not evidence. Anything that re-runs CI goes red until this lands — #331 already has.

`h2` arrives transitively via `hyper`, so there is no manifest change. The lockfile is the entire fix.

## Why this is hand-edited

`cargo update -p h2` on my machine produced the h2 bump **plus nine unrelated rewrites**, all downgrades:

| crate | from | to | pulled in by |
|---|---|---|---|
| `syn` | 2.0.117 | 1.0.109 | `data-encoding-macro-internal` |
| `windows-sys` | 0.61.2 | 0.52.0 | `errno` (×3) |
| `socket2` | 0.6.3 | 0.5.10 | (×3) |

`--precise 0.4.16` produced the identical set, so it is not collateral re-resolution from an unpinned update — it is local crates.io index drift against whatever generated `main`'s lockfile. Those downgrades have nothing to do with the advisory and do not belong in a security bump, so this commit changes only h2's `version` and `checksum` lines.

That `cargo check --locked` accepts the result is the proof that 0.4.16 introduces no dependency changes of its own — a hand-edited lockfile that got that wrong would fail `--locked` immediately.

Worth someone re-running `cargo update` on a Linux box to see whether those nine entries are real drift in `main`'s lockfile or an artifact of my toolchain. Either way it is a separate question from this fix.

## Verification

- `cargo check --locked --workspace` — exit 0
- `cargo tree --locked -p h2` — resolves `h2 v0.4.16`
- Full suite on a branch carrying the Windows test-gating fix: `gitlawb-core` 104, `gl` 328, `git-remote-gitlawb` 58, all passing

`cargo check --locked --workspace --all-targets` cannot run from `main` on Windows — the unix-only test code does not compile there, which is exactly what #330's first commit fixes. Linux CI covers that path.

## Scope

Deliberately kept to the advisory alone rather than folded into #331, so it can land on its own and unblock every open PR at once.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3542,9 +3542,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.4.13"
+version = "0.4.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2f44da3a8150a6703ed5d34e164b875fd14c2cdab9af1252a9a1020bde2bdc54"
+checksum = "a9f37a958b41b3b19ee2707c06439c0e9e547e847223eb791ecb0cb821c65e27"
 dependencies = [
  "atomic-waker",
  "bytes",


Vasanthdev2004:
Adding the blast radius, since the green ticks elsewhere make this look less urgent than it is.

**Every open PR is affected. Most just have not noticed yet.**

`cargo audit` re-evaluates against the live advisory database on each run, so a PR's result reflects when it last ran, not the state of its code. The advisory landed today between two runs:

| | |
|---|---|
| last green run | #285, `2026-08-18T05:44:10Z` |
| first red run | #327, `2026-08-18T08:35:05Z` |

Two PRs have re-run `cargo audit` since. Both went red, both on `RUSTSEC-2026-0258`:

- #327 — `08:35:05Z`
- #331 — `09:59:20Z`

The other 21 open PRs still show `SUCCESS`, all from runs that predate the advisory — some from July. Those are cached results, not passing checks. Each one flips red the next time anything triggers CI: a push, a rebase, a re-run, a base update. Nothing about their code has to change.

So the count of visibly-broken PRs will keep climbing on its own, and each author will independently hit a red check that has nothing to do with their work and costs them time to diagnose. That is the real cost here, more than the vulnerability itself — `h2` reaches us transitively through `hyper`, and the DoS surface is a matter for deployed nodes rather than for CI.

**The change is two lines**, both inside `h2`'s lockfile entry — `version` and `checksum`. No manifest change, no code change. `cargo check --locked` passing is itself the proof that 0.4.16 needs no dependency changes of its own; a wrong hand-edit would fail `--locked` immediately.

18/18 checks green here, including `cargo audit`.

Happy to answer anything on the hand-edit rationale (the nine unrelated downgrades `cargo update` wanted to bring along are in the description). If someone would rather do this differently — a full `cargo update -p h2` on a Linux box, or folding it into a dependency sweep — that is fine by me and I will close this; the goal is that it lands, not that it lands this way.

Vasanthdev2004:
@kevincodex1 — could you take a look when you get a moment?

Two-line lockfile diff, 18/18 green. It is blocking #327 and #331 today, and the remaining 21 open PRs go red on their next CI trigger regardless of their own changes (detail in the comment above).

No rush if you would rather it went in differently — happy to close this in favour of another approach.

beardthelion (approved):
Verified the two changed lines rather than the CI ticks.

The checksum is the real one: sha256 of the h2-0.4.16 tarball from static.crates.io is `a9f37a958b41b3b19ee2707c06439c0e9e547e847223eb791ecb0cb821c65e27`, byte-identical to the line this adds, and the sparse index agrees. 0.4.16 is the advisory's minimum patched version, so this is the smallest bump that clears RUSTSEC-2026-0258. h2's normal dependency set is unchanged between 0.4.13 and 0.4.16, and the one requirement that moved (`http ^1` to `^1.1`) is satisfied by the locked http 1.4.0, so leaving the entry's dependency block untouched is correct.

I ran it both ways. Reverting the two lines takes `cargo audit` to exit 1 on RUSTSEC-2026-0258; restoring them, exit 0. Corrupting one hex digit of the new checksum takes `cargo fetch --locked` to exit 101 with a checksum error, and it does that with the crate warm in the registry cache and with no network at all. So your `cargo check --locked` was already sufficient proof that the hand-edited checksum matches what the registry publishes.

On your open question: it is not your toolchain. From main's lockfile on a Linux box, `cargo update -p h2 --precise 0.4.16` gives the h2 bump byte-identical to your hand edit plus exactly the same seven rewrites. They are not crate downgrades, which is why they look worse than they are. syn, socket2 and windows-sys each already exist at both versions in main's lockfile, and what moves is which duplicate a consumer's edge points at. rustix, tempfile and errno declare `windows-sys ">=0.52, <0.62"` and data-encoding-macro-internal declares `syn ">=1, <4"`, so both choices are legal.

They are also not caused by your bump. Running `cargo update -p h2 --precise 0.4.13`, the version main already pins, produces the identical seven changes with h2 itself untouched. So main's committed lockfile is not a fixed point of the current resolver and index, and any command that rewrites it re-normalizes those edges; the crate you name is incidental. Both lockfiles pass `cargo metadata --locked`. Nothing is drifting, hand-editing was the right call, and there is no separate issue to file.

One scope note, no change needed here: merging this re-runs nothing on the other open PRs, so their stale green ticks stay stale until each pushes or enters the merge queue. Nothing can land on a red audit regardless, since the checks run again on the merge group.

Worth knowing that CodeRabbit skipped this rather than cleared it, since `Cargo.lock` sits behind its default `!**/*.lock` filter.

---

*Edited to correct two things in my original text. I had written that "cargo re-unifies downward when it rewrites the file" as the explanation for the seven rewrites; I never verified that and have retracted it rather than replacing it with another guess. What is verified is the no-op-update result above, which is the more useful fact anyway. I had also framed the dependency block as correct "not only because `--locked` accepted it" and described purging the registry cache as part of the checksum test; both undersold your own evidence, since `--locked` catches a bad checksum warm and offline.*

beardthelion:
@kevincodex1 LGTM

Vasanthdev2004:
Four days on, an update on the count — it has grown, and the growth is the point.

**Six open PRs are now red on `cargo audit`, every one on `RUSTSEC-2026-0258`:**

| PR | author | first red |
|---|---|---|
| #327 | euxaristia | 2026-08-18 08:35 |
| #363 | andreolf | 2026-08-18 17:07 |
| #365 | andreolf | 2026-08-18 15:48 |
| #325 | beardthelion | 2026-08-18 22:07 |
| #366 | andreolf | 2026-08-21 01:01 |
| #331 | (mine) | 2026-08-22 09:02 |

I checked the job logs rather than assuming: #325, #327, #363 and #331 all report `ID: RUSTSEC-2026-0258` with `error: 1 vulnerability found!`. Not a coincidence of unrelated failures.

When I first commented there were two. The other four arrived simply by touching their branches — no code of theirs changed to cause it.

**28 more open PRs still show `SUCCESS`**, all from runs predating the advisory. Those are cached results, not passing checks. Every one flips the next time anything triggers CI.

So the shape of this is: five authors who have nothing to do with `h2` are now each looking at a red check on their own work, and the number goes up on its own. #366 in particular is a careful fix — it preserves the distinct-signer counting from `3993fd1` and adds a test that a forged maintainer signature is not counted — and it is blocked on a dependency bump.

Still two lines, both inside `h2`'s lockfile entry. 18/18 green here, approved by @beardthelion since 14:19 on the 18th.

@kevincodex1 — no pressure on the approach; if you would rather this went in as part of a wider dependency sweep, or run from a Linux box with a fresh index, I will close this in favour of that. It is only the sitting-still that costs anything.

beardthelion:
Closing as superseded on `main`.

@Vasanthdev2004 flagged RUSTSEC-2026-0258 when `cargo audit` started failing repo-wide, and the hand-edited lockfile approach here (bump only `h2`'s version and checksum, no unrelated resolver drift) was the right fix. That analysis held up in review.

The advisory is cleared on `main` via `861cced8` (`chore: update h2 0.4.13 → 0.4.18 to resolve RUSTSEC-2026-0258`), which landed in merge commit `21c74c65` (#223). `cargo audit` is green on current `main` (`73fd7473`). This branch targeted 0.4.16; `main` is already on 0.4.18, so there is nothing left to merge from here.

Thanks for getting this unblocked.

Issue #367, opened 2026-08-18
Write a language-agnostic protocol specification (PROTOCOL.md)


andreolf:
## Summary

There's no language-agnostic protocol specification, so a node can only be implemented by reading the Rust source. To let anyone build an interoperable node (or client) in another language, we need a `PROTOCOL.md` (or `docs/PROTOCOL.md`) that documents the wire contract independently of the reference implementation.

## Why

"Anyone can run a node" is only true if the protocol is specified separately from `gitlawb-node`. Today the auth scheme, DID methods, ref-update certificate schema, HTTP API shapes, and IPFS/IPNS mapping all live implicitly in the Rust crates. A spec:

- lets a second implementation (Go, TypeScript, Python…) interoperate with existing nodes;
- pins down what is **protocol** (must match across implementations) vs. **implementation detail** (free to differ);
- gives reviewers and integrators a single source of truth.

## Proposed scope

A first `PROTOCOL.md` covering, sourced from the current code:

1. **Identity & DIDs** — `did:key` / `did:web` / `did:gitlawb`; key type (Ed25519); DID → verifying-key resolution.
2. **Authentication** — RFC 9421 HTTP Signatures. The exact `Signature-Input` covered components (`@method`, `@path`, `content-digest`), `keyid` = signer DID, `alg="ed25519"`, `created`, and `Content-Digest` construction. Which routes require signatures.
3. **iCaptcha (proof-of-intelligence)** — the `403 icaptcha_proof_required` flow, `x-icaptcha-url` / `x-icaptcha-level` / `x-icaptcha-proof` headers, which writes are gated.
4. **Ref-update certificate** — the frozen v1 schema (`gitlawb/ref-update/v1`): body fields, canonical signing bytes, signature entries, threshold/countersignature semantics.
5. **Attestations** — the external-provenance attestation format bound to a cert hash, and verification policy (`RequireAll` leniency).
6. **HTTP API surface** — the `/api/v1/*` resources (repos, refs, certs, agents, tasks, bounties, peers, resolve, stats) with request/response shapes and status/error conventions (incl. 404-shaped denials).
7. **Git transport** — smart-HTTP endpoints (`/{owner}/{repo}/info/refs`, `git-upload-pack`) and the `gitlawb://` remote-helper URL scheme.
8. **Content addressing / storage** — git SHA-256 → IPFS CID mapping, branch refs as IPNS records, `/ipfs/{cid}` retrieval. (Which parts are normative vs. optional.)
[...]

Issue #371, opened 2026-08-20
Language-agnostic protocol specification (PROTOCOL.md) so nodes can be built in any language


andreolf:
## Summary

There's no language-agnostic protocol specification, so a node can only be implemented by reading the Rust source. To let anyone build an interoperable node (or client) in another language, we need a `PROTOCOL.md` (or `docs/PROTOCOL.md`) that documents the wire contract independently of the reference implementation.

## Why

"Anyone can run a node" is only true if the protocol is specified separately from `gitlawb-node`. Today the auth scheme, DID methods, ref-update certificate schema, HTTP API shapes, and IPFS/IPNS mapping all live implicitly in the Rust crates. A spec:

- lets a second implementation (Go, TypeScript, Python…) interoperate with existing nodes;
- pins down what is **protocol** (must match across implementations) vs. **implementation detail** (free to differ);
- gives reviewers and integrators a single source of truth.

## Proposed scope

A first `PROTOCOL.md` covering, sourced from the current code:

1. **Identity & DIDs** — `did:key` / `did:web` / `did:gitlawb`; key type (Ed25519); DID → verifying-key resolution.
2. **Authentication** — RFC 9421 HTTP Signatures. The exact `Signature-Input` covered components (`@method`, `@path`, `content-digest`), `keyid` = signer DID, `alg="ed25519"`, `created`, and `Content-Digest` construction. Which routes require signatures.
3. **iCaptcha (proof-of-intelligence)** — the `403 icaptcha_proof_required` flow, `x-icaptcha-url` / `x-icaptcha-level` / `x-icaptcha-proof` headers, which writes are gated.
4. **Ref-update certificate** — the frozen v1 schema (`gitlawb/ref-update/v1`): body fields, canonical signing bytes, signature entries, threshold/countersignature semantics.
5. **Attestations** — the external-provenance attestation format bound to a cert hash, and verification policy (`RequireAll` leniency).
6. **HTTP API surface** — the `/api/v1/*` resources (repos, refs, certs, agents, tasks, bounties, peers, resolve, stats) with request/response shapes and status/error conventions (incl. 404-shaped denials).
7. **Git transport** — smart-HTTP endpoints (`/{owner}/{repo}/info/refs`, `git-upload-pack`) and the `gitlawb://` remote-helper URL scheme.
[...]

Issue #253, opened 2026-07-27
Signed requests are replayable for ~600s and are not bound to a target host


beardthelion:
`require_signature` verifies a signature and keeps no record that it did. There is no nonce, no spent-signature ledger, and no per-request state, so a captured `Signature-Input` + `Signature` + body triple is a bearer credential that works repeatedly until the clock-skew check expires it. Separately, nothing in the signing string names the target host, so the same triple is accepted by any node serving that path.

Filing this publicly as a deliberate call despite `SECURITY.md`, because the exposure is bounded to actors the threat model already treats as untrusted (see Reachability), and the fix needs design discussion in the open.

## What was measured

All against `origin/main` at 111cff7, driving the real `require_signature` middleware over a migrated Postgres, using signatures produced by `gitlawb_core::http_sig::sign_request` exactly as `gl` and the git helper produce them.

1. **Replay.** One signature over `POST /api/v1/tasks`, submitted three times unmodified: `201`, `201`, `201`, and three `agent_tasks` rows with three distinct UUIDs. The route mints its id server-side (`api/tasks.rs:106`), so replays cannot collide.
2. **No host binding.** The same signature sent with `Host: node-a.gitlawb.com`, then `node-b.example.org`, then `attacker.invalid`. All three `201`.
3. **The window is two-sided.** `check_created` compares `(now - created).abs()` against 300 (`gitlawb-core/src/http_sig.rs:117-126`). A hand-built signature dated 250s in the *future* is accepted, so one signature can be valid across roughly a 600s span.
4. **The `Signature` header is malleable.** The exact header, plus leading-space, trailing-space and leading-tab variants: four distinct byte strings, all four verify. `HttpSignature::parse` trims internally (`http_sig.rs:97`) but the middleware never uses the trimmed value.

Reproducing (1) takes about fifteen lines: mount any handler behind `crate::auth::require_signature`, sign one request, send the identical bytes twice.

## Why there is no defense

`require_signature` is `middleware::from_fn` and never receives `AppState` (`auth/mod.rs:51`), so it structurally cannot record what it has seen. Its checks are parse, `check_created`, `missing_components`, alg, Ed25519 verify, content-digest compare, then `next.run`.

[...]

PR #372, opened 2026-08-20
docs: implement language-agnostic PROTOCOL.md wire spec

ronr47:
Addresses #367 by defining identity, RFC 9421 auth, iCaptcha gates, and ref-update certificate schemas.

## Summary by CodeRabbit

* **Documentation**
  * Added the Gitlawb Wire Protocol Specification v1.
  * Documented supported identities, HTTP authentication, request integrity, iCaptcha proof requirements, ref-update certificates, and Smart-HTTP/Git storage mappings.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -1,23 +1,77 @@
-# Gitlawb Wire Protocol Specification (v1)
+# Gitlawb Wire Protocol Specification (v1, alpha)
+
+This document describes the wire behavior a second implementation can interop
+with **today**. Anything the node does not ship yet is explicitly marked
+*planned*; nothing below is aspirational unless it says so.
 
 ## 1. Identity & DIDs
-- **Key Types:** Ed25519 (`ed25519`).
-- **Methods:** `did:key`, `did:web`, `did:gitlawb`.
-- **Resolution:** Deterministic public key extraction from multicodec prefixes.
+
+- **Key type:** Ed25519.
+- **Method (alpha):** `did:key` only. Resolution is deterministic public-key
+  extraction from the multicodec prefix (`z6Mk…`, multicodec `0xed01`).
+- Signed requests presenting any other DID method are rejected; the node's
+  auth layer answers with a hint that only `did:key` is supported in alpha.
+- *Planned:* `did:web` and `did:gitlawb` resolution. Clients MUST NOT assume
+  either is accepted anywhere yet.
 
 ## 2. Authentication (RFC 9421)
-- **HTTP Signatures:** Requires `Signature-Input` covering `@method`, `@path`, and `content-digest`.
-- **Signer Identity:** `keyid` points to the actor DID; algorithm is `alg="ed25519"`.
-- **Integrity:** SHA-256 `Content-Digest` verification on write endpoints.
+
+- **HTTP signatures:** `Signature-Input` covers `@method`, `@path`, and
+  `content-digest`:
+  `sig1=("@method" "@path" "content-digest");keyid="did:key:…";alg="ed25519";created=<unix>`
+- **Signer identity:** `keyid` is the actor DID (`did:key` only, see §1).
+- **Integrity:** `Content-Digest` is SHA-256 over the request body and is
+  verified when the header is present.
 
 ## 3. Proof of Intelligence (iCaptcha)
-- **Gate:** Enforced via `403 icaptcha_proof_required` responses.
-- **Headers:** Clients must present `x-icaptcha-url`, `x-icaptcha-level`, and `x-icaptcha-proof`.
+
+- **Mode:** controlled by `ICAPTCHA_MODE` = `off` | `shadow` | `enforce`,
+  **default `off`**. A node with the gate off never demands proof.
+- **Scope:** the gate applies to a small set of write endpoints (agent
+  registration and repo create/fork), not to all writes.
+- **Wire shape:** the client-supplied header is `x-icaptcha-proof` — that is
+  the only iCaptcha header a node reads from a request. When proof is
+  required and missing/invalid, the node answers
+  `403 icaptcha_proof_required` and sets `x-icaptcha-url` and
+  `x-icaptcha-level` **on the response** (also mirrored as JSON fields) so
+  the client can go solve a challenge and retry.
 
 ## 4. Ref-Update Certificate
-- **Schema:** `gitlawb/ref-update/v1`.
-- **Payload:** Canonical JSON bytes defining target repo, commit OID, previous OID, and actor DID.
+
+Two distinct certificate-shaped payloads exist; this section documents the
+**core signing schema**, which is what `gitlawb/ref-update/v1` names:
+
+- **Type string:** `gitlawb/ref-update/v1`.
+- **Body (canonical JSON, the signed bytes):** `type`, `repo` (the
+  repository DID), `ref_name`, `from` (previous hash, 64 hex chars,
+  all-zeros for a new ref), `to` (target hash, 64 hex chars), `seq`
+  (monotonically increasing, replay prevention), `timestamp` (RFC 3339),
+  `nonce`.
+- **Signatures:** a list of `{signer: did:key, sig: base64url-unpadded
+  Ed25519}` entries appended outside the signed body; thresholds are counted
+  over distinct signers that verify.
+
+The node's REST API separately serves **push receipts** — a different,
+node-issued JSON payload attesting that a push was processed. Receipts are
+not `gitlawb/ref-update/v1` documents and are not interchangeable with them;
+they are documented with the node's HTTP API, not here.
 
 ## 5. Storage & Git Transport
-- **Smart-HTTP:** Endpoints at `/{owner}/{repo}/info/refs` and `git-upload-pack`.
-- **Content Addressing:** Git SHA-256 mapped to IPFS CID chunks with IPNS-backed branch pointers.
+
+- **Object format:** bare repositories are created with
+  `--object-format=sha1`; production OIDs are 40 hex chars. *Planned:*
+  SHA-256 object format.
+- **Smart-HTTP endpoints:**
+  - `GET /{owner}/{repo}/info/refs` — ref advertisement (fetch and push
+    service discovery).
+  - `POST /{owner}/{repo}/git-upload-pack` — clone/fetch.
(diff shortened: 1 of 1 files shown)


kevincodex1:
hi @ronr47 thanks a lot for initiating this. I'm thinking to put the protocol  and specs in another repo for it to be clean . will ping you once its good

Vasanthdev2004 (commented):
Useful thing to have — a wire spec is exactly what a second implementation needs, and #371 has been open a while. I checked each claim against the code, since a spec's value is entirely in whether an implementer can build against it. Three don't match today's behaviour.

**§1 — `did:web` and `did:gitlawb` do not resolve the way this says.**

> **Methods:** `did:key`, `did:web`, `did:gitlawb`.
> **Resolution:** Deterministic public key extraction from multicodec prefixes.

All three are constructible (`Did::web`, `Did::gitlawb`), but `to_verifying_key` refuses everything except `did:key` — `crates/gitlawb-core/src/did.rs:75-81` returns `expected did:key, got did:{method}`, and there are tests named `to_verifying_key_fails_for_did_web` and `to_verifying_key_fails_for_did_gitlawb` pinning it. Deterministic multicodec extraction only applies to `did:key`; `did:web` needs a fetch and `did:gitlawb` needs a DHT lookup the module docstring says is not live yet ("migrates to `did:gitlawb` once DHT anchoring is live").

This is the one I would most want fixed before merge: an implementer reading it would accept a `did:web` signer that every gitlawb node rejects.

**§5 — objects are SHA-1, not SHA-256.**

> **Content Addressing:** Git SHA-256 mapped to IPFS CID chunks…

Repos are created with an explicit `--object-format=sha1` (`crates/gitlawb-node/src/git/store.rs:15`). Some local variables are named `sha256_hex` further down that file, which is misleading, but the `git init` is unambiguous.

**§5 — IPNS is not implemented.**

> …with IPNS-backed branch pointers.

The only occurrence of IPNS in the tree is a comment calling an endpoint "IPNS-**style**" (`crates/gitlawb-node/src/api/repos.rs:2605`). Nothing publishes or resolves an IPNS name. Worth either dropping or marking explicitly as planned — a spec that mixes shipped and intended behaviour without distinguishing them is hard to implement against.

**Correct as written:** the RFC 9421 covered components match `COVERED_COMPONENTS = ["@method", "@path", "content-digest"]` exactly (`http_sig.rs:24`); `gitlawb/ref-update/v1` matches `CERT_TYPE` (`cert.rs:20`); the iCaptcha header names match what `gl` sends.

**One gap:** §5 lists `info/refs` and `git-upload-pack` but not `git-receive-pack`. That is the entire write path, and it is where the interesting protocol requirements live — signatures are mandatory there, and `X-Ucan` rides along for delegated pushes. A wire spec that documents only fetch leaves out the half a second implementation is most likely to get wrong.

Suggestion, and take it or leave it: a short "Status" column or a **Shipped / Planned** marker per section would let this land now and grow, instead of needing to be exactly right about a moving target. Happy to help fill in the receive-pack section if that is useful — I have been in that path recently.

Not a maintainer, so this is a review rather than a gate.

beardthelion (dismissed):
Useful direction for #371. I read PROTOCOL.md at head 3b00725f against the tree and checked identity resolution, git object format, smart-HTTP routes, RFC 9421 components, certificate types, and iCaptcha wiring. Several sections read as shipped node behavior when the code is alpha-scoped, uses SHA-1 repos, or implements a different schema than the sketch names.

## Findings

- [P2] Scope identity resolution to did:key for alpha
`PROTOCOL.md:6`
Section 1 lists did:web and did:gitlawb with deterministic multicodec extraction. to_verifying_key only accepts did:key (did.rs:75-86), with tests pinning failure for web and gitlawb (did.rs:270-286). HTTP auth rejects other keyids with "only did:key is supported in alpha" (auth/mod.rs:149). An interop client that signs as did:web will be rejected on every gated route.

- [P2] Correct storage claims to SHA-1 repos and drop or mark IPNS as planned
`PROTOCOL.md:23`
Bare repos are created with --object-format=sha1 (store.rs:15). Section 5's "Git SHA-256 mapped to IPFS CID chunks with IPNS-backed branch pointers" overstates both object format and IPNS. IPNS appears only in an "IPNS-style" comment on the branch refs endpoint (repos.rs:2585). Either label IPNS planned or describe the REST /refs CID metadata that actually ships.

- [P2] Document git-receive-pack and the fetch-vs-push auth split
`PROTOCOL.md:22`
Section 5 lists info/refs and git-upload-pack only. The node also serves POST git-receive-pack with mandatory auth (git_receive_pack in repos.rs, routed in server.rs). Fetch advertisement can 404 for withheld repos while receive-pack expects 401 on unsigned push. A spec covering only fetch omits the write path a second implementation is most likely to get wrong.

- [P2] Fix iCaptcha wire semantics and deployment scope
`PROTOCOL.md:15`
Section 3 says clients present x-icaptcha-url, x-icaptcha-level, and x-icaptcha-proof. The node reads only x-icaptcha-proof from clients (icaptcha.rs:32). It sets url and level on 403 responses (error.rs:108-134). ICAPTCHA_MODE defaults to off (icaptcha.rs:10-13) and the gate applies to a small endpoint set, not all writes.

- [P2] Complete or split ref-update certificate schemas
`PROTOCOL.md:18`
Section 4 names gitlawb/ref-update/v1 and a minimal payload sketch. Core's RefUpdateCert includes ref_name, seq, timestamp, nonce, and signatures (cert.rs:32-50). Node push receipts use a different JSON payload (gitlawb-node/src/cert.rs:29-38) served from the API. Say which schema this section documents, or split push authorization certs from node issuance receipts.

One process note, not a finding: kevincodex1 noted on the timeline that protocol docs may move to a separate repo. Worth aligning with the contributor before this lands on main if that plan is still active.

Not an ask, recorded only: RFC 9421 covered components and the gitlawb/ref-update/v1 type string match shipped core definitions. Content-Digest body verification runs only when the header is present (auth/mod.rs:221-237; gap tracked separately in #306).

beardthelion (changes requested):
Round 1's five findings all landed. I checked each against source at 31343d25: `did:key` scoping, `--object-format=sha1`, the receive-pack auth split, the `x-icaptcha-proof` wire shape, and the cert/receipt schema split are each accurate now. The doc got substantially more honest. What is left is one wrong claim and one section that should not be here.

## Findings

- [P1] Say that `@path` carries the query string, not RFC 9421's path-only `@path`
`PROTOCOL.md:19-21`
`auth/mod.rs` binds `@path` from `parts.uri.path_and_query()`. RFC 9421 defines `@path` as the absolute path with `@query` as a separate component, so an implementation that reads this line and follows the RFC builds a different signing string and gets a signature failure. It fires immediately: `?service=` is mandatory on `GET /info/refs` (`crates/gitlawb-node/src/api/repos.rs:557-559`, 400 without it), so the one signed endpoint the spec names is query-bearing. Either state the non-standard semantics explicitly, or treat this as a code bug and add `@query` to the covered set.

- [P1] Mark section 4 as planned or drop it; `gitlawb/ref-update/v1` has no wire presence
`PROTOCOL.md:39-52`
`git grep 'RefUpdateCert\|RefUpdateBody'` at head returns nothing under `gitlawb-node`, `gl`, or `git-remote-gitlawb`. The types live in `gitlawb-core` and `gitlawb-attest`, and `gitlawb-attest` is a workspace member that no crate depends on. The doc's own premise says nothing below is aspirational unless it says so, and this section says nothing. The schema is also unconstructible as specified: lines 46-47 require `from`/`to` at 64 hex (`crates/gitlawb-core/src/cert.rs:158-166` enforces exactly that), while line 62 states production OIDs are 40 hex.

- [P2] State what `content-digest` holds on a body-less request
`PROTOCOL.md:23-24`
The covered-components set is mandatory (`crates/gitlawb-core/src/http_sig.rs:24`; a missing component is a 400 `incomplete_signature`), so every signed request must sign over `content-digest`, including a body-less GET. The node reads an absent header as the empty string. "Verified when the header is present" is true, but it leaves an implementer guessing what to sign for `info/refs`.

- [P2] Document the smart-HTTP requirements a client cannot guess
`PROTOCOL.md:64-69`
The endpoint list omits the mandatory `?service=` parameter and its accepted values, pkt-line framing, and the `application/x-git-upload-pack-result` / `application/x-git-receive-pack-result` content types. Section 5 is the part a second implementation most needs and currently the least buildable from.

I left three CodeRabbit asks alone rather than passing them on: the iCaptcha token internals (a client forwards a proof, it does not mint one, and the gate defaults off), the 404/401 per-surface nuance, and the CIDv1 derivation formula. Those are omissions, not wrong claims, and I would rather keep this round short.

One note on CI: `PR Checks` is sitting at `action_required` awaiting approval on this head, so the only green check is quality-signal triage. Worth knowing that the suite has not actually run against your changes yet.

beardthelion (commented):
One more for the next round, found while fixing the client-side version of it. My
earlier request for changes still stands; this adds to it rather than replacing it.

## Findings

- [P2] State the transport requirement in the RFC 9421 section
  `PROTOCOL.md` section 2 (Authentication), with a pointer from section 5
  The spec pins the covered components, `keyid`, `alg`, `created` and
  `Content-Digest`, which is the right level of detail, but it never says the
  request must be sent over TLS. RFC 9421 authenticates a request; it does not
  encrypt one. A second implementation that follows this document exactly can
  therefore ship a client that puts the `Signature` header and the request body
  on the wire in cleartext and still be spec-compliant.
  This is not hypothetical: our own `git-remote-gitlawb` did exactly that until
  #414, because nothing told it otherwise. Combined with #253, where a signed
  request is replayable for roughly 600s and is not bound to a target host, a
  cleartext hop hands a passive observer a usable credential for that window
  rather than requiring a compromised endpoint.
  A sentence or two is enough: `https://` required for any non-loopback node,
  plaintext acceptable only to loopback for local development. Worth saying
  explicitly that the signature is not a substitute for transport encryption,
  since that is the specific inference a reimplementer is likely to make.

Checked at `31343d25`: no line in the document mentions TLS, `https`, or
transport confidentiality, and section 5 covers the smart-HTTP routes and the
`gitlawb://` URL scheme without touching it either.

Separately, and purely so it does not get lost: @kevincodex1 mentioned above that
the spec may move to its own repo. If it does, this finding should travel with the
document rather than close with this PR.

PR #376, opened 2026-08-23
fix(api,sync): Delimit positional arguments in git clone mirror invocations

### LINKED ISSUE #374: fix(api,sync): Delimit positional arguments in git clone mirror invocations
## Description
In `crates/gitlawb-node/src/api/repos.rs` (`fork_repo`) and `crates/gitlawb-node/src/sync.rs` (`clone_repo`), `git clone --mirror` passes positional arguments without preceding `--` delimiters.

## Proposed Fix
Add `--` before positional path/URL arguments in `git clone --mirror` invocations across `repos.rs` and `sync.rs`.

euxaristia:
## Summary
Passes `--` before positional path and URL arguments in `git clone --mirror` subprocess invocations to ensure positional inputs cannot be interpreted as command-line flags.

## Changes
- `crates/gitlawb-node/src/api/repos.rs`: Add `--` before positional arguments in `fork_repo`.
- `crates/gitlawb-node/src/sync.rs`: Add `--` before positional arguments in `clone_repo`.
- `crates/gl/src/mirror.rs`: Add `--` before positional arguments in `gl mirror`.

## Test plan
- `cargo check --workspace`
- `git diff HEAD --check`

Fixes #374

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cloning, mirroring, forking, and synchronization reliability for repository paths or URLs beginning with a hyphen.
  * Prevented Git from misinterpreting repository inputs as command-line options.
  * Improved recovery behavior when refreshing repository data or processing failed blob and manifest sources.
  * Improved handling of overloaded repository write operations and disconnected push processes.
  * Sanitized warning output for safer error reporting.

* **Tests**
  * Added regression coverage for safe repository inputs, recovery failures, stalled operations, and write-operation handling.

### DIFF
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -3071,6 +3071,7 @@ pub async fn fork_repo(
         .args([
             "clone",
             "--mirror",
+            "--",
             source_path.to_str().unwrap_or(""),
             disk_path.to_str().unwrap_or(""),
         ])
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -3000,6 +3000,17 @@ pub struct ForkRepoRequest {
 }
 
 /// POST /api/v1/repos/:owner/:repo/fork
+/// Argv for the fork's mirror clone, with `--` before the first positional (#374).
+///
+/// Both positionals here are server-derived paths rather than caller-supplied
+/// URLs, so this is defence in depth rather than a reachable injection. It is
+/// still the same argv contract as the other clone sites, and it is extracted for
+/// the same reason: so the delimiter's placement is pinned by a test instead of
+/// resting on review.
+fn fork_clone_args<'a>(source: &'a str, dest: &'a str) -> [&'a str; 5] {
+    ["clone", "--mirror", "--", source, dest]
+}
+
 pub async fn fork_repo(
     State(state): State<AppState>,
     Extension(auth): Extension<AuthenticatedDid>,
@@ -3068,13 +3079,10 @@ pub async fn fork_repo(
 
     // Clone the source repo as a mirror
     let output = std::process::Command::new("git")
-        .args([
-            "clone",
-            "--mirror",
-            "--",
+        .args(fork_clone_args(
             source_path.to_str().unwrap_or(""),
             disk_path.to_str().unwrap_or(""),
-        ])
+        ))
         .output()
         .map_err(|e| AppError::Git(format!("git clone --mirror failed: {e}")))?;
 
@@ -3336,6 +3344,29 @@ fn dedupe_canonical_repos(rows: Vec<(RepoRecord, i64)>) -> Vec<(RepoRecord, i64)
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    /// #374: the fork path builds its own `git clone --mirror` argv. Driven
+    /// through real git so a missing or misplaced `--` fails, matching the
+    /// coverage on the `gl clone`, `gl mirror`, and sync sinks.
+    #[test]
+    fn fork_clone_never_parses_the_source_as_a_git_option() {
+        let td = tempfile::TempDir::new().unwrap();
+        let dest = td.path().join("fork.git");
+        let injected = "--upload-pack=false";
+
+        let out = std::process::Command::new("git")
+            .args(fork_clone_args(injected, dest.to_str().unwrap()))
+            .output()
+            .unwrap();
+        let stderr = String::from_utf8_lossy(&out.stderr);
+
+        assert!(!out.status.success(), "an option-shaped source must fail");
+        assert!(
+            stderr.contains(&format!("repository '{injected}'")),
+            "git must name the source as the repository, not consume it as an option: {stderr}"
+        );
+    }
+
     use crate::auth::caller_authorized_to_push;
     use crate::error::AppError;
     use gitlawb_core::identity::Keypair;
--- a/crates/gitlawb-node/src/sync.rs
+++ b/crates/gitlawb-node/src/sync.rs
@@ -1041,6 +1041,51 @@ mod tests {
         );
     }
 
+    /// #374: `fetch_repo` interpolates the peer's origin URL into
+    /// `git remote set-url`, a second argv construction with the same contract as
+    /// the clone. It needs its own proof: the clone test cannot reach this path.
+    ///
+    /// The assertion is on stored state rather than on the error string, and that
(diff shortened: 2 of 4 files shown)


beardthelion (approved):
The delimiter is right, and the regression test earns its place. I gutted `args.push("--")` and the
test goes red on its own assertion message, so it is pinning the behavior rather than passing
alongside it. `git clone --mirror -- <src> <dst>` still clones cleanly, the Promisor arm keeps
`--filter=blob:limit=10g` ahead of the delimiter, and the sync module is 42 green on this head.

One correction to the framing rather than the code. I could not reach the execution this closes.
Peer origins are composed as `http(s)://...` and `is_public_http_url` rejected every dash-leading
form I drove through it, so `remote_url` cannot begin with `-`. Independently of that gate, even a
hostile value fails: consuming it as an option leaves the destination as the only positional, and
`process_batch` calls `clone_repo` only when that path does not exist, so git dies at
`fatal: repository '<dest>' does not exist` before any transport opens. I confirmed `--upload-pack`
really does execute in other shapes, just not in this one. So this is hardening, which is still
worth taking, and I have relabelled #374 to sev:low to match.

Approving. Three follow-ups, none of them blocking.

## Findings

- [P3] Carry the delimiter into `fetch_repo`
  `crates/gitlawb-node/src/sync.rs:733`
  `remote set-url origin <remote_url>` takes the same string you just delimited in `clone_repo`,
  still undelimited. `process_batch` branches on whether the mirror path exists, so `clone_repo`
  runs once per repo and this runs on every sync after it. I drove option-shaped values through the
  sink and they all fail loudly with a usage or unknown-option error and leave the stored URL
  untouched, so this is consistency rather than a hole. `git remote set-url origin -- <url>` is
  accepted, so it is the same one-line change.

- [P3] Give `gl clone` the same treatment as `gl mirror`
  `crates/gl/src/clone.rs:113`, `:114`, `:120`
  The URL and the `--branch` value are undelimited positionals at the same trust level as the
  `mirror.rs` source you fixed here. I ran all three of those argv shapes with `--` inserted and git
  accepts each one. Fixing one command and leaving its sibling is the gap worth closing while the
  change is fresh.

- [P3] Correct the error string the test docstring quotes
  `crates/gitlawb-node/src/sync.rs:1017`
  The comment says the undelimited form fails with "Could not read from remote repository". Run both
  ways it gives `fatal: repository '<dest>' does not exist`, because git resolves the remaining
  positional as a local path and never opens a transport. The conclusion the docstring draws is
  correct and the test is unaffected, so only the quoted string needs changing. While you are in
  there, "arbitrary execution on the node" claims more than this call shape can do.

The red `cargo audit` is not yours. RUSTSEC-2026-0258 landed against h2 0.4.13, which `main` carries
too, and this branch does not touch `Cargo.lock`. It needs its own bump on main.

kevincodex1 (approved):

kevincodex1:
@euxaristia please address the cargo audit and test

jatmn (changes requested):

## Overall guidance

There are two remaining items, not a broad set of unrelated defects. They have a common reviewability root cause: this security-hardening change was developed on a branch whose merge resolution did not preserve a current-main API migration, and its regression proof was added for one implementation of the repeated Git-argv pattern rather than for every newly changed CLI sink.

Please avoid another piecemeal update. Reconstruct or rebase the branch on the current `main` first, resolve and inspect the resulting diff, then validate the complete intended change as one unit:

1. **Start from the live target, not the old merge base.** Preserve the current `read_body_capped(...).await.text` contract in `gl whoami` before addressing delimiter behavior. A green compile is the prerequisite for trusting any test result.
2. **Treat every changed `Command::new("git")` argv construction as one security contract.** For each path changed by this PR, identify which values are Git options and which are untrusted or variable positionals. Keep all real Git options before `--`; place `--` immediately before the first positional; do not move option values such as `--branch <branch>` across the delimiter.
3. **Make the tests prove the property, not merely successful cloning.** For each changed CLI command shape, use an option-shaped remote/source (for example `--upload-pack=false`) and assert Git treats the entire value as the nonexistent repository. The assertion should fail if `--` is removed or placed after the positional. Keep the existing happy-path tests because they independently prove ordinary clone, branch selection, and sparse/promisor behavior still work.
4. **Run the required checks from the rebased head.** At minimum, run the focused `gl` and sync tests that exercise these paths, followed by the repository-required format, clippy, build, and test checks. Do not interpret a targeted delimiter test as passing while `gl` cannot compile.

This keeps the fix narrow: no command redesign is needed, no new input policy is requested, and the existing public clone, selected-branch, sparse/promisor, sync, fork, and mirror semantics should remain unchanged.

## Merge readiness

- [P1] Rebase onto current `main` and resolve the `whoami` API conflict
  `crates/gl/src/whoami.rs:63`
  This is target-branch drift rather than a defect in the delimiter work. `read_body_capped()` now returns `CappedBody`, whose decoded response body is in `.text`; current `main` already uses `read_body_capped(resp, 8 * 1024).await.text` at this call site. The PR head retained the older expression without `.text`, then passes the `CappedBody` struct to both `serde_json::from_str` (which requires `&str`) and `Option<String>::unwrap_or` (which requires `String`). That produces E0308 and prevents every `gl` build/test target from compiling, including the new delimiter tests.

  Please rebase or reconstruct the branch on current `main` and resolve this merge against the current `read_body_capped` contract by extracting the body text before JSON parsing and fallback selection. Then rerun the required Rust checks. The root cause is an incomplete/stale merge resolution: the live-target API migration was not carried into the PR head.

## Findings

- [P3] Add regression coverage for every changed CLI delimiter sink
  `crates/gl/src/clone.rs:160`
  This is a regression-coverage gap, not a reproduced runtime exploit: direct Git probes confirm that the current delimiter makes an option-shaped value such as `--upload-pack=false` be treated as the repository. But the PR claims to harden clone/mirror positional arguments across the command paths it changes, while only `sync::clone_repo` has a pre-fix-red test for that behavior. The new `gl mirror` invocation and the changed `gl clone` plain, selected-branch, and sparse/promisor invocations have happy-path tests using ordinary URLs only. Removing or misplacing `--` in any of those CLI sinks can therefore leave the suite green.

  Please add focused subprocess-facing regressions for the changed CLI paths. Use an option-shaped source/remote and assert that Git reports that exact value as a nonexistent repository rather than reporting option parsing or a different positional error. Cover `gl mirror` and the `gl clone` argument shapes changed here; keep the existing ordinary URL, branch-selection, and sparse/promisor tests. The root cause is that the hardening contract is duplicated across several direct Git argv constructions, but its negative-path test only pins the sync implementation.

jatmn (approved):
LGTM

beardthelion:
@kevincodex1 LGTM

PR #377, opened 2026-08-24
fix(gl): derive View: URL from node instead of hardcoding gitlawb.com (#370)

### LINKED ISSUE #370: gl repo create prints a View: URL on gitlawb.com that 404s for any repo not on node.gitlawb.com
## What happens

`gl repo create` prints a `View:` URL that 404s for any repo not hosted on `node.gitlawb.com`.

Creating a repo against a self-hosted node:

```
$ GITLAWB_NODE=https://origin.gitlawb.com gl repo create my-repo
✓ Created repository: my-repo
  Clone: git clone gitlawb://did:key:z6Mk…/my-repo
  HTTP:  https://origin.gitlawb.com/z6Mk…/my-repo.git
  View:  https://gitlawb.com/z6Mk…/my-repo          ← 404
```

`Clone:` and `HTTP:` are both correct and both work. `View:` is built against `gitlawb.com` regardless of which node the repo was just created on:

```
https://gitlawb.com/<did>/<repo>
  → 307 → https://explorer.gitlawb.com/repos/<did>/<repo>
  → 404
```

The explorer is backed by the main network, so it can only resolve repos that node knows about. Confirmed by asking each node directly for the same repo:

| node | `/api/v1/repos/<did>/<repo>` |
|---|---|
| `origin.gitlawb.com` | 200 |
| `node.gitlawb.com` | 404 |
| `node2.gitlawb.com` | 404 |
| `node3.gitlawb.com` | 404 |

## Why it matters

This is the first thing a self-hosting operator sees after their first successful command, and it is the one line of the three that is wrong. The failure is also silent in the unhelpful direction: the URL is well-formed and the 307 makes it look like it resolved somewhere real, so the natural conclusion is "my repo did not get created" rather than "this link was never going to work".

Anyone posting or sharing that link — which is what a `View:` line invites — shares a 404.

## Two things tangled together

1. **The URL is unconditional.** `View:` should be derived from the node the repo was created on, or omitted when that node has no known web view. Printing a link to a service that cannot see the repo is worse than printing nothing.

2. **`origin.gitlawb.com` has no peers**, so nothing it hosts can propagate anywhere the explorer could find it:
   ```
   $ curl https://origin.gitlawb.com/api/v1/peers
   {"count":0,"peers":[]}
   ```
   That may be intentional for that particular node, but it means (1) is not a display bug alone — there is genuinely no URL that would work for this repo today. Worth deciding separately whether that node should be peered and added to `bootstrap-peers.json`.

## Suggested direction

For (1): only print `View:` when the target node is one with a known web view — or make the web view host a property the node advertises on `GET /`, alongside `did` and `version`, and print the line only when the node supplies one. That keeps the CLI from hardcoding a topology it cannot verify.

## Environment

- `gl` @ `fad75a6`, reports `gl 0.7.1`
- node `origin.gitlawb.com`, v0.7.1, `network: alpha`

Gravirei:
## Summary

gl repo create and gl mirror print a `View:` link unconditionally pointed at `gitlawb.com`, which 404s for repos on self-hosted nodes.

## Motivation & context

Closes #370

After `gl repo create` on a self-hosted node, the `View:` URL resolves through `gitlawb.com` → `explorer.gitlawb.com`, which can only see repos on the main network. Self-hosted repos silently 404. The fix lets nodes advertise an optional `web_url` so the CLI only prints a working link.

## Kind of change

- [x] Bug fix

## What changed

- **gitlawb-node/config.rs**: Added `web_url: Option<String>` config field (`GITLAWB_WEB_URL` env var). When set, the node advertises it in `GET /`.
- **gitlawb-node/server.rs**: `node_info` conditionally includes `web_url` in the response.
- **gitlawb-node/main.rs**: `DegradedState`, `build_degraded_router`, `run_degraded_server`, and `degraded_node_info` propagate and include `web_url`.
- **gl/repo.rs**: After repo creation, fetches `GET /` and only prints `View:` when the node supplies `web_url`.
- **gl/mirror.rs**: Same pattern — only prints `View:` when `web_url` is present.

## How a reviewer can verify

```sh
# 1. Node without GITLAWB_WEB_URL — View: line should NOT appear
cargo run -p gitlawb-node &
GITLAWB_NODE=http://localhost:7545 gl repo create test-repo
# (no View: in output)

# 2. Node with GITLAWB_WEB_URL — View: line should use the configured URL
GITLAWB_WEB_URL=https://gitlawb.com cargo run -p gitlawb-node &
GITLAWB_NODE=http://localhost:7545 GITLAWB_WEB_URL=https://gitlawb.com gl repo create test-repo
# View:  https://gitlawb.com/<owner>/<repo>
```

## Before you request review

- [x] `cargo test --workspace` passes locally (pre-existing failures in events/arweave tests unrelated)
- [ ] New behavior is covered by tests (node_info has no existing test suite; CLI tests pass)
- [x] Docs / `.env.example` updated if behavior or config changed (or N/A) — new env var, no docs break

## Protocol & signing impact

## Notes for reviewers

- `web_url` is entirely opt-in: nodes that don't set `GITLAWB_WEB_URL` omit it from `GET /`, and the CLI silently skips the `View:` line. No behavior change for existing deployments.
- The `GET /` endpoint is already public (no auth), consistent with how node identity info is served.

## Summary by CodeRabbit

- **New Features**
  - Added optional web frontend URL configuration via `GITLAWB_WEB_URL` or `--web-url`.
  - Node information and degraded-mode responses include the configured URL when available.
  - Mirror and repository creation commands display a “View” link when supported.

- **Bug Fixes**
  - Removed hardcoded web links and suppress links when metadata is unavailable or invalid.
  - Web URLs must be absolute HTTP(S) addresses without query strings or fragments.
  - Improved warnings for unreachable nodes and invalid advertised URLs.

- **Documentation**
  - Added configuration guidance and examples for `GITLAWB_WEB_URL`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,6 +10,12 @@ GITLAWB_KEY=/data/keys/identity.pem
 # Publicly reachable URL of this node (used in peer announcements)
 GITLAWB_PUBLIC_URL=https://your-node.example.com
 
+# Base URL for the web view of repos on this node (used by `gl` to print a
+# working View: link after repo creation). Omit for nodes with no web front-end.
+# Distinct from GITLAWB_PUBLIC_URL: that is API reachability for peers;
+# this is browser reachability for humans.
+# GITLAWB_WEB_URL=https://gitlawb.com
+
 # ── Server ────────────────────────────────────────────────────────────────
 GITLAWB_HOST=0.0.0.0
 GITLAWB_PORT=7545
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2630,7 +2630,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ab67060fc6b8ef687992d439ca0fa36e7ed17e9a0b16b25b601e8757df720de"
 dependencies = [
  "data-encoding",
- "syn 2.0.117",
+ "syn 1.0.109",
 ]
 
 [[package]]
@@ -3002,7 +3002,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -3544,9 +3544,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.4.13"
+version = "0.4.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2f44da3a8150a6703ed5d34e164b875fd14c2cdab9af1252a9a1020bde2bdc54"
+checksum = "ef8e5e5a340588f4452631496976cf8636d4a7ecf600239fdc27615d2530bc16"
 dependencies = [
  "atomic-waker",
  "bytes",
@@ -3865,7 +3865,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "tokio",
  "tower-service",
  "tracing",
@@ -5433,7 +5433,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash",
  "rustls",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
@@ -5470,7 +5470,7 @@ dependencies = [
  "cfg_aliases",
  "libc",
  "once_cell",
- "socket2 0.6.3",
+ "socket2 0.5.10",
  "tracing",
  "windows-sys 0.52.0",
 ]
@@ -5902,7 +5902,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -6796,7 +6796,7 @@ dependencies = [
  "getrandom 0.4.2",
  "once_cell",
  "rustix",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2181,7 +2181,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2630,7 +2630,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ab67060fc6b8ef687992d439ca0fa36e7ed17e9a0b16b25b601e8757df720de"
 dependencies = [
  "data-encoding",
- "syn 1.0.109",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -3002,7 +3002,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3866,7 +3866,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "tokio",
  "tower-service",
  "tracing",
@@ -5434,7 +5434,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash",
  "rustls",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
@@ -5471,7 +5471,7 @@ dependencies = [
  "cfg_aliases",
  "libc",
  "once_cell",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "tracing",
  "windows-sys 0.52.0",
 ]
@@ -5903,7 +5903,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -6797,7 +6797,7 @@ dependencies = [
  "getrandom 0.4.2",
  "once_cell",
  "rustix",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3483,6 +3483,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "unicode-normalization",
+ "url",
  "uuid",
  "zstd",
 ]
@@ -3508,6 +3509,7 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
+ "url",
  "urlencoding",
  "uuid",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,12 +51,11 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 chrono = { version = "0.4", features = ["serde"] }
 # uuid
 uuid = { version = "1", features = ["v4"] }
-# URL parsing (absolute browser URLs for GITLAWB_WEB_URL / node web_url)
-url = "2"
 # http client
 reqwest = { version = "0.12", features = ["blocking", "json", "multipart", "rustls-tls"], default-features = false }
 # URL parsing (what reqwest::Url re-exports, so the shared redirect predicate can
-# take a parsed URL without pulling reqwest into gitlawb-core)
+# take a parsed URL without pulling reqwest into gitlawb-core; also used for the
+# absolute-browser-url validation on GITLAWB_WEB_URL in the node and CLI)
 url = "2"
 # HMAC
 hmac = "0.12"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,8 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 chrono = { version = "0.4", features = ["serde"] }
 # uuid
 uuid = { version = "1", features = ["v4"] }
+# URL parsing (absolute browser URLs for GITLAWB_WEB_URL / node web_url)
+url = "2"
 # http client
 reqwest = { version = "0.12", features = ["blocking", "json", "multipart", "rustls-tls"], default-features = false }
 # HMAC
--- a/crates/gitlawb-node/Cargo.toml
+++ b/crates/gitlawb-node/Cargo.toml
@@ -23,6 +23,7 @@ tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
 chrono = { workspace = true }
 uuid = { workspace = true }
+url = { workspace = true }
 axum = { version = "0.8", features = ["http1", "http2", "json", "ws"] }
 async-graphql = { version = "7", features = ["chrono", "uuid", "tracing"] }
 async-graphql-axum = "7"
--- a/crates/gitlawb-node/src/config.rs
+++ b/crates/gitlawb-node/src/config.rs
@@ -1527,6 +1527,8 @@ mod tests {
         let mut cfg = Config::parse_from(["gitlawb-node"]);
         cfg.web_url = Some("http://localhost:8080/ui".into());
         assert!(cfg.validate().is_ok(), "port+path web_url must validate");
+    }
+
     /// The DECLARED default, read off the parser rather than out of a parse.
     ///
     /// `Config::parse_from` consults the process environment, so on a host that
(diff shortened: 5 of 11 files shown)


beardthelion (dismissed):
I read the diff on head 9ca4fcff, ran `cargo test -p gl repo::tests::test_cmd_create_success` (green, no View: because GET / is unmocked), and premise-mutated both sides: removing the server web_url insert left the full gitlawb-node suite green; removing the CLI View block or restoring the hardcoded gitlawb.com link also left test_cmd_create_success green. The core design is sound: stop hardcoding gitlawb.com, advertise web_url from the node, print View: only when present, trim trailing slashes on the CLI. CodeRabbit's normalization ask is already on head. I am declining their silent-omit ask: this matches the existing fail-soft pattern for optional post-success hints (repo.rs replica count).

One process note, not a finding: open PR #325 also touches crates/gitlawb-node/src/config.rs and is likely to land first. Expect a rebase conflict there, not a design rework.

## Findings

- [P2] Pin the View-URL behavior with load-bearing tests
  `crates/gl/src/repo.rs:838`
  test_cmd_create_success mocks only POST /api/v1/repos and never asserts stdout. Restoring the hardcoded gitlawb.com View line still passes. Add a mock GET / returning {"web_url":"https://example.com",...}, capture stdout, assert the View: line uses that base; add a case with no web_url (or failed GET /) asserting no View: and no gitlawb.com. Mirror should get the same treatment or share a small helper under test.

- [P2] Reject empty or whitespace-only GITLAWB_WEB_URL at boot
  `crates/gitlawb-node/src/config.rs:574`
  Clap maps GITLAWB_WEB_URL="" to Some("") not None, and the node advertises it whenever is_some(). Whitespace-only values produce a broken View: line. Add a Config::validate() check (trim, reject if empty) or normalize to None.

- [P3] Document GITLAWB_WEB_URL beside GITLAWB_PUBLIC_URL
  `.env.example:11`
  The struct field and clap help exist, but .env.example only documents GITLAWB_PUBLIC_URL. Operators need a one-line distinction: API reachability vs browser View base.

Not an ask, recorded only: gl peer add reading public_url from GET / is a pre-existing contract gap, not introduced here. Fleet deploy templates setting only GITLAWB_PUBLIC_URL means View links stay hidden until someone sets GITLAWB_WEB_URL separately; that is a deployment follow-up, not a blocker on this PR.

beardthelion (dismissed):
Re-read head 195a0179. CI on this head is green except cargo audit (fleet advisory). Your three asks from my first round are on head: six load-bearing `fetch_node_web_url` unit tests, `Config::validate` rejects empty/whitespace, and `.env.example` documents `GITLAWB_WEB_URL`. `test (stable)` passed on 195a0179. A cross-model refute pass on this head found one gap the first round missed.

One process note, not a finding: open PRs #324, #325, and #330 also touch `config.rs`; expect a rebase conflict when those land.

## Findings

- [P2] Sanitize and cap the node-advertised web_url before printing it
  `crates/gl/src/repo.rs:231`
  `fetch_node_web_url` parses `GET /` with unbounded `.json().await` and returns `web_url` for `println` without stripping control characters. `peer add` already uses `read_body_capped` plus `sanitize_node_msg` before terminal output for untrusted remote bytes (`peer.rs` around 198 and 160). A hostile node the operator pointed `gl` at can embed ANSI or newlines in `web_url` and they render on the `View:` line. Cap the `GET /` body before parse and run the same `sanitize_node_msg` pass before return or print. Add a unit test with control bytes in the mock `web_url` asserting they do not reach the returned string.

beardthelion (dismissed):
Re-read head 57276ca. CI is green except cargo audit (fleet advisory). The #370 core is in place: nodes advertise optional web_url, the CLI prints View: only when present, and validate_web_url rejects empty, scheme-less, and non-http(s) values at boot. 57276ca also adds transport-error warnings on GET / and rejects query/fragment web_urls on both sides. I ran a rustc probe against validate_web_url: "https://git lawb.com" fails parse, but "https://example.com/\x1b[31m" still passes with control bytes in the stored string.

Open PRs #285, #324, #325, #173, and #330 also touch config.rs; expect a rebase conflict when those land, not a design rework.

## Findings

- [P2] Cap and sanitize web_url before terminal output
  `crates/gl/src/repo.rs:249`
  fetch_node_web_url still calls info_resp.json().await on the full GET / body and returns the raw JSON string for println without sanitize_node_msg. peer.rs already treats caller-chosen node replies as the least trusted bytes in gl: read_body_capped(resp, 8 * 1024) then defang before the terminal (peer.rs:198-202, http.rs:209-236). A hostile node can pass http(s) url::Url parse with ANSI, bell, or bidi controls in the path; those bytes reach View: in both cmd_create and mirror.rs. Mirror the peer pattern here: cap the body, parse from the cap, run sanitize_node_msg on the accepted base before return/print, and add a unit test with control bytes in the mock web_url asserting they do not reach the returned string. The malformed-web_url stderr path should sanitize trimmed the same way.

Not an ask, recorded only: test_cmd_create_success still does not mock GET / or assert stdout; the helper unit tests cover #370 at the function level.

beardthelion (dismissed):
Re-read head 6c40812. CI is green. The cap-and-sanitize round from my last review is on head: `fetch_node_web_url` caps GET / at 8 KiB (`read_body_capped`), rejects C0 controls and bidi-format chars, and `cargo test -p gl fetch_node_web_url` is 11/11 green including `test_fetch_node_web_url_rejects_control_bytes`. Node boot validation, `.env.example`, and the shared helper wiring in `mirror.rs` all look right for #370.

One gap remains in the same terminal-safety class.

## Findings

- [P2] Reject U+2028 and U+2029 in advertised web_url before printing View:
  `crates/gl/src/repo.rs:274`
  `fetch_node_web_url` gates on `is_control()` and `is_bidi_format()`, but U+2028 LINE SEPARATOR and U+2029 PARAGRAPH SEPARATOR are neither (`rustc`: both `is_control=false`, not in `is_bidi_format`). I mocked GET / with `web_url` containing U+2028; the helper returned `Some("https://example.com/\u{2028}evil")`, which would reach `println!("  View: ...")` in both `cmd_create` and `mirror.rs`. Those code points break the terminal line the same way a newline would. Extend the rejection predicate to cover them and add a unit-test case; optionally mirror the same rule in `validate_web_url` at boot so the node does not advertise them on GET /.

One process note, not a finding: open PRs #285, #324, and #325 also touch `crates/gitlawb-node/src/config.rs`. Expect a rebase conflict when those land.

Not an ask, recorded only: I am declining CodeRabbit's warn-on-invalid-JSON thread again. A 200 with unparseable JSON is the same fail-soft contract as a missing `web_url` field for this optional hint. Node boot `validate_web_url` does not yet reject control/bidi bytes the client already blocks; defense in depth only, not blocking here. The `read_body_capped` call is present but not proven load-bearing by an oversized-body case in the helper tests; the implementation matches the `peer.rs` precedent.

beardthelion (dismissed):
Re-read head a6113fd8 after the main merge. CI is red workspace-wide: `crates/gitlawb-node/src/config.rs` is missing a closing `}` after `web_url_rejects_non_absolute_or_non_browser_urls` (~line 1529), so the next test nests inside it and the crate does not compile. I counted braces in `mod tests`: 47 open, 46 close. Prior head 6c40812 was green on the node `web_url` tests; this is a merge regression, not a logic revert.

When the tree compiles, the #370 core from my last round is still there: optional `GITLAWB_WEB_URL` with boot validation, `GET /` advertises it, `fetch_node_web_url` caps the body at 8 KiB, warns on transport/status/malformed advertisement, and rejects control/bidi bytes before any `View:` line. Query/fragment rejection is symmetric between node `validate_web_url` and the CLI helper.

## Findings

- [P1] Close the missing brace in the config test module
`crates/gitlawb-node/src/config.rs:1529`
Add `}` after the `port+path web_url must validate` assert so `web_url_rejects_non_absolute_or_non_browser_urls` closes before `enforce_owner_push_is_declared_true_independent_of_the_environment`. Without it `cargo check -p gitlawb-node` and the full workspace CI matrix fail. This landed in the a6113fd8 merge, not in the feature commits.

- [P2] Pin the #370 premise with a cmd_create integration test
`crates/gl/src/repo.rs:904`
`test_cmd_create_success` mocks only `POST /api/v1/repos` and never `GET /`. If `cmd_create` went back to hardcoding `https://gitlawb.com` for `View:`, that test would still pass. The ten `fetch_node_web_url_*` tests cover the helper, not the command path. Add a test that mocks both routes: with `web_url` advertised, assert stdout contains `View: https://.../{owner}/{name}`; without it, assert no `View:` line. That is the load-bearing check for issue #370.

Not an ask, recorded only: CodeRabbit's two unresolved threads (warn on invalid JSON / non-string `web_url`; reject bidi in node `validate_web_url`) are reasonable polish. Invalid JSON on a 200 `GET /` is already fail-soft with no warning by design; bidi in operator config is trusted and the CLI already rejects bidi on the advertised value before print. Neither blocks merge once CI is green.

One process note, not a finding: rebase may conflict with #285/#324/#325 on `config.rs`; resolve against whichever of those merges first.

jatmn (dismissed):

## Findings

- [P2] Normalize the configured URL before advertising it
  `crates/gitlawb-node/src/config.rs:782`

  `validate_web_url` intentionally accepts surrounding whitespace by parsing `raw.trim()`, but it does not normalize `Config.web_url`. Both `node_info` and `degraded_node_info` then serialize the original string. As a result, `GITLAWB_WEB_URL=" https://git.example "` starts successfully even though `GET /` advertises a value with leading/trailing whitespace rather than the valid absolute URL that was actually checked. The current `gl` helper hides this inconsistency by trimming a second time, but any other node-info consumer that parses `web_url` directly can reject it or generate a broken browser link.

  Please fix the root cause at the configuration boundary: define one canonical representation for `web_url`, store that normalized representation after validation, and use it for both normal and degraded node-info responses. Alternatively, reject any input that is not already canonical. Keep the existing support for valid HTTP(S) bases with paths and ports, and retain the CLI’s defensive validation for remote nodes.

- [P2] Reject credentials in advertised web URLs
  `crates/gitlawb-node/src/config.rs:790`

  The new validation accepts URL userinfo because it only checks scheme, query, and fragment. For example, `https://user:secret@example.com` is accepted as `GITLAWB_WEB_URL`; the node publishes that exact value from public `GET /`, and `fetch_node_web_url` returns the original string for both `gl repo create` and `gl mirror` to print in their `View:` lines. That exposes an accidentally configured password or token in public metadata, terminal scrollback, CI logs, and any captured command output.

  Please address this at the root cause by treating `web_url` as a credential-free browser origin/path, not an arbitrary HTTP URL. Reject URLs whose parsed representation contains a username or password during node startup validation, and enforce the same constraint when the CLI consumes an advertisement from a remote node. Add regression coverage for user-only and user-plus-password forms, and preserve support for ordinary host, port, and path base URLs.

beardthelion (dismissed):
Re-read head 4c485112. CI is 12/12 green. The merge-regression items from my last round are fixed: the config test module compiles, and `cmd_create_view_url` integration test pins the #370 premise (hardcoding `gitlawb.com` fails it RED). The core design still looks right: optional `GITLAWB_WEB_URL`, advertised on `GET /`, CLI prints `View:` only when present.

Two gaps remain at the configuration boundary before this can merge.

## Findings

- [P2] Store and advertise a canonical web_url
  `crates/gitlawb-node/src/config.rs:782`
  `validate_web_url` validates `raw.trim()` but leaves `Config.web_url` as the operator typed it. `node_info` and `degraded_node_info` then serialize that raw string on public `GET /`. I set `web_url` to `"  https://git.example  "` in a config test and validation passes while the advertised value still carries padding. The CLI trims again in `fetch_node_web_url`, which hides the bug on the View line, but any other consumer that parses `web_url` directly gets a non-canonical base. Normalize to one stored form after validation (or reject non-canonical input outright) and use that everywhere you advertise it.

- [P2] Reject userinfo in web_url values
  `crates/gitlawb-node/src/config.rs:790`
  Validation checks scheme, query, and fragment only. `https://user:secret@example.com` passes boot, is published verbatim on `GET /`, and `fetch_node_web_url` returns the original string for `View:` output. That puts an accidentally configured password into public metadata, terminal scrollback, and CI logs. Treat `web_url` as a credential-free browser base: reject URLs with a username or password at node startup and in the CLI when consuming a remote advertisement. Add regression tests for user-only and user-plus-password forms.

Not an ask, recorded only: `cmd_create_view_url.rs` should call `assert_async()` on its mockito mocks when you touch tests again; the house pattern elsewhere in `gl` does.

beardthelion (approved):
Re-read head 10da6f6b. CI is 12/12 green on this head. I ran `cargo test -p gl fetch_node_web_url` (12 pass), `cargo test -p gl --test cmd_create_view_url` (1 pass), and `cargo test -p gitlawb-node web_url` (3 pass). Premise mutation on `cmd_create` (hardcoded `gitlawb.com` instead of `fetch_node_web_url`) turned `cmd_create_view_url_tracks_node_advertisement` RED; restoring the fetch path turned it green again. Vacuity on the `validate_web_url` userinfo guard turned `web_url_rejects_non_absolute_or_non_browser_urls` RED.

jatmn's two P2s from `4c485112` are on head: `Config::validate` now rewrites `web_url` to the canonical `url::Url::to_string()` form, and both node boot validation and `fetch_node_web_url` reject userinfo. Normal and degraded `GET /` both emit the same post-validate string. `web_url` is browser metadata, not server egress, so skipping `is_public_http_url` is correct for self-hosted front-ends.

Open PRs #285, #325, and #324 will move `config.rs` before this lands; expect a rebase, not a design redo.

Not an ask, recorded only: node-side `validate_web_url` could mirror the CLI's control/bidi rejection so misconfigured `GITLAWB_WEB_URL` never publishes terminal-dangerous bytes on public `GET /` for non-CLI consumers. The `View:` path is already protected client-side.

Not an ask, recorded only: a 200 `GET /` with invalid JSON still omits `View:` without a stderr warning (transport and HTTP failures do warn). Cosmetic degrade, not a regression from #370.

jatmn (approved):
LGTM

beardthelion:
@kevincodex1 LGTM

PR #381, opened 2026-08-25
fix(gl): read whoami error bodies through CappedBody.text

beardthelion:
## Summary

Fix the `gl` compile break on `main` by reading `.text` from `read_body_capped`'s `CappedBody` in `whoami`, and pin `enforce_owner_push` off in three receive-pack lock tests that use non-owner proof DIDs.

## Motivation & context

`main` is red (#380). `#173` changed `read_body_capped` to return `CappedBody`; `peer.rs`, `sync.rs`, and `ipfs_cmd.rs` were updated, but `#223` merged whoami error handling that still treated the return as a bare `String`.

Once compile is fixed, CI also exposed three `#173` lock-path tests that inherit the new default-on owner-push gate and fail before receive-pack runs.

Closes #380

## Kind of change

- [x] Bug fix
- [ ] Feature (N/A)
- [ ] Security fix (N/A)
- [ ] Docs (N/A)
- [x] Tests / CI
- [ ] Refactor (no behavior change)
- [ ] Breaking or protocol change (issue required first)

## What changed

- `crates/gl/src/whoami.rs`: use `read_body_capped(...).await.text` on the non-404 error path, matching `peer.rs` and `sync.rs`.
- `crates/gitlawb-node/src/api/repos.rs`: set `enforce_owner_push = false` in the three receive-pack lock tests (disconnect hold, success release, lock-pool exhaustion).

## How a reviewer can verify

```sh
cargo check --workspace --all-targets
cargo test -p gl whoami
cargo test -p gitlawb-node --bin gitlawb-node receive_pack_success_reclaims
cargo test -p gitlawb-node --bin gitlawb-node receive_pack_disconnect_holds
cargo test -p gitlawb-node --bin gitlawb-node receive_pack_lock_pool
```

## Before you request review

- [x] Scope is one logical change; no unrelated churn
- [x] `cargo test --workspace` passes locally
- [x] New behavior is covered by tests (required for fixes)
- [x] `cargo fmt --all` and `cargo clippy --workspace --all-targets -- -D warnings` are clean
- [x] Commit titles use Conventional Commits (`feat(...)`, `fix(...)`, `docs(...)`)
- [x] Docs / `.env.example` updated if behavior or config changed (or N/A)
- [x] Checked existing PRs so this isn't a duplicate

## Protocol & signing impact

N/A (client error-display path only; no wire or signing change).

- [ ] Touches DID / `did:key`, Ed25519 / RFC 9421 signatures, UCAN, ref certs, or P2P wire formats
[...]

PR #244, opened 2026-07-23
fix(node): implement reconciliation sweep as durability backstop (#218)

Gravirei:
## Summary

Implements the periodic reconciliation sweep the replication path already assumes as a durability backstop. Previously, every path that drops a pin or recovery copy (mid-drain panic, node crash/seal, client disconnect at the receive-pack tail) resulted in data loss with no safety net.

## Motivation & context

Closes #218

The codebase justified tolerating dropped post-push replication work by pointing at a reconciliation sweep that did not exist. This made "lost forever" literal rather than conservative phrasing, violating the project's stated promise that "once code is pushed to the network, it should not disappear because one server went down."

## Kind of change

- [ ] Bug fix
- [x] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI
- [ ] Refactor (no behavior change)
- [ ] Breaking or protocol change (issue required first)

## What changed

- **crates/gitlawb-node/src/reconciliation.rs** (new): Periodic sweep that re-derives the set of objects a repo should have pinned/sealed under current visibility rules
  - Runs hourly, capped at 100 repos per pass with a cursor to prevent O(repos) amplification
  - For each announceable repo: full-scans all git objects, applies fail-closed visibility filtering, pins missing objects to IPFS and Pinata, re-seals encrypted recovery copies, and anchors sealed manifests to Arweave
  - Reuses existing fresh-resolution pipeline (list_all_objects, replicable_blob_set, replicable_objects_fail_closed, ipfs_pin::pin_new_objects, pinata::pin_new_objects, encrypted_pin::encrypt_and_pin)
  - Skips non-announceable repos (private / mode A / undetermined)
  - Respects graceful shutdown signal
  - The mid-scan visibility re-filter uses its own fresh deadline, so a repo whose budget was nearly spent still completes its re-filter instead of dying on a stale, already-expired deadline

- **crates/gitlawb-node/src/metrics.rs**: Added gitlawb_reconciliation_gaps_found_total and gitlawb_reconciliation_gaps_filled_total counters

- **crates/gitlawb-node/src/main.rs**: Registered reconciliation module and spawned the background sweep task

[...]

Issue #136, opened 2026-06-30
list_pins / list_anchors serve stale metadata for repos made private after push (no index reconciliation on visibility downgrade)


beardthelion:
## Summary

`GET /api/v1/ipfs/pins` and `GET /api/v1/arweave/anchors` serve metadata for repos that were public when pushed but later made private. The visibility gate on the write side is evaluated once at push time and is never reconciled when a repo's visibility is tightened, so the index rows persist and the read endpoints return them with no current-visibility filter.

This is the real mechanism behind #121 (and what PR #134 only partially addresses): the leak is not that private repos are indexed (they are not, the write path is correctly gated), it is that visibility is mutable after indexing and nothing purges or re-filters the index.

## Mechanism (verified)

Write side is gated at push, point-in-time:
- `pinned_cids` is written only via `pin_new_objects` (`crates/gitlawb-node/src/ipfs_pin.rs:133`, `crates/gitlawb-node/src/pinata.rs:114`), called only inside the `withheld.is_some()` blocks of the push handler (`crates/gitlawb-node/src/api/repos.rs:1055`, `:1167`).
- `arweave_anchors` is written only via `record_arweave_anchor` (`crates/gitlawb-node/src/api/repos.rs:1249`), inside `if announce && !irys_url.is_empty()` (`repos.rs:1231`).
- `announce`/`withheld` come from `replication_withheld_set` = `listable_at_root(rules, is_public, owner_did, None)` (`repos.rs:49`), the anonymous root-read decision evaluated during the push async task.

Nothing reconciles the index on visibility change:
- There is no `DELETE FROM pinned_cids` or `DELETE FROM arweave_anchors` anywhere in the tree.
- `set_visibility` (`crates/gitlawb-node/src/api/visibility.rs:82`) adds/updates rules and touches neither table.
- The read queries are unfiltered: `list_pinned_cids` is `SELECT ... FROM pinned_cids ORDER BY pinned_at DESC` (`db/mod.rs:2038`), and the global `list_arweave_anchors(None,..)` is a plain `SELECT ... FROM arweave_anchors` with no visibility predicate (`db/mod.rs:2344`).

## Repro sequence

1. Create a public repo (`is_public=true`, no rules).
2. Push to it: `announce=true`, so `arweave_anchors` and `pinned_cids` rows are written (with IPFS/Irys configured).
[...]

PR #382, opened 2026-08-26
fix(node): drop withheld-subtree trees from the replication pin set

### LINKED ISSUE #172: Replication/pin path exports withheld-subtree tree objects to public IPFS (write-side of #135)
## Summary

The replication/pin path passes every non-blob object through unconditionally, so a withheld subtree's **tree object** is pinned to IPFS on push (Kubo or Pinata). Once pinned, the CID is announced to the public IPFS DHT and served by bitswap to any peer, disclosing the withheld subtree's child filenames and child oids off-node — bypassing the `get_by_cid` gate #135 adds. This is the write-side counterpart of #135 (which covers only the on-node `GET /ipfs/{cid}` read).

## Where

- `crates/gitlawb-node/src/git/visibility_pack.rs:353` — `replicable_objects_fail_closed` filters `!all_blob_oids.contains(oid) || allowed_blobs.contains(oid)`; trees are never in `all_blob_oids`, so they always pass. The doc comment states the now-wrong assumption: "commits and trees are structural, never content-withheld."
- Candidate set includes trees: `crates/gitlawb-node/src/git/push_delta.rs:178` `list_all_objects` (`git cat-file --batch-all-objects`); the delta path uses `rev-list --objects`.
- Both pin backends consume the filtered set: `crates/gitlawb-node/src/ipfs_pin.rs` (Kubo), `crates/gitlawb-node/src/pinata.rs` (Pinata).
- The pin CID is content-sha256 (`crates/gitlawb-core/src/cid.rs:32`), independent of repo object format.

## Why it matters

This is live independent of the reachability caveat that makes `get_by_cid` dormant: repos init `--object-format=sha1` (`crates/gitlawb-node/src/git/store.rs:15`), so `get_by_cid` cannot resolve a sha256 CID today — but the pin path pins by content-sha256 regardless, so the withheld tree bytes are fetchable by CID on the public IPFS network. Exposure is off-node and persists until GC/unpin; via Pinata it reaches public gateways. The confidentiality boundary is the same one `get_tree` and #135 protect (subtree child names + oids).

Confirmed: a withheld subtree's tree oid survives `replicable_objects_fail_closed` into the replicable/pinned set, while blob withholding is unaffected.

## Scope

Write-side of the same KTD3 divergence as #135. The fix reuses #135's tree-reachability machinery: gate the replicable set on an anonymous `replicable_tree_set` (the `caller = None` allowed-tree-set), dropping withheld-subtree tree oids from replication exactly as blobs are handled via `replicable_blob_set`. Best done after #135 lands `tree_paths`.

Live only when a pin backend (`GITLAWB_IPFS_API` or `GITLAWB_PINATA_JWT`) is configured; both default empty.

beardthelion:
## Summary

A tree that exists only under a denied path no longer gets pinned to public IPFS or Pinata. The replication filters treated trees as structural, so a public CID still named child files and blob oids.

This is the write-side of the same withhold #173 already applies on GET /ipfs/{cid}.

## Motivation & context

Closes #172.

## Kind of change

- [x] Bug fix
- [ ] Feature
- [x] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

- `gitlawb-node`: `replication_withheld_set` now walks blobs and withheld-subtree trees together. All three delta pin sites already take that set, including a first push, which runs `rev-list --objects` without `--not`.
- Full-scan fail-closed now takes a tree allow-list and a typed blob/tree universe, under the same scan deadline as the blob walk.
- Pinning test `fail_closed_keeps_nonblobs_and_allowed_blobs_only` is flipped on purpose. Its premise (trees are never content-withheld) was falsified by #135/#173. Kubo and Pinata still consume the already-filtered list; there is no second filter.

## How a reviewer can verify

```sh
cargo test -p gitlawb-node --bin gitlawb-node --locked -- \
  fail_closed_keeps_commits_allowed_blobs_and_allowed_trees_only \
  withheld_object_set_contains_withheld_subtree_tree \
  first_push_delta_drops_withheld_subtree_tree \
  full_scan_drops_withheld_and_dangling_trees \
  withheld_object_walk_fails_closed_on_non_commit_ref \
  full_scan_shares_one_deadline_with_tree_allowed_set_walk
```

The first-push and full-scan tests assert the secret tree is in the candidate universe before they assert it is dropped. Reverting the tree clauses, swapping the wrapper back to blob-only, dropping the non-commit-ref guard, or handing the later full-scan phases a fresh timeout each goes red.

`cargo test --workspace` is red on `gl` from #330 / #381 (`CappedBody.text`). This PR does not touch `gl`. The crate gate above was 1093 passed; the three `receive_pack_*` failures are the same owner-push default from #330, not this diff.

## Before you request review

- [ ] `cargo test --workspace` passes locally

## Protocol & signing impact

Does not touch DID / did:key, signatures, UCAN, ref certs, or P2P wire formats.

## Notes for reviewers

Already-pinned withheld trees stay on public IPFS/Pinata until unpin or GC. This PR stops new pins only. Distinct from #136 and from #218 / #244.

Live only when `GITLAWB_IPFS_API` or `GITLAWB_PINATA_JWT` is set; both default empty.

Does not change git upload-pack. A path-scoped clone still receives withheld-subtree trees.

Parent allowed trees still name withheld children. That is the SHA-intact tradeoff, same as `get_tree`.

`replicable_objects` still trusts the withheld set it is handed. A future blob-only set would reopen this class. Production delta sites all go through `replication_withheld_set`.

#244 touches the same files and will need a rebase onto this.

Related: #173

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved replication and pinning to exclude withheld, dangling, or unreachable blobs and trees.
  * Added fail-closed handling for invalid references, incomplete scans, traversal errors, timeouts, and malformed paths.
  * Improved reliability when processing large repositories and nested tags.
  * Ensured full scans apply a shared deadline across all processing stages.
* **Documentation**
  * Clarified that withheld subtree trees, as well as blobs, are excluded from replication and pinning.

### DIFF
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -21,24 +21,24 @@ use crate::webhooks;
 /// The git all-zeros object id — the create/delete sentinel in a ref update.
 const ZERO_SHA: &str = "0000000000000000000000000000000000000000";
 
-/// The set of blob OIDs withheld from **anonymous** replication for a repo, or
-/// `None` when the repo must not replicate at all (private / mode A /
-/// undetermined — fail closed). This is the anonymous replication gate:
-/// `caller` is hard-coded to `None` and there is intentionally no caller
-/// parameter, which distinguishes it from the per-caller read-serve projection
-/// in `git_upload_pack` (which passes the real caller). Both the push pin path
-/// and the reconciliation sweep call this helper so the two cannot drift on
-/// what is withheld. `rules` is the already-fetched visibility-rule snapshot
+/// The set of blob and withheld-subtree tree OIDs withheld from **anonymous**
+/// replication for a repo, or `None` when the repo must not replicate at all
+/// (private / mode A / undetermined, fail closed). This is the anonymous
+/// replication gate: `caller` is hard-coded to `None` and there is intentionally
+/// no caller parameter, which distinguishes it from the per-caller read-serve
+/// projection in `git_upload_pack` (which passes the real caller). Both the push
+/// pin path and the reconciliation sweep call this helper so the two cannot drift
+/// on what is withheld. `rules` is the already-fetched visibility-rule snapshot
 /// (callers fetch once and may reuse it, e.g. for encrypt-then-pin).
 ///
 /// Returns `(announce, withheld)`: `announce` is whether the repo may be
 /// announced/replicated to the anonymous public at all (also gates gossip and
-/// Arweave anchoring downstream), and `withheld` is the anonymous withheld blob
-/// set when announceable (`None` when not announceable). A failed/panicked
-/// withheld walk fails closed on both axes: `announce` is forced false and
-/// `withheld` is `None`, so an unvetted push neither replicates blobs nor
-/// announces. Returning both keeps the gate's announce decision a single
-/// source rather than recomputing it at each call site.
+/// Arweave anchoring downstream), and `withheld` is the anonymous withheld object
+/// set (blobs and withheld-subtree trees) when announceable (`None` when not
+/// announceable). A failed/panicked withheld walk fails closed on both axes:
+/// `announce` is forced false and `withheld` is `None`, so an unvetted push
+/// neither replicates objects nor announces. Returning both keeps the gate's
+/// announce decision a single source rather than recomputing it at each call site.
 ///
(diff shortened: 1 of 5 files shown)


jatmn (approved):
LGTM

beardthelion:
@kevincodex1 LGTM

PR #224, opened 2026-07-20
fix(node): harden Arweave anchoring and add verification (#26)

Gravirei:
## Summary

Anchor ref updates and encrypted-blob manifests to Arweave as **signed ANS-104 data items**, and add `/api/v1/arweave/verify/{tx_id}` which fetches the anchor from the gateway and validates the embedded certificate chain (node Ed25519 signature, SHA-256 `prev` linkage, pusher RFC 9421 proof). Also harden the Arweave surface: spec-correct ANS-104 preimages, tamper rejection, fail-closed corroboration, credential redaction on every public output, structural URL handling that rejects fragments, an explicit verified bundler-funding configuration, and durable post-receive jobs persisted before the push is acknowledged.

Durability scope is deliberate and narrow: **push accounting, per-ref certificates, and the Arweave anchor are the durable unit** — a crash between the pack landing and that bookkeeping is recovered by the startup drain. The replication tail (Pinata/IPFS pinning, gossip, GraphQL push, peer notify) remains best-effort and is not part of the durable unit.

Closes #26.

## Kind of change

- [x] Feature
- [x] Security fix
- [x] Tests / CI

## What changed

**gitlawb-node**

- **ans104.rs** — build and verify ANS-104 signed data items (ed25519, recursive deepHash). The **tags preimage is the FLAT serialized tag stream** (`item.rawTags` in the published `arbundles` `getSignatureData`), not a nested `[[name, value], ...]` list — the nested form is what Arweave layer-one transactions use, and a data item's signature would not reproduce with it. Zero tags is an empty blob, not `deepHash([])`. The empty-tags reference vector and a 3-tag interop fixture are produced by the independent `arbundles` package (`createData` + `sign`) and pinned as hex in tests; the node's own signer produces items this verifier accepts.
[...]

PR #383, opened 2026-08-28
docs: state that GITLAWB_ENFORCE_OWNER_PUSH defaults to true

Gravirei:
## Why

Reviewer 2 closed #224 on 2026-08-28 with a directive: split the work into four narrow PRs. This is **Split PR 4** (configuration + operator policy) and addresses the P2 finding the reviewer assigned to it.

Two operator-facing docs contradicted the code that PR #330 landed. `Config::enforce_owner_push` declares `default_value_t = true`, `.env.example` sets `=true`, the README env-var table says "Defaults to `true`", and the runtime behavior matches. But `SECURITY.md` line 57 (the "Repository write authorization defaults" section) and `README.md` line 69 (the "Known limitations" prose) both still said the default was `false`. The runtime, .env.example, env-var table, and security prose were lying to operators in opposite directions.

## What this PR changes

- `SECURITY.md` — rewrite the owner-push limitation to lead with "defaults to `true`" and describe `GITLAWB_ENFORCE_OWNER_PUSH=false` as the explicit rolling-upgrade compatibility override.
- `README.md` — same correction in the Known-limitations bullet.
- `crates/gitlawb-node/tests/owner_push_default_docs.rs` — new regression test that reads the four surfaces at test time and asserts each independently says the default is `true`. Reverting either doc edit or flipping `default_value_t` back to `false` turns a named test red.

## Boundaries covered

- **Producer:** the clap `default_value_t = true` declaration in `Config` and the `.env.example` shipping default.
- **Persistence:** none — no state.
- **Restart/restore:** none — no restart concern.
- **Consumer/effect:** operator reads SECURITY.md and README to learn the default.
- **Cleanup/failure:** if a future doc edit reintroduces the wrong wording, the named test goes red on the named surface.

Reviewer 2: *"Use the tested configuration default as the source of truth. State that owner enforcement is enabled by default and describe `GITLAWB_ENFORCE_OWNER_PUSH=false` only as the explicit rolling-upgrade compatibility override. Consolidate the duplicate limitation sections."* The reviewer called this out as the P2 finding for the policy split, not the post-receive or transport split. It does not need the outbox work, the ANS-104 work, or any of the cert/CLI changes to land.

## Overlap with open PRs

- **#134** (anchors auth) — excludes. This PR does not touch any auth gate.
- **#285** (advisory-lock session affinity) — excludes. No lock or session work.
- **#306** (Content-Digest on signed requests) — excludes. No signature work.
- **#314** (small-order Ed25519) — excludes. No key validation.
- **#324** (libp2p keypair persistence) — excludes. No p2p identity.
- **#325** (gossip ref-update auth) — excludes. No gossip or wire format.
- **#382** (replication withheld-subtree trees) — excludes. No replication.

This PR is safe to land independently. It does not require any sibling PR. No new migration. No new dependency. No new env var. No code path change.

## Verification

```sh
cargo test -p gitlawb-node --test owner_push_default_docs
cargo fmt --all -- --check
cargo clippy -p gitlawb-node --tests -- -D warnings
```

The regression test was load-bearing-checked before committing: stashing the SECURITY.md edit made `security_md_states_owner_push_default_is_true` go red on the exact pre-fix prose; stashing the README edit made `readme_known_limitations_states_owner_push_default_is_true` go red on the exact pre-fix prose. The other two tests (`.env.example` row and the existing env-var table) stayed green on each revert, confirming each surface is checked independently.

## Summary by CodeRabbit

* **Documentation**
  * Updated security guidance to clarify that repository writes require owner authorization by default.
  * Documented that authentication alone does not grant push access, including for delegated or CI keys.
  * Clarified that the compatibility override temporarily disabling owner enforcement should only be used during upgrades.
* **Tests**
  * Added checks to ensure the owner-enforcement default is consistently documented and configured across repository guidance and environment examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Good today:
 
 Known limitations:
 
-- Repository write authorization is not secure by default: `GITLAWB_ENFORCE_OWNER_PUSH` defaults to `false` for compatibility, so a valid HTTP Signature identifies a pusher but does not enforce owner-only pushes.
+- Repository write authorization is secure by default: `GITLAWB_ENFORCE_OWNER_PUSH` defaults to `true`, so a valid HTTP Signature identifies a pusher and the node additionally requires that pusher to be the repository owner. Set `GITLAWB_ENFORCE_OWNER_PUSH=false` only for the duration of a rolling upgrade whose pushers are not yet the repo owner. Authentication is not authorization on its own — anyone can mint a `did:key` and sign — so owner enforcement is the only thing that prevents every signed caller from pushing to every repository, private ones included.
 - UCAN proof chains are validated when supplied, but UCAN capabilities are not consulted by write authorization and the root issuer is not independently trust-anchored. UCANs therefore do not yet grant scoped collaborator access.
 - Agent lifecycle revocation is not enforced by HTTP Signature authorization; do not rely on removing or revoking an agent record to block a compromised signer.
 - Read visibility is not a blanket data-classification boundary: task, IPFS-pin, and Arweave-anchor listings are not repository-gated; withheld path names can be visible to a root reader; and later visibility changes cannot retract content already announced or externally anchored.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -54,9 +54,9 @@ We will acknowledge receipt within 48 hours and aim to release a fix within 14 d
 These are documented limitations of the current live release. They should be prioritized without breaking existing nodes during rolling upgrades.
 
 ### Repository write authorization defaults
-- `git-receive-pack` verifies HTTP Signatures, but `GITLAWB_ENFORCE_OWNER_PUSH` defaults to `false` for compatibility during rollout.
-- **Impact:** With the default setting, a valid signature authenticates the pusher but does not require that DID to be the repository owner.
-- **Mitigation:** Set `GITLAWB_ENFORCE_OWNER_PUSH=true` on nodes where owner-only pushes are required. Confirm that every legitimate pusher uses the owner DID before enabling it.
+- `git-receive-pack` verifies HTTP Signatures and, with the default configuration, also requires the pusher DID to be the repository owner: `GITLAWB_ENFORCE_OWNER_PUSH` defaults to `true`.
+- **Impact:** With the default setting, a valid signature authenticates the pusher *and* the pusher must be the repository owner. Authentication is not authorization on its own — anyone can mint a `did:key` and sign — so owner enforcement is the only thing that prevents every signed caller from pushing to every repository, including private ones.
+- **Mitigation:** `GITLAWB_ENFORCE_OWNER_PUSH=false` is the explicit rolling-upgrade compatibility override. Disable it only for the duration of an upgrade whose pushers are not yet the repo owner, then turn it back on. Delegated and CI keys count as non-owners: a UCAN `git/push` capability is verified but not yet honored for authorization, so they cannot push while this is on.
 
 ### UCAN delegation and revocation
 - The middleware validates a supplied UCAN's complete proof chain, but a root token is accepted without an independently trusted issuer anchor. `Ucan::can` is not yet used by write handlers, so a UCAN does not grant scoped repository access.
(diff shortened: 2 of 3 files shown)


beardthelion (dismissed):
The doc edits match the live default. I ran `cargo test -p gitlawb-node --test owner_push_default_docs` (4/4 green) and revert probes on SECURITY.md, the README limitations bullet, and `.env.example`: each named test went RED on the pre-fix wording. `enforce_owner_push_is_declared_true_independent_of_the_environment` in `config.rs` already pins the clap default independently of env.

## Findings

- [P2] Pin the clap default in the integration guard or narrow the module doc
  `crates/gitlawb-node/tests/owner_push_default_docs.rs:15`
  The module header lists four surfaces and says the integration layer re-asserts `Config::enforce_owner_push` so unit-test removal is caught, but none of the four tests read the parser default. The unit test covers it today (`cargo test -p gitlawb-node enforce_owner_push` passed both cases). Either scrape `Config::command()` for `default_value_t = true` here, or drop Config from the stated contract.

- [P2] Bound the Known limitations slice before the next section
  `crates/gitlawb-node/tests/owner_push_default_docs.rs:66`
  `section_after(&readme, "Known limitations:")` sets `header_level = 0`, so the stop condition never fires and the slice runs to EOF, including the env-var table row. Reverting the limitations bullet to the old `defaults to false` wording fails correctly, but removing the bullet fails only because `.find()` lands on the table row and the lowercase `defaults to \`true\`` assert mismatches `Defaults to \`true\``. Stop at `See:` (line 77) or the next `#` heading so this test guards the limitations prose, not the table row the sibling test already covers.

One process note, not a finding: mechanical README overlap with several open PRs; expect a rebase conflict, not a redone review.

jatmn (dismissed):

## Findings

- [P2] Keep the Known limitations assertion within its actual block
  `crates/gitlawb-node/tests/owner_push_default_docs.rs:66`
  This guard passes `Known limitations:` to `section_after`, but that helper derives
  its boundary from Markdown `#` characters. Because this marker is plain text,
  `header_level` is zero and `level > 0 && level <= header_level` can never hold;
  the returned slice extends from the marker through EOF. Its subsequent `.find()`
  is therefore not scoped to the limitations bullet and can select the separate
  `GITLAWB_ENFORCE_OWNER_PUSH` environment-variable table row later in README.md.
  A future documentation edit that removes or relocates the security limitation can
  consequently leave this test green if that table uses the accepted lowercase
  wording. Address the root cause by using a selector whose boundaries match the
  README structure—e.g. take text from `Known limitations:` through `See:` (or parse
  the bullet block explicitly)—then select the env-var line within that bounded
  region. Keep the table test independent so it cannot satisfy this prose test.

- [P3] Make the Config default coverage match the guard's stated contract
  `crates/gitlawb-node/tests/owner_push_default_docs.rs:11`
  The module documentation and PR describe this target as a four-surface guard that
  independently reasserts `Config::enforce_owner_push`'s declared
  `default_value_t = true`, including protection against removal of the existing
  unit test. The implementation does not do that: its four tests read SECURITY.md,
  two README locations, and `.env.example`; none imports Config, calls
  `Config::command()`, or inspects Clap's declared default. The root cause is that
  the test's stated invariant was broader than its actual inputs, creating a second
  test target that can remain green after a default flip if the separate
  `config.rs` test is removed or weakened. Either add the advertised narrow
  declaration assertion (following the existing `Config::command()` test) or revise
  the module and PR wording to say this target covers only documentation and the
  example file. The existing `config.rs` unit test protects the runtime default
  today, so this is a coverage/maintenance mismatch rather than a present runtime
  regression.

beardthelion (approved):
The doc edits match the live default on head 96260a0e. I ran `cargo test -p gitlawb-node --test owner_push_default_docs` (4/4) and `cargo test -p gitlawb-node --bin gitlawb-node enforce_owner_push_is_declared_true` (1/1). Revert probes on SECURITY.md, the README limitations bullet, and `.env.example` each RED the named test; removing the limitations bullet while keeping the env-var table row still fails because the slice no longer mentions the env var. The `section_after` plain-text fix stops at `## Quickstart`, so the limitations guard cannot be satisfied by the table row alone. The module doc now honestly splits doc surfaces from the `config.rs` parser test.

One process note, not a finding: expect mechanical README rebase conflicts with several open PRs; that is overlap, not a reason to re-audit the diff.

Not an ask, recorded only: `docs/RUN-A-NODE.md` and `docs/OSS-READINESS-AUDIT.md` say "on by default" but are outside this guard's four-surface contract; wording is correct today.

jatmn (approved):
lgtm

beardthelion:
@kevincodex1 LGTM

PR #384, opened 2026-08-28
fix(node): durable post-receive outbox for receive-pack (#26 split 1/4)

Gravirei:
## Why

Reviewer 2 closed PR #224 on 2026-08-28 with a directive: split the work into four narrow PRs. This is **Split PR 1** (durable post-receive lifecycle).

The pre-outbox crash window the reviewer flagged: `smart_http::receive_pack` can apply a ref to disk and return Ok, and a process exit, a dropped future, or a DB failure before the bookkeeping at `crates/gitlawb-node/src/api/repos.rs:2361` (push event + cert + webhook) loses the recovery record. The startup drain enumerates only sources written from that bookkeeping, so it cannot reconstruct the missing work. The partial fallback that re-derives from a row present in the bookkeeping substitutes `did:key:recovered` and an empty attestation — not equivalent to the original authenticated push.

The fix is to persist the authentic intent **before** the receive-pack call lands the ref, then flip the row's state based on the outcome. The drain reads only `applied` rows, so a row that never reaches the post-Ok branch stays in `prepared` (handler crash / dropped future) or `cancelled` (receive-pack Err) and is never promoted.

## What this PR changes

- **Migration v27:** new `pending_ref_transitions` table (state machine: `prepared` → `applied`/`cancelled`) and new `anchor_jobs` table (per-transition upload queue for PR 2 to consume). Both with the unique indexes that make recovery re-derivation idempotent.
- **DB methods on `Db`:** `insert_pending_ref_transitions`, `mark_pending_ref_transitions_applied` / `_cancelled`, `list_pending_ref_transitions_applied`, `delete_pending_ref_transition`, plus the idempotent `record_push_with_id`, `insert_ref_certificate_idempotent`, and `insert_anchor_job_idempotent`. The deterministic id helpers `push_event_id_for`, `ref_cert_id_for`, `anchor_job_id_for`, and the underlying `deterministic_id` (SHA-256 with an ASCII Unit Separator so two distinct tuples can never collide on prefix overlap).
- **Handler refactor in `git_receive_pack`:** at the last possible moment before `smart_http::receive_pack`, the handler now generates a `request_id`, captures the raw `Signature` / `Signature-Input` / `Content-Digest` headers, and writes one `prepared` row per ref update. After the call: on Ok, `mark_applied`; on Err, `mark_cancelled`. A process crash between the post-Ok `mark_applied` and the bookkeeping is the exact window recovery closes.
[...]

Issue #26, opened 2026-06-04
Harden Arweave anchoring and add verification


kempsterrrr:
*Code references are against `main` @ `0b056f3`.*

## The problem

The architecture page presents Arweave (Tier 3) as the layer that lets anyone verify repo history without trusting a gitlawb node. Two things stand in the way today:

- **The anchored object is unsigned.** On each push the node already builds a `RefCertificate` (`crates/gitlawb-node/src/cert.rs`) that is signed with the node's Ed25519 key and carries the real `old_sha`, `new_sha`, `pusher_did`, and `node_did`. But that certificate is only stored locally. What actually goes to Arweave is a separate `RefAnchor` (`crates/gitlawb-node/src/arweave.rs`) — unsigned, with `old_sha` hardcoded to zeros, sent to Irys with `https://arweave.net/<tx_id>` as the read URL. The permanent record is a weaker, unsigned copy of facts the node has already signed.
- **Nothing verifies it.** `GET /api/v1/arweave/anchors` reads the local Postgres mirror; it never fetches from Arweave or checks a signature. So "verify without trusting a node" isn't possible against the current implementation.

A few related facts worth noting:

- Anchoring is off by default (`GITLAWB_IRYS_URL` defaults to `""`), runs fire-and-forget in a `tokio::spawn`, and fires on every push.
- The push *is* cryptographically verified at ingest (`auth::require_signature`: RFC 9421, Ed25519, content-digest checked), but the verified pusher signature is not persisted.
- Object IDs are SHA-1 (the node shells out to system `git`), while the anchor's zero placeholder is 64 hex chars — a width mismatch to clean up alongside the `old_sha` fix.

This is also a good moment to act: Irys has moved to its own L1 datachain and is winding down Arweave bundling, so the current write path is on a deprecating route — but since anchoring is off by default, nothing breaks in the meantime.

## What we'd like to verify, and what each requires

A verifier should be able to confirm history using only signed artifacts, a public immutable anchor, and hashes it recomputes — never by trusting a node's answer. That needs four things:

1. **Identity** — verify the Ed25519 signature against the signer's key (`did:key` is self-certifying; `did:web`/`did:gitlawb` add a resolution step).
2. **Authorized transition** — the real `old→new`, plus a pusher signature (not just a `pusher_did` field), plus maintainer countersignatures to the repo's threshold.
3. **Continuity** — a per-ref chain so transitions can't be inserted, deleted, or reordered without breaking a signature.
[...]

PR #385, opened 2026-08-28
fix(node): ANS-104 codec and recovery policy types (#26 split 2/4)

Gravirei:
## Why

Reviewer 2 closed PR #224 on 2026-08-28 with a directive: split the work into four narrow PRs. This is **Split PR 2**.

The P1 finding the reviewer assigned to this split: the legacy `anchor_item_present` mapped `BAD_REQUEST` and `GONE` to `Ok(false)`, and the recovery code interpreted false as permission to reclaim the row and pay for another upload. Neither 400 nor 410 proves a previously paid item was never accepted: 410 can describe an artifact that existed but is no longer served, and 400 can be produced by gateway, proxy, or routing failure. The concrete failure sequence the reviewer named: bundler accepts the item, the response or terminal DB write is lost, recovery probes the persisted item_id, the gateway returns 400 or 410, and the node pays for a second immutable item for the same transition.

Fix: model the probe result as three outcomes, and authorize a re-upload only on a trustworthy, protocol-defined absence.

Round-5 direction (reviewer 2): land here only the standard-compatible ANS-104 primitives and the pure three-outcome policy types. The gateway probe, the `verify_anchor` path, and the public verify endpoint are deferred until the uploader, retrieval proof, and recovery consumer can be proven as one vertical slice against a real provider contract (same-network upload/read pair, redirect-safe client, envelope-supplying read API, provider-documented absence evidence).

## What this PR changes

- **New module `ans104`:** ANS-104 data item (de)serialization, deep-hash matching `arbundles@0.10.x` `getSignatureData` (8-element fold, verified byte-exact against the JS implementation), Ed25519 sign/verify (strict), full two-byte signature-type validation, raw-tag-payload preservation across parse/hash/encode. 16 unit tests including a real arbundles-signed Ed25519 golden vector (verified without re-signing), multi-block and size-prefixed tag blocks, and a non-canonical-sigtype rejection test.
- **New module `arweave_v2` (policy types only):** the `ProbeOutcome` model the reviewer demanded — `Present` / `DefinitivelyAbsent` / `Indeterminate` — plus the `permits_reupload` rule, with 1 unit test. No gateway client, no endpoint.
- **Golden-vector scripts** (`scripts/ans104_golden_ed25519.mjs`, output doc) capturing the interop fixture from the pinned `arbundles` dependency.

Deferred to the vertical slice (explicitly NOT in this PR): the gateway probe (`probe_anchor_item`), the `verify_anchor` path, `GET /api/v1/arweave/anchors/verify/{item_id}`, the `irys_tx_id` lookup method and index migration, and the gateway/rate-limit config knobs. There is therefore no provider response table here: `DefinitivelyAbsent` is authorized only by whatever evidence the future provider contract defines, and every ambiguous response stays retryable.

## The recovery policy (the reviewer's named rule)

`ProbeOutcome::permits_reupload` is the policy the reviewer demanded: only `DefinitivelyAbsent` authorizes a paid re-upload. `Indeterminate` keeps the outbox non-terminal; the next probe (or a future retry) gets another chance to give a trustworthy answer. The exact evidence that establishes `DefinitivelyAbsent` is defined by the provider contract in the vertical slice that introduces the probe — never by a mock-only body shape.

## Required proof (named tests shipped here)

`crates/gitlawb-node/src/ans104.rs::tests`:

- `dataitem_matches_arbundles_golden_vector` — real arbundles-signed Ed25519 item: shape pin, id pin, deep-hash pin against `getSignatureData`, verify without re-signing against an independently pinned expected key, byte-exact binary round-trip.
- `signed_binary_parse_and_verify_multi_block_and_negative_count` — non-canonical tag encodings hash the original payload bytes and verify without re-signing.
- `decode_tags_accepts_negative_block_count_with_size` — legal size-prefixed Avro form parses.
- `from_binary_rejects_non_canonical_signature_type_high_byte` — wire `02 01` (258) is rejected, never aliased onto Ed25519 type 2.
- `binary_round_trip` — signature slot, canonical owner, and payload survive `to_binary`/`from_binary` with the signature intact (no re-sign after parse).
- `sign_then_verify_round_trips`, `flipped_signature_byte_fails_verify`, `wrong_expected_pubkey_fails_verify`, `mutated_data_fails_verify`, `wire_shape_round_trip`, `deep_hash_is_stable`, `deep_hash_empty_tags_is_distinct_from_one_tag`, `data_item_id_is_base64url_of_sha256_of_signature` — round-trip, tamper, and derivation pins.
- `external_reference_vectors` (3 tests) — SHA-384 deep-hash primitives against the reference suite.

`crates/gitlawb-node/src/arweave_v2.rs::tests`:

- `only_definitively_absent_authorizes_reupload` — the policy: collapsing any ambiguous outcome to absent spends a second immutable upload for the same transition.

Each test names the invariant it pins. Reverting the covered line turns the assertion red.

PR 2 owns ANS-104 serialization/verification and recovery outcome classification. The bundler upload itself (the handler that builds a DataItem and POSTs to the bundler), the provider-backed probe, and the public verification endpoint live in later slices. The ANS-104 module is fully tested and ready for the bundler upload to call into. PR 1's recovery drain consumes `ProbeOutcome::permits_reupload`; the probe that produces the outcome arrives with the provider contract.

## Overlap with open PRs (declared per the reviewer's instruction)

- **Split 1 (#384, durable post-receive)** — composes. It owns migration versions 27–33; this PR ships no migration.
- **Split 3 (cert compat)** — composes. It owns migration version 37; this PR ships no migration.
- **Split 4 (config policy)** — composes. No shared knobs; this PR adds no config.
- **#134** (anchors auth) — independent. The list endpoint is untouched.
- **#285** (advisory-lock session affinity) — independent. No DB locking here.
- **#306** (Content-Digest on signed requests) — independent. No request signing here.
- **#314** (small-order Ed25519) — independent. No key parsing changes here.
- **#324** (libp2p keypair persistence) — independent.
- **#325** (gossip ref-update auth) — independent.

## Safety to land standalone

- It compiles, runs, and passes its focused tests by itself. No sibling PR required.
- No migration shipped, and no released migration edited. The `arweave_anchors` table is untouched.
- No endpoint, route, config knob, or rate-limiter state added. The public API surface is unchanged.
- Purely additive: two new modules plus golden-vector scripts. Nothing existing is modified except `main.rs`/`server.rs` module wiring and a comment.

## Verification

```sh
cargo test -p gitlawb-node --bin gitlawb-node ans104
cargo test -p gitlawb-node --bin gitlawb-node arweave_v2
cargo fmt --all -- --check
```

Focused suites: **ans104 16/16, arweave_v2 1/1**. Full workspace suite runs on CI.


beardthelion (dismissed):
I read the diff on head `a53a63a`, ran `cargo test -p gitlawb-node --bin gitlawb-node` locally (1113 passed), and gut-checked the premise arm (neutering `return ProbeOutcome::Indeterminate` made `probe_400_is_indeterminate_not_absent` fail). CI on GitHub shows 12/13 green with test (stable) red on the same head; I could not read that job log, but the blocking issues below are in the diff regardless.

Prior art checked: enumerate-all-readers-of-a-data-class-when-gating.md (global-id readers invisible to route-shape guards), a-fixture-cannot-refute-the-model-that-produced-it.md (mockito/sign-round-trip does not prove bundler interop), distinguish-unknown-from-empty-and-fail-closed.md (three-outcome model is right; empty must not collapse to absent without evidence).

## Findings

- [P1] Look up anchors by `irys_tx_id`, not the UUID primary key

  `crates/gitlawb-node/src/db/mod.rs:3892`

  `record_arweave_anchor` assigns a random UUID to `id` and stores the gateway item id in `irys_tx_id` (lines 3814-3827). `get_arweave_anchor_by_item_id` filters `WHERE id = $1`. Production push passes only `irys_tx_id`. Endpoint tests seed `id = item_id`, which masks the mismatch. Real callers pass the Arweave/Irys tx id and always get 404 before verify runs.

- [P1] Gate the verify endpoint on repo read before returning payload

  `crates/gitlawb-node/src/api/arweave.rs:61`

  `verify_anchor` takes no caller identity and never calls `authorize_repo_read`. After DB lookup it returns `data_payload` with repo/ref/SHA fields to any holder of `item_id`. `arweave_routes` in `server.rs` has no `optional_signature` layer (unlike `get_cert`, which gates even on cert id). Map denials to the existing 404 shape and add anonymous plus unauthorized-authenticated denial tests through `build_router`. `list_anchors` on the same router is still ungated; #134 tracks that surface, but this PR introduces verify and must not ship it open.

- [P1] Implement the standard ANS-104 deep-hash before bundler interop

  `crates/gitlawb-node/src/ans104.rs:166`

  `deep_hash` chains SHA-256 over UTF-8 `target`/`anchor` strings. ANS-104 section 2 requires the Arweave 2.0 deep-hash (SHA-384 blob/list framing in the reference deepHash.ts). Items signed with this module will not verify on a real bundler/gateway. Add a reference vector with non-empty tags and optional fields from outside this repo; keep a versioned path if you must retain the current encoder for tests only.

- [P2] Carry `ProbeOutcome` in the verify result instead of parsing error text

  `crates/gitlawb-node/src/api/arweave.rs:92`

  HTTP status classification matches the substring `"never served"` inside `result.error`. Any wording change in `arweave_v2.rs:256` silently reclassifies definitive absence as `indeterminate`. Add `outcome: ProbeOutcome` (or equivalent) to `AnchorVerifyResult` and match on it here.

- [P2] Treat empty 404 bodies as indeterminate, not definitive absence

  `crates/gitlawb-node/src/arweave_v2.rs:128`

  The module header limits `DefinitivelyAbsent` to a protocol-defined 404 body, but `classify_404` returns `DefinitivelyAbsent` on an empty body, and `permits_reupload()` authorizes paid re-upload from that alone. Proxies and misconfigured gateways emit bodyless 404s; that is the same double-payment failure mode the three-outcome model exists to prevent. Keep `DefinitivelyAbsent` for the recognized JSON shape only and flip `probe_404_with_empty_body_is_definitively_absent` to expect `Indeterminate`.

One process note, not a finding: rebasing will conflict with several open PRs (#134, #384, #386, #285, and others); mechanical EXPECT-REBASE only.

Not an ask, recorded only: CodeRabbit nitpicks on redundant gateway GET on the indeterminate path and missing `StatusCode::OK` assert in one endpoint test are fair follow-ups once the above land.

jatmn (dismissed):

## Merge readiness

- [P1] Resolve the failing required `test (stable)` check before merge

  GitHub reports `test (stable)` failed on `a53a63a5`, while the other listed checks passed. The job log was unavailable to classify it, so this is currently an unresolved merge blocker rather than a source-attributed finding.

## Findings

- [P1] Query anchors by the persisted external item id
  `crates/gitlawb-node/src/db/mod.rs:3892`

  `record_arweave_anchor` generates a UUID for `id` and stores the Irys/Arweave transaction id in `irys_tx_id`, but this new reader looks up `id`. A real caller supplies the id from the gateway URL/listed `irys_tx_id`, so the new endpoint returns 404 before verification. The tests mask this by seeding both columns with the same value. Resolve the reader against the persisted externally-addressable id without changing the primary-key contract.

  The root cause is that the endpoint has conflated two identities: the database row's internal UUID and the externally routable transaction id. Keep the UUID as the row identity, make the lookup use the external-id column (with an appropriate index/uniqueness contract if required), and add an integration test that records an anchor through the production writer before calling `/verify/{irys_tx_id}`. The test fixture should deliberately use distinct values for `id` and `irys_tx_id`.

- [P1] Implement the actual ANS-104 binary format and deep hash
  `crates/gitlawb-node/src/ans104.rs:158`

  The new encoder is a JSON/base64 representation with a custom SHA-256 hash, and the verifier parses gateway responses as JSON. ANS-104 DataItems are binary and use the Arweave 2.0 recursive deep-hash construction; a conforming item served by a gateway therefore cannot be parsed or verified here, while items produced here will not interoperate with a bundler. Implement the standard wire/signing rules and pin them with an external protocol vector rather than a self-round-trip fixture.

  The root cause is treating an application-level JSON projection as the ANS-104 wire format. The verification boundary must consume the same binary bytes that a gateway/bundler serves, and signing must use the standard data-item field framing and deep-hash algorithm. Use a reference implementation or published vectors to establish byte-for-byte compatibility, then add positive real-vector verification and negative malformed-binary cases. Tests that create and verify with this module alone only prove internal consistency.

- [P1] Bind a successful gateway response to the requested item id
  `crates/gitlawb-node/src/arweave_v2.rs:78`

  `probe_anchor_item` treats any item signed by the persisted node key as `Present`; it never derives or compares that item's id with `req.item_id`. A stale or malicious mirror can serve a different valid item from the same node for `GET /<requested-id>`, and `/verify/<requested-id>` will attest that substitute payload as verified. Verify the artifact's standard-derived id against the persisted requested id before reporting success; the existing different-owner test does not cover this case.

  The root cause is using owner identity as a substitute for artifact identity. A node key authorizes many DataItems, so a valid signature establishes only who signed the response—not that it is the item the caller asked to verify. Derive the protocol-defined id from the returned signature after parsing the standard wire format and require equality with the requested/persisted id on both the probe and payload-return path. Add a regression test that serves a different valid same-owner item for the requested URL and asserts `Indeterminate`/no payload.

- [P1] Enforce the response cap while streaming, before buffering
  `crates/gitlawb-node/src/arweave_v2.rs:144`

  `read_capped_body` only rejects an oversized declared `Content-Length`; otherwise `resp.bytes()` buffers the complete response before its length is checked. A chunked or HTTP/2 response without that header can force unbounded allocation on the public verifier (which can fetch twice), so the advertised 1 MiB protection does not hold. Read incrementally and abort once the cumulative limit is crossed.

  The root cause is enforcing a logical size limit after delegating buffering to `reqwest`. Treat `Content-Length` only as an optimization: consume the response stream chunk by chunk, track the cumulative byte count, and stop/drop the response immediately once it exceeds the cap. Keep the same indeterminate outcome, and add a no-`Content-Length` streaming test that sends more than the limit so the test proves the reader aborts before collecting the entire body.

- [P1] Do not expose a verifier that cannot read the anchors this head writes
  `crates/gitlawb-node/src/arweave_v2.rs:101`

  The unchanged production writer still calls the legacy raw-JSON `anchor_ref_update` path, while this verifier accepts only the new signed `DataItem` representation. Even after the lookup is corrected, every anchor produced by this standalone head is classified indeterminate rather than verified; the endpoint tests bypass that integration by seeding a synthetic new-format item. Add a compatible verification path or land the producer and verifier together so the advertised endpoint works for persisted anchors.

  The root cause is splitting a producer/consumer format transition across PRs while exposing the consumer as a completed public feature. Decide on one deployment-compatible boundary: either teach the verifier to recognize and accurately report the legacy persisted format, or defer/register the public route until the ANS-104 producer writes matching rows. In either case, add an end-to-end test from the actual anchor-writing path through lookup and verification; synthetic database rows do not cover this compatibility contract.

beardthelion (dismissed):
Re-review on `f2658b47` after `fe7d2cb`. I read the diff against base `bfc44f92`, ran focused tests in a worktree (`ans104` 11/11, `arweave_v2` probe suite, `record_then_lookup_round_trips_via_irys_tx_id`, verify-endpoint gate tests), gut-checked the premise (flipping non-404 `Indeterminate` to `DefinitivelyAbsent` RED on `probe_400_is_indeterminate_not_absent`), and confirmed CI is 12/12 green on this head.

Most of the round-1 P1s landed: `irys_tx_id` lookup, repo-read gating with `optional_signature`, structured `ProbeOutcome`, empty 404 as indeterminate, artifact-id binding, streaming body cap, and v1/v2 dual verify. Two correctness gaps remain on the signing path and the new lookup column.

Prior art checked: `a-fixture-cannot-refute-the-model-that-produced-it.md`, `distinguish-unknown-from-empty-and-fail-closed.md`, `backfill-join-key-must-match-write-side-key.md`.

## Findings

- [P1] Pass raw field bytes into the ANS-104 list fold, not pre-hashed blobs

`crates/gitlawb-node/src/ans104.rs:221`

`DataItem::deep_hash` computes `deep_hash_blob` for each of the eight signature-data fields, then passes those 48-byte digests into `deep_hash_list`, which blob-hashes each element again. ANS-104 folds raw field bytes (only the tags slot is already `deepHash(tags)`). I reproduced the mismatch in Python (`correct != wrong` for the same field set). Items signed here will not verify on a standard bundler/gateway. The `external_reference_vectors` tests inline SHA-384 and never call `deep_hash_blob`, `deep_hash_list`, or `DataItem::deep_hash`, so they stay green while production interop is wrong. Fix the fold, add a vector that calls `DataItem::deep_hash` against an independent reference (arbundles or published bytes), and version if you must keep a legacy encoder for already-signed test artifacts.

- [P2] Index `arweave_anchors.irys_tx_id` for the verify lookup

`crates/gitlawb-node/src/db/mod.rs:3896`

The reader now filters `WHERE irys_tx_id = $1` instead of the PK. `arweave_anchors` still only indexes `repo` and `new_sha` (lines 659-660). Every verify call, including misses, is a sequential scan over anchor history. Append migration v27 with `CREATE INDEX ... ON arweave_anchors(irys_tx_id)`; check existing rows before choosing unique vs non-unique.

- [P2] Use the same opaque 404 message for every deny path on verify

`crates/gitlawb-node/src/api/arweave.rs:89`

Missing-row denial uses `AppError::RepoNotFound(format!("anchor {item_id}"))` while `authorize_repo_read` denial uses `repository '{owner}/{name}' not found`. Both return `error: repo_not_found`, but the `message` field lets a caller distinguish "unknown item id" from "private repo I cannot read". The comment at line 90 says the item id is never surfaced; line 92 contradicts that. Collapse all three deny paths (no row, malformed repo slug, gate deny) to the same message shape the private-repo test already expects.

One process note, not a finding: rebasing onto current `main` will conflict with several open PRs that touch the same files (#134 on `server.rs`/`api/arweave.rs`, #384 on `db/mod.rs`/`main.rs`, #285 on `config.rs`, and others). Resolve those mechanically when you rebase.

Not an ask, recorded only: `list_anchors` remains ungated (#134 tracks that surface). The verify route has no per-IP rate limit and can issue two gateway GETs per success; fair follow-up once the P1/P2 items land.

jatmn (dismissed):

## Findings

- [P1] Implement the actual ANS-104 wire and signing contract
  `crates/gitlawb-node/src/ans104.rs:63`

  The public gateway serves ANS-104 DataItems in the standard binary frame, but this module only attempts to decode a JSON/base64 projection. Its signing path also builds a different message: ANS-104 deep-hashes the nested byte structure `["dataitem", "1", owner, target, anchor, tags, data]`, whereas this code adds a signature-type element and passes already-hashed fields into `deep_hash_list`, which hashes them again. `deep_hash_tags` makes the same nested-list mistake. A standard item therefore becomes `indeterminate` at the new endpoint, and an item signed here cannot be verified by a conforming bundler or gateway.

  The root cause is treating a convenient test representation as the protocol wire format and testing helpers/self-round-trips rather than an entire external artifact. Please make the DataItem parser/encoder consume and produce the binary layout, construct the signature data directly from the standard's nested byte/list structure, and add complete golden vectors (binary input, expected ID, signature verification, and payload) from an independent ANS-104 implementation. Keep legacy v1 handling isolated rather than using it to relax v2 parsing.

- [P2] Keep second-read failures within the advertised three outcomes
  `crates/gitlawb-node/src/arweave_v2.rs:293`

  `verify_anchor` classifies the first gateway response, then issues a second GET solely to obtain the bytes for payload extraction. If that second request loses its connection, times out, or exceeds the body cap, the `?` propagation reaches `api::arweave::verify_anchor`, which turns it into an internal-error 500. That contradicts this endpoint's stated `verified` / `definitively_absent` / `indeterminate` result model: the gateway state is ambiguous, not an application fault.

  The root cause is splitting validation and consumption across two independently mutable network reads. Prefer returning the capped, validated bytes from the probe and parsing those exact bytes once. If retaining a second request is unavoidable, convert all of its transport, status, and cap failures to an `AnchorVerifyResult` with `ProbeOutcome::Indeterminate`; add a test where the first response succeeds and the second fails.

- [P2] Make all verification denials genuinely opaque
  `crates/gitlawb-node/src/api/arweave.rs:87`

  The handler intends to collapse missing, malformed, and unauthorized anchors into the same 404. Instead, a missing ID constructs `RepoNotFound("anchor {item_id}")`, while an existing private row reaches `authorize_repo_read` and constructs `RepoNotFound("{owner}/{repo}")`. `AppError::IntoResponse` serializes the supplied string into `message`, so an unauthenticated caller can distinguish a nonexistent ID from a private anchor and recover the private repository slug. The current tests assert only status/code, which masks the observable difference.

  The root cause is delegating one deny branch to a helper whose otherwise-correct repo-specific error is exposed by the shared response formatter. Normalize all deny branches at this route boundary to one constant opaque response (including malformed stored slugs), and add a table-driven test that compares complete bodies for missing, private, malformed, and anonymous cases.

- [P2] Add a forward index for the public transaction-ID lookup
  `crates/gitlawb-node/src/db/mod.rs:3894`

  The new verifier resolves its public path parameter through `WHERE irys_tx_id = $1`, but `arweave_anchors` has indexes only for `repo` and `new_sha`. Every verification request—including arbitrary misses before any gateway work—therefore takes a sequential scan over all retained anchor history. This turns a public endpoint into growing database work and compounds the route's external-fetch cost.

  The root cause is changing the reader's access path without evolving the deployed schema. Append a new, versioned migration creating a non-unique `irys_tx_id` index (unless existing data and writer invariants demonstrate uniqueness), and exercise the migration catalogue rather than modifying v1, which existing installations have already applied.

- [P2] Bound the anonymous gateway-verification work
  `crates/gitlawb-node/src/server.rs:247`

  The new route accepts anonymous callers, and valid IDs are exposed by the still-public anchor list. Each request can open one gateway request for classification and a second one for extraction, each held up to the shared HTTP-client timeout. Unlike the comparable public IPFS path, the route has neither an IP admission limit nor a concurrency bound. An attacker can therefore hold arbitrary request tasks and outbound connections while amplifying gateway egress.

  The root cause is adding an externally blocking read route outside the repository's existing expensive-work admission-control patterns. Put a narrowly scoped IP and/or concurrency limiter around the verification router before the handler performs database or gateway work; make its budget and response behavior explicit, and test that an over-limit request does not reach the gateway. Do not apply a broad limiter to unrelated Arweave listing traffic.

jatmn (dismissed):

## Findings

- [P1] Implement the actual ANS-104 wire and signing contract
  `crates/gitlawb-node/src/ans104.rs:60`

  This module invents a JSON/base64 envelope for DataItems, while ANS-104 defines a binary frame. Its signing preimage also adds `signature_type`, encodes an Ed25519 owner as a padded 64-byte value, and flattens nested tag lists into digests which are blob-hashed again. The standard preimage is the seven-element nested structure `["dataitem", "1", owner, target, anchor, tags, data]`.

  The failure path is direct: a standard gateway artifact arrives as binary bytes (or a gateway serves the data payload), `serde_json::from_slice::<DataItem>` fails, and the probe reports `indeterminate`; conversely, the next upload slice will sign bytes a conforming bundler cannot verify. The root cause is treating the test-friendly Rust projection as protocol wire data, then representing recursive deep-hash nodes as byte blobs. Make the parser/encoder operate on the standard binary layout, preserve `tags` and tag pairs as nested deep-hash lists, and use the signature configuration's actual owner length. Add a complete binary fixture from an independent implementation that asserts parsing, ID derivation, signature verification, and payload extraction—not only helper/self-round trips.

- [P1] Do not attest legacy JSON as a verified anchor without an identity proof
  `crates/gitlawb-node/src/arweave_v2.rs:101`

  A 2xx body with `schema: "gitlawb/ref-update/v1"` is immediately treated as present, and `verify_v1` returns `verified: true` when five copied fields match the database row. There is no signature, transaction/content-address check, or binding to the requested `item_id`.

  Thus a gateway or proxy can synthesize those public row fields, and the endpoint emits `verified: true` plus attacker-chosen JSON as the purported permanent anchor. The root cause is treating agreement with server-side metadata as proof of the remote artifact: it proves only that the response copied known values. Preserve legacy compatibility only behind an identity proof that binds the requested transaction/item ID to immutable content; otherwise return `indeterminate` rather than a successful verification result. Add a negative test that serves matching unsigned JSON for the requested URL and proves it cannot be reported verified.

- [P2] Bind the item ID before classifying a probe as present
  `crates/gitlawb-node/src/arweave_v2.rs:113`

  The recovery-facing probe validates the owner and signature but never derives `DataItem::id()` or compares it with `req.item_id`. A stale or malicious gateway can therefore return a different valid item from the same node and receive `Present`, even though the requested item is absent. `verify_v2` does perform this check later, but the probe has its own three-outcome contract and is the state that governs recovery.

  The root cause is using signer identity as a substitute for artifact identity: one node key legitimately signs many items. A false `Present` suppresses the retry path for the persisted item, contradicting the PR's stated different-item boundary. Derive the protocol ID immediately after successful signature validation and require equality before returning `Present`; use `Indeterminate` on mismatch. Add the same-owner/different-item regression case at the probe level, not only through the later HTTP verifier.

- [P2] Keep the second gateway read inside the three-outcome result model
  `crates/gitlawb-node/src/arweave_v2.rs:297`

  After a successful probe, payload extraction performs another independent GET and propagates transport and capped-body failures with `?`. The handler maps that error to `AppError::Internal`, so a gateway that succeeds once and then resets, times out, or sends an oversized second body produces a 500 instead of the documented `indeterminate` result.

  The root cause is classifying one mutable network response and consuming a different one. The first response does not guarantee the second will be available or contain the same artifact, so its failure is gateway ambiguity rather than an application fault. Prefer returning the capped, validated bytes from the probe and parsing exactly those bytes once. If two reads remain necessary, convert every second-read transport/status/body-cap failure into `ProbeOutcome::Indeterminate` and add a first-success/second-failure test.

- [P2] Bound anonymous verification work before gateway I/O
  `crates/gitlawb-node/src/server.rs:247`

  The new route accepts anonymous callers and installs only `optional_signature`. Item IDs are available from the still-public anchor list, and every valid-ID request can hold two outbound gateway requests (with an additional request for indeterminate results), yet the route has neither a per-IP limiter nor a concurrency/work bound.

  The failure path is inexpensive inbound requests turning into unbounded outbound connections and gateway traffic until the shared HTTP-client timeout; a single public item ID is enough to sustain the load. The root cause is adding a blocking public read outside the repository's existing expensive-work admission pattern—the nearby `/ipfs/{cid}` router attaches its limiter before request processing. Add a narrowly scoped per-IP and/or concurrency admission bound around only the verify router, ahead of gateway work, and prove with a route-level test that an over-limit request never reaches the gateway.

jatmn (dismissed):

## Findings

- [P1] Do not report legacy anchors verified from copied row fields
  `crates/gitlawb-node/src/arweave_v2.rs:548`
  `verify_v1` returns `verified: true` when an unsigned gateway response copies five values from the database, but it never binds that response to the requested `item_id` or verifies immutable content. A stale or hostile gateway can serve a different JSON object with those public values at the requested URL and make the endpoint attest a ref update that was not proven by the persisted artifact. Preserve legacy compatibility only with a non-forgeable item/content binding; otherwise classify this response as indeterminate.

- [P1] Retrieve a verifiable ANS-104 envelope instead of parsing gateway content as a `DataItem`
  `crates/gitlawb-node/src/arweave_v2.rs:126`
  The route fetches `GET /{item_id}` and parses its body as a JSON `DataItem`. Gateways normally resolve a data-item ID and return the item’s content, while the ANS-104 signature/owner/header live in the enclosing binary frame. Thus a real signed v2 anchor yields its payload (or other raw content), fails this JSON projection, and can never be cryptographically verified here. Obtain the complete item/bundle frame with a verifiable location/provenance path, then parse and verify that frame with an independently produced fixture.

- [P2] Preserve the signed fields in `DataItem::to_binary`
  `crates/gitlawb-node/src/ans104.rs:546`
  A freshly signed Ed25519 item hashes the 64-byte owner produced by `new_unsigned`, whereas this encoder writes only 32 owner bytes and always fills the signature slot with zeros. Consequently `new_unsigned -> sign_data_item -> to_binary -> from_binary -> verify_data_item` does not preserve a valid signature; an upload caller would publish an unverifiable item. Canonicalize the owner before signing and serialize the populated signature, then cover the complete sign/encode/parse/verify path without re-signing after parse.

- [P2] Support the valid size-prefixed Avro tag block form
  `crates/gitlawb-node/src/ans104.rs:579`
  ANS-104’s Avro tag-array encoding permits a negative block count followed by the block byte length. `decode_tags` rejects every negative count, so a conforming DataItem using that legal encoding cannot be parsed or verified. Consume and validate the size field while retaining the existing bounds checks.

beardthelion (dismissed):
Re-review on `f4c2340`. I read the diff against base `bfc44f92`, installed this PR's own pinned `arbundles` 0.10.1 from `scripts/package.json` and compared its signing preimage against the committed fixture, ran the generator script both as committed and with one line added, and ran a mutation matrix in a scratch worktree (premise, the deep-hash vectors, the gate's caller, the node_did guard, the route mount). CI is green on this head.

Credit first, because two of my round-3 asks landed cleanly. The second gateway GET is gone: the probe hands its buffered bytes to the `Present` arm, so the 500-on-second-read contract break is closed. And the verify route now has a real per-IP brake with its own bucket, a bounded key map that rejects before inserting, rightmost-XFF only under an explicit trusted-proxy mode, `0` meaning disabled rather than block-all, and a test that a second request from the same peer is shed rather than served.

## Findings

- [P1] Sign what `arbundles` signs, and pin it with a signed item

  `crates/gitlawb-node/src/ans104.rs:288`

  I installed the exact dependency this PR pins and read `getSignatureData` in `arbundles/build/node/cjs/src/ar-data-base.js`. It deep-hashes eight elements: `"dataitem"`, `"1"`, `signatureType.toString()`, `rawOwner`, `rawTarget`, `rawAnchor`, `rawTags`, `rawData`, where `rawTags` is the serialized Avro buffer folded as a flat blob. This head folds seven, drops the signature type, and passes tags as a nested list. On the golden vector's own inputs arbundles produces `35a24c0e1c1923fe...`; the expected array committed here is `3ad967a7...`. That is the third fold in three rounds and it is still not the one bundlers run.

  Two more breaks sit underneath it, either of which is independently fatal. `new_unsigned` builds a 64-byte owner and `deep_hash` folds all 64, while `to_binary` writes `owner_size(2)`, which is 32. The comment at line 950 calls that "a documented gitlawb convention", but the deep-hash is the signing preimage, so the signature computed in memory does not belong to the frame that gets published. And `to_binary` writes the signature slot as zeros unconditionally, so nothing this module emits is publishable at all and `id()` over it is `sha256(zeros)`.

- [P1] Sign the golden vector; the generator never calls `sign`

  `scripts/ans104_golden.mjs:14`

  `createData(...)` returns an unsigned item and the script reads `getRaw()` and `id` straight off it. I ran it as committed: it prints `mM5C3u9R1AJp1UL1MUvvLHRo1AGtXYUWi_q0wBCPdfc`, which is the id pinned in the test and is exactly `base64url(sha256(65 zero bytes))`. Its own `idCheck !== id` guard passes vacuously because both sides hash the same zeros. Adding `await item.sign(signer)` produces a real signature and `item.isValid()` returns true; I ran that too. The capture is also an Ethereum item, signature type 3, while the node only ever handles Ed25519, so even once signed it would not reach the Ed25519 verify path.

  This is why none of the three breaks above is visible. `binary_round_trip` re-signs the parsed item before verifying it, which re-signs over the 32-byte owner and hides the width mismatch; the golden's `assert_eq!(bin2, binary)` passes only because that fixture's signature is already zeros. Gutting `deep_hash_blob` entirely leaves all three `external_reference_vectors` tests green while the rest of `ans104::` goes red, because they rebuild the fold from raw `sha384` calls and never touch the production primitives.

- [P2] Drive the verify route through `build_router` in at least one test

  `crates/gitlawb-node/src/api/arweave.rs:952`

  Every test builds its own `Router::new().route(...)` and re-applies the layers by hand. I removed `rate_limit_by_ip` from `server.rs` and the suite stayed green; I then removed `.merge(arweave_verify_routes)` entirely, unmounting the endpoint, and it was still green. So the route's presence, its `optional_signature` layer, the `IpRateLimiter` extension the code's own comment warns must be attached, and the layer order are all unpinned. `ipfs_route_ip_rate_limit_is_attached` is the template.

- [P2] Test the gate with an authenticated caller

  `crates/gitlawb-node/src/api/arweave.rs:122`

  Hardwiring `caller` to `None` leaves the endpoint suite green; hardwiring it to the private fixture's owner turns it red. The gate runs, but only its anonymous arm is exercised, so an authenticated non-collaborator has never been driven through it. This has been an open review thread since round 2.

- [P2] Drop the `node_did` comparison or make it compare two different things

  `crates/gitlawb-node/src/arweave_v2.rs:516`

  `expected_pk` is derived from `row.node_did` in the handler and `owner_did` is derived back from `expected_pk`, so the two sides are the same value. Removing the branch changes no test; inverting it reddens every accepting test, which confirms it executes on every valid verify and simply cannot fail. The comment says a mismatch means someone re-keyed and the row is stale. It cannot detect that.

- [P2] Let database errors out of the read gate

  `crates/gitlawb-node/src/api/arweave.rs:129`

  `.map_err(|_| AppError::RepoNotFound(...))` turns a connection failure into "anchor not found". `get_cert`, the pattern this handler's comment cites, uses a bare `?`, and `list_anchors` in this same file has a test asserting a closed pool yields 503. Match the denial variants and re-raise the rest.

- [P3] Housekeeping on the new knobs and the capture

  `crates/gitlawb-node/src/config.rs:143`

  `GITLAWB_ARWEAVE_GATEWAY_URL` and `GITLAWB_ARWEAVE_VERIFY_RATE_LIMIT` are in neither README nor `.env.example`; `GITLAWB_IPFS_RATE_LIMIT`, the knob this one is modelled on, is in both. `scripts/ans104_golden_output.txt` disagrees with the fixture it documents: its hex carries one byte more, so it does not parse as a data item at all, and it reports `owner_len = 512` for a type-3 item whose owner field is 65 bytes wide. `scripts/ans104_golden.ts` exits without emitting anything and should go. `ans104.rs:816` is the only non-strict `verify` in the workspace. And `verify_endpoint_public_repo_anonymous_200` asserts the status only, while its message promises "200 with payload".

## On the shape of this round

Four of the seven items above are consequences of round 2 and round 3 fixes rather than of the original diff: the golden vector, the binary encoder, the `node_did` guard, and the hand-built routers. Each round has answered the previous round's finding with new code that carries the next one, and the interop question from round 1 is still open. That is the signal to change approach rather than run a fifth round on the same shape.

The design call is mine and I am settling it rather than leaving it open. The public endpoint must not report `verified` for a lane with no cryptographic proof, and this PR should not ship an ANS-104 signer that cannot interoperate. The cleanest cut is to land the transport and the three-outcome probe here, which are genuinely good and which I would take today, and hold the public verify endpoint and the signing half until there is a producer and a signed external vector to test them against. If you would rather keep it whole, then the fold, the owner width and the signature slot all need to be right together, proven by a real Ed25519 item that `arbundles` signed and this code verifies without re-signing.

beardthelion (dismissed):
Checked the verify and ans104 test suites on this head (all green on CI), premise RED on the artifact-id guard in verify_v2, and the verify trust boundary (opaque 404, gate-before-gateway-fetch, expected_pk from persisted row.node_did). Second-model refute pass ran; one finding survives verification.

## Findings

- [P2] Compare derived item id to the requested id inside probe_item before returning Present
  `crates/gitlawb-node/src/arweave_v2.rs:145`
  probe_item verifies owner and signature but not artifact identity. A gateway answering GET /wrong-id with any other valid same-owner ANS-104 item returns Present. verify_v2 catches the mismatch for the HTTP verify endpoint, but split 1's recovery drain will call probe_anchor_item directly and never hits verify_v2. probe_2xx_with_valid_signed_item_is_present currently expects Present with item_id "abc" while serving an item whose derived id differs, which encodes the gap. Add item.id() != req.item_id -> Indeterminate in probe_item and fix that test to use the real derived id or assert Indeterminate on mismatch.

Not an ask, recorded only: permits_reupload stays dead_code until split 1 wires recovery; build_http_client already applies a 10s outbound timeout; list_anchors public visibility unchanged (#134).

beardthelion (dismissed):
I read the diff on head `1e0a0ef`, ran focused tests (ans104 15/15, arweave_v2 20/20, verify_endpoint 11/11), premise RED on the 400→indeterminate arm, and a second-model refute pass (`gpt-5.5-2026-04-23`). CI is 12/12 green. The trust-boundary work from prior rounds holds: `irys_tx_id` lookup, repo gate with opaque 404, arbundles Ed25519 golden vector, single GET, v1 never `verified`, rate limit on verify router. The second-model timeout finding refuted (`build_http_client` already applies a 10s outbound cap).

## Findings

- [P2] Require signature verification before `probe_item` returns `Present`

  `crates/gitlawb-node/src/arweave_v2.rs:157`

  When `ProbeRequest.expected_owner_pk` is `None`, `probe_item` skips `ans104::verify_data_item` but still returns `Present` if `item.id()` matches `req.item_id`. A gateway can serve a DataItem with arbitrary owner bytes and a random 64-byte signature, set the requested `item_id` to `base64url(sha256(signature))`, and the probe reports confirmed presence with no cryptographic proof. The HTTP verify path always passes `Some(pk)`, but split 1's recovery drain will call `probe_anchor_item` directly and uses `Present` for paid re-upload decisions. Require `expected_owner_pk: Some` before returning `Present`, or treat `None` as `Indeterminate` even on id match. Add a test: `probe_anchor_item` with `expected_owner_pk: None` and an invalid-but-id-matching DataItem must be `Indeterminate`, not `Present`.

Not an ask, recorded only: second-model payload-vs-row tuple binding gap (verify checks sig + artifact id + `node_did`, not payload fields against the DB row); `list_anchors` still ungated (#134).

beardthelion (dismissed):
Re-reviewed head `91a8ac6e` after the second-model refute pass. Focused tests still green (ans104 15/15, arweave_v2 20/20, verify_endpoint 11/11), premise RED holds on the 400, empty-404, and artifact-id arms, and the trust-boundary work from prior rounds (irys_tx_id lookup, repo gate with opaque 404, arbundles golden, streaming cap, v1 never verified) still holds. Two items below need fixing before merge.

## Findings

- [P2] Return indeterminate for invalid persisted node_did instead of 500

  `crates/gitlawb-node/src/api/arweave.rs:143`

  The handler parses `row.node_did` into a `did:key` before calling `arweave_v2::verify_anchor`, even though v1 anchors carry no signature and the verifier is supposed to report `Indeterminate`. I seeded a public-repo row with `node_did=''`, mocked a v1 gateway body, and got HTTP 500 with zero outbound gateway requests. Legacy or malformed rows should reach the v1 indeterminate path (or return the advertised `{status:"indeterminate", verified:false}` envelope) rather than `AppError::Internal`.

- [P2] Align the PR description with the empty-404 policy the code implements

  `crates/gitlawb-node/src/arweave_v2.rs:208`

  The PR body table and test inventory still say a bodyless 404 is `DefinitivelyAbsent` and name `probe_404_with_empty_body_is_definitively_absent`. The code and the committed test `probe_404_with_empty_body_is_indeterminate` treat empty 404 as `Indeterminate`. Update the description and named-test list to match the implementation so the recovery contract is not contradicted in prose.

Not an ask, recorded only: split 1 (#384) still owns wiring `permits_reupload` into the live recovery drain; that is out of scope for split 2/4. The default-planner EXPLAIN half of `migration_v27_irys_tx_id_index_is_used_by_the_lookup` is planner-fragile; the `enable_seqscan=off` lookup half is the load-bearing proof.

jatmn (dismissed):
I rechecked the current head as a whole. The previously reported gateway-envelope defect remains, and the adjacent failures all come from the same root problem: the tests define a convenient mock gateway contract that the configured production provider does not implement. I am consolidating those symptoms below so they can be fixed together rather than producing another sequence of one-line review patches.

## Overall guidance

This PR currently spans four boundaries at once: an ANS-104 codec, a remote-provider adapter, a recovery classification policy, and a public HTTP endpoint backed by persisted rows. The repeated findings are not mainly caused by missing isolated guards. They are caused by the absence of one executable end-to-end contract connecting those boundaries.

In particular:

1. The positive tests return a locally constructed binary/JSON `DataItem`, but the configured gateway URL serves decoded data and redirects before doing so. The mock has therefore become the specification even though it does not model production.
2. The absence test invents a JSON response shape and calls it protocol-defined, but neither ANS-104 nor the selected gateway defines that response as authoritative proof that an upload was never accepted.
3. Upload and read configuration are independent. The checked-in example enables Irys devnet upload while verification defaults to an Arweave mainnet gateway, so even a correct response parser can query the wrong network.
4. The producer and live recovery consumer are deferred to sibling PRs. That is a valid way to split implementation, but it means this PR needs especially strong boundary fixtures; otherwise locally consistent code can compile and pass while no real producer/provider/consumer path can use it.
5. The PR description and named-test inventory have fallen behind the implementation during review. That makes it difficult to tell whether a later patch restored the intended contract or merely changed the code to satisfy the latest comment.

Before another patch round, please choose one coherent landing shape:

- **Narrow foundation:** land only the standard-compatible ANS-104 primitives and the pure three-outcome policy types here, and defer the provider-backed public endpoint/probe until the uploader, retrieval proof, and recovery consumer can be proven as one vertical slice; or
- **Complete provider adapter:** keep the endpoint in this PR, but define one supported upload/read network and an exact provider API for retrieving authenticated envelope material and authoritative absence evidence. Exercise that adapter using the same redirect policy and response representation production uses.

Either choice is within the existing split intent. This is not a request for a new feature or a provider redesign. It is a request that the transport and verification behavior already claimed by this PR be supported by one real, testable contract. A convincing closure test should start from an independently signed Ed25519 item, use a response captured from or faithful to the selected provider API, drive the production client/route, and prove all of: known-present verification, wrong-item rejection, ambiguous failure remaining indeterminate, and the one exact provider response that safely authorizes recovery.

## Merge readiness

- [P1] Allocate a migration version outside the active sibling range
  `crates/gitlawb-node/src/db/mod.rs:1126`
  This branch assigns `arweave_anchors_irys_tx_id_index` version 27. Active split PR #384 already uses versions 27 through 33, beginning with `pending_ref_transitions_durable_outbox`. The accepted split direction requires distinct migration versions and an order-independent final schema; as written, the branches cannot compose safely, regardless of which one lands first. Please rebase against the actual series order and allocate a version outside the sibling range. Do not merely change this to 28: the whole 27-33 range is already occupied on #384, and the final number should also be checked against the other split branches before pushing.

## Findings

- [P1] Replace the mock-only gateway assumptions with one real provider contract
  `crates/gitlawb-node/src/arweave_v2.rs:92`
  The current default path has four independent ways to remain permanently `Indeterminate`:

  1. `.env.example` enables upload through `https://devnet.irys.xyz`, while the new verifier defaults to the Arweave mainnet gateway. There is no validation that the upload receipt and read gateway belong to the same network.
  2. `probe_anchor_item` requests `https://arweave.net/<item_id>`, but the shared production client has `Policy::none()` because it is also used for attacker-influenceable peer URLs. The default gateway currently answers item URLs with a 302 to an ID-scoped host, so production rejects the response before reading its body. The tests use direct mock URLs and do not exercise this client/default combination.
  3. Even if that redirect is followed safely, the bare-ID API returns the item's decoded data. The [Arweave HTTP API](https://docs.arweave.org/developers/arweave-node-server/http-api) explicitly describes `GET /{id}` as “Get the decoded data from a transaction,” and [Irys documents](https://docs.irys.xyz/build/d/features/txids) its transaction-ID URL as a data download. The probe instead requires the complete ANS-104 binary frame or the test-only JSON `DataItem` projection. A real v2 ref-update therefore returns ordinary ref-update JSON, fails both parsers, and never reaches `Present`/`verified`. This is the same envelope-retrieval defect previously reported on this PR; it remains unresolved on the current head.
  4. On the missing-item path, `classify_404` recognizes only `{"status":"not found"}`/`not_found`. That shape is not defined by ANS-104 or the selected public gateway. The default gateway's current redirected route returns HTML for an unknown ID, while documented gateway APIs may use an empty 404; both are deliberately classified `Indeterminate`. Consequently the only outcome that permits recovery is not reachable through the shipped provider contract.

  Please address these together at the provider boundary. Select and document a compatible uploader/network/read API; use a dedicated HTTP client or narrowly allowlisted redirect policy rather than weakening the shared SSRF guard; retrieve the complete data-item envelope or authenticated bundle location/proof through an endpoint that actually supplies it; and define `DefinitivelyAbsent` from provider-documented evidence rather than a Mockito-only body. Add provider-realistic fixtures for the redirect, successful envelope retrieval, decoded-data response, unknown/indexing response, and authoritative absence response, and drive at least one through the real router with the production client construction. Fixing only the redirect, parser, 404 arm, or example URL will leave the other failure modes intact.

- [P2] Classify legacy rows before requiring v2 identity material
  `crates/gitlawb-node/src/api/arweave.rs:143`
  The handler parses the persisted `node_did` into an Ed25519 key before it fetches or classifies the gateway response. A legacy/public row whose `node_did` is empty, malformed, or not a usable `did:key` therefore returns HTTP 500 without making the gateway request. That contradicts this PR's stated compatibility rule: v1 has no cryptographic proof and must return the structured `indeterminate` result rather than being verified or failing the endpoint internally. This failure has also been reproduced in the current public review by seeding an invalid legacy `node_did`.

  Please separate format classification from v2 owner verification. For example, the retrieval result can identify legacy content before an expected Ed25519 key is required, or invalid persisted v2 identity can be represented as an indeterminate verification result. Preserve the strict v2 owner/signature check; the requested outcome is only that unusable legacy identity state follows the already-advertised v1 result contract instead of escaping as 500. Add an endpoint test with a public v1 row and invalid/empty `node_did` that asserts HTTP 200, `status: "indeterminate"`, `verified: false`, and no accidental relaxation for v2.

- [P2] Validate the complete two-byte signature type before selecting an algorithm
  `crates/gitlawb-node/src/ans104.rs:421`
  ANS-104 stores the signature type as a two-byte little-endian value, but `from_binary` immediately casts the parsed `u16` to `u8`. As a result, the malformed header `02 01` (wire value 258) aliases supported Ed25519 type 2. The parser then uses Ed25519 field widths, reconstructs the deep hash using the normalized value `2`, accepts the unchanged signature and item ID, and `to_binary` silently rewrites the header as canonical `02 00`. A strict implementation rejects the artifact while this verifier reports it present, so the new cryptographic boundary is malleable outside the signature check.

  Please keep the wire value as `u16`, reject every value outside the explicitly supported ANS-104 set, and only then convert to an internal algorithm identifier if needed. Add a regression test that starts from the signed Ed25519 golden frame, changes only the high type byte, and requires parsing/probing to fail as indeterminate. Also keep parse/serialize type widths aligned so no unsupported wire value can be normalized into a supported signed artifact.

- [P2] Make the PR description match the final recovery contract
  `PR description — “The three-outcome probe” and “Required proof” sections`
  The description currently says an empty 404 is `DefinitivelyAbsent` and lists `probe_404_with_empty_body_is_definitively_absent`, while current code and the committed test classify an empty 404 as `Indeterminate`. It also describes the `{"status":"not found"}` body as the trustworthy absence case without identifying a provider specification that emits it. This contradiction is already part of the current public review and remains unresolved.

  After settling the provider contract above, update the response table, named-test inventory, configuration examples, and standalone-safety claims in one pass. The description should identify the supported provider/network pair, the exact response or proof that establishes presence, the exact evidence—if any—that establishes definitive absence, and which ambiguous responses remain retryable. This documentation is part of the recovery safety contract: future reviewers and sibling PRs should not have to infer a paid re-upload decision from stale prose.

## Definition of done for this review

To avoid another drip round, please verify the complete set together before requesting re-review:

- the sibling migration range is reconciled;
- uploader and reader configuration are guaranteed to target the same network;
- the production client's redirect behavior is intentionally supported and tested without weakening the shared SSRF boundary;
- the selected provider API supplies the envelope/proof representation the parser expects;
- present, ambiguous, malformed, wrong-item, and authoritative-absence provider responses are covered with provider-realistic fixtures;
- legacy invalid identity state returns structured indeterminate, while v2 still requires the expected signer;
- the full two-byte signature type is validated before hashing or verification; and
- the PR description, configuration examples, named tests, and standalone-safety statement describe the behavior that actually ships.

The row-to-payload tuple comparison and generic non-Ed25519 signing behavior are intentionally not requested here: the former depends on producer/database association semantics deferred to the next slice, and the latter has no supported production caller on this head. Keeping them out is deliberate so this review stays on the verified responsibilities of split 2/4.

beardthelion (dismissed):
The codec logic is correct: the deep-hash fold matches arbundles, `from_binary` parses the frame properly, `to_binary` round-trips byte-exact, `verify_data_item` checks the owner against an external expected key before constructing the VerifyingKey, and signature_type aliasing (wire 258 to Ed25519 type 2) is rejected. The golden vector pins the expected owner key independently from the seed, not extracted from the item. The `ProbeOutcome` policy test is load-bearing. Four cleanup asks before this lands.

## Findings

- [P1] Fix the `signature_size` doc to match the code
  `crates/gitlawb-node/src/ans104.rs:96-97`
  The doc says "Unknown types fall back to the Ed25519 width with a debug-visible 0" but the code returns 0 (`_ => 0` at line 104), not 64. There is no Ed25519 fallback. The doc misdescribes the function on a crypto path that the next slice will build on.

- [P2] Fix the `owner_pubkey_ed25519` doc to match the code
  `crates/gitlawb-node/src/ans104.rs:252-253`
  The doc says "The owner field carries 32 pubkey bytes + 32 zero bytes" but Ed25519 owner is exactly 32 bytes (`owner_size` returns 32 at line 112). The 64-byte owner is the Arweave/RSA format (sigtype 1). The comment conflates signature types and will mislead the trust-boundary slice. Also fix the `sign_data_item` comment at lines 895-898: it says "default to Ed25519 if it was zeroed" but the code unconditionally overwrites sigtype at line 903, and `deep_hash()` bails on sigtype 0 before that line runs. Also swap the labels in the `verify_data_item` error at lines 920-923: "ref sigtype" holds the item's value, not the expected value.

- [P3] Remove the duplicate bounds check in `to_binary`
  `crates/gitlawb-node/src/ans104.rs:558-564`
  Lines 534-540 and 558-564 are identical `if owner_bytes.len() < own_len` checks. `owner_bytes` is not modified between the two, so the second is unreachable dead code.

- [P4] Use saturating_sub for tag length bounds checks
  `crates/gitlawb-node/src/ans104.rs:714,728`
  `pos + name_len > payload.len()` can overflow `usize` on 32-bit when `name_len_i as usize` truncates. Use `name_len > payload.len().saturating_sub(pos)` instead. The block_size path at line 697 already uses this pattern correctly.

The `external_reference_vectors` tests (lines 1407-1486) call `sha384` directly and re-derive the expected output in the test body. They do not exercise `deep_hash_blob`, `deep_hash_list`, or `deep_hash_chunk`. The golden vector test (line 1174) does exercise the real deep-hash path and pins the output against arbundles, so the interop pin is sound. The reference vectors are testing the SHA-384 primitive, not the wrappers.

`verify_data_item` does not check `item.id()` against a requested artifact ID. The module docs (lines 26-30) already call this out as a responsibility of the future endpoint, not the verifier. Correct for this slice.

No non-Ed25519 signature type coverage in tests. The PR only implements Ed25519 verification, so this is expected. When RSA/Ethereum/Solana verification lands, those tests should come with it.

jatmn (dismissed):

## Required changes (do exactly this)

### 1. Code change — one line in `ans104.rs`

**File:** `crates/gitlawb-node/src/ans104.rs`  
**Line:** 960 (`mod tests` doc comment)

**Current (wrong):**
```rust
    //! `scripts/ans104_golden.mjs`. The team memory
```

**Change to:**
```rust
    //! `scripts/ans104_golden_ed25519.mjs` (Ed25519 interop pin; the legacy
    //! Ethereum script `scripts/ans104_golden.mjs` is illustrative only). The team memory
```

That is the **only code change** this review requests. Do not change golden hex, tests, dependencies, or add new files.

**Why:** Line 960 tells maintainers to run the wrong regeneration script. The real interop pin is `ans104_golden_ed25519.mjs` (already named correctly at lines 1154–1156 and in `scripts/ans104_golden_output.txt`).

---

### 2. Process change — unblock merge (no code)

**Action:** Re-request review from beardthelion on head `c9292d8`, or have a maintainer dismiss the stale `CHANGES_REQUESTED` on `0271f49b`.

**Why:** beardthelion’s four codec cleanup items from `0271f49b` are already fixed in `c9292d8`. GitHub still shows `CHANGES_REQUESTED` from the older commit, which blocks merge even though the code is ready.

**No further codec work is needed** for beardthelion’s prior P1–P4 items — they landed in `c9292d8`.

---

## Status of the codec (nothing else to fix here)

On `c9292d8`, the split 2/4 codec is otherwise ready:

- beardthelion (`0271f49b`): “The codec logic is correct” — deep-hash matches arbundles, binary round-trip works, verify checks external owner key.
- `c9292d8` fixed his four doc/bounds items (`signature_size` doc, `owner_pubkey` doc, duplicate bounds check, `saturating_sub`).
- Tests: `ans104` 16/16, `arweave_v2` 1/1, clippy clean, CI 13/13 green.
- Interop pin: `dataitem_matches_arbundles_golden_vector` passes against real arbundles-signed Ed25519 bytes.

---

## Findings (detail)

- [P3] Fix wrong golden script name in tests module doc  
  `crates/gitlawb-node/src/ans104.rs:960`  
  See **Required changes §1** above.

- [P2] Clear stale `CHANGES_REQUESTED` after `c9292d8`  
  See **Required changes §2** above. This is merge process, not a code defect.

---

## Do not fix in this PR (closed)

Prior review rounds raised many items on code that was **removed** in `8ffe22e` or **deferred** to later splits. Do not expand this PR to address them — that is what caused the endless review loop.

| Do not add / do not chase | Reason |
|---|---|
| Verify HTTP endpoint, gateway probe, `verify_anchor` | Removed; deferred to vertical slice |
| `verify_data_item` vs URL `item_id` | Deferred; beardthelion accepted on `0271f49b` |
| DB lookup, migration v27 | Not in this split (#384 owns migrations) |
| `ProbeOutcome` evidence / transport wiring | Policy types only on this head |
| JSON `raw_tags` round-trip | Binary is canonical; no JSON persistence here |
| `new_unsigned` plain-text anchor interop | Empty-anchor golden is the pin; bundler slice owns constructor API |
| npm range / CI regen / non-Ed25519 verify | Dev maintenance or later sigtype work |

---

## Why review kept going in circles

1. **Scope shrank mid-flight.** Early heads had verify endpoint + probe + DB. `8ffe22e` removed that; many GitHub threads still point at deleted code.
2. **Reviewers looked at different heads.** jatmn’s integration findings on `91a8ac6e` are stale after narrow-down. beardthelion’s latest codec review is on `0271f49b` (fixed in `c9292d8`).
3. **Docs lagged fixture changes.** Ethereum script → Ed25519 pin; most comments were updated, but the `mod tests` header at L960 was missed.
4. **GitHub review state didn’t clear** after `c9292d8`, so merge stayed blocked and invited more passes.
5. **Bots posted on every push** regardless of narrowed scope.

**After the one-line doc fix + re-review on `c9292d8`, this split should be done.**

---

## Validation on head `c9292d8`

```sh
cargo test -p gitlawb-node ans104::
cargo test -p gitlawb-node arweave_v2::
cargo clippy -p gitlawb-node -- -D warnings
node scripts/ans104_golden_ed25519.mjs   # reproduces pinned hex/deephash/id
```

All pass on current head.

jatmn (approved):
@Vasanthdev2004 off to you

beardthelion (approved):
Re-review on `a3d7903c`. The deep-hash fold matches `arbundles@0.10.x` `getSignatureData`: I gutted the 8-element fold in a mutation tree and the golden vector test went red, then restored it and the test went green. The golden vector is a real `arbundles`-signed Ed25519 item with the expected owner key pinned independently in the test, not derived from the item. Binary round-trip is byte-exact against the committed 192-byte artifact. The recovery policy is the correct three-outcome split: only `DefinitivelyAbsent` authorizes re-upload, and `Indeterminate` stays non-terminal. The full suite passes on CI.

The four prior cleanup asks from my round on `0271f49b` are fixed: `signature_size` doc, `owner_pubkey_ed25519` doc, duplicate bounds check removal, and overflow-safe tag length checks. The one-line doc fix requested by jatmn is in. The four unresolved CodeRabbit threads are stale from earlier heads where this PR carried the verify endpoint, DB lookup, and gateway probe; those surfaces are gone and the threads do not apply to the current diff.

## Findings

- [P2] Reject non-canonical owner lengths in `owner_pubkey`, `deep_hash`, and `to_binary`
  `crates/gitlawb-node/src/ans104.rs:240`
  `owner_pubkey` (line 240), `deep_hash` (line 335), and `to_binary` (line 535) each check `owner_bytes.len() < need` but not `> need`. All three silently truncate to `owner_bytes[..need]`. A `DataItem` with `owner = base64url(pubkey || junk)` passes `verify_data_item` because the key comparison and the hash both see only the first 32 bytes. I confirmed this with a probe: signed an item, appended 10 bytes of junk to the owner, called `verify_data_item` and it returned `Ok(())`. `from_binary` reads exactly `owner_size` bytes so parsed items are always canonical, but directly constructed or post-mutation items with non-canonical owners pass verification and `to_binary` silently drops the extra bytes on re-encode. The fix is a `>` check in all three locations: `if owner_bytes.len() != need { bail!(...) }`. These functions are `#[allow(dead_code)]` with no production caller on this head, so this is not blocking, but the next slice will call them.

- [P3] Set `signature_type` to Ed25519 before computing `deep_hash` in `sign_data_item`
  `crates/gitlawb-node/src/ans104.rs:894`
  `sign_data_item` calls `item.deep_hash()` at line 894, which folds `self.signature_type.to_string()` as the sigtype ASCII element, then sets `item.signature_type = SIGNATURE_TYPE_ED25519` at line 897. For items built via `new_unsigned` (which sets sigtype 2 at construction) this is a no-op. But all fields on `DataItem` are `pub`, so a caller can construct an item with sigtype 1 and call `sign_data_item`: the hash folds `"1"`, the signature is computed over that hash, then sigtype is overwritten to 2. `verify_data_item` computes the hash with `"2"` and the signature fails. The comment at lines 889-892 says "Signing always produces Ed25519" but the hash uses the pre-existing sigtype. The fix is one line: move the `signature_type` assignment before the `deep_hash` call. This function is `#[allow(dead_code)]` with no production caller on this head.

Not an ask, recorded only: `verify_data_item` constructs `VerifyingKey::from_bytes(&owner_pk)` from the item's owner at line 943, after comparing `owner_pk` against the externally supplied `expected_pubkey` at line 934. The comparison makes verification correct today. Constructing the verifying key from `expected_pubkey` instead would be defense-in-depth: if a future refactor drops the comparison, verification stays anchored to the external key rather than becoming vacuous. The current code is correct.

beardthelion:
@kevincodex1 LGTM

PR #386, opened 2026-08-28
fix(node): version ref_certificates for forward compat (#26 split 3/4)

Gravirei:
## Why

Reviewer 2 closed PR #224 on 2026-08-28 with a directive: split the work into four narrow PRs. This is **Split PR 3** (certificate / API / CLI compatibility). No findings are assigned to it; it is required because the protocol/client work is independently reviewable and should not remain coupled to post-receive recovery.

The wire-format problem: the ref-cert payload is a fixed 7-field JSON blob the node signs. Future changes (additional pusher attestations, content-binding fields, secondary signatures) would either break every existing cert or be impossible to ship without a flag day. Versioning the cert makes the format forward-compat and explicit.

## What this PR changes

- **Migration v37:** `ref_certificates` gains `version INTEGER NOT NULL DEFAULT 1`. `DEFAULT 1` means an existing row reads as v1 with no backfill; `NOT NULL` means a missing value is a hard error rather than a silent v0.
- **`RefCertificate.version: u32`:** v1 is the pre-versioning 7-field payload — no `version` key in the signed JSON, byte-for-byte identical to today's cert. v2+ certs will include a `version` key and any new fields.
- **API:** `list_certs` and `get_cert` include `version` in the JSON response. `issue_ref_certificate` sets `version: 1` on every new cert.
- **gl cert show:** reads the `version` field; missing field defaults to 1 (forward-compat with old servers). v1 certs go through the unchanged verify path. v2+ certs are explicitly rejected with an "upgrade the client" message rather than silently guessing the payload shape.
- **gl cert list (round 4):** the list endpoint now runs through `error_for_status()` before parsing JSON so a gated 404 surfaces as an error, not an empty list with exit 0.
- **gl cert show anchor (round 4):** `--verify` without `--expect-node` refuses to mint a trust anchor from the queried node's self-reported DID. This is a breaking change; the canonical command is `gl cert show <repo> <id> --verify --expect-node <did>`. The most recent commit on this branch carries the canonical form in its `BREAKING CHANGE:` footer for release-note scrapers.

The reviewer said PR 3 must own certificate payload versions, API representation, CLI display and verification, and legacy compatibility fixtures. New fields must be optional for readers, and both old-server/new-client and new-server/old-client combinations must remain safe. The split is required because this protocol/client work is independently reviewable and should not remain coupled to post-receive recovery.

## The v1 → v2 forward-compat table (the reviewer's "both combinations must remain safe")

| | Old client (pre-PR-3) | New client (this PR) |
| --- | --- | --- |
| **Old server (pre-PR-3)** | ✅ v1 cert, no `version` field, signature verifies | ✅ v1 cert, missing `version` defaults to 1, v1 verify path |
| **New server (this PR)** | ✅ v1 cert with `version: 1`, old client ignores the field, signature verifies | ✅ v1 cert, v1 verify path. v2+ cert refused with "upgrade the client" |

The "old client" cells rely on the v1 payload being byte-for-byte identical to the pre-PR-3 shape — that is what `v1_payload_matches_frozen_canonical_form` pins.

## Why "v1" is the current default (and what v2 will look like)

A v1 cert is the pre-versioning shape, unchanged. v2 will add fields (the design is out of scope for this PR — the spec said "New fields must be optional for readers" but did not name any). When v2 ships, the `version` field will be set to 2 in the signed JSON, and the field order will be a superset of v1's so a v1 client still verifies the v1 fields it understands. A v2+ client reading a v2 cert reconstructs the v2 payload shape and verifies. A v2+ client reading a v1 cert reconstructs the v1 payload shape (no `version` key in the signed JSON) and verifies — same path as today.

## Required proof (load-bearing tests)

The reviewer said each invariant must be load-bearing: removing it must turn a test red.

In `crates/gitlawb-node/src/cert.rs`:

- `v1_payload_matches_frozen_canonical_form` — pins the v1 payload shape byte-for-byte. A regression that adds a `version` key to the v1 signed JSON breaks this test and is forbidden.
- `v1_ref_certificate_structure_is_well_formed` — pins the v1 RefCertificate round-trip. The reconstructed payload must be byte-identical to the signed one.
- `v1_signing_payload_has_no_version_key` — pins that the v1 signed payload is the pre-versioning 7-field shape with no `version` key.
- `v2_signing_payload_binds_version_and_resists_downgrade` — pins the v2 forward-compat shape and the signature downgrade.
- `issuer_stamps_v1_over_v1_payload` — drives the live `issue_ref_certificate` path and asserts the cert it issues agrees with the v1 payload it signed. A regression that stamps a different `version` over the v1 payload (or vice versa) breaks this test.

In `crates/gitlawb-node/src/db/mod.rs`:

- `v37_ref_certificates_version_applies_on_upgrade` — pins the v37 migration: the new `version` column is added with `DEFAULT 1 NOT NULL`, and existing rows read as v1 with no backfill.

In `crates/gl/src/cert.rs`:

- `payload_serialization_matches_frozen_canonical_form` — gl's payload reconstruction is byte-identical to the node's. A feature flag that flips the JSON serializer's key order breaks every existing cert; this test catches it.
- `missing_version_defaults_to_1_and_verifies` — forward-compat: a v1 cert from an old server (no `version` field) is read by a new client and verified.
- `explicit_version_1_takes_v1_path` — explicit `version: 1` is the v1 verify path.
- `parse_cert_version_truth_table` — pins the parser that distinguishes a missing-key (legacy v1) from a present-but-malformed value. Collapsing `null`, a string, a float, or a u32 overflow onto v1 would let a server advertise "version 2" while the client runs the v1 signature path; this test catches every collapsing shape.
- `cmd_show_verify_rejects_v2_even_with_valid_v1_signature` — drives `cmd_show` end to end with a v1-signed cert that declares `version: 2`. The command must refuse with the version named, not silently verify.
- `cmd_show_verify_without_expect_node_refuses_self_asserted_anchor` — drives `cmd_show` with `--verify` but no `--expect-node`. A node that returns its own self-reported DID must not be laundered into a trust anchor; this test pins the refusal (and asserts both mocks were hit so a regression that skips the `/` fetch is caught).
- `cmd_list_surfaces_gated_404_as_err` — a gated 404 on the cert list endpoint must surface as an Err, not as "No ref certificates …" with exit 0. Pins the round-4 `error_for_status()` gate.

## Overlap with open PRs (declared per the reviewer's instruction)

- **#134** (anchors auth) — independent. The cert API is unchanged in shape; auth gates apply as before.
- **#285** (advisory-lock session affinity) — independent.
- **#306** (Content-Digest on signed requests) — independent.
- **#314** (small-order Ed25519) — independent. v1 certs verify through the same path, which already enforces the small-order check from #314 on the embedded node DID.
- **#324** (libp2p keypair persistence) — independent.
- **#325** (gossip ref-update auth) — independent. PR 1's signed HTTP push is the durable-intent producer; this PR owns the cert shape. The two PRs read and write the same cert table but do not step on each other.
- **#382** (replication withheld-subtree trees) — independent.

The split-PR siblings (#384, #385, #386, #383) land in their own order; this PR's contract is the cert shape and CLI, and it does not require any of them to land first or last. Migration v37 is reserved for this branch on the current head; check the open-PR list before any push that lands between #384 and #386.

## Safety to land standalone

- It compiles, migrates, runs, and passes its focused tests by itself. No sibling PR required.
- It reads the existing `ref_certificates` table and adds one column. The new column has a `DEFAULT 1`, so no backfill migration is needed.
- It does not change a serialized payload for v1 certs: the signed JSON is byte-for-byte identical to the pre-PR-3 shape. Old clients and old servers both work.
- Migration version 37 is reserved for this branch on the current head.

## Verification

```sh
cargo test -p gitlawb-node --bin gitlawb-node
cargo test -p gl --bin gl cert
cargo fmt --all -- --check
cargo clippy -p gl --bin gl --tests -- -D warnings
```

The full focused suites (the tests named under **Required proof** above) pass on this head. Total counts in `gitlawb-node` and `gl` move with unrelated merges; check the CI run rather than the PR description for the live totals.

## Summary by CodeRabbit

- **New Features**
  - Certificate API responses now include certificate version information.
  - Certificates retain version 1 compatibility and support version-aware signing formats.

- **Bug Fixes**
  - Certificate listing now reports HTTP errors instead of treating failed responses as empty results.
  - Verification failures are reported as inability to verify when version validation prevents verification.

- **Documentation**
  - Updated the certificate verification example to include the repository argument.

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,7 +311,7 @@ metadata        local disk / optional S3
 | DID | A user, agent, or node identity derived from an Ed25519 public key. |
 | HTTP Signature | RFC 9421 signature proving control of the DID key for write requests. |
 | Ref certificate | Signed record of a ref update. Useful for audit and replication. |
-| Ref certificate versioning | The wire format is versioned; v2 certs (the current issuance) require `gl cert show <id> --verify --expect-node <did>`, and the queried node's self-reported DID is **not** accepted as a trust anchor. The v1 path remains the only signature shape currently produced; the v2 stamp is the issuer's forward-compat claim. |
+| Ref certificate versioning | The wire format is versioned; new certificates are currently issued as v1. `gl cert show <id> --verify` requires `--expect-node <did>`, and the queried node's self-reported DID is **not** accepted as a trust anchor. v2 is reserved for a future payload shape that carries the version inside the signed bytes; this client refuses to verify v2+ rather than guessing. |
 | UCAN | Delegation token for future capability-based workflows. |
 | Peer announce | Node-to-node HTTP announcement of DID + public URL. |
 | Gossipsub | libp2p topic for ref-update events. |
--- a/README.md
+++ b/README.md
@@ -311,6 +311,7 @@ metadata        local disk / optional S3
 | DID | A user, agent, or node identity derived from an Ed25519 public key. |
 | HTTP Signature | RFC 9421 signature proving control of the DID key for write requests. |
 | Ref certificate | Signed record of a ref update. Useful for audit and replication. |
+| Ref certificate versioning | The wire format is versioned; v2 certs (the current issuance) require `gl cert show <id> --verify --expect-node <did>`, and the queried node's self-reported DID is **not** accepted as a trust anchor. The v1 path remains the only signature shape currently produced; the v2 stamp is the issuer's forward-compat claim. |
 | UCAN | Delegation token for future capability-based workflows. |
 | Peer announce | Node-to-node HTTP announcement of DID + public URL. |
 | Gossipsub | libp2p topic for ref-update events. |
--- a/README.md
+++ b/README.md
@@ -311,7 +311,7 @@ metadata        local disk / optional S3
 | DID | A user, agent, or node identity derived from an Ed25519 public key. |
 | HTTP Signature | RFC 9421 signature proving control of the DID key for write requests. |
 | Ref certificate | Signed record of a ref update. Useful for audit and replication. |
-| Ref certificate versioning | The wire format is versioned; new certificates are currently issued as v1. `gl cert show <id> --verify` requires `--expect-node <did>`, and the queried node's self-reported DID is **not** accepted as a trust anchor. v2 is reserved for a future payload shape that carries the version inside the signed bytes; this client refuses to verify v2+ rather than guessing. |
+| Ref certificate versioning | The wire format is versioned; new certificates are currently issued as v1. `gl cert show <repo> <id> --verify` requires `--expect-node <did>`, and the queried node's self-reported DID is **not** accepted as a trust anchor. v2 is reserved for a future payload shape that carries the version inside the signed bytes; this client refuses to verify v2+ rather than guessing. |
 | UCAN | Delegation token for future capability-based workflows. |
 | Peer announce | Node-to-node HTTP announcement of DID + public URL. |
 | Gossipsub | libp2p topic for ref-update events. |
--- a/crates/gitlawb-node/src/api/certs.rs
+++ b/crates/gitlawb-node/src/api/certs.rs
@@ -52,6 +52,7 @@ pub async fn list_certs(
                 "node_did":   c.node_did,
                 "signature":  c.signature,
                 "issued_at":  c.issued_at,
+                "version":    c.version,
             })
         })
         .collect();
@@ -92,5 +93,6 @@ pub async fn get_cert(
         "node_did":   cert.node_did,
         "signature":  cert.signature,
         "issued_at":  cert.issued_at,
+        "version":    cert.version,
     })))
 }
--- a/crates/gitlawb-node/src/api/events.rs
+++ b/crates/gitlawb-node/src/api/events.rs
@@ -437,6 +437,7 @@ mod ref_updates_feed_tests {
             node_did: "did:key:z6MkNode".into(),
             signature: "sig".into(),
             issued_at: Utc::now().to_rfc3339(),
+            version: 1,
         }
     }
 
--- a/crates/gitlawb-node/src/cert.rs
+++ b/crates/gitlawb-node/src/cert.rs
@@ -11,6 +11,38 @@ use uuid::Uuid;
 use crate::db::RefCertificate;
 use crate::state::AppState;
 
+/// Build the v1 signing payload — the EXACT bytes the node signs and
+/// gl's `verify_signature` reconstructs. This is the single source
+/// of truth shared by `issue_ref_certificate` and the frozen
+/// canonical-form tests below; drift here is what would break every
+/// existing cert.
+///
+/// Field order and key set are fixed: a default `serde_json::Value`
+/// serializes `Map<String, Value>` with sorted keys, so any change to
+/// the literal is observable as a different byte sequence on the
+/// wire. Adding a key (in particular `version`) for v2+ will break
+/// the v1 verify path, which is exactly the contract the version
+/// field exists to enforce.
+pub(crate) fn v1_signing_payload(
+    repo_id: &str,
+    ref_name: &str,
+    old_sha: &str,
+    new_sha: &str,
+    pusher_did: &str,
+    node_did: &str,
+    issued_at: &str,
+) -> serde_json::Value {
+    serde_json::json!({
+        "repo_id": repo_id,
+        "ref":     ref_name,
+        "old":     old_sha,
+        "new":     new_sha,
+        "pusher":  pusher_did,
+        "node":    node_did,
+        "ts":      issued_at,
+    })
+}
+
 /// Issue a signed ref-update certificate for a successful push.
 ///
 /// Builds a canonical JSON payload, signs it with the node's Ed25519 key,
@@ -26,16 +58,18 @@ pub async fn issue_ref_certificate(
     let node_did = state.node_did.to_string();
     let issued_at = Utc::now().to_rfc3339();
 
-    // Build the canonical signing payload.
-    let payload = serde_json::json!({
-        "repo_id": repo_id,
-        "ref":     ref_name,
-        "old":     old_sha,
-        "new":     new_sha,
-        "pusher":  pusher_did,
-        "node":    node_did,
-        "ts":      issued_at,
-    });
+    // Build the canonical signing payload via the shared builder so
+    // the frozen-vector test below and the live signer cannot drift
+    // apart — a regression in either side fails both.
+    let payload = v1_signing_payload(
+        repo_id,
+        ref_name,
+        old_sha,
+        new_sha,
+        pusher_did,
+        &node_did,
+        &issued_at,
+    );
     let payload_bytes = serde_json::to_vec(&payload)?;
 
     let signature = state.node_keypair.sign_b64(&payload_bytes);
@@ -76,6 +110,13 @@ mod v1_payload_tests {
     //! determinism per the v1 version, and the v2 read forward
     //! compat (an unknown version refuses to verify rather than
     //! guessing).
+    //!
+    //! Reviewer 1 finding: the live signer (`issue_ref_certificate`)
+    //! and the frozen-vector test (`v1_payload_matches_frozen_canonical_form`)
+    //! previously held two separate `serde_json::json!` literals,
+    //! so a regression in the signer could not fail this test.
+    //! Both now go through the shared `v1_signing_payload` builder.
+    use super::v1_signing_payload;
     use crate::db::RefCertificate;
     use gitlawb_core::identity::Keypair;
 
@@ -84,17 +125,21 @@ mod v1_payload_tests {
     /// includes a `version` key) is what the versioned format
     /// forbids: an old client reading a v1 cert must not see a
     /// `version` key in the signed JSON.
+    ///
+    /// The literal is built via the shared `v1_signing_payload`
+    /// helper that `issue_ref_certificate` also uses, so the live
+    /// signer and the frozen vector share one source of truth.
     #[test]
     fn v1_payload_matches_frozen_canonical_form() {
-        let payload = serde_json::json!({
-            "repo_id": "repo-1",
-            "ref":     "refs/heads/main",
-            "old":     "0".repeat(40),
-            "new":     "a".repeat(40),
-            "pusher":  "did:key:z6MkPusher",
-            "node":    "did:key:z6MkNode",
-            "ts":      "2026-07-22T00:00:00+00:00",
-        });
+        let payload = v1_signing_payload(
+            "repo-1",
+            "refs/heads/main",
+            &"0".repeat(40),
+            &"a".repeat(40),
+            "did:key:z6MkPusher",
+            "did:key:z6MkNode",
+            "2026-07-22T00:00:00+00:00",
+        );
         let frozen = format!(
             r#"{{"new":"{}","node":"did:key:z6MkNode","old":"{}","pusher":"did:key:z6MkPusher","ref":"refs/heads/main","repo_id":"repo-1","ts":"2026-07-22T00:00:00+00:00"}}"#,
             "a".repeat(40),
@@ -106,19 +151,23 @@ mod v1_payload_tests {
     /// Round-trip: sign the v1 payload with a keypair, build a
     /// RefCertificate with version: 1, and verify the structure.
     /// This is the construction the live code does on every push.
+    ///
+    /// Uses the shared builder for both halves so a regression in
+    /// the live signer (adding/removing/renaming a field) cannot
+    /// pass this test by accident.
     #[test]
     fn v1_ref_certificate_structure_is_well_formed() {
         let kp = Keypair::generate();
         let node_did = kp.did().as_str().to_string();
-        let payload = serde_json::json!({
-            "repo_id": "repo-1",
-            "ref":     "refs/heads/main",
-            "old":     "0".repeat(40),
-            "new":     "a".repeat(40),
-            "pusher":  "did:key:z6MkPusher",
-            "node":    node_did,
-            "ts":      "2026-07-22T00:00:00+00:00",
-        });
+        let payload = v1_signing_payload(
+            "repo-1",
+            "refs/heads/main",
+            &"0".repeat(40),
+            &"a".repeat(40),
+            "did:key:z6MkPusher",
+            &node_did,
+            "2026-07-22T00:00:00+00:00",
+        );
         let sig = kp.sign_b64(&serde_json::to_vec(&payload).unwrap());
         let cert = RefCertificate {
             id: "cert-id".into(),
@@ -136,15 +185,15 @@ mod v1_payload_tests {
         // The signed payload reconstructs identically: gl's
         // verify_signature would build the same JSON, hash the
         // same bytes, and verify the same signature.
-        let reconstructed = serde_json::json!({
-            "repo_id": cert.repo_id,
-            "ref":     cert.ref_name,
-            "old":     cert.old_sha,
-            "new":     cert.new_sha,
-            "pusher":  cert.pusher_did,
-            "node":    cert.node_did,
-            "ts":      cert.issued_at,
-        });
+        let reconstructed = v1_signing_payload(
+            &cert.repo_id,
+            &cert.ref_name,
+            &cert.old_sha,
+            &cert.new_sha,
+            &cert.pusher_did,
+            &cert.node_did,
+            &cert.issued_at,
+        );
         let reconstructed_bytes = serde_json::to_vec(&reconstructed).unwrap();
         let payload_bytes = serde_json::to_vec(&payload).unwrap();
         assert_eq!(
@@ -152,6 +201,41 @@ mod v1_payload_tests {
             "the round-trip serialization must be byte-identical; \
              this is what makes gl's verify_signature succeed"
         );
+    }
+
+    /// Reviewer 1 finding: the live signer must share the
+    /// canonical-form literal so a regression in either side
+    /// fails both. This test pins the shared builder's output
+    /// against the literal a future v2 byte stream would have
+    /// to break — if someone adds a `version` key to
+    /// `v1_signing_payload`, every existing cert breaks verify,
+    /// and this test is one of the canaries that fires.
+    #[test]
+    fn v1_signing_payload_has_no_version_key() {
+        let payload = v1_signing_payload(
+            "repo-1",
+            "refs/heads/main",
+            "oldsha",
+            "newsha",
+            "did:key:z6MkPusher",
+            "did:key:z6MkNode",
+            "2026-07-22T00:00:00+00:00",
+        );
+        let obj = payload.as_object().expect("payload is a JSON object");
+        assert!(
+            !obj.contains_key("version"),
+            "v1 signing payload must not carry a `version` key — adding \
+             one changes the bytes and breaks every existing cert's \
+             signature. v2 certs build a different payload, not this one"
+        );
+        // Pin the exact set so a future field addition is caught.
+        let expected: std::collections::BTreeSet<&str> = [
+            "new", "node", "old", "pusher", "ref", "repo_id", "ts",
+        ]
+        .into_iter()
+        .collect();
+        let actual: std::collections::BTreeSet<&str> = obj.keys().map(String::as_str).collect();
+        assert_eq!(actual, expected, "v1 payload key set is frozen");
     }
 
     /// The v1 RefCertificate shape with version: 2 is a
--- a/crates/gitlawb-node/src/cert.rs
+++ b/crates/gitlawb-node/src/cert.rs
@@ -20,9 +20,13 @@ use crate::state::AppState;
 /// Field order and key set are fixed: a default `serde_json::Value`
 /// serializes `Map<String, Value>` with sorted keys, so any change to
 /// the literal is observable as a different byte sequence on the
-/// wire. Adding a key (in particular `version`) for v2+ will break
-/// the v1 verify path, which is exactly the contract the version
-/// field exists to enforce.
+/// wire.
+///
+/// v1 is the unversioned legacy shape: no `version` key in the signed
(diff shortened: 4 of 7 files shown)


beardthelion (dismissed):
The core design holds: v1 signed JSON stays the 7-field canonical form with no `version` key in the bytes the node signs, the DB column defaults to 1 for existing rows, and `gl cert show` takes the v1 verify path when `version` is missing or 1 while refusing v2+ instead of guessing. I ran the cert test modules locally (`cargo test -p gl cert`, `cargo test -p gitlawb-node v1_payload`) and confirmed CI is green on this head.

This sits in the #26 split stack (sibling to #384/#385); rebasing onto current main will likely conflict on `db/mod.rs` and `cert.rs`.

## Findings

- [P2] Drop the duplicate version ALTER from the applied v1 migration block
  `crates/gitlawb-node/src/db/mod.rs:668`
  The same `ALTER TABLE ref_certificates ADD COLUMN ... version` appears inside the v1 bundle and again in v28. Upgraded nodes never re-run v1, so only v28 matters for production; the v1 copy is dead code and violates append-only migration discipline. Keep v28 only.

- [P2] Add a v28 upgrade-path test matching the v25/v26 pattern
  `crates/gitlawb-node/src/db/mod.rs:7200`
  `ref_certificate_tests` has `v25_repos_created_at_id_index_applies_on_upgrade` and `v26_discovery_continuation_applies_on_upgrade` but nothing for `ref_certificates_version`. Seed a pre-v28 `ref_certificates` table without `version`, run migrations through v27, and assert the column exists with DEFAULT 1.

- [P2] Make the gl forward-compat tests actually call verify
  `crates/gl/src/cert.rs:431`
  `missing_version_defaults_to_1_and_verifies` only parses `version` defaulting to 1; it never calls `verify_signature`. `explicit_version_1_takes_v1_path` only parses JSON. The PR body lists both as load-bearing proof that old-server/new-client verification works; rename or extend them so a well-signed v1 cert without a `version` field round-trips through the v1 verify path.

- [P2] Pin the v2+ rejection branch with a regression test
  `crates/gl/src/cert.rs:191`
  `cmd_show` returns Err for `version > 1`, and the inline comment asks for a test that this branch returns Err, not Ok. No test covers `version: 2` through the verdict path. Add one.

- [P2] Wire the v1 payload premise to the production signer
  `crates/gitlawb-node/src/cert.rs:30`
  I gutted `issue_ref_certificate` by adding `"version": 1` to the signing payload; `cargo test -p gitlawb-node v1_payload` stayed 3/3 green because `v1_payload_matches_frozen_canonical_form` tests a standalone literal, not the live signer. Either exercise `issue_ref_certificate` in a test or share one canonical builder so a regression in the signer cannot slip past a decoupled frozen vector.

One process note, not a finding: rebasing onto main will likely conflict with #384 on migration ordering; v27 is already taken by split PR 1 per the PR body.

Not an ask, recorded only: the repo events `local_cert` feed omits `version` while the certs API includes it (`api/events.rs:243`). Fine for v1-only rollout, but consumers of the events stream will not see the field until a follow-up adds it.

jatmn (changes requested):

## Findings

- [P2] Keep the version column out of migration v1
  `crates/gitlawb-node/src/db/mod.rs:668`
  This adds the column to the already-applied v1 migration and then adds the same DDL in v28. The migration runner skips every recorded version, so an existing deployment never executes the new v1 statement; a fresh deployment does execute it and then makes v28 a no-op. Both paths happen to reach the same schema today because of `IF NOT EXISTS`, but they no longer share one immutable migration history—the exact condition the migration catalogue and contributor rules forbid.

  The root cause is treating the bootstrap schema bundle as a place to keep the current schema complete. It is historical input once any deployment can record it. Remove the v1 addition and keep the column definition exclusively in v28. Then make the protection load-bearing with an upgrade test that marks the pre-v28 migrations applied, runs the real migration runner, and proves both the `version` column and its default of `1` are present for an existing `ref_certificates` row.

- [P2] Reject malformed explicit certificate versions instead of treating them as v1
  `crates/gl/src/cert.rs:164`
  The fallback conflates a genuinely absent field from an old server with an explicitly malformed field: `null`, a string, a float, or another non-integer all become `1`. The narrowing cast is lossy too—`4294967297` becomes `1`. In either case the client prints “Version: 1” and runs the v1 signature verifier. With `--verify`, an otherwise valid v1 signature and trusted node DID can therefore yield a successful command even though the response explicitly declared a format this client cannot represent. That contradicts the new promise to refuse unknown versions rather than guessing their payload shape.

  The root cause is using one `Option` fallback for two semantically different states (missing and invalid), followed by an unchecked integer narrowing. Parse the field so that only a missing key selects legacy v1; a present value must be a losslessly representable integer and exactly a version this client supports, otherwise produce the existing unsupported-version verification failure. Add cases for a missing field, explicit `1`, explicit `2`, and an overflow/non-numeric value so the distinction cannot regress.

beardthelion (dismissed):
CI is green again on b44f740a and I confirmed both fixes locally: `cargo fmt --check` and `cargo clippy -p gl --bin gl --tests -- -D warnings` are clean, and `v28_ref_certificates_version_applies_on_upgrade` passes now that the helper uses `fetch_optional` and flattens.

Five of the six round-1 asks landed and landed well. The duplicate `ALTER` is gone from the v1 bundle, v28 is the only place the column is defined, `v1_signing_payload` is now shared between the live signer and the frozen vector, and `parse_cert_version` is genuinely sound: I drove missing, `null`, `"1"`, `true`, `[1]`, `{}`, `0`, `-1`, `2`, `1.0`, `1e0`, `4294967295/6/7` and `u64::MAX` through it, and only a missing key and integer `1` reach the v1 path. The upsert threads `version` through the same `issued_at` predicate as every other column, so there is no downgrade hole there either.

What is left is that the two guards this round added do not guard anything, and I can show it rather than argue it. On a throwaway checkout of b44f740a I made two changes to production code: `cmd_show`'s v2 arm became `Ok(_v) => Ok(())`, silently accepting a v2 certificate, and the issuer anchor became `.or(Some(node_did))`, pinning the certificate to its own self-asserted DID. All six gl cert tests stayed green through both.

## Findings

- [P2] Drive the v2 rejection through `cmd_show`, and delete the arm it cannot reach
  `crates/gl/src/cert.rs:679`
  `verdict_branch_rejects_v2_even_with_valid_v1_signature` copies `cmd_show`'s three-arm match into its own body instead of calling `cmd_show`, so it cannot see the real match drift. It is also pointed at an arm that cannot execute: `parse_cert_version` returns `Ok` only from `None => Ok(1)` and the final `Ok(1)`, so a `version: 2` response lands on `Err(reason)`, not `Ok(v)`. You can see it in the runtime message, which comes back double-wrapped as `cert declared a version this client cannot represent (this client supports cert version 1 only; server returned 2)`. Drop the `Ok(v)` arm and test the command.

- [P2] Anchor `--verify` to something the queried node cannot assert
  `crates/gl/src/cert.rs:262`
  With no `--expect-node`, the anchor falls back to the DID the node reports at `/`, which is the same node that served the certificate. I minted a keypair, self-signed a v1 cert with it, served that cert and a matching `/` DID, and `gl cert show --verify` printed `VALID`, printed `Issuing node DID matches the node being queried`, and exited 0 on a wholly forged certificate. Everything else in that block fails closed correctly: a different `/` DID exits 1, a 500 exits 1, a missing `did` key exits 1, a mismatched `--expect-node` exits 1. So the fix is narrow. Either require `--expect-node` before claiming a trusted issuer, or stop calling the fallback an anchor; `--expect-node`'s doc comment at line 48 describes it as "a DID you trust", which the queried node's self-report is not.

- [P2] Give the version path a test that fails when the version is wrong
  `crates/gitlawb-node/src/cert.rs:241`
  `v1_is_the_default_version` builds a `RefCertificate` literal with `version: 1` and asserts `cert.version == 1`, which holds no matter what the signer does. In `v1_ref_certificate_structure_is_well_formed`, `payload` and `reconstructed` both come from `v1_signing_payload` with argument-identical inputs, so the byte comparison is true for any builder body including one that returns `{}`; `sig` is computed and never verified, though the comment says otherwise. Verifying `sig` against `reconstructed_bytes` would fix the second one. Nothing else covers the production path either: no test reaches the `version: 1` literal in the signer, the `bind(cert.version as i32)`, or the `"version"` key in either certs-API response, and `gl` refuses to verify anything but 1, so a regression there breaks verification quietly.

`v1_payload_matches_frozen_canonical_form` and `v1_signing_payload_has_no_version_key` are the two that are load-bearing by construction, and now that they sit on the same builder the signer uses, the round-1 drift gap is genuinely closed.

Two smaller things, not asks. The bail under `--verify` is hardcoded `certificate signature did not verify: {reason}`, but on a version rejection the signature was never checked, which will send someone hunting a key mismatch. And the certs API now emits `version` while the repo events feed does not; that feed already omits `signature` so it was never a verification surface, but the two views of the same row now disagree.

Rebasing onto current main will conflict on `db/mod.rs` and `cert.rs` with the sibling splits. That is mechanical, not another review round.

jatmn (approved):
lgtm

beardthelion (dismissed):
Round 2 closed both escapes I demonstrated last round, and I confirmed that by mutation rather than by reading: gutting `parse_cert_version`'s `n != 1` rejection turns three gl tests red and the v2 cert prints "Version: 1 / VALID" again, and replacing the `--expect-node` refusal with a no-op fails `cmd_show_verify_without_expect_node_refuses_self_asserted_anchor`. Both guards drive `cmd_show` end to end now, and the anchor test carries a positive control, so it binds the anchor and not the signature. CI is 12/12 on this head.

## Findings

- [P1] Renumber the migration above every version an open PR already claims
  `crates/gitlawb-node/src/db/mod.rs:1134`
  #384 and #327 both claim 28 as well. The runner selects on the version number alone and `continue`s on anything already recorded in `schema_migrations`, with no name-mismatch abort, so whichever of the three merges second and third has its DDL skipped in silence. On an upgraded node the `version` column then never exists and every statement naming it 500s, with a clean migration log. Sweeping the open PRs today, 36 is the highest claimed (#244), so 37 is the current floor; it is worth re-checking right before merge since that floor moves. The PR body's "PR 1 used 27" is stale.

- [P2] Pin the version the live issuer stamps
  `crates/gitlawb-node/src/cert.rs:87`
  `issued_version_claim_matches_the_v1_payload_it_signs` is a real improvement over the tautology it replaced, but it builds the `RefCertificate` by hand, so it never observes `issue_ref_certificate`. I flipped the stamp at :87 to `version: 2` and all four `cert::` tests stayed green; the function has one production caller and zero test callers. `test_support::test_state(pool)` under `#[sqlx::test]` gives you the `AppState` it needs, or factor the sign-and-stamp sequence into a helper both the issuer and the test call.

- [P2] Document the `--verify` contract change
  `crates/gl/src/cert.rs:285`
  `gl cert show <id> --verify` without `--expect-node` used to verify against the node's self-reported DID and now hard-fails. That is the right call and the help text covers it, but anyone verifying against their own node has a working command break on upgrade and nothing outside the help text says so. CHANGELOG.md is release-please generated, so don't hand-edit it: a `BREAKING CHANGE:` footer on the commit gets it into the release notes, and the docs should carry the new anchor requirement.

- [P3] Assert the root mock is consumed
  `crates/gl/src/cert.rs:813`
  mockito 1.7's `assert_on_drop` is off by default. I repointed that mock at `/never-visited` and the test still passed, so the forger's matching self-report, which is the scenario the test is named for, is not actually required for it to go green. `.expect(1)` plus `.assert_async()` keeps the setup honest.

On the open CodeRabbit thread at `crates/gl/src/cert.rs:207`: `verify_signature` still derives the key from the certificate's own `node_did`, so the letter of that comment is unimplemented, but the `--expect-node` gate closes the escape it describes and I proved that gate load-bearing above. Not carrying it as an ask.

beardthelion (changes requested):
Round 3 closed three of the four round-2 asks, and I confirmed each by mutation rather than by reading. The migration moved to v37, which clears every version an open PR claims today (#244 is next at 36). The root-mock fix is real: repointing it at `/never-visited` now turns the anchor test red, where in round 2 it stayed green. And the stamp is genuinely bound to the live issuer, flipping `cert.rs:93` turns `issuer_stamps_v2_over_v1_payload` red.

The problem is what that binding was achieved by. My round-2 text described flipping the stamp to 2 as the mutation I used to show the old test was vacuous, and it was read as the fix. Production now stamps every new certificate v2, so this head ships a node issuing certificates its own shipped client refuses.

I drove the whole chain rather than composing two green tests. A real `issue_ref_certificate` call under `#[sqlx::test]`, then the certs API as an anonymous caller on a public repo, returns 200 with `"version":2`. Feeding that verbatim response, real signature and real node DID, to `gl cert show --verify` with `--expect-node` supplied so the anchor could not be the failure:

```
certificate signature did not verify: cert declared a version this client cannot
represent (this client supports cert version 1 only; server returned 2); refusing to verify
```

Existing certificates are fine, v37 defaults them to 1. The break is forward-only and hits every push after upgrade.

## Findings

- [P1] Stamp new certificates version 1 until a v2 verifier ships
  `crates/gitlawb-node/src/cert.rs:93`
  Every newly issued cert is stamped 2, and `parse_cert_version` rejects anything but 1 before it ever checks the signature, so the only shipped verifier refuses every fresh certificate. Two green tests hide it because they live in different crates: the node side pins `cert.version == 2`, the gl side pins that 2 is refused, and nothing exercises the composition. Revert the stamp to 1, or land the v2 verify path in the same release. The version test can bind to the live issuer without changing what it stamps.

- [P2] Sign the version, or stop calling it a format version
  `crates/gitlawb-node/src/cert.rs:26`
  The signed payload is byte-identical for v1 and v2, so the version travels outside the signature as an unsigned sibling. Today that is harmless, since flipping the field changes nothing cryptographically. It stops being harmless the moment v2 has its own payload shape, because a downgrade to v1 then verifies cleanly. My call is that the version belongs in the signed bytes from v2 onward, with v1 remaining the unversioned legacy shape.

- [P2] Correct the three places that still say new certs are version 1
  `crates/gitlawb-node/src/db/mod.rs:1141`, `README.md:314`, PR body
  The migration comment says "The current version is 1". The PR body says `issue_ref_certificate` sets `version: 1` on every new cert, and its compat table repeats it; it also still says migration 28 is reserved when the head is v37. The README is the one that will cost someone real time: it tells users v2 certs require `gl cert show <id> --verify --expect-node <did>`, and I ran that exact command shape against a v2 cert with `--expect-node` supplied, and it fails on the version.

- [P3] Verify the signature the well-formed test computes
  `crates/gitlawb-node/src/cert.rs:171`
  Carried over from round 2 and unchanged. The test signs a payload and asserts byte-equality of the reconstruction, but never calls `identity::verify`, while its own comment says gl "would build the same JSON, hash the same bytes, and verify the same signature". Verifying `sig` against the reconstructed bytes closes it.

- [P3] Fix the comment claiming a database read-back
  `crates/gitlawb-node/src/cert.rs:264`
  The comment says the test reads the row back by `(repo_id, ref_name)`. It asserts on the in-process return value and never queries. Either add the read-back, which would also cover the persisted column, or fix the comment.

One more when you touch the P1: the failure message leads with "certificate signature did not verify" when the signature was never checked. I raised it as a minor point in round 2; the v2 stamp turns it into the message on every certificate.

Rebasing will still conflict with the sibling splits on `db/mod.rs` and `cert.rs`. That is mechanical, not another round. The v37 floor moves as other PRs land, so it is worth re-checking right before merge.

beardthelion (changes requested):
Round 3 fixes hold on this head. I re-ran the cert test filters on `8789d556` and confirmed `issue_ref_certificate` stamps v1, v37 is the only version-column DDL, and the gl guard tests pass. Two items remain before merge.

## Findings

- [P2] Route cert list through error_for_status before parsing JSON
  `crates/gl/src/cert.rs:113`
  `cmd_list` still does `get_authed(...).await?.json()` and treats a missing `certificates` key as an empty list, so a gated 404 from `/api/v1/repos/{owner}/{repo}/certs` prints "No ref certificates" with exit 0. `cmd_show` and `resolve_cert_id` in the same file already call `error_for_status()` first. Wire list through the same status check before reading JSON.

- [P3] Align the PR description with migration v37
  `crates/gitlawb-node/src/db/mod.rs:1134`
  The PR body still says "Migration v28" and "Migration version 28 is reserved" while this branch adds version 37 (`ref_certificates_version`) and the upgrade test is `v37_ref_certificates_version_applies_on_upgrade`. The in-tree migration comment is already correct. Update the PR description so the claimed version matches what ships and states which open-branch range v37 clears.

jatmn (changes requested):
I found two documentation consistency issues that need to be addressed before this is ready. The certificate implementation and compatibility tests themselves hold on this head; the remaining work is to make the operator and reviewer-facing claims match what now ships.

## Findings

- [P3] Publish the complete `cert show` verification command
  `README.md:314`
  The new README text gives `gl cert show <id> --verify`, then says `--expect-node <did>` is required. The shipped parser requires two positional arguments—`gl cert show [OPTIONS] <REPO> <ID>`—so the documented form binds `<id>` as the repository and fails because the certificate ID is missing. The same shortened form appears in the `BREAKING CHANGE:` commit footer, which is intended to feed release notes, so correcting only one copy can leave users with the same broken upgrade instruction elsewhere.

  Please make the canonical example `gl cert show <repo> <id> --verify --expect-node <did>` everywhere this PR publishes the breaking contract, and verify the final wording against `gl cert show --help`. The root cause is that the command was copied as prose without preserving the parser's required positional shape. Keep the fix at the documentation/release-note boundary; the CLI arguments and certificate verification behavior do not need redesign for this issue.

- [P3] Reconcile the PR description with the current head
  `PR description: “Migration v28”`
  The description accumulated facts from earlier revisions and no longer describes `8789d556` consistently. It says the new column is migration v28 and later says v28 is reserved, while the code adds `ref_certificates_version` as v37 and tests it with `v37_ref_certificates_version_applies_on_upgrade`. It also names `v1_is_the_default_version`, which no longer exists, reports earlier focused-test counts, and describes sibling-PR issuance/idempotence details that have changed as those branches evolved. Because migration numbers and compatibility proof are exactly what reviewers use to judge whether these split PRs can land independently, stale evidence here obscures the otherwise-correct implementation and invites another review round.

  Please do one source-of-truth pass over the whole description against the current head: use v37 consistently, list only tests that exist now, refresh or remove obsolete counts, and restate sibling overlap in terms that are true of the current branches. The root cause is incremental description edits after migration renumbering and test rewrites. Prefer stable statements of the owned contract over volatile exact test totals or moving open-branch floors; this keeps the description useful without making it stale after the next unrelated merge.

## Close-out guidance

The repeated feedback on this PR is no longer evidence that the certificate design has many independent defects. Most churn came from fixing one layer while leaving a copied claim, test description, command example, or overlap note tied to an older head. Before requesting the next review, freeze the intended head and do one consistency sweep across these paired surfaces: CLI help ↔ README/release wording; `MIGRATIONS` ↔ PR migration claims; current test symbols/results ↔ the “Required proof” section; and current sibling heads ↔ overlap statements. Remove details that are not necessary to establish the compatibility contract if they are likely to move again.

Please keep that close-out narrow. The v1 payload, v37 upgrade behavior, API representation, legacy fallback, unsupported-version refusal, and final `--expect-node` failure behavior passed the current review and focused validation. Changing those working paths while correcting the prose would create new review surface and restart the exact cycle this cleanup is meant to end.

jatmn (approved):
lgtm

beardthelion (approved):
Round 5 on `848ccfa5`. The two incremental commits address the round-4 asks: `cmd_list` now routes through `error_for_status()` before parsing JSON, and the README `cert show` example carries both positionals (`<repo> <id> --verify`). The PR description is reconciled with migration v37 and the current test set.

I verified each fix is load-bearing by mutation. Removing `error_for_status()` from `cmd_list` turns `cmd_list_surfaces_gated_404_as_err` red: the 404 body falls into the empty-list branch and prints "No ref certificates for o/r" with exit 0, which is the exact denial-renders-as-success escape the fix closes. Restoring the call returns the test to green.

I re-ran the prior-round guards on this head, each by flipping the guard and observing red, then restoring and observing green. Gutting the `n != 1` rejection in `parse_cert_version` turns both `parse_cert_version_truth_table` and `cmd_show_verify_rejects_v2_even_with_valid_v1_signature` red (a version-2 cert prints VALID with a v1 signature). Flipping the `version: 1` stamp in `issue_ref_certificate` to `version: 2` turns `issuer_stamps_v1_over_v1_payload` red (left: 2, right: 1). Replacing the `--expect-node` refusal with a no-op turns `cmd_show_verify_without_expect_node_refuses_self_asserted_anchor` red (a forged self-signed cert passes VALID with exit 0). All four guards are live on `848ccfa5`.

Migration v37 is the sole source of `ref_certificates.version`: the v1 bootstrap `CREATE TABLE` carries no `version` column, and no other migration entry touches it. v37 is the highest version in `MIGRATIONS` (1-11, 17-26, 37), and no other open PR claims 37 (the closest is #244 at 36). The upgrade test drops the column and the schema_migrations record, re-runs migrations, and asserts the column defaults to 1 and a legacy row reads back as version 1. All four production SQL queries (insert, list, list_by_prefix, get) include `version`, and the `ON CONFLICT` upsert carries `version` forward only when the incoming cert is newer.

The full suite passes on CI (13/13 green).

Not an ask, recorded only: `RefCertificate.version` is `u32` but the column is Postgres `INTEGER` (signed 32-bit), bound and read through `as i32` / `as u32`. The round-trip is bit-preserving, so values read back correctly, but a raw SQL consumer would see a negative integer for any version above `i32::MAX`. Only version 1 is stamped today and v2 is the next planned shape, so this is latent. Worth switching to `i32` or `BIGINT` if version numbers ever approach that range, but not blocking now.

One process note, not a finding: this branch overlaps mechanically with sibling #384 on `cert.rs` (split 1/4 vs split 3/4). Rebase conflicts are expected when whichever lands first; neither is a reason to hold this review.

beardthelion:
@kevincodex1 LGTM

PR #388, opened 2026-08-30
fix(node): scope the refs listing by the canonical repo slug

### LINKED ISSUE #387: GET /repos/{owner}/{repo}/refs returns an empty 200 when the owner is in full did:key form
`GET /api/v1/repos/{owner}/{repo}/refs` returns an empty list with a 200 when the owner is given in the full `did:key:` form. The same repository under the bare key form returns its refs. The request is authorized either way, so the caller gets a successful response that is missing data rather than an error explaining why.

## Cause

`branch_cids.repo` is written as `{normalize_owner_key(owner_did)}/{name}`. The push path builds that slug from the repo record in `crates/gitlawb-node/src/api/repos.rs` before calling `upsert_branch_cid`.

`list_refs` in the same file builds its filter from the raw URL path segments instead:

```rust
let (_record, _rules) =
    crate::api::authorize_repo_read(&state, &owner, &repo, caller, "/").await?;

let repo_slug = format!("{owner}/{repo}");
let refs = state.db.list_branch_cids(&repo_slug).await?;
```

`get_repo` normalizes the owner before matching, so `did:key:zX` and bare `zX` both resolve to the same repository and both pass the read gate. `list_branch_cids` then does an exact `WHERE repo = $1`, which the full-DID string never matches. The authorization step normalizes and the query step does not.

## Reproduction

Seed a repository owned by `did:key:zX` and record a branch CID under the slug the push path writes (`zX/name`), then request both forms:

```
GET /api/v1/repos/zX/name/refs            -> 200 {"count":1,...}
GET /api/v1/repos/did:key:zX/name/refs    -> 200 {"count":0,"refs":[]}
```

Confirmed against `bfc44f92` with a `#[sqlx::test]` driving both forms through `build_router`.

## Notes

Present since the initial public release. `list_refs` has no test coverage today, which is why it went unnoticed.

The same shape was recently fixed in the Arweave anchors handler, and `api/events.rs` already builds its slug from the authorized record. Fixing this the same way keeps the three consistent.

beardthelion:
Closes #387.

`list_refs` built its `branch_cids` filter from the raw URL path segments while the only production writer stores the canonical `normalize_owner_key(owner_did)/name` slug. `get_repo` normalizes the owner before matching, so `did:key:zX` and bare `zX` resolve to the same repository and both pass the read gate, but only one of them matches the exact `WHERE repo = $1` that follows. The full-DID form came back `200 {"count":0,"refs":[]}`.

The fix binds the record `authorize_repo_read` already returns and builds the filter from it, so the resource that gets queried is the one that was authorized. `api/events.rs` and this file's own push path already do exactly this, so the three now agree.

## Tests

`list_refs` had no coverage at all before this, which is why the bug survived since the initial release. Two tests, each confirmed to fail without the line it guards rather than assumed to:

- `list_refs_resolves_both_owner_did_forms_to_the_canonical_slug` drives both owner forms through `build_router` and asserts they return the same refs, with a second repository whose refs must not appear. Restoring the raw path slug turns the positive half red; widening the query to match any owner turns the negative half red.
- `list_refs_denies_an_anonymous_caller_on_a_private_repo` covers the deny direction. Replacing `authorize_repo_read` with a bare `get_repo` lookup turns it red on its own message. Without this test the first one passes with the read gate removed entirely, since every repository it seeds is public.

Full suite 1090 passed, 0 failed. `cargo clippy -D warnings` clean, `cargo fmt --check` clean, and `cargo +1.91 check --all-targets --locked` clean against the pinned MSRV.

## Scope

Only `list_refs` changes. The raw `format!("{owner}/{repo}")` shape appears in plenty of other handlers, but everywhere else it feeds a `RepoNotFound` message or a JSON echo field. As a database query key it occurred once, and this is it.

Reader and writer now hold the slug formula in two matching expressions kept in step by a comment. A shared helper would make that hold by construction, which is worth doing across every slug-keyed reader at once rather than piecemeal here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Repository branch references now resolve consistently whether the owner is provided as a full or shortened DID.
  * Prevented references from being returned for the wrong repository owner.
  * Private repositories no longer reveal branch information or pinned content identifiers to anonymous requests.

* **Tests**
  * Added coverage for equivalent owner formats, cross-owner isolation, and private repository access controls.

### DIFF
--- a/crates/gitlawb-node/src/api/repos.rs
+++ b/crates/gitlawb-node/src/api/repos.rs
@@ -2886,10 +2886,21 @@ pub async fn list_refs(
     auth: Option<Extension<AuthenticatedDid>>,
 ) -> Result<Json<serde_json::Value>> {
     let caller = auth.as_ref().map(|e| e.0 .0.as_str());
-    let (_record, _rules) =
+    let (record, _rules) =
         crate::api::authorize_repo_read(&state, &owner, &repo, caller, "/").await?;
 
-    let repo_slug = format!("{owner}/{repo}");
+    // Build the filter from the canonical stored slug, NOT from the raw path
+    // segments. `branch_cids.repo` is written as
+    // `{normalize_owner_key(owner_did)}/{name}` (the push path builds the same
+    // slug before calling `upsert_branch_cid`), and
+    // `get_repo` resolves both `did:key:zX` and bare `zX` to the same record, so
+    // a request in the full-DID form authorizes and then matches zero rows,
+    // returning an empty page with a 200.
+    let repo_slug = format!(
+        "{}/{}",
+        crate::db::normalize_owner_key(&record.owner_did),
+        record.name
+    );
     let refs = state.db.list_branch_cids(&repo_slug).await?;
 
     Ok(Json(
--- a/crates/gitlawb-node/src/test_support.rs
+++ b/crates/gitlawb-node/src/test_support.rs
@@ -206,6 +206,136 @@ mod tests {
         }
     }
 
+    /// `list_refs` must resolve both owner-DID forms to the SAME refs.
+    /// `branch_cids.repo` is written as `{normalize_owner_key(owner_did)}/{name}`
+    /// (`api/repos.rs:2643` -> `:2754`), and `get_repo` resolves `did:key:zX` and
+    /// bare `zX` to one repo, so both path forms authorize. Reading the table with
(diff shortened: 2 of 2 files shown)


jatmn (approved):
LGTM

beardthelion:
@kevincodex1 LGTM

PR #390, opened 2026-09-01
fix(node): canonicalize and 404-on-no-match in remove_label (#344)

### LINKED ISSUE #344: remove_label skips the canonicalization add_label applies, and reports removed:true whether or not anything was deleted
`add_label` (`api/labels.rs:24`) canonicalizes before storing:

```rust
let label = req.label.trim().to_lowercase();
```

`remove_label` (`api/labels.rs:56-70`) passes the raw path segment straight through:

```rust
Path((owner, name, label)): Path<(String, String, String)>,
...
state.db.remove_label(&record.id, &label).await?;
Ok(Json(serde_json::json!({ "label": label, "removed": true })))
```

So creating `Bug` stores `bug`, and `DELETE /api/v1/repos/{owner}/{repo}/labels/Bug` matches nothing.

The second half is the part worth fixing carefully. `db::remove_label` (`db/mod.rs:1928-1935`) is
`Result<()>` and discards the row count:

```rust
sqlx::query("DELETE FROM repo_labels WHERE repo_id = $1 AND label = $2")
    .bind(repo_id).bind(label).execute(&self.pool).await?;
Ok(())
```

The handler then returns 200 with `{"removed": true}` unconditionally. A delete that removed nothing
is indistinguishable from one that worked, which is the no-op-rendered-as-success shape we have been
trying to stamp out on the client side.

Both handlers are correctly owner-gated (`require_repo_owner`), so this is not an authorization issue.

## Fix direction

Extract the trim-and-lowercase plus the charset validation into a shared helper in `api/labels.rs` and
call it from both handlers. Have `db::remove_label` return the affected row count, and let the handler
report `removed` from it (or 404 when nothing matched).

Verified by reading both handlers and the query at `origin/main` 50d3cbbe. Not driven end to end.

Ayush7614:
Closes #344.

## What

`add_label` (`api/labels.rs:24`) stored labels trimmed and lowercased, but `remove_label` passed the path segment through as-is, so a label stored as `bug` did not match `DELETE /repos/{owner}/{repo}/labels/Bug`.

Worse, `db::remove_label` discarded the row count and the handler always returned 200 + `{"removed": true}`, including when the delete matched nothing — the no-op-rendered-as-success shape we have been trying to stamp out on the client side.

## Fix

- Move trim + lowercase and the charset/length validation into a shared `canonicalize_label` helper used by both handlers.
- `db::remove_label` returns the affected row count (`Result<bool>`) so the handler can return 404 when no label was deleted.
- `add_label` semantics are unchanged (still idempotent; re-adding an existing label returns 200 with `added: false`).

The only externally observable change for clients that were already calling `remove_label` with the canonical form is that one matching nothing now returns 404 instead of `removed: true`.

## Why not break authz

Both handlers are owner-gated (`require_repo_owner`). A non-owner with a real signed request against an existing label still gets 403 — the owner-gated write keeps the existence-leak shape separate from the not-found shape, consistent with `add_label` and the gate rules in `AGENTS.md`. Tested below.

## Tests

Two new tests in `crates/gitlawb-node/src/test_support.rs`, both passing locally against `cargo test -p gitlawb-node --tests label`:

- `remove_label_canonicalizes_and_404s_on_no_match` — mixed-case path segment matches the stored label; trimmed path segment matches; empty-after-trim returns 400; disallowed charset returns 400; a second delete of an already-deleted label returns 404, not `removed: true`.
- `remove_label_denies_non_owner` — non-owner gets 403, the owner-gated mutation shape; the label is still there afterwards.

## Verification

```
cargo fmt --all -- --check     # clean
cargo check -p gitlawb-node --tests --all-targets   # OK
cargo test -p gitlawb-node --tests label            # 5 passed; 0 failed
```

(Ignored the unrelated pre-existing `clippy::duplicated_attributes` at `crates/gitlawb-node/src/api/ipfs.rs:2169` — present on `upstream/main` before this PR.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Label handling now consistently trims whitespace and normalizes capitalization.
  * Invalid labels—including empty, overlong, or incorrectly formatted values—are rejected with clear validation errors.
  * Removing a label that does not exist now returns a not-found response instead of succeeding silently.
  * Unauthorized label removal continues to be rejected, preserving existing labels.
  * Label removal now behaves consistently regardless of label capitalization or surrounding whitespace.

### DIFF
--- a/crates/gitlawb-node/src/api/labels.rs
+++ b/crates/gitlawb-node/src/api/labels.rs
@@ -14,14 +14,13 @@ pub struct LabelRequest {
     pub label: String,
 }
 
-/// POST /api/v1/repos/:owner/:repo/labels
-pub async fn add_label(
-    State(state): State<AppState>,
-    Extension(auth): Extension<AuthenticatedDid>,
-    Path((owner, name)): Path<(String, String)>,
-    Json(req): Json<LabelRequest>,
-) -> Result<(StatusCode, Json<serde_json::Value>)> {
-    let label = req.label.trim().to_lowercase();
+/// Canonicalize and validate a label string. `add_label` and `remove_label`
+/// must agree on the form stored in the DB: `add_label` already trims and
+/// lowercases before storing, so `remove_label` has to do the same to find
+/// the row. Validation is shared so neither handler can drift on charset or
+/// length (#344).
+fn canonicalize_label(raw: &str) -> std::result::Result<String, AppError> {
+    let label = raw.trim().to_lowercase();
     if label.is_empty() || label.len() > 50 {
         return Err(AppError::BadRequest("label must be 1–50 characters".into()));
     }
@@ -33,6 +32,17 @@ pub async fn add_label(
             "label must contain only alphanumeric characters, hyphens, and colons".into(),
         ));
     }
+    Ok(label)
+}
+
+/// POST /api/v1/repos/:owner/:repo/labels
+pub async fn add_label(
+    State(state): State<AppState>,
+    Extension(auth): Extension<AuthenticatedDid>,
+    Path((owner, name)): Path<(String, String)>,
+    Json(req): Json<LabelRequest>,
+) -> Result<(StatusCode, Json<serde_json::Value>)> {
+    let label = canonicalize_label(&req.label)?;
 
     let record = state
         .db
@@ -59,6 +69,8 @@ pub async fn remove_label(
     Extension(auth): Extension<AuthenticatedDid>,
     Path((owner, name, label)): Path<(String, String, String)>,
 ) -> Result<Json<serde_json::Value>> {
+    let label = canonicalize_label(&label)?;
+
     let record = state
         .db
         .get_repo(&owner, &name)
@@ -66,7 +78,10 @@ pub async fn remove_label(
         .ok_or_else(|| AppError::RepoNotFound(format!("{owner}/{name}")))?;
     crate::api::require_repo_owner(&record, &auth.0)?;
 
-    state.db.remove_label(&record.id, &label).await?;
+    let removed = state.db.remove_label(&record.id, &label).await?;
+    if !removed {
+        return Err(AppError::NotFound(format!("label '{label}'")));
+    }
     Ok(Json(serde_json::json!({ "label": label, "removed": true })))
 }
 
--- a/crates/gitlawb-node/src/db/mod.rs
+++ b/crates/gitlawb-node/src/db/mod.rs
@@ -2216,13 +2216,13 @@ impl Db {
         Ok(result.rows_affected() > 0)
     }
 
-    pub async fn remove_label(&self, repo_id: &str, label: &str) -> Result<()> {
-        sqlx::query("DELETE FROM repo_labels WHERE repo_id = $1 AND label = $2")
+    pub async fn remove_label(&self, repo_id: &str, label: &str) -> Result<bool> {
+        let result = sqlx::query("DELETE FROM repo_labels WHERE repo_id = $1 AND label = $2")
             .bind(repo_id)
             .bind(label)
             .execute(&self.pool)
             .await?;
-        Ok(())
+        Ok(result.rows_affected() > 0)
     }
 
     pub async fn list_labels(&self, repo_id: &str) -> Result<Vec<String>> {
--- a/crates/gitlawb-node/src/test_support.rs
+++ b/crates/gitlawb-node/src/test_support.rs
@@ -14359,6 +14359,201 @@ mod tests {
         assert!(resp.status().is_success());
     }
 
+    /// #344: `remove_label` must canonicalize the path segment the same way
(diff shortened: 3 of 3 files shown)


beardthelion (changes requested):
The fix is right and the tests behind it hold up. I checked the guard is load-bearing rather than
taking the green run for it: gutting the `if !removed` branch flips
`remove_label_canonicalizes_and_404s_on_no_match` red (200 against an expected 404), inverting the
condition to `if removed` also goes red the other way, and gutting `require_repo_owner` flips
`remove_label_denies_non_owner` red (200 against an expected 403). All three restore green. The label
tests pass at head and CI is green on the current head.

Two things before I approve.

## Findings

- [P2] Pin the 50-character arm of the new helper.
  `crates/gitlawb-node/src/api/labels.rs:24`
  Nothing in the suite exercises a label longer than 50 characters. The only bad-request arms under test are
  empty-after-trim and disallowed charset, so deleting the `label.len() > 50` clause leaves every
  test green. The clause predates this PR, but this is the change that lifts it into a shared helper
  two handlers now depend on, so it should carry its own case. A 51-character DELETE asserting 400
  plus a 50-character one asserting 404 covers it, and the second has the side benefit of decoupling
  the no-match assertion from the delete that precedes it.

- [P3] The description understates the contract change.
  `crates/gitlawb-node/src/api/labels.rs:83`
  The body says the only externally observable change is 404-on-no-match. A DELETE carrying a
  malformed label now returns 400 where it previously matched nothing and returned
  `{"removed": true}`. Any client cleaning up historically malformed label names sees a new status
  class. Worth naming in the description.

## On the CodeRabbit thread

I looked at it and I am not asking you to act on it. The ordering it describes is real:
`canonicalize_label` runs before `require_repo_owner`, so an authenticated non-owner sending a
malformed label gets 400 rather than 403. But that 400 is a function of the caller's own path
segment and tells them nothing about whether the repo exists, who owns it, or whether the label is
there. The owner gate still sits after the record load and before the mutation, which is where it
belongs.

The bigger reason to leave it: `add_label` on `main` already validates before the repo lookup.
Reordering only `remove_label` would reintroduce exactly the drift between the two handlers that
this PR exists to remove. If we want authorization ahead of input validation as a general rule here,
that is a separate change touching both handlers and it is mine to make, not yours to absorb in this
one.

Push those two and I will take another look.

Ayush7614:
Thanks for the review @beardthelion 

Addressed review feedback (pushed `24e869f`):

**P2 – Pin `len > 50` arm `crates/gitlawb-node/src/api/labels.rs:24`:** Done. `remove_label_canonicalizes_and_404s_on_no_match` (`crates/gitlawb-node/src/test_support.rs:1495`) now also asserts:
- `51`x`a` → `400` (proves `canonicalize_label` rejects over-length)
- `50`x`a` → `404` (passes validation, proves `len==50` is allowed and decouples the no-match `404` from the double-delete above, so the helper cannot be gutted without a red test). Verified deleting the `len > 50` clause now breaks the new case.

**P3 – Description understates contract change:** Ack – a `DELETE` with a malformed label that previously fell through to `{removed:true}` now returns `400` via the shared helper. The externally observable change is therefore `404` on no-match **plus** `400` on malformed labels. Will update PR body on next push if you want it reworded; current code behavior is intentional (shared validation).

**On the CodeRabbit thread `crates/gitlawb-node/src/api/labels.rs:72`:** Not reordering. `canonicalize_label` before `require_repo_owner` is deliberate per your note – keeps `add_label`/`remove_label` from drifting ("add" already validates before the repo lookup on `main`) and the `400` reveals only the caller-supplied path segment, not repo/label existence. Owner gate stays after `get_repo` and before the DB mutation where it belongs.

`cargo check -p gitlawb-node --tests --all-targets` OK; `cargo fmt` clean; existing `clippy::duplicated_attributes` on `ipfs.rs:2169` is pre-existing on `main` (ignored per PR).

beardthelion (approved):
Re-checked head `24e869f`. The len arm you added is load-bearing: gutting `label.len() > 50` in `canonicalize_label` flips the 51-char case red (404 instead of 400). The premise guard and the owner gate still pin as before: gutting `if !removed` goes red on the double-delete (200 vs 404), gutting `require_repo_owner` goes red on the stranger case (200 vs 403). `remove_label_` tests pass, `authz_guard` passes (including the `remove_label` / `require_repo_owner(` row), CI is green on the current head.

Not an ask, recorded only: the PR description still understates the contract change from round 1 P3. A DELETE with a malformed label now returns 400 where it previously returned `removed: true`. Worth a one-line note if you edit the body, not a blocker.

On the CodeRabbit thread: unchanged from round 1. `canonicalize_label` before `require_repo_owner` means a signed non-owner with a malformed path gets 400 not 403, but that happens before any repo lookup and reveals nothing about existence. `add_label` already validates before the record load; reordering only `remove_label` would reintroduce the drift this PR closes. I am not asking you to act on it.

jatmn (approved):
LGTM

beardthelion:
@kevincodex1 LGTM

Issue #393, opened 2026-09-01
gl doctor prints the node URL unsanitized, so ANSI and bidi characters from --node reach the terminal


beardthelion:
## Summary

`gl doctor` prints the node URL into three of its result rows without sanitizing it. The URL comes from `--node` or `GITLAWB_NODE`, so ANSI escapes, BEL, and a Unicode right-to-left override embedded in that value reach the operator's terminal verbatim. `crates/gl/src/doctor.rs` calls no sanitizer at all, while `sync.rs`, `clone.rs`, `peer.rs`, `ipfs_cmd.rs` and `whoami.rs` all route node-supplied strings through `sanitize_node_msg`.

This is the same terminal-injection class as #123 (read side) and #187 (write side), but a different input: those two are about the node's error `message`, this is about the node URL the user was handed. #187's inventory of 44 `bail!` sites does not include doctor, and doctor surfaces the value on its success path too, not only on failure.

## Where

On `main`, `crates/gl/src/doctor.rs`:

- `:178` the node pass row, `format!("{} — v{version} ({short_did}…)", args.node)`
- `:197` the non-2xx row, `format!("{} returned HTTP {}", args.node, resp.status())`
- `:204` the unreachable row, `format!("{} unreachable: {e}", args.node)`

## Reproduced by execution

Built `gl` and ran doctor with a `--node` value carrying ESC, BEL, and U+202E, then read the raw bytes of the output:

```
$ gl doctor --node $'http://127.0.0.1:1/\x1b[31m\x07'
  ✗  node   http://127.0.0.1:1/<ESC>[31m<BEL> unreachable: GET http://127.0.0.1:1/<ESC>[31m<BEL>/
```

`od -c` on that line shows `033`, `\a` and `342 200 256` present unmodified. The value appears twice per row, once from `args.node` and once inside the reqwest error's own URL.

## Why it matters

Lower severity than #187, and worth saying so plainly: the value is one the user typed or exported rather than one a node chose, so the delivery path is a copy-pasted install command, a README in an untrusted repo, or a shell profile someone else wrote, not a hostile node answering a request. What makes it worth closing anyway is that doctor is the command a confused user runs, its whole job is to be believed, and a right-to-left override can reorder a printed URL so the host reads as something else. Doctor also prints on success, so unlike the `bail!` sites there is no failure required to reach the sink.

## Fix direction

[...]

PR #394, opened 2026-09-01
fix(gl): make doctor diagnose the node the git transport will use

beardthelion:
## What

`gl doctor` reported a healthy node and exited 0 on a stock install whose `git push` and `git clone` could not reach anything. `gl`'s `--node` defaults to `https://node.gitlawb.com` while `git-remote-gitlawb` fell back to `http://127.0.0.1:7545`, so with `GITLAWB_NODE` unset the check validated a URL the transport never contacts. An explicit `--node` outranks the environment, so the two could disagree even when the variable was set, and that state produced no signal at all.

Observed on `main` before the change, same shell, variable unset both times:

```
gl doctor:  ✓  node   https://node.gitlawb.com — v0.7.0 (did:key:z6Mkicjkc95VcFx38Xg2Sv…)   EXIT=0
helper:     repo_base: http://127.0.0.1:7545/z6Mk.../myrepo
```

A real `git clone gitlawb://` with the variable unset never left the machine (`connecting to 127.0.0.1:7545`, then `error sending request`). The same clone with `GITLAWB_NODE=https://node.gitlawb.com` reached the node and returned a genuine `anonymous info/refs returned 404`.

## Fix

`gitlawb_core::resolve_transport_node` is now the single resolution both binaries call. Doctor probes whatever it returns whenever that differs from `args.node`, and folds the result into the existing `GITLAWB_NODE` row rather than adding a second row for one condition.

Three properties the probe holds that a naive reachability check does not:

1. **It requires identity, not a 200.** The row reports the transport usable only when the response carries a `did`. A local service answering 200 on that port otherwise reads as a working transport, which is the same false green this change exists to remove. A stub serving `<html>not a node</html>` now produces `something is listening but it is not a gitlawb node`.
2. **It uses a dedicated client.** `no_proxy`, 3s timeout. The shared `NodeClient` routes a loopback probe through `HTTP_PROXY` (verified: a proxy listener received `GET http://127.0.0.1:7545/`), which reports a running local node as dead, and its 30s timeout stalled every stock run behind a hung listener. Measured 30s before, 3s after.
3. **It sanitizes.** The URL is caller-supplied and the remedy line is printed for the user to paste, so the value goes through `sanitize_node_msg` and is single-quoted.

A blank or whitespace-only `GITLAWB_NODE` now resolves to the local default in both binaries. It previously gave the helper an empty base, so `repo_base` came out as `/z6Mk.../repo` with no scheme or host.

Tiering follows who chose the broken value. An unset variable stays advisory, because `gl` itself still works and #357 requires a stock install to keep exiting 0. A variable the user set to something unusable fails.

## Verification

Every state driven against the built binaries:

| state | row |
| --- | --- |
| unset | `⚠ git push/clone will use http://127.0.0.1:7545 (unreachable: ...); gl targets ...` |
| blank or whitespace | same, and the helper now resolves the same URL |
| `--node` overriding a set variable | `✗` naming the variable's node (previously no row) |
| variable and flag agree | `✓`, no divergence clause |
| loopback | `✓ ... (local node, intentional for self-hosting/dev ...)`, unchanged |
| non-gitlawb 200 on the port | `something is listening but it is not a gitlawb node` |

Seven mutations, run one at a time, each observed RED: dropping the identity check, disabling the divergence gate, treating blank as configured, removing sanitization, the helper reverting to a private literal, the help text going stale, and mis-tiering the row. One initially came back green and exposed that the test's own stub never delivered a usable 200; the stub was fixed and the assertion made positive before it was trusted.

`cargo fmt --all -- --check` clean. `cargo test --locked` green for `gl` (363 + 8), `gitlawb-core` (101), `git-remote-gitlawb` (53 + 8). `cargo clippy --all-targets --locked -- -D warnings` clean. No lockfile change.

## Notes for review

- **Conflicts with #391.** That PR edits the same `GITLAWB_NODE` arm to re-tier it to Warn and add a non-zero exit on Fail rows. Whichever lands second needs a rebase. The tiering here is deliberately compatible: a stock install produces a Warn row, so #391's exit gate keeps it at 0.
- The unchanged `node` row still prints the URL unsanitized. That is pre-existing and now filed as #393; the control-character test here is scoped to the rows this change owns rather than reaching into adjacent code.
- `DEFAULT_LOCAL_NODE` is an IPv4 literal, so a node bound only to `[::1]` reads as unreachable. The row is truthful, since the helper shares the constant and fails identically, but the remedy steers to the public node instead of `export GITLAWB_NODE=http://[::1]:7545`. Left as is; changing it changes transport behaviour, not the diagnostic.

## Summary by CodeRabbit

- **New Features**
  - Added consistent node resolution for Git transport, including a local default at `http://127.0.0.1:7545`.
  - Blank or surrounding whitespace in node configuration is handled automatically.
  - `gl doctor` now checks the node used by Git transport and reports differences from the CLI node.
  - Added direct connectivity checks that distinguish Gitlawb nodes from unrelated services.

- **Bug Fixes**
  - Improved diagnostics for missing, invalid, or unreachable transport node configuration.
  - Sanitized URLs and errors shown in diagnostic output.

### DIFF
--- a/crates/git-remote-gitlawb/src/main.rs
+++ b/crates/git-remote-gitlawb/src/main.rs
@@ -25,6 +25,7 @@
 use anyhow::{bail, Context, Result};
 use gitlawb_core::http_sig::sign_request;
 use gitlawb_core::identity::Keypair;
+use gitlawb_core::{resolve_transport_node, DEFAULT_LOCAL_NODE};
 use std::io::{self, BufRead, Read, Write};
 
 fn main() -> Result<()> {
@@ -63,9 +64,9 @@ fn main() -> Result<()> {
 
     let (_, short_owner, repo_name) = parse_gitlawb_url(url)?;
 
-    // v0.1: default to localhost. Override with GITLAWB_NODE env var.
-    let node_base =
-        std::env::var("GITLAWB_NODE").unwrap_or_else(|_| "http://127.0.0.1:7545".to_string());
+    // v0.1: default to localhost. Override with GITLAWB_NODE env var. Shared with
+    // `gl doctor` so the diagnostic reports the URL this actually resolves.
+    let node_base = node_base_from_env();
     let repo_base = format!("{}/{}/{}", node_base, short_owner, repo_name);
     tracing::debug!("repo_base: {repo_base}");
 
@@ -76,6 +77,12 @@ fn main() -> Result<()> {
     let keypair = load_keypair();
 
     run_helper(&repo_base, keypair.as_ref())
+}
+
+/// The node base URL this helper will contact, resolved through the same function
+/// `gl doctor` reports on so the diagnostic cannot describe a different node.
+fn node_base_from_env() -> String {
+    resolve_transport_node(std::env::var("GITLAWB_NODE").ok().as_deref())
 }
 
 // ── CLI argument handling ──────────────────────────────────────────────────────
@@ -123,7 +130,7 @@ fn help_text() -> String {
          \x20   git clone gitlawb://did:key:z6Mk.../<repo>\n\
          \n\
          ENVIRONMENT:\n\
-         \x20   GITLAWB_NODE   Node base URL (default: http://127.0.0.1:7545)\n\
+         \x20   GITLAWB_NODE   Node base URL (default: {DEFAULT_LOCAL_NODE})\n\
          \x20   GITLAWB_KEY    Identity PEM path for signed fetch/push (default: ~/.gitlawb/identity.pem)\n\
          \x20   GITLAWB_LOG    Log filter (default: warn)\n\
          \n\
@@ -2014,7 +2021,19 @@ mod tests {
         assert!(help.contains("--version"));
         assert!(help.contains("--help"));
         assert!(help.contains("GITLAWB_NODE"));
+        // The advertised default must be the value the helper actually resolves.
+        assert!(help.contains(DEFAULT_LOCAL_NODE));
         assert!(help.ends_with('\n'));
+    }
+
+    /// The helper's own resolution must stay the shared one; a site walking back to
+    /// a private literal is exactly the drift the shared function exists to stop.
+    #[test]
+    fn node_base_matches_the_shared_resolver() {
+        // No env manipulation: assert the call site delegates, for the default case.
+        let expected = resolve_transport_node(std::env::var("GITLAWB_NODE").ok().as_deref());
+        assert_eq!(node_base_from_env(), expected);
+        assert_eq!(resolve_transport_node(None), DEFAULT_LOCAL_NODE);
     }
 
     // ── #117 multi-round fetch negotiation ───────────────────────────────────
--- a/crates/gitlawb-core/src/lib.rs
+++ b/crates/gitlawb-core/src/lib.rs
@@ -17,5 +17,50 @@ pub mod sanitize;
 pub mod scan_token;
 pub mod ucan;
 
+/// Node URL the git transport falls back to when `GITLAWB_NODE` is unset.
+///
+/// `gl` defaults to the public node instead, so the two disagree on an install
+/// that never sets the variable.
+pub const DEFAULT_LOCAL_NODE: &str = "http://127.0.0.1:7545";
+
+/// The node `git clone` and `git push` will contact, given the raw `GITLAWB_NODE`
+/// value (`None` when the variable is absent).
+///
+/// `git-remote-gitlawb` calls this to pick its base URL and `gl doctor` calls it
+/// to report that URL, so a diagnostic cannot describe a node the transport will
+/// not use. A blank or whitespace-only value is not a configured node: treating
+/// it as one gave the helper an empty base and every clone URL a missing scheme
+/// and host.
+pub fn resolve_transport_node(env_value: Option<&str>) -> String {
+    match env_value.map(str::trim) {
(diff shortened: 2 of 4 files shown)


jatmn (changes requested):

## Findings

- [P2] Separate shell serialization from terminal sanitization
  `crates/gl/src/doctor.rs:178`
  CodeRabbit's unresolved request here is valid. `gl_node` comes from `sanitize_node_msg`, which removes terminal control characters but deliberately does not escape shell syntax. Embedding that value between single quotes means an apostrophe closes the assignment: for example, a node ending in `'; command; #'` makes the paste-ready remedy run `command` when the user follows it. The same sanitizer caps output at 200 characters, so using it as command data can also turn a long valid URL into a different, truncated value.

  The root cause is that one representation is being used for two different trust boundaries: safe terminal display and lossless shell argument encoding. Please define the canonical value that is safe to place back into configuration, encode it with a dedicated shell-quoting routine (or avoid emitting a paste-ready assignment), and apply the 200-character cap only to explanatory prose. Add regression cases containing an apostrophe, whitespace, shell metacharacters, rejected/stripped control characters, and an otherwise-valid value longer than the display cap; for accepted values, the rendered command should set exactly one variable to the complete canonical value and execute nothing else.

- [P2] Build the probe from the transport's redirect contract
  `crates/gl/src/doctor.rs:365`
  CodeRabbit's unresolved redirect request is valid. The dedicated probe keeps reqwest's default redirect behavior, whereas `git-remote-gitlawb` applies the shared same-origin, same-request-target policy with a bounded chain. A configured endpoint can therefore redirect `/` to another origin that returns a `did`; doctor reports `reachable`, but clone/push stop at the original 3xx. The probe can also contact a redirect-selected origin that the transport deliberately refuses.

  The root cause is duplicating HTTP-client construction without carrying over the transport semantics this diagnostic is supposed to model. Please reuse the shared redirect decision and chain bound—or extract a common client-policy builder if that is the maintainable boundary—while retaining the probe's shorter timeout and credential-free request. A regression test should prove that cross-origin and request-target-changing redirects are not followed, while any redirect shape intentionally supported by the transport remains supported by doctor too.

- [P2] Make proxy behavior part of the shared transport policy
  `crates/gl/src/doctor.rs:366`
  `.no_proxy()` applies to every probe URL, but the actual remote helper retains reqwest's normal environment/system proxy discovery. The two processes can consequently test different routes: a corporate-proxy-only node can work for clone/push while doctor marks it unreachable, and a directly reachable listener can look healthy even though the helper's configured proxy route fails or reaches another service. This is especially important because the row claims to describe what git push/clone will use, not merely whether the origin is directly reachable.

  The root cause is solving the loopback-proxy symptom only in the diagnostic, which creates a second network policy instead of aligning the producer and observer. Please define the intended proxy rule once and apply it consistently to both the helper and probe. If loopback must bypass proxies, update or explicitly configure both paths rather than special-casing only doctor. Cover at least a proxy-required remote node and a loopback node under a configured proxy, asserting that doctor's verdict matches the helper's effective route in both cases.

- [P2] Derive the suggested fix from normalized state, not the raw blank value
  `crates/gl/src/doctor.rs:178`
  With blank or whitespace-only `GITLAWB_NODE` and no explicit `--node`, the shared resolver correctly maps the transport to the local default, but clap leaves the raw blank value in `args.node`. The new remedy is built from that raw/display value, so doctor prints `export GITLAWB_NODE='   '`. Pasting it recreates the broken configuration instead of fixing it; before this PR, the branch at least suggested a concrete public-node URL.

  The root cause is mixing three states—raw env/CLI input, normalized effective transport configuration, and display text—when constructing recovery guidance. Please choose the intended valid recovery target from normalized configuration and explicit-flag precedence, then pass that value through the shell serializer described above. Add an end-to-end case that invokes the binary with only blank/whitespace `GITLAWB_NODE` (no synthetic `--node` override) and asserts that the suggested command changes the state to a valid target rather than reproducing the blank input.

- [P3] Apply a response-size budget before parsing probe JSON
  `crates/gl/src/doctor.rs:381`
  `resp.json::<serde_json::Value>()` buffers the complete success response. The three-second timeout bounds elapsed time but not the number of bytes a local or configured endpoint can deliver, so a fast large response can consume excessive memory before parsing fails. This is a newly added untrusted response path, and the crate already has `read_body_capped` to enforce the same hostile-node boundary elsewhere.

  The root cause is treating the timeout as the probe's full resource budget. Please read only a small bounded root document, reject truncation/read failure as unusable, and parse the bounded bytes into the narrow response shape the probe needs. Include an oversized-body regression proving that allocation/read work stops at the cap and cannot be mistaken for a valid identity response.

- [P3] Give subprocess tests a seam for every external dependency
  `crates/gl/tests/doctor_transport_node.rs:35`
  The helper comment says the subprocess is kept off the unrelated iCaptcha and GitHub networks, but only iCaptcha is redirected. Every `doctor()` invocation still reaches `check_version(current, "https://api.github.com")`; offline or filtered environments therefore acquire real external side effects and can wait on the five-second timeout even though release lookup is unrelated to these assertions.

  The root cause is testing the full command through hard-coded production dependencies without a controllable boundary. Please inject the release API base/checker into the doctor runner, provide a narrowly scoped test override, or otherwise route it to a local fixture; production should continue using the real release endpoint. Add an assertion at the fixture boundary so these focused tests fail if any request escapes to the public API instead of silently succeeding or timing out.

Issue #395, opened 2026-09-02
Agent-task read surfaces are not gated behind repo/task visibility rules


euxaristia:
Task read endpoints (`GET /tasks`, GraphQL subscriptions) serve task data without checking repo visibility or task-level access rules. This allows unauthenticated callers to read task bodies on private repos.

Needed:
- Gate task reads behind `authorize_repo_read`
- Decouple open-claim eligibility from read visibility
- Apply rate limiting to GraphQL WebSocket subscriptions
- Enforce hostile-node response limits in the CLI and MCP client

PR #396, opened 2026-09-04
fix(node)!: Gate agent-task reads behind visibility rules

euxaristia:
## Summary
Gates agent-task read surfaces behind repo/task visibility rules, decouples open-claim eligibility from read visibility, resets WebSocket field budgets per operation, and aligns task claim tests with the opaque 404 existence-hiding contract.

Refs #395

## Changes
- Implement `TaskReadBrakeExtension` in `crates/gitlawb-node/src/graphql/mod.rs` to reset the 5-field task read budget per WebSocket operation while preserving connection-level per-IP rate limits.
- Register `/graphql/ws` under `optional_signature` and thread `AuthenticatedDid` into GraphQL connection data.
- Gate unresolved task claims behind `is_repo_quarantined` to prevent leaks on quarantined mirror repos.
- Validate `task_list` `limit` parameter in `gl` MCP strictly as integer or default 50.
- Drain WebSocket frames through terminal frame (`complete`/`error`) by operation ID in test helpers.
- Add forged-signature WebSocket handshake rejection regression test asserting HTTP 401 when declared DID does not match signing keypair.
- Align `claim_task_does_not_steal_preassigned_assignee` with opaque 404 expectations and preserve direct SQL guard coverage.
- Align GraphQL task claim race tests with opaque 404 for rival and fixed conflict mapping for lost write races.

## Prior reviewer feedback addressed
- CodeRabbit: Enforce `is_repo_quarantined` on task claim fallback, authenticate WebSocket queries, strictly validate MCP limit argument, and add forged-signature WebSocket rejection test.
- CI: Aligned `task_write_conflict` assertion message in `claim_task_does_not_steal_preassigned_assignee`.

## Test plan
- `cargo check --workspace --all-targets`
- `cargo clippy --workspace --bins -- -D warnings`
- `cargo test -p gitlawb-node graphql_ws_authenticated_query_accesses_private_task`
- `cargo test -p gl --bin gl mcp::tests`

## Summary by CodeRabbit

- **New Features**
  - Added cursor-based pagination for task listings in GraphQL, CLI, and MCP.
  - Added resumable CLI listings with `--cursor` and incomplete-result notices.
  - Task read responses no longer expose sensitive task tokens.
  - Authenticated clients can access authorized private tasks.
  - Added configurable rate limits for task reads.

- **Bug Fixes**
  - Improved handling of task conflicts, stale updates, invalid requests, and unavailable repositories.
  - CLI and MCP commands now report HTTP failures clearly.
  - Hidden tasks consistently return not-found responses.

- **Security**
  - Protected cursors are caller- and filter-bound, expiry-limited, and node-specific.


beardthelion (changes requested):
Checked head ea5469b2 against origin/main. CI is green on the SHA (https://github.com/Gitlawb/node/actions/runs/33930967342/job/101209295281). I ran the task-read regression module and the GraphQL WebSocket auth/rate-limit tests locally on this head, and read the shared REST/GraphQL collectors. The anonymous and stranger read gate looks solid. One quarantine hole remains on task reads.

## Findings

- [P2] Hard-drop quarantined-repo tasks before the delegator/assignee short-circuit
  `crates/gitlawb-node/src/api/tasks.rs:161`
  `task_visible` returns true for the delegator or assignee before it looks at `repo_id` or quarantine state. `get_visible_task` and `collect_visible_tasks` only call `task_visible`, so a task on a quarantined canonical repo is still readable by its parties through `GET /api/v1/tasks`, `GET /api/v1/tasks/{id}`, and GraphQL `tasks`/`task`. `get_claimable_task` already hard-drops quarantined repos at line 460; the read path does not. Ref-update feeds withhold quarantined rows even from the repo owner. Check `is_repo_quarantined` on `task.repo_id` before the party-match returns, and add a regression that quarantines a repo, seeds a task with that owner as delegator, then asserts REST 404 and GraphQL null on list/get.

Not an ask, recorded only: `fail_task`/`failTask` could mirror the visible-non-assignee tests that `complete_task` already carries; that is separate from this quarantine gap.

beardthelion (changes requested):
Checked head 0c1d5e4d against origin/main. The quarantine read gap from round 1 is fixed: `task_visible` hard-drops quarantined repos before party checks, and `quarantined_repo_task_withheld_from_owner_on_rest_and_graphql` passes. I ran `visible_tasks_tests` (32/32) and a revert probe that turned the quarantine guard RED. A second-model pass on this head surfaced two items I want addressed before merge.

## Findings

- [P2] Fail closed for claim eligibility on mirror rows and unresolved canonical repo IDs
  `crates/gitlawb-node/src/api/tasks.rs:470`
  `task_claimable` returns `true` for slash-form repo IDs and for canonical IDs with no local deduped record, so a signed stranger who knows an unassigned task ID can claim through `get_claimable_task` and receive `task_to_json`, including `payload` and `ucan_token`. Read surfaces correctly hide mirror-repo tasks (`mirror_only_repo_task_is_hidden_from_anonymous_reads`), but the claim path treats those repo states as open. Seed an unassigned task on a mirror id or a nonexistent canonical repo id, POST `/api/v1/tasks/{id}/claim` as an unrelated DID, and assert opaque 404 instead of 200 with the task body.

- [P3] Document `GITLAWB_TASK_READ_RATE_LIMIT` beside the other operator rate-limit knobs
  `crates/gitlawb-node/src/config.rs:715`
  The clap field help is present, but `.env.example` and the README rate-limit table list `GITLAWB_IPFS_RATE_LIMIT` and peers without this knob. Add the variable with default 1200 and the same `0` disables semantics the code implements.

Not an ask, recorded only: the scan-ceiling continuation probe at `tasks.rs:374-387` is deliberate bounded disclosure pinned by `scan_ceiling_continuation_discloses_only_a_terminal_page`; I am not asking to remove it on this round.

jatmn (changes requested):

The runtime work for #395 / #268 is on this head: REST and GraphQL task reads share `collect_visible_tasks` / `get_visible_task`, reads omit `ucan_token`, quarantine drops before party checks, `/graphql/ws` is under `optional_signature`, task-read IP brakes are wired, and `gl` / MCP follow opaque cursors with hostile-node caps. What did not move with that change is the operator-facing contract. Both findings below are the same miss: clap, `main.rs`, collectors, and tests describe the new gate; `README.md`, `.env.example`, and `SECURITY.md` still describe the old hole or omit the new brake. That is why this is still open after the code rounds. It is not because claim, cursors, WebSocket auth, or quarantine need another pass.

Please treat this comment as the closed set for merge. Fix the two docs items in one commit. Do not retouch working gates to “look complete.”

## Close-out (so the next commit can be the last)

Do:

- Add `GITLAWB_TASK_READ_RATE_LIMIT` next to the other per-IP rate-limit knobs in the README settings table and `.env.example`, copying the semantics already implemented (default 1200, `0` disables, keyed via `GITLAWB_TRUSTED_PROXY`).
- Rewrite the published “task listings are not repository-gated / include a UCAN token” sentences in `SECURITY.md` and the README known-limitations bullet so they match gated, redacted reads. Leave pin and anchor listing limitations in those same sentences.

Do not:

- Fail-closed unassigned claim for slash-form mirror ids or unresolved canonical repo ids. #395 asked to decouple open-claim eligibility from read visibility. `task_visible` already withholds mirror ids from reads; `task_claimable` leaves unassigned claim open when no local canonical row can run `listable_at_root`. CodeRabbit’s accepted claim-side fix was quarantine-deny plus preserve-open for unresolved remotes, including slash-form. Repo-less open claim is pinned by `open_repoless_task_create_claim_complete_lifecycle`. Hosted private repos already 404 via `claim_task_on_private_repo_task_returns_404`. Returning `payload` / `ucan_token` on a successful claim is the #268 delivery path, not a leftover list leak. Recoupling those claim paths to the read 404 is out of scope and will bounce.
- Change the 1200 default, the limiter wiring, REST vs GraphQL brake placement, or `0` disables.
- Put `ucan_token` back on GET list/get or GraphQL `tasks` / `task`.
- Rewrite pin, anchor, sparse-clone, or “visibility cannot retract already-announced content” limitations.
- Change the `[0, 200]` task `limit` clamp, cursor MAC, scan-ceiling probe, or CLI/MCP caps. Those are either documented here or owned by #401 / #405.
- Reopen WS auth, forged-signature tests, MCP integer `limit`, or the preassigned SQL claim guard. Those are done on this head.

## Merge readiness

- Overlapping work exists but does not replace a review of this head. Closed #327 is the prior attempt at the same title. Open #275 also honors pre-assigned assignees on claim; this branch already includes the SQL `(assignee_did IS NULL OR normalized key)` guard and opaque 404 for strangers. Open #401 and #405 retarget #399 (task `limit` clamp); this branch clamps to `[0, 200]` and documents GraphQL negatives as clamp-to-zero, while those PRs want `[1, 200]`. Resolve or close the duplicates after this lands so they do not fight the collector.

- Mergeable against current `main` (`bfc44f926d08c0bf774e2c05dd76b245871294f1`); no rebase drift. CI on head `0c1d5e4df255a71e6958b059801ffb7c84a4e3d9` is green (fmt/clippy, tests, audit, MSRV, Docker smoke). Merge is blocked on review.

## Findings

- [P3] Document `GITLAWB_TASK_READ_RATE_LIMIT` beside the other operator rate-limit knobs
  `crates/gitlawb-node/src/config.rs:715`
  `README.md:414`
  `.env.example:273`

  Root cause: the PR added a new operator security knob (`task_read_rate_limit`, env `GITLAWB_TASK_READ_RATE_LIMIT`, default 1200, `0` disables) and wired it on REST `GET /api/v1/tasks` / `GET /api/v1/tasks/{id}` plus GraphQL/WS `TaskReadBrake`, with a startup warning in `main.rs` when it is `0`. The catalogs operators actually copy from were not updated. `.env.example` and the README “Important node settings” table already list `GITLAWB_IPFS_RATE_LIMIT`, `GITLAWB_CREATE_RATE_LIMIT`, `GITLAWB_PEER_WRITE_RATE_LIMIT`, and `GITLAWB_SYNC_TRIGGER_RATE_LIMIT`. An operator grepping those files cannot discover, tune, or knowingly disable the task-read brake, including that `0` turns it off.

  Please add the variable in both places, next to `GITLAWB_IPFS_RATE_LIMIT`, with the contract the code already implements:

  - Default 1200 requests per client IP per hour (above the `/ipfs` 600 budget because a list page plus per-task reads is a normal client pattern).
  - `0` disables.
  - Keyed on the resolved client IP via `GITLAWB_TRUSTED_PROXY`, same as the sibling brakes.
  - Applies to the anonymous task-read routes (`GET /api/v1/tasks`, `GET /api/v1/tasks/{id}`) and the GraphQL/WS task-read brake that shares that budget.

  Copy the clap comment at `config.rs:706-714` into `.env.example` the way the IPFS/create/peer blocks are written. One README table row matching the `GITLAWB_IPFS_RATE_LIMIT` row shape is enough.

  Do not change the default, the limiter, REST-vs-GraphQL layering, or any other rate-limit knob. No code change is required. After the edit, `rg GITLAWB_TASK_READ_RATE_LIMIT README.md .env.example` should hit both files.

- [P3] Align published security limitations with gated task reads
  `SECURITY.md:76`
  `README.md:72`

  Root cause: this PR closed the #268 / #395 task-read hole (visibility collectors + no `ucan_token` on read projections) but left the Known Limitations text that existed to disclose that hole. Operators reading `SECURITY.md` or the README limitations list are still told the old exposure is live. The three-dot diff did not include those files; the code change made the published sentences false, so this PR owns the update.

  Two sentences, one fact:

  - `SECURITY.md:76` currently: `GET /api/v1/tasks`, `/api/v1/ipfs/pins`, and `/api/v1/arweave/anchors` are not repository-gated. Task records include a UCAN token; pin and anchor listings expose object and ref metadata.
  - `README.md:72` currently: task, IPFS-pin, and Arweave-anchor listings are not repository-gated; …

  On this branch, `GET /api/v1/tasks`, `GET /api/v1/tasks/{id}`, and GraphQL `tasks` / `task` go through `collect_visible_tasks` / `get_visible_task`. REST reads use the projection that omits `ucan_token` (`task_to_json` is the write/claim path). Pins and anchors are unchanged.

  Please rewrite only the task-read half of those two sentences so they say task list/get is repository/task-gated and does not include `ucan_token`. Keep the pin and anchor clauses, the sparse-clone withheld-path note, and the “visibility cannot retract already-announced content” note exactly as they are. Do not expand `SECURITY.md` into a design doc, do not mention claim, and do not claim pins/anchors are gated.

  Example shape for `SECURITY.md:76` (wording can vary; the facts cannot): task list/get (REST and GraphQL) is gated by the same visibility collectors as other reads and omits `ucan_token`; `/api/v1/ipfs/pins` and `/api/v1/arweave/anchors` are not repository-gated and still expose object and ref metadata.

  Example shape for `README.md:72`: drop “task” from the ungated-listing list; keep IPFS-pin and Arweave-anchor listings, withheld path names, and the retractability clause.

  After the edit, `rg -n 'not repository-gated' SECURITY.md README.md` should no longer attach that phrase to task reads. Leave `docs/OSS-READINESS-AUDIT.md` alone; it is a historical snapshot that already defers to `SECURITY.md`.

euxaristia:
Addressed the two findings in [review 5123664742](https://github.com/Gitlawb/node/pull/396#pullrequestreview-5123664742) in commit 41fef3a, and the [follow-up wording correction](https://github.com/Gitlawb/node/pull/396#discussion_r3945981191) in commit d9aded2.

- Added `GITLAWB_TASK_READ_RATE_LIMIT` beside the IPFS rate-limit setting in `README.md` and `.env.example`, including the 1200 requests/IP/hour default, `0` disables, trusted-proxy IP resolution, and shared REST/GraphQL/WS task-read budget.
- Updated `README.md` and `SECURITY.md` to describe gated task list/get reads that omit `ucan_token`, preserving the pin/anchor, withheld-path, and already-announced-content limitations.
- Corrected `.env.example` and the matching CLI help comment in `crates/gitlawb-node/src/config.rs`: the list route returns a visibility-filtered page; the individual-task route returns an opaque 404 when the task is hidden or missing. Both can incur visibility-query costs even when no task is returned.

These commits update documentation and CLI help wording only; runtime gates, defaults, and limiter wiring are unchanged. Local validation passed: the requested documentation searches, `cargo fmt --all -- --check`, and Git diff whitespace checks. Runtime tests were not rerun locally for these edits.

At the time of this update, CI on d9aded2 has passed formatting/clippy, audit, MSRV, Windows tests, and the other completed checks; stable/beta tests, the release build, and Docker smoke testing are still running. Current results: [PR checks](https://github.com/Gitlawb/node/pull/396/checks).

beardthelion (changes requested):
Checked head d9aded2c against origin/main. CI is green (12/12). The two docs findings from the prior round are addressed: GITLAWB_TASK_READ_RATE_LIMIT is in .env.example and the README settings table, and the SECURITY.md / README.md limitation sentences now describe gated task reads that omit ucan_token. I re-ran the prior load-bearing checks on this head: visible_tasks_tests (32/32) and the GraphQL WS auth tests (4/4). A revert probe that removes the quarantine check from task_visible goes RED on quarantined_repo_task_withheld_from_owner_on_rest_and_graphql, and a probe that adds ucan_token back to task_to_read_json goes RED on ucan_token_never_appears_in_read_responses. The gate, the quarantine drop, and the token omission are all load-bearing.

The visibility gate is sound across both surfaces: REST and GraphQL share collect_visible_tasks / get_visible_task, task_visible hard-drops quarantined repos before party checks, mirror repos and unhosted repo ids fail closed, and ucan_token appears only on the write/claim delivery path. The cursor is SIV-encrypted and caller/filter-bound, so a token holder cannot read or forge the position it carries. The WS handshake rejects forged signatures with 401, and the per-operation field budget reset is tested. One amplification issue on the anonymous read path, plus two minor doc gaps.

## Findings

- [P2] Batch the quarantine check in collect_visible_tasks
  `crates/gitlawb-node/src/api/tasks.rs:288`
  The scan loop calls db.is_repo_quarantined(repo_id) per distinct repo id per batch. With 200 distinct repos per batch and up to 5 batches (scan ceiling 1000), one anonymous GET /api/v1/tasks triggers up to 1000 individual SELECT quarantined FROM repos WHERE id = 1 queries. The route is anonymous-reachable (optional_signature), and the rate limit bounds frequency but not per-request cost. list_repos_deduped_by_ids already batches the repo record fetch but does not select the quarantined column, so the per-repo loop is the only unbatched query in the scan. Replace it with a single SELECT id FROM repos WHERE id = ANY(1) AND quarantined = TRUE populating the same HashSet, or add quarantined to the existing batched SELECT and RepoRecord. The gate order is preserved either way: task_visible checks quarantined_repos before the delegator/assignee short-circuit.

- [P3] Mention the GraphQL/WS budget sharing in the --help docstring
  `crates/gitlawb-node/src/config.rs:706`
  The clap field help describes only the two REST routes. .env.example and the README table both say the GraphQL/WS task-read brake shares this budget. An operator tuning from --help alone would not know that GraphQL tasks / task queries and WS task queries debit the same per-IP bucket. Add one sentence matching the .env.example wording.

- [P3] Update the stale /graphql/ws mounting comment in subscription.rs
  `crates/gitlawb-node/src/graphql/subscription.rs:14`
  The comment says /graphql/ws is "mounted outside the optional_signature layer." This PR moved it under optional_signature (server.rs:107), and graphql_ws_handler now threads Option<Extension<AuthenticatedDid>> into connection data. The subscription resolvers do not read caller identity, so the write-side gating safety model is unchanged and correct. The comment is factually wrong about the mounting and could mislead a future developer into thinking the WS handshake is unauthenticated when it now rejects forged signatures with 401.

One process note, not a finding: open #401 and #405 retarget the task limit clamp to [1, 200] while this branch clamps to [0, 200]. Resolve or close those after this lands so they do not fight the collector.

jatmn (changes requested):

Treat this comment as the **closed merge set** for the runtime work on head `d9aded2`. The security and product contracts for #268 / #395 are in place on this head; what remains is one implementation gap in the list scanner, plus two maintainer-doc updates that should have landed with the same operator-facing pass that updated README and `.env.example`. Please address **only** the three findings below in **one** follow-up commit. Do not reopen claim eligibility, cursor semantics, WS auth wiring, quarantine gate order, rate-limit defaults, or limit clamps to “look complete” — those paths are load-bearing and tested.

## What is already verified on this head

- REST and GraphQL task reads share `collect_visible_tasks` / `get_visible_task`; reads omit `ucan_token`.
- Quarantined repos are hard-dropped before delegator/assignee short-circuit in `task_visible`.
- `/graphql/ws` is under `optional_signature`; forged signatures get HTTP 401; authenticated WS queries can read private tasks.
- Open-claim eligibility remains decoupled from read visibility for unassigned tasks (#395); mirror/unhosted claim behavior is intentional, not a defect.
- `GITLAWB_TASK_READ_RATE_LIMIT` is documented in README and `.env.example`; SECURITY/README limitation text matches gated reads.
- CI: 12/12 checks green on `d9aded2ca0957d4c34e04d2f41a608220342a5a6`.

Revert probes on this head go red when quarantine is removed from `task_visible` or when `ucan_token` is restored on read projections — the gates are load-bearing.

## Why review kept returning (and why this list is short)

This PR is large (~7k lines) and touched **every layer** of one contract at once: visibility collectors, opaque cursors, rate limits, GraphQL/WS auth, CLI/MCP pagination, and operator docs. That shape naturally produces **drip review** unless each round closes a single surface completely:

1. **Security and product rounds came first.** Early passes correctly focused on visibility holes (#268), claim vs read semantics (#395), cursor binding, WS authentication, and MCP parsing. Those required code changes across `tasks.rs`, `server.rs`, `task_cursor.rs`, and `gl/`. Each fix shifted behavior that later reviewers re-read from scratch.

2. **Operator docs lagged the code.** The runtime gained `GITLAWB_TASK_READ_RATE_LIMIT` and changed what SECURITY.md must say about task reads, but the first doc pass updated README/`.env.example` only. Reviewers correctly blocked merge until those files caught up (jatmn round). That round is **done** on `d9aded2`, but the **same** operator contract was not copied into clap `--help` and `AppState` field docs — a partial doc pass leaves a second doc-only round.

3. **Routing changes outpaced comments.** Moving `/graphql/ws` under `optional_signature` was the right fix, but the `ref_updates` module comment still describes the **pre-change** mount order (accurate on merge-base `main`, wrong on this head). Stale comments invite the next reviewer to re-litigate WS auth as if it were still broken.

4. **A new hot path was added without matching an existing batching pattern.** Quarantine enforcement in `collect_visible_tasks` correctly calls `is_repo_quarantined` per repo, but the sibling ref-update collector and the batched `list_repos_deduped_by_ids` call in the **same loop** already show the intended “one round trip per batch” shape. The per-repo loop is the last unbatched query in that scan — easy to miss in a diff this size, but it is the remaining **runtime** issue.

5. **Automated and parallel reviewers re-raise intentional choices.** Mirror/unhosted open claim, node-local cursors, GraphQL `TaskPageType` breaking shape, and anonymous publish gating for `task_events` are **design decisions** pinned by tests and maintainer direction. Treating each automated comment as a new merge blocker would never end.

**Root cause for authors on contracts like this:** when a PR changes behavior, document **every operator entry point in one commit** (`--help`, `AppState` comments, README, `.env.example`, SECURITY) using the same sentence template; when a PR changes routing, grep for stale mount/auth comments in the same commit; when a PR adds per-row DB work inside a public list scanner, compare against the nearest sibling collector (`events.rs`) before merge.

## Merge readiness

- Mergeable against current `main` (`bfc44f926d08c0bf774e2c05dd76b245871294f1`); no rebase drift observed.
- All 12 required PR checks passed on head `d9aded2`.
- Open #401 and #405 retarget task `limit` clamp to `[1, 200]` while this branch uses `[0, 200]`. Resolve or close those **after** this lands; do not fold that product choice into this close-out commit.
- Prior jatmn documentation findings are addressed on this head.

## Findings

### [P2] Batch the quarantine check in `collect_visible_tasks`

`crates/gitlawb-node/src/api/tasks.rs:281-300`  
`crates/gitlawb-node/src/db/mod.rs:1642-1662` (batched repo fetch — extend or mirror)

**What happens today**

Inside the `while scanned < MAX_TASK_SCAN_CANDIDATES` loop, each keyset batch:

1. Dedupes `repo_id` values from up to 200 tasks (`referenced`).
2. Loops `referenced` and awaits `db.is_repo_quarantined(repo_id)` — **one `SELECT quarantined FROM repos WHERE id = $1` per distinct id**.
3. Then calls `list_repos_deduped_by_ids(&referenced)` and `list_visibility_rules_for_repos` — already batched.

With `MAX_VISIBLE_TASKS` = 200 per batch and `MAX_TASK_SCAN_CANDIDATES` = 1,000, a single anonymous `GET /api/v1/tasks` or GraphQL `tasks` query can issue **up to 1,000** quarantine lookups. GraphQL aliases debit the same per-IP bucket per field but still run the full collector per field (capped at 5 fields/request). `GITLAWB_TASK_READ_RATE_LIMIT` limits **requests per hour**, not **queries per request**.

**Root cause**

Quarantine was added correctly to the visibility gate (`task_visible` checks `quarantined_repos` before party match) but implemented with the convenient single-row helper instead of aligning with:

- `api/events.rs:66` — `list_quarantined_repos()` once per request for ref updates, or
- the batched `list_repos_deduped_by_ids` call immediately below in the same loop.

This is **not** a missing security gate; it is **amplification on a public read path** introduced by this PR’s quarantine enforcement.

**Requested outcome (pick one; do not change gate semantics)**

**Option A (preferred — extends existing batch):** Add `quarantined` to the `SELECT` in `list_repos_deduped_by_ids` (or add `Db::quarantined_ids_among(&[String]) -> HashSet<String>` used only here). Build `quarantined_repos` from rows where `quarantined == true`. Drop the `for repo_id in &referenced` loop entirely.

**Option B:** One query per batch: `SELECT id FROM repos WHERE id = ANY($1) AND quarantined = TRUE` with `referenced` as the array bind.

**Invariants to preserve (do not drift)**

- `task_visible` must still test quarantine **before** delegator/assignee short-circuit (`tasks.rs:167-170`).
- `get_visible_task` single-repo quarantine early-return stays as-is.
- Do not switch to loading **all** quarantined repos on every list request unless you measure repo-table size; batching to `referenced` ids is enough.

**Verification**

- Existing `visible_tasks_tests` and `quarantined_repo_task_withheld_from_owner_on_rest_and_graphql` must stay green.
- Add or extend a test that seeds many distinct quarantined repo ids on one page and asserts behavior unchanged (optional: count DB round-trips in a unit test if the repo has a pattern for that).

---

### [P3] Complete the operator contract for `GITLAWB_TASK_READ_RATE_LIMIT`

`crates/gitlawb-node/src/config.rs:706-716`  
`crates/gitlawb-node/src/state.rs:326-332`

**What happens today**

Runtime debits **one** `task_read_rate_limiter` for:

- REST: `GET /api/v1/tasks`, `GET /api/v1/tasks/{id}` via `rate_limit_by_ip` on `task_read_routes`.
- GraphQL HTTP: `TaskReadBrake` in `graphql_handler` (`server.rs:45-49`).
- GraphQL WS: same limiter/key in `graphql_ws_handler` (`server.rs:65-68`), per-operation field budget reset in `TaskReadBrakeExtension` (`graphql/mod.rs:115-130`).

README (`.env.example:285`, `README.md:415`) already says GraphQL/WS share the budget. Clap `--help` and `AppState::task_read_rate_limiter` doc comment name **only** the two REST routes and say the limiter is “layered on `task_read_routes` via `rate_limit_by_ip`,” which is **incomplete** for GraphQL/WS.

**Root cause**

Operator-facing text was updated where jatmn explicitly asked (README + `.env.example`) but not at the other two places operators and maintainers actually read when tuning (`gitlawb-node --help`, rustdoc on `AppState`). Partial doc passes cause a second review round for the same knob.

**Requested outcome**

Copy the **GraphQL/WS sharing** sentence from `.env.example:285` into:

1. The clap field doc on `task_read_rate_limit` in `config.rs` (after the REST route description).
2. The `task_read_rate_limiter` field comment in `state.rs` (note shared limiter across REST + GraphQL/WS, not only `task_read_routes`).

**Template (wording may vary; facts must not):**

> The same per-IP hourly budget applies to GraphQL `tasks` / `task` queries and WebSocket task queries via `TaskReadBrake`.

**Do not change**

- Default 1200, `0` disables, `GITLAWB_TRUSTED_PROXY` keying, limiter wiring, or REST vs GraphQL layering.

**Verification**

- `rg 'GraphQL/WS' README.md .env.example crates/gitlawb-node/src/config.rs crates/gitlawb-node/src/state.rs` should hit all four surfaces.

---

### [P3] Fix stale `/graphql/ws` mounting comment in `subscription.rs`

`crates/gitlawb-node/src/graphql/subscription.rs:14-22`  
`crates/gitlawb-node/src/server.rs:103-107`

**What happens today**

The `ref_updates` subscription doc says `/graphql/ws` is mounted **outside** `optional_signature` and the resolver has **no** caller identity. On merge-base `main` that was true (`/graphql/ws` was registered after the auth layer). On this head, `graphql_routes` registers `/graphql/ws` **before** `.layer(optional_signature)`, so signed handshakes are verified and `AuthenticatedDid` can be inserted into WS connection data. Forged-signature rejection and authenticated private-task query tests pin this.

The subscription **resolver** still does not read caller identity or filter per subscriber — same as before. Safety still rests on write-side gating (`announce` for ref updates, `announce_task_event` for tasks).

**Root cause**

A routing fix landed without updating the module-level comment that future readers use as the security model spec. That comment now asserts false facts about mounting and implies WS auth was never added.

**Requested outcome**

Rewrite the **first two sentences** of the `ref_updates` doc comment to state:

- `/graphql/ws` is under `optional_signature`; signed upgrades attach `AuthenticatedDid` to connection data.
- Subscription resolvers still **do not** gate per subscriber; visibility safety remains on the **write** side (keep the existing `if announce` / `announce_task_event` invariant paragraph).

**Do not change**

- Subscription relay logic, `announce_task_event` anonymous gate, or WS handler wiring.
- Do not add per-subscriber filtering in this close-out — that would be a product change, not a comment fix.

**Verification**

- Comment should not contain “mounted outside the `optional_signature` layer.”
- `server.rs:105-107` and the updated comment should agree.

---

## Close-out — do not expand scope

**Do in the one follow-up commit:**

- Batch quarantine lookup in `collect_visible_tasks` (finding 1).
- Align clap + `AppState` docs with README/`.env.example` for the shared task-read budget (finding 2).
- Fix the `subscription.rs` mounting/auth comment (finding 3).

**Do not (will bounce or reopen settled rounds):**

- Fail-close unassigned claim for mirror ids or unresolved canonical repo ids.
- Change `GITLAWB_TASK_READ_RATE_LIMIT` default, wiring, or disable semantics.
- Put `ucan_token` back on GET list/get or GraphQL read types.
- Rewrite pin/anchor/sparse-clone limitations in SECURITY.md.
- Change `[0, 200]` limit clamp, cursor MAC, scan-ceiling probe, or MCP/CLI page caps (#401 / #405 own limit semantics).
- Add per-subscriber filtering to `task_events` or `ref_updates` subscriptions.
- “Fix” node-local cursors for load-balanced deploys in this PR (document operationally elsewhere if needed; not a merge blocker here).

After these three items, the runtime contract for #268 / #395 on this branch is complete from my review. Further rounds should be reserved for **new head regressions**, not re-argument of intentional #395 claim semantics or breaking GraphQL pagination shape.

jatmn (changes requested):
One P3 documentation item from my previous closed merge set remains incomplete. The requested quarantine batching and subscription mounting-comment changes are present, and the configuration comment now describes the shared budget. The outstanding change is the corresponding sentence in `AppState`.

This is the complete requested follow-up from this review. I am not asking for another runtime, security-policy, or test-hardening round.

## Findings

### [P3] Complete the previously requested shared-budget sentence in AppState

`crates/gitlawb-node/src/state.rs:326-332`

**What is incomplete**

My previous review explicitly named two locations for the same documentation correction: the `task_read_rate_limit` configuration documentation and the `AppState::task_read_rate_limiter` field comment. The configuration comment now says that GraphQL `tasks` / `task` queries, including WebSocket operations, share the per-IP budget with REST task reads. The state comment still describes only the REST routes and says the limiter is layered on `task_read_routes` through `rate_limit_by_ip`.

The runtime already shares this limiter. Both `graphql_handler` and `graphql_ws_handler` pass a clone of `state.task_read_rate_limiter` through `TaskReadBrake`, and the task-query resolvers debit it. README and `.env.example` also describe this sharing. The missing sentence therefore leaves the documentation at the state field incomplete; it does not demonstrate a missing runtime brake.

**Why it matters**

A maintainer reading the state field in isolation sees only its REST use. That description omits a material property of the setting: REST and GraphQL task queries consume the same per-IP budget. Because the earlier request explicitly included this location, updating only configuration leaves that request partially completed.

**Root cause and requested change**

The observable cause is a partial documentation update across the two named locations: one received the shared-budget sentence and the other did not. Please finish that same correction by adding the following sentence to the existing state-field comment:

```rust
/// The same per-IP hourly budget applies to GraphQL `tasks` / `task`
/// queries and WebSocket task queries via `TaskReadBrake`.
```

Equivalent wording is fine. Keep the scope precise: this describes task queries over HTTP and WebSocket, not every GraphQL operation or event subscription. The sentence can be appended to the existing comment; no broader comment rewrite is required.

**Verification and completion criterion**

Read the updated state comment alongside `config.rs:706-719`, `.env.example:285`, and `README.md:415`. All four should describe the shared task-read budget consistently. The three already-correct locations do not need another edit. This documentation-only correction does not require new regression tests or runtime changes; normal repository checks are sufficient.

Once that sentence is present and accurate, this finding is addressed.

## Why the review has kept returning

There is not a large set of outstanding findings on this head. There is one incomplete item from the agreed follow-up. The current changes implement the other requested corrections, and I have not established a new production regression in this follow-up.

This PR spans several representations of the same behavior: implementation, configuration help, state comments, README, and example environment settings. A correction can be technically accurate where applied but leave a specifically named companion location untouched. That is what happened with this documentation request. The practical way to close it is to check every location named in the request before marking the request complete, and submit the complete correction together.

Review discipline also matters. My previous feedback explicitly established a closed merge set and reserved further rounds for new-head regressions. Additional observations about already-reviewed behavior should not automatically become new merge requirements. Repeatedly expanding that set would make the author chase a moving target; responsibility for avoiding that expansion belongs to the review as well.

For this follow-up, use the original three-item request as the completion checklist: quarantine batching is implemented, the subscription mounting comment is corrected, and the shared-budget documentation needs only its state-field sentence. This is guidance for completing the existing request, not a request to audit or rewrite every documentation surface in the repository, add a documentation framework, or open another implementation pass.

## Scope and stopping point

Please make the single state-comment correction in one small follow-up. Preserve the accepted runtime behavior, including limiter defaults and wiring, claim eligibility, cursor and scan-ceiling behavior, subscription relay logic, and client limits.

With the requested sentence added, the closed merge set is complete from this review. Any genuinely new regression introduced by a later change should be reported with its specific causal change and failure evidence; unrelated improvements should be handled separately rather than extending this PR's obligations.

beardthelion (changes requested):
Checked head `80ba7a33` against origin/main. CI is green (12/12). The incremental change since the last reviewed head (`ab054fb3`) is two lines: a shared-budget sentence added to the `AppState::task_read_rate_limiter` field comment, completing the documentation correction requested in the prior round. All four companion locations now describe the shared per-IP task-read budget consistently: `state.rs:333-334`, `config.rs:716-717`, `.env.example:285`, and `README.md:415`.

Re-ran the prior load-bearing checks on this head. `visible_tasks_tests` passed 32/32. GraphQL WebSocket tests passed 4/4, including `graphql_ws_authenticated_query_accesses_private_task`, the per-IP rate-limit enforcement, the per-request field cap, and the per-operation budget reset. Two revert probes confirmed the read gates are load-bearing: removing the quarantine check from `task_visible` goes RED on `quarantined_repo_task_withheld_from_owner_on_rest_and_graphql`, and restoring `ucan_token` to `task_to_read_json` goes RED on `ucan_token_never_appears_in_read_responses`. The runtime shares the limiter across REST and GraphQL as the new comment states.

All six inline review threads are resolved.

## Findings

- [P2] Fail closed for claim eligibility when a task's repo cannot be resolved to a locally hosted repo with established visibility.
  `crates/gitlawb-node/src/api/tasks.rs:468`
  `task_claimable` returns `true` for non-mirror repo IDs that are missing from `repos_by_id` (line 468) and for slash-form mirror repo IDs (line 465), so `claim_task` lets any signed DID claim an unassigned task whose repo visibility was never established. The claim response uses `task_to_json` (line 673), which still includes `ucan_token` (line 103), turning the claim path into a data egress for tasks that are invisible to reads. I seeded a task with `repo_id = "nonexistent-repo"` (no repo row in the database) and `ucan_token = "SECRET-UCAN-TOKEN"`, then signed `POST /api/v1/tasks/t-unresolved/claim` as an unrelated DID: the response was 200 OK with `"ucan_token":"SECRET-UCAN-TOKEN"`. Meanwhile `task_visible` returns `false` for the same unresolved repo (line 193), so `GET /api/v1/tasks/t-unresolved` returns 404. The read gate and the claim gate disagree on unresolved repos: the read gate fails closed, the claim gate fails open. Quarantine is checked in `get_claimable_task` (line 491), so quarantined repos are blocked; the gap is unresolved and mirror repos. The fix is to return `false` from `task_claimable` when the repo ID is not in `repos_by_id` (or to return `None` from `get_claimable_task` when the repo lookup fails), matching the fail-closed behavior of `task_visible`.

Not an ask, recorded only: the cursor token is content-opaque but not length-opaque. The SIV construction encrypts JSON containing variable-length `created_at` and `id` fields with an XOR keystream, which preserves plaintext length, and the base64 wire form preserves that length variation. I minted two tokens with the same timestamp but different ID lengths: a 2-character ID produced a 116-character token, a 47-character ID produced a 165-character token. When the scan ceiling is hit, `next_position` is set to the last examined candidate, which may be a hidden row, so an anonymous list caller can distinguish hidden task ID lengths from the token length. The PR body says the cursor is opaque and the caller "learns nothing from it", which overstates the guarantee: the content is opaque but the length is not. Padding the plaintext to a fixed width before encryption would close the side channel.

One process note, not a finding: open #405 retargets the `GET /api/v1/tasks` limit clamp to `[1, 200]` while this branch clamps to `[0, 200]`. Resolve or close it after this lands so the two do not fight the collector.

jatmn (approved):
lgtm

jatmn (changes requested):
I found two implementation issues on head `80ba7a33`: one database-work amplification issue to address before merge, and one low-risk MCP input-validation issue worth fixing in the same follow-up. The task-read security work has concrete value: REST and GraphQL share the visibility collectors, read responses omit UCAN tokens, quarantine takes precedence over party access, and authenticated WebSocket queries use the shared task-read brake.

Please treat the two findings below as the complete requested change set for this reviewed head. Address their root causes and regression coverage together. They do not require another redesign of task visibility, claims, pagination, or client limits.

## Merge readiness

The reviewed head is mergeable against current `main` (`bfc44f92`), with no rebase drift and all current checks passing. GitHub reports merging blocked; there is no source conflict to resolve.

The renewed request to fail-close unresolved/mirror claims conflicts with the accepted #395 open-claim contract. That is not part of this change request. Successful claim delivery of `payload`/`ucan_token` is distinct from the gated, redacted read contract.

#275 overlaps task-assignment changes but contains additional work, and #405 proposes a competing `[1, 200]` clamp. Coordinate those branches when they land; neither requires changing this PR's accepted `[0, 200]` contract. Closed #327 is the predecessor, not a replacement for this PR.

## Findings

### [P2] Bound the repository work performed by the scoped task lookup

`crates/gitlawb-node/src/db/mod.rs:1552-1556`

**Failure and impact**

`list_repos_deduped_by_ids` returns only the requested repositories, but its query still visits unrelated repository rows. With the production schema and all production repository indexes, requesting one repository among 100,000 produces both custom and generic PostgreSQL plans that discard 99,999 rows. The task get path invokes this lookup, and the list collector can repeat it across five candidate batches. Anonymous REST reads and GraphQL task queries reach this work.

The 1,000-task scan ceiling bounds task candidates; it does not bound this separate repository-table scan. The per-IP brake limits request frequency, not the work within each request. As unrelated repositories accumulate, a task read becomes more expensive even when the number of task references remains fixed. This is a demonstrated scaling defect, not a claim that a production outage has occurred.

**Root cause**

The shared CTE combines unscoped and scoped access in one predicate:

```sql
AND ($2::text[] IS NULL OR EXISTS (
    SELECT 1 FROM requested_groups requested
    WHERE requested.owner_key = (...) AND requested.name = repos.name
))
```

The requested IDs are used efficiently to construct `requested_groups`, but the outer repository access applies group membership as a filter over the broader scan. In the measured plans, the outer scan removes 99,999 rows; only the inner requested-ID lookup has the bounded primary-key index condition. The final `WHERE d.id = ANY($2)` restricts the result, after the expensive work.

This explains why a test asserting “only the requested repo was returned” passes while the operational problem remains. It also explains why batching the quarantine query, although correct and useful, does not resolve this different source of amplification. The accepted #327 request to scope repository loading needs to cover the access work as well as the returned rows.

At the merge base, task get reads `agent_tasks` directly without this repository query. The PR introduces the cost through its new visibility lookup. The existing unscoped repository-listing cost is outside this finding.

**Requested outcome and implementation guidance**

Make the scoped lookup's repository access follow the requested logical groups. A suitable approach is to give the scoped call a dedicated query shape, without the nullable “all repositories” alternative, and resolve matching group members through the existing owner-key/name index. Keep the unscoped callers on their existing all-repository path. Another formulation is acceptable if its measured access work and results satisfy the same requirements; this is not a requirement to add an index or reorganize all repository queries.

Preserve the order of operations that protects deduplication semantics: resolve the requested IDs to logical groups, consider the relevant canonical/mirror members, choose the existing survivor, then apply the final requested-ID restriction. Simply filtering the candidate rows to `id = ANY(...)` before choosing the survivor could change which row wins and is not an equivalent fix.

**Verification for this fix**

- Use a populated repository fixture with many unrelated groups and a small requested set. Run `ANALYZE`, then inspect `EXPLAIN (ANALYZE, BUFFERS)` for the actual scoped query with both custom and generic prepared plans. The reproduced case uses 100,000 groups and one requested ID.
- Verify that unrelated repository growth does not cause the scoped lookup to scan and discard the entire unrelated population. Assert the relevant access behavior rather than an exact elapsed-time threshold or one complete planner output; small-table sequential scans alone are not the bug.
- Keep the existing deduplication and visibility regressions green: canonical/mirror pairs, private canonical precedence, DID-method distinctions, deterministic ties, quarantine handling, requested-only output, and empty input. Check both existing unscoped CTE callers if shared SQL construction changes.

Authorization, quarantine-first precedence, group survivor rules, and task pagination semantics must remain unchanged.

### [P3] Reject a present non-string MCP cursor before issuing a request

`crates/gl/src/mcp.rs:1076`

**Failure and impact**

The new cursor extraction is:

```rust
args.get("cursor").and_then(|v| v.as_str())
```

A supplied number, object, array, or boolean becomes `None`. For example:

```json
{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"task_list","arguments":{"limit":1,"cursor":123}}}
```

The built MCP server sends `/api/v1/tasks?limit=1`, with no cursor. When the node returns a terminal first page, MCP reports successful first-page data with `complete: true`, rather than rejecting the invalid resume argument. The same behavior is reproduced with `{}`, `[]`, and `true` as the cursor value.

This is limited to malformed MCP calls. Valid string cursors work, and there is no demonstrated authorization bypass. I would not block merge on this P3 alone.

**Root cause**

The advertised input schema says the cursor is a string, but `tools/call` passes the arguments directly to `call_tool`; it does not validate that schema. The conversion then collapses two different states—“cursor absent” and “cursor supplied with the wrong type”—into the same first-page request. Once that distinction is lost, `fetch_tasks` cannot recover it.

The adjacent `limit` parser already preserves the distinction between absent/null and invalid present values. Apply that narrow boundary rule to this new cursor parameter. This does not call for a general MCP schema-validation framework or cleanup of pre-existing argument coercions elsewhere.

**Requested outcome and verification**

Parse the optional cursor before calling `fetch_tasks`: absent or null remains `None`, a string is passed through unchanged, and every other present type returns an error. Do not stringify objects/numbers, trim or reinterpret valid strings, or change server-side cursor validation.

Add coverage at the MCP dispatch boundary, where the mistake occurs:

- Number, object, array, and boolean cursors return an error and issue zero task HTTP requests.
- Absent and null cursors retain first-page behavior.
- A string cursor is forwarded unchanged as a URL-encoded query value with the existing filters. Invalid string-token contents continue to be handled by the node.

Keep the current limit, pagination, identity-error, and continuation tests passing. A test of `fetch_tasks` alone cannot catch this issue because it receives an already typed `Option<&str>`.

## Root-cause guidance and close-out scope

The common pattern in these two findings is that the existing checks cover the visible result without covering the boundary that determines its meaning or cost. The repository lookup returns the correct IDs while doing excessive underlying work. The MCP tool advertises the correct type while its runtime parser silently accepts a different one. Fix the boundary and add a regression that fails on the current implementation; adding more happy-path assertions around the final output will not close either gap.

This does not mean the entire feature is unstable or needs another hardening campaign. The visibility, quarantine, cursor, WebSocket, CLI/MCP, migration, and deduplication tests exercised for this review passed. These two missing assertions explain why those passing checks did not catch these particular issues. They do not establish that the same cause explains every previous review round.

Some review churn has also come from requests to revisit settled behavior. Keep those choices out of this follow-up:

- Preserve open unassigned claims for repo-less, unresolved, and slash-form mirror references, with quarantine denial and the existing locally hosted private-repository gate.
- Preserve the atomic preassigned-assignee guard and successful claim token delivery; keep UCAN tokens omitted from reads.
- Preserve `[0, 200]`, cursor contents/bindings/expiry, the scan-ceiling continuation behavior, and all existing CLI/MCP page and byte caps.
- Preserve the task-read limiter's default, disable semantics, and REST/GraphQL/WebSocket sharing. Keep publish-side subscription filtering and the existing identity-error behavior.
- Leave unrelated pin/anchor limitations and competing PR scope outside this patch.

Please deliver both fixes and their focused regression evidence together. For the SQL change, include the populated custom/generic plan evidence and preserved dedup behavior. For the MCP change, include wrong-type rejection before network access and unchanged valid/absent cursor behavior. Run the existing affected suites after both changes are in place so the shared consumers are checked together.

Any additional request on a later head should identify a concrete regression or newly verified defect, its causal change, and its smallest required correction. It should not reopen a settled product choice under a new description. The requested work here is these two bounded fixes, not an open-ended mandate to redesign adjacent behavior.

jatmn (approved):
lgtm

beardthelion (changes requested):
Verified `a0ea4c78` end to end. The scoped lookup is bounded the way it should be: `EXPLAIN (ANALYZE, BUFFERS)` on the query `list_repos_deduped_by_ids` issues over a 4000-repo fixture resolves requested groups by primary key and joins candidates through `idx_repos_owner_key_name` with no mass discard, and the same holds under a real generic plan (`PREPARE` with `force_generic_plan`). Dedup semantics are unchanged: mirror-id and quarantined-id lookups still return nothing, requesting canonical+mirror returns the canonical once, and all three callers (list, get, claim) route through the scoped path. The MCP side rejects a non-string `cursor` before any request is issued; reverting it in a scratch checkout makes the new test fail with the request actually reaching the mock server, which also proves the rejection is pre-dispatch. The quarantine and `ucan_token` revert checks from prior rounds still go red on this head, and the task-read suites are green.

On my last finding: claim eligibility for unresolved and mirror repo ids is the accepted #395 contract per jatmn's review, so I am not re-litigating it.

## Findings

- [P2] Make the plan test able to fail on the defect it certifies
  `crates/gitlawb-node/src/db/mod.rs:6743`
  `assert_scoped_repo_plan` accepts any index node on `repos` and rejects only a `Seq Scan`, but the removed defect plans as an `Index Scan` on `idx_repos_owner_key_name` that discards 3999 of 4000 rows through a filter (a hashed subplan over the group-membership `EXISTS`). I restored that exact shape and the test stayed green. Assert the index name on the join node, or bound `Rows Removed by Filter` under `EXPLAIN (ANALYZE)`. The generic half is also inert: `SET plan_cache_mode` never reaches a literal `EXPLAIN`, which always plans a custom plan, so both halves produce byte-identical output. Use `EXPLAIN (GENERIC_PLAN)` (CI's postgres:16 supports it) or `PREPARE` plus `EXPLAIN EXECUTE` instead; I ran the real generic plan and index access holds, so only the test needs to change.

- [P3] Drop the stray escapes in the dedup doc comment
  `crates/gitlawb-node/src/db/mod.rs:1602`
  The line now reads `per-caller \"/\"` with literal backslashes; the parent had `per-caller `"/"``.

Not an ask, recorded only: `.expect(2)` on the mock in `test_task_list_null_or_absent_cursor_accepted` is not enforced by mockito without a matching `.assert()`; the test still proves dispatch through the `unwrap` on the mock's 200, so it is cosmetic.

beardthelion (approved):
Verified on b360f3fe. This round's delta is the plan-test hardening plus the doc fix; I re-ran the surrounding suites rather than trusting the diff.

The plan guard now fails on the shape it exists to catch. Restoring the old membership-filter scan produces an Index Scan on idx_repos_owner_key_name that discards 3999 of 4000 rows through a hashed SubPlan filter, and the test rejects it; the previous node-type-only check passed that same plan. The custom/generic split is real this round: the test PREPAREs the production query, runs `EXPLAIN EXECUTE` under each `plan_cache_mode`, and asserts the `pg_prepared_statements` counters, so a literal EXPLAIN can no longer stand in for a generic plan. Both modes pass locally on Postgres 16.15, the dedup suite (canonical survivor, quarantine withholding, empty input) is green, and the stray escapes in the dedup doc comment are gone. CI is green on this head.

Not an ask, recorded only: the discarded-rows bound sums `Rows Removed by Filter` but not its siblings `Rows Removed by Join Filter` and `Rows Removed by Index Recheck`, and nothing asserts the plan carries `ANALYZE` instrumentation, so an edit that dropped `ANALYZE` would quietly reduce the guard to a node-type check. Two more field names and a presence assert would close it whenever the file is next touched. Also recorded, also not an ask: `create_task` still returns `e.to_string()` on a db failure while the sibling write handlers in this module now go through the opaque AppError envelope; worth the same treatment whenever the write path is next touched.

jatmn (approved):
LGTM

Issue #400, opened 2026-09-05
Silent empty results on clone recovery and changelog fetch mask failures as success


Gravirei:
## Summary

Several code paths return an empty list / empty response when a downstream call fails, without surfacing the error. The user sees a "Done" or "Recovered 0 files" message — or a 200 `{ events: [], count: 0 }` from `/changelog` — and has no way to tell whether the upstream is healthy. In one case (`clone.rs:832-841`) the file was just rewritten specifically to add a warning on the same anti-pattern; the line right below it still has the unfixed version.

### 1. `crates/gl/src/clone.rs:320-326` — first encrypted-blobs fetch

```rust
let resp = match client
    .get_signed(&format!("/api/v1/repos/{owner}/{name}/encrypted-blobs"))
    .await
{
    Ok(r) if r.status().is_success() => r,
    _ => return Ok(vec![]),
};
```

Any non-2xx (or transport error) is dropped silently. The per-blob stage in the same function (`clone.rs:368-375`) calls `warn_skip(...)`, so this fetch is the inconsistent outlier — the function otherwise surfaces per-blob failures but treats "couldn't fetch the list at all" as "no blobs to recover".

Suggested fix: match the per-blob warning style — `eprintln!("warning: encrypted-blobs fetch failed: {e}")` or equivalent — and return `Ok(vec![])` only when the response was a definite empty list.

### 2. `crates/gl/src/clone.rs:842-851` — Arweave/IPFS recovery

```rust
let from_arweave = recover_from_arweave(...)
    .await
    .unwrap_or_default();
```

This is the exact anti-pattern the lines above (`clone.rs:832-841`) were just rewritten to fix. The comment on the previous block explicitly calls out that `unwrap_or_default()` "silently swallows it into 'no paths'". The Arweave path is missing the equivalent fix:

```rust
let from_arweave = recover_from_arweave(...)
    .await
    .unwrap_or_else(|e| {
        eprintln!("warning: arweave/ipfs gateway recovery failed: {e}");
        Vec::new()
    });
```

### 3. `crates/gitlawb-node/src/api/changelog.rs:46,63`

```rust
let commits = store::log(&disk_path, &head_ref, limit).unwrap_or_default();
let prs = state.db.list_prs(&record.id).await.unwrap_or_default();
```

[...]

PR #405, opened 2026-09-05
Fix #399: GET /api/v1/tasks: clamp limit to match siblings; negative limit returns 500 with raw DB error

### LINKED ISSUE #399: GET /api/v1/tasks: clamp limit to match siblings; negative limit returns 500 with raw DB error

jialfaro:
## Summary

Fixes #399: Clamps the `GET /api/v1/tasks` query `limit` parameter to `[1, 200]` before querying the database, preventing negative values from triggering raw 500 database errors and preventing oversized limits from causing unbounded table scans.

## Changes

- **`crates/gitlawb-node/src/api/tasks.rs`**:
  - Introduced `MAX_TASK_LIMIT` (200).
  - Clamped `q.limit` to `[1, MAX_TASK_LIMIT]` (defaulting to 50 when omitted).
  - Added comprehensive `sqlx::test` test cases:
    - `tasks_negative_limit_clamped`: Seeds multiple tasks and asserts negative limit clamps to 1.
    - `tasks_limit_ceiling_clamped_to_200`: Tests boundary with `i64::MAX` (`9223372036854775807`) and asserts ceiling is clamped to 200.
    - `tasks_omitted_limit_defaults_to_50`: Asserts omitted limit returns exactly 50 tasks.

## Verification

- Tested with `sqlx::test` harness against Postgres test database. All 3 boundary tests pass.

Closes #399
/claim #399

### DIFF
--- a/crates/gitlawb-node/src/api/tasks.rs
+++ b/crates/gitlawb-node/src/api/tasks.rs
@@ -323,27 +323,29 @@ mod tests {
     #[sqlx::test]
     async fn tasks_negative_limit_clamped(pool: PgPool) {
         let state = test_state(pool).await;
-        // seed a task
+        // seed two tasks to verify clamp to 1 instead of default 50
         let now = Utc::now().to_rfc3339();
-        state
-            .db
-            .create_task(&crate::db::AgentTask {
-                id: "task-negative".to_string(),
-                repo_id: None,
-                kind: "build".to_string(),
-                status: "pending".to_string(),
-                delegator_did: OWNER.to_string(),
-                assignee_did: None,
-                capability: "repo:write".to_string(),
-                ucan_token: None,
-                payload: None,
-                result: None,
-                created_at: now.clone(),
-                updated_at: now,
-                deadline: None,
-            })
-            .await
-            .unwrap();
+        for id in &["task-negative-1", "task-negative-2"] {
+            state
+                .db
+                .create_task(&crate::db::AgentTask {
+                    id: (*id).to_string(),
+                    repo_id: None,
+                    kind: "build".to_string(),
+                    status: "pending".to_string(),
+                    delegator_did: OWNER.to_string(),
+                    assignee_did: None,
+                    capability: "repo:write".to_string(),
+                    ucan_token: None,
+                    payload: None,
+                    result: None,
+                    created_at: now.clone(),
+                    updated_at: now.clone(),
+                    deadline: None,
+                })
+                .await
+                .unwrap();
+        }
 
         let router = Router::new()
             .route("/api/v1/tasks", axum::routing::get(super::list_tasks))
@@ -360,7 +362,7 @@ mod tests {
         let bytes = axum::body::to_bytes(resp.into_body(), usize::MAX).await.unwrap();
         let body: Value = serde_json::from_slice(&bytes).unwrap();
         assert_eq!(body["count"].as_u64().unwrap(), 1);
-        assert!(body["tasks"].as_array().unwrap().len() > 0);
+        assert_eq!(body["tasks"].as_array().unwrap().len(), 1);
     }
 
     #[sqlx::test]
@@ -395,7 +397,7 @@ mod tests {
 
         let req = Request::builder()
             .method(Method::GET)
-            .uri("/api/v1/tasks?limit=5000")
+            .uri("/api/v1/tasks?limit=9223372036854775807")
             .body(Body::empty())
             .unwrap();
 
--- a/crates/gitlawb-node/src/api/tasks.rs
+++ b/crates/gitlawb-node/src/api/tasks.rs
@@ -126,14 +126,21 @@ pub async fn create_task(
     Ok((StatusCode::CREATED, Json(task_to_json(&task))))
 }
 
+const MAX_TASK_LIMIT: i64 = 200;
+
 /// GET /api/v1/tasks
 pub async fn list_tasks(
     State(state): State<AppState>,
     Query(q): Query<ListTasksQuery>,
 ) -> Result<Json<Value>, (StatusCode, Json<Value>)> {
+    // Clamp client-provided limit before it reaches SQL. Negative values and
+    // absurdly large values are both bounded here to avoid DB errors and
+    // unbounded table scans / serialization.
+    let limit = q.limit.clamp(1, MAX_TASK_LIMIT);
+
     let tasks = state
         .db
-        .list_tasks(q.status.as_deref(), q.assignee_did.as_deref(), q.limit)
+        .list_tasks(q.status.as_deref(), q.assignee_did.as_deref(), limit)
         .await
         .map_err(|e| {
             (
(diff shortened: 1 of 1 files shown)


jialfaro:
Updated test boundary assertions in `crates/gitlawb-node/src/api/tasks.rs`:
- `tasks_negative_limit_clamped` now seeds multiple tasks and asserts exactly 1 task returned, verifying clamping to 1 rather than the default limit of 50.
- `tasks_limit_ceiling_clamped_to_200` now tests against `9223372036854775807` (`i64::MAX`) verifying clamping to 200.

jatmn (changes requested):
The functional fix matches #399. The handler clamps the limit before database access, preserves the default of 50, and all three HTTP/database regression tests pass. The exact negative-limit and `i64::MAX` assertions also address CodeRabbit's test request.

There is one verified lint finding and a separate formatting failure in the added tests. I found no functional defect in the clamp. Both repairs are mechanical and can be addressed together without changing the implementation or expanding this PR's scope.

## Merge readiness

**Format the added test module.** `cargo fmt --all -- --check` exits 1 on `crates/gitlawb-node/src/api/tasks.rs`. The reported differences are:

- An extra blank line before `#[cfg(test)]` around line 307.
- Import ordering at lines 311–320.
- Line wrapping for the three `axum::body::to_bytes(...).await.unwrap()` expressions at lines 362, 406, and 451.

These are the formatter's prescribed changes, not requests to restructure the tests. Run `cargo fmt --all`, inspect the resulting diff, and confirm the formatting check passes. The immutable base version of this file passes formatting; the failures are in this PR's additions.

## Findings

### [P2] Remove the unused import so warnings-as-errors validation passes

`crates/gitlawb-node/src/api/tasks.rs:311`

The new test module contains `use super::*;`, but it does not consume any names through that import. Types and traits such as `Body`, `Request`, `StatusCode`, `Router`, `Utc`, `Value`, `PgPool`, and `ServiceExt` are imported explicitly. Each route refers to the handler through the qualified name `super::list_tasks`.

That combination leaves the wildcard import unused. The focused test build reports it as a warning, and this command fails with exit code 101:

```sh
cargo clippy --locked -p gitlawb-node --tests -- -D warnings
```

The diagnostic identifies the precise cause:

```text
error: unused import: `super::*`
  --> crates/gitlawb-node/src/api/tasks.rs:311:9
  = note: `-D unused-imports` implied by `-D warnings`
```

The PR workflow runs Clippy over all workspace targets with the same `-D warnings` policy, so it includes this test target. This is a validation blocker, not a runtime or security defect.

**Root-cause fix:** remove the unused `use super::*;` line and retain the explicit imports and qualified handler calls. There is no need for an `#[allow(unused_imports)]` annotation, a relaxed CI setting, or changes to the tests' assertions. Formatting alone will not remove an unused import, so both repairs are needed.

## Address both issues in one pass

The validation results explain how this can happen even when the regression tests pass: `cargo test` permits this compiler warning, formatting is checked separately, and Clippy promotes the warning to an error under CI's policy. In addition, the workflow runs formatting before Clippy, so a formatting failure can prevent that job from reaching the second diagnostic. A passing test run alone does not establish that these other checks pass.

Please remove the unused import, format the additions, and run the following checks on the resulting patch from the workspace root:

```sh
cargo fmt --all
cargo fmt --all -- --check
cargo clippy --locked --workspace --all-targets -- -D warnings
cargo test --locked -p gitlawb-node api::tasks::tests --no-fail-fast
```

The test command requires the project's PostgreSQL test setup. Let the normal PR workflow run its full test/build checks as well, and include the command results with the revision. This follows the repository's existing validation requirements; no additional workflow or testing framework is needed.

Preserve the regression tests' existing strength:

| Request | Seeded tasks | Expected returned tasks and count |
|---|---:|---:|
| `limit=-1` | 2 | 1 |
| `limit=9223372036854775807` | 201 | 200 |
| Omitted limit | 60 | 50 |

Those seed sizes distinguish the intended behavior from an unclamped limit or an accidental fallback to the default. Keep the tests exercising the real HTTP handler and database, along with their successful-status and response-count assertions.

The requested revision is limited to the unused import and formatter output. Keep the `[1, 200]` clamp, default of 50, filters, response shape, and test behavior intact. These findings do not call for a broader task API refactor or new functionality.

beardthelion:
Heads up on two things:

The issue this PR targets (#399) is now closed as a duplicate of #317. #317 covers the broader class: all eight `e.to_string()` map_err sites in `tasks.rs`, the unbounded `limit`, the anonymous-reachable read routes, and the 409/conflict-vs-db-failure confusion. Your limit clamp partially addresses #317 (the unbounded limit half), so I'd suggest re-scoping this PR to #317 rather than the closed #399. #317 stays open for the remaining seven handlers.

CI: `fmt + clippy` is failing. Everything else is green (test, build, audit, MSRV, docker). Can you fix the fmt+clippy issue so this can proceed? Once that's green I'll do a full review.

beardthelion (changes requested):
The clamp is right and I could not break it. Twenty-one hostile values for `limit` either fail deserialization with a 400 or land inside [1, 200], `i64::MIN` does not wrap, and gutting the clamp turns two of the three new tests red, so the fix is load-bearing rather than decorative. It mirrors `bounties.rs` closely, constant name included, and the tests drive the real handler against a real database instead of asserting on `clamp()` arithmetic.

Blocking on the CI failures jatmn already named, plus the items below.

## Findings

- [P2] Decide the negative-limit contract against #396 before either merges
  `crates/gitlawb-node/src/api/tasks.rs:364`
  #396 asserts `count == 0` for `/api/v1/tasks?limit=-1`. This asserts `count == 1`. Same URL, opposite behavior, each pinned by its own test, so whoever rebases second has to delete one of them. That call is mine, so treat it as settled: REST keeps the floor at 1, matching `bounties.rs:125` and `repos.rs:359`, and #396 conforms when it rebases. The GraphQL resolver keeps `clamp(0, 200)`, because its field description documents that floor as contract.

- [P2] Assert the 200 boundary, not only `i64::MAX`
  `crates/gitlawb-node/src/api/tasks.rs:369`
  The three tests bind that a ceiling exists, not that it is 200. An implementation capping at some higher threshold satisfies all of them while `?limit=201` still returns 201 rows. The ceiling test already seeds 201 rows, so one more request against that same fixture closes it.

- [P3] Point the description at #317 rather than the closed #399
  #399 is closed as a duplicate. #317 stays open and covers the rest of the class, including the eight sites in this file that still put raw database text into a response body. Nothing here closes those, so the description should not read as though it does.

- [P3] Use a conventional-commit prefix on the title and first commit
  `fix(node): clamp tasks query limit to [1, 200]` would do it. Releases are automated off those prefixes, so as written this lands without a version bump. Your second commit is already correct. While you are in there, drop the trailing `/claim #399` from the body.

- [P3] Rename or drop `tasks_omitted_limit_defaults_to_50`
  `crates/gitlawb-node/src/api/tasks.rs:414`
  It stays green even if the clamp line is deleted outright, because it exercises the serde default rather than anything this PR changes. It reads like clamp coverage and seeds 60 rows to prove something the clamp cannot affect.

On the clippy error specifically: `use super::*` is not a style problem, and six other api test modules use the same glob. Yours is flagged only because everything it would bring in is already named explicitly on the lines below it.

Two things I checked and am deliberately not asking you to fix, both #317's scope rather than yours: `agent_tasks` carries no index on `created_at`, so the sort makes every request a full scan whatever the limit is, and the route has no rate limiter. Worth knowing, since together they bound what this fix can buy.

beardthelion:
Correcting my own finding above, the first P2. I had it backwards.

I said REST keeps the floor at 1 and #396 conforms on rebase, reasoning from `bounties.rs` and `repos.rs`. That reasoning does not survive reading #396. It makes REST and GraphQL call one shared `collect_visible_tasks`, and that function owns the clamp, so after it lands there is no separate REST clamp for sibling precedent to apply to. One collector, one contract, `[0, 200]`, with a cursor for reading past the ceiling.

So: #396 keeps `[0, 200]` and this PR conforms to it, not the other way round. Sorry for the churn.

That has a consequence worth being straight with you about. #396 also gates these reads behind visibility rules, drops `ucan_token` from list responses, and adds the per-IP brake on the task-read routes. Between that and its own clamp, it covers what this PR does and a good deal more, and it is green and mergeable today. If it lands first, the remaining delta here is close to nothing.

Your call whether to keep this open. If you would rather not spend more time on it, closing it is reasonable and no reflection on the work: the fix was correct, the tests drive the real handler against a real database, and the review round it prompted is what surfaced the contract conflict in the first place. If you do want to keep it, the other findings above still stand and the floor becomes `[0, 200]`.

The offer from my earlier comment holds either way: `clone.rs:320-326` in #400 is unclaimed and nothing overlaps it.

Issue #95, opened 2026-06-24
Document a minimum supported Git version (blob:limit filter is version-sensitive)


beardthelion:
## Problem

`clone_repo()` in `crates/gitlawb-node/src/sync.rs` runs `git clone --mirror --filter=blob:limit=10g` in promisor mode. The `blob:limit` filter syntax is sensitive to the local Git version, and the repo pins no minimum Git version anywhere (README, `docs/RUN-A-NODE.md`, or CI). CI only ever exercises it against ubuntu-latest's Git, so any version skew is invisible to the test suite.

Surfaced during review of #67: jatmn's Git 2.53 (Windows) rejected the `--filter=blob:limit=10g` invocation, while ubuntu-latest accepts it. The failure is unrelated to #67's diff, but it exposed a real portability gap.

## What to do

- Determine the actual minimum Git version that supports the `--mirror --filter=blob:limit=<size>` form we rely on, and confirm the size-suffix (`10g`) parsing floor.
- Document that minimum in README / `docs/RUN-A-NODE.md` as a node runtime requirement.
- Consider a startup or CI check that verifies the local Git meets the floor, so a too-old (or differently-behaving) Git fails loudly instead of at clone time.

## Notes

- Code: `crates/gitlawb-node/src/sync.rs` (`clone_repo`, promisor branch).
- Last touched by #48/#69 and #40; not introduced by #67.

PR #408, opened 2026-09-07
fix(node): bound GraphQL query cost

euxaristia:
## Summary
Anonymous GraphQL documents can repeat DB-backed root fields through aliases, multiplying unpaginated database reads without a request-level work budget. This change rejects overly complex or deeply nested documents during validation, before resolvers start.

Refs #446

Note: The Windows promisor filter fix previously reverted will be tracked / landed in a follow-up PR referencing #95.

## Changes
- cap document complexity at 400 and nesting depth at 14
- assign a base complexity cost to every DB-backed query root so aliases consume the budget
- paginate visible repositories (`reposPage`) ordered by normalized owner and name with COLLATE "C"
- anti-correlate pagination fixture IDs and insertion order with full terminal page assertions

## Test plan
- `cargo test -p gitlawb-node --bin gitlawb-node graphql::`
- `cargo clippy --workspace --all-targets -- -D warnings`

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ Good today:
 
 - Local or Docker node startup.
 - Postgres-backed repo metadata.
+- Bounded GraphQL queries with [repository pagination](docs/graphql-pagination.md).
 - Bare git repository storage.
 - Git smart-HTTP clone/fetch/push.
 - RFC 9421-signed writes.
--- a/crates/gitlawb-node/src/db/mod.rs
+++ b/crates/gitlawb-node/src/db/mod.rs
@@ -6,6 +6,9 @@ use std::time::Duration;
 use tracing::info;
 use uuid::Uuid;
 
+/// Maximum visible repositories per page, plus one internal look-ahead row.
+pub(crate) const MAX_VISIBLE_REPO_PAGE_SIZE: usize = 200;
+
 // ── Public data types ─────────────────────────────────────────────────────────
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -1537,6 +1540,57 @@ impl Db {
             .fetch_all(&self.pool)
             .await?;
 
+        Ok(rows.into_iter().map(row_to_repo).collect())
+    }
+
+    /// A bounded, mirror-deduplicated page ordered by owner key and repository
+    /// name. Apply root visibility before LIMIT: private rows must not consume
+    /// page slots or influence continuation metadata. The root-rule predicate
+    /// mirrors `visibility::listable_at_root`; a differential test pins it.
+    /// Cursors only select a position and never confer read authority.
+    pub async fn list_visible_repos_page(
+        &self,
+        caller: Option<&str>,
+        after: Option<(&str, &str)>,
+        limit: usize,
+    ) -> Result<Vec<RepoRecord>> {
+        let sql = format!(
+            "{}
+             SELECT d.id, d.name, d.owner_did, d.description, d.is_public,
+                 d.default_branch, d.created_at, d.updated_at, d.disk_path,
+                 d.forked_from, d.machine_id
+             FROM deduped d
+             LEFT JOIN LATERAL (
+                 SELECT reader_dids FROM visibility_rules v
+                 WHERE v.repo_id = d.id AND v.path_glob ~ '^/*([*][*])*$'
+                 ORDER BY v.path_glob DESC LIMIT 1
+             ) root_rule ON TRUE
+             WHERE (
+                 ($2::text IS NOT NULL AND ({key}) = $2)
+                 OR CASE WHEN root_rule.reader_dids IS NULL THEN d.is_public
+                    WHEN jsonb_typeof(root_rule.reader_dids::jsonb) = 'array' THEN
+                        COALESCE(root_rule.reader_dids::jsonb ? $3::text, FALSE)
+                        AND NOT EXISTS (
+                            SELECT 1 FROM jsonb_array_elements(root_rule.reader_dids::jsonb) reader
+                            WHERE jsonb_typeof(reader) <> 'string'
+                        )
+                    ELSE FALSE END
+             )
+             AND ($4::text IS NULL OR (({key}), d.name) > ($4, $5::text))
+             ORDER BY ({key}), d.name
+             LIMIT $6",
+            Self::dedup_cte(),
+            key = OWNER_KEY_CASE_SQL,
+        );
+        let rows = sqlx::query(&sql)
+            .bind(None::<&str>)
+            .bind(caller.map(normalize_owner_key))
+            .bind(caller)
+            .bind(after.map(|(owner, _)| normalize_owner_key(owner)))
+            .bind(after.map(|(_, name)| name))
+            .bind(limit.min(MAX_VISIBLE_REPO_PAGE_SIZE + 1) as i64)
+            .fetch_all(&self.pool)
+            .await?;
         Ok(rows.into_iter().map(row_to_repo).collect())
     }
 
--- a/crates/gitlawb-node/src/graphql/mod.rs
+++ b/crates/gitlawb-node/src/graphql/mod.rs
@@ -181,11 +181,199 @@ mod tests {
         let pool = sqlx::postgres::PgPoolOptions::new()
             .connect_lazy("postgres://localhost/unused")
             .unwrap();
+        production_test_schema_with_db(Arc::new(Db::for_testing(pool)))
+    }
+
+    fn production_test_schema_with_db(db: Arc<Db>) -> GitlawbSchema {
         build_schema(
-            Arc::new(Db::for_testing(pool)),
+            db,
             tokio::sync::broadcast::channel(1).0,
             tokio::sync::broadcast::channel(1).0,
         )
+    }
+
+    #[tokio::test]
+    async fn ref_update_subscription_alias_budget() {
+        assert_subscription_alias_budget("refUpdates { repo }").await;
+    }
+
+    #[tokio::test]
+    async fn task_event_subscription_alias_budget() {
+        assert_subscription_alias_budget("taskEvents { taskId }").await;
+    }
+
+    async fn assert_subscription_alias_budget(field: &str) {
+        use futures::StreamExt;
+        use std::time::Duration;
+        let pool = sqlx::postgres::PgPoolOptions::new()
+            .connect_lazy("postgres://localhost/unused")
+            .unwrap();
+        let (ref_tx, _) = tokio::sync::broadcast::channel(16);
+        let (task_tx, _) = tokio::sync::broadcast::channel(16);
+        let schema = build_schema(
+            Arc::new(Db::for_testing(pool)),
+            ref_tx.clone(),
+            task_tx.clone(),
+        );
+        for count in [7, 8] {
+            let fields = (0..count)
+                .map(|n| format!("r{n}: {field}"))
+                .collect::<Vec<_>>()
+                .join(" ");
+            let query = format!("subscription {{ {fields} }}");
+            let mut responses = Box::pin(schema.execute_stream(query));
+            if count == 8 {
+                let response = tokio::time::timeout(Duration::from_secs(1), responses.next())
+                    .await
+                    .expect("over-budget subscriptions must reject before waiting for events")
+                    .expect("the validation error must be returned");
+                assert_eq!(response.errors.len(), 1);
+                assert_eq!(response.errors[0].message, "Query is too complex.");
+                assert_eq!(ref_tx.receiver_count() + task_tx.receiver_count(), 0);
+                continue;
+            }
+            // Poll the accepted operation far enough to register its receivers.
+            assert!(
+                tokio::time::timeout(Duration::from_millis(30), responses.next())
+                    .await
+                    .is_err()
+            );
+            assert_eq!(ref_tx.receiver_count() + task_tx.receiver_count(), 7);
+            let _ = ref_tx.send(RefUpdateBroadcast {
+                repo: "owner/repo".into(),
+                ref_name: "refs/heads/main".into(),
+                old_sha: "0".repeat(40),
+                new_sha: "1".repeat(40),
+                pusher_did: "did:key:test".into(),
+                node_did: "did:key:node".into(),
+                timestamp: "2026-01-01T00:00:00Z".into(),
+                owner_did: "did:key:test".into(),
+            });
+            let _ = task_tx.send(TaskEventBroadcast {
+                task_id: "task".into(),
+                old_status: "pending".into(),
+                new_status: "claimed".into(),
+                by_did: "did:key:test".into(),
+                at: "2026-01-01T00:00:00Z".into(),
+            });
+            let response = tokio::time::timeout(Duration::from_secs(1), responses.next())
+                .await
+                .unwrap()
+                .unwrap();
+            assert!(response.errors.is_empty(), "{:?}", response.errors);
+            assert_ne!(response.data, Value::Null);
+            drop(responses);
+            assert_eq!(ref_tx.receiver_count() + task_tx.receiver_count(), 0);
+        }
+    }
+
+    #[sqlx::test]
+    async fn production_mutation_budget_accepts_seven_and_rejects_eight(pool: sqlx::PgPool) {
+        let db = Arc::new(Db::for_testing(pool));
+        db.run_migrations().await.unwrap();
+        let caller = "did:key:reviewer";
+        let now = chrono::Utc::now().to_rfc3339();
+        for index in 0..8 {
+            db.create_task(&crate::db::AgentTask {
+                id: format!("task-{index}"),
+                repo_id: None,
+                kind: "test".into(),
+                status: "pending".into(),
+                delegator_did: caller.into(),
+                assignee_did: None,
+                capability: "test".into(),
+                ucan_token: None,
+                payload: None,
+                result: None,
+                created_at: now.clone(),
+                updated_at: now.clone(),
+                deadline: None,
+            })
+            .await
+            .unwrap();
+        }
+        let schema = production_test_schema_with_db(db.clone());
+        for count in [8, 7] {
+            let fields = (0..count)
+                .map(|n| {
+                    format!("r{n}: claimTask(id: \"task-{n}\", assigneeDid: \"{caller}\") {{ id }}")
+                })
+                .collect::<Vec<_>>()
+                .join(" ");
+            let response = schema
+                .execute(
+                    async_graphql::Request::new(format!("mutation {{ {fields} }}"))
+                        .data(crate::auth::AuthenticatedDid(caller.into())),
+                )
+                .await;
+            if count == 8 {
+                assert_eq!(response.errors.len(), 1);
+                assert_eq!(response.errors[0].message, "Query is too complex.");
+                assert_eq!(
+                    db.list_tasks(Some("pending"), None, 8).await.unwrap().len(),
+                    8
+                );
+                continue;
+            }
+            assert!(response.errors.is_empty(), "{:?}", response.errors);
+            assert_eq!(
+                response
+                    .data
+                    .into_json()
+                    .unwrap()
+                    .as_object()
+                    .unwrap()
+                    .len(),
+                7
+            );
+            assert_eq!(
+                db.list_tasks(Some("claimed"), Some(caller), 8)
+                    .await
+                    .unwrap()
+                    .len(),
+                7
+            );
+            assert_eq!(
+                db.get_task("task-7").await.unwrap().unwrap().status,
+                "pending"
+            );
+        }
+    }
+
+    #[sqlx::test]
+    async fn production_list_cost_scales_with_the_same_alias_count(pool: sqlx::PgPool) {
+        let db = Arc::new(Db::for_testing(pool));
+        db.run_migrations().await.unwrap();
+        let schema = production_test_schema_with_db(db);
+        for (field, selection) in [
+            ("refUpdates", "repo"),
+            ("tasks", "id"),
+            ("reposPage", "nodes { name }"),
+        ] {
+            for limit in [1, 200] {
+                let fields = (0..2)
+                    .map(|n| format!("r{n}: {field}(limit: {limit}) {{ {selection} }}"))
+                    .collect::<Vec<_>>()
+                    .join(" ");
+                let response = schema.execute(format!("{{ {fields} }}")).await;
+                if limit == 200 {
+                    assert_eq!(response.errors.len(), 1);
+                    assert_eq!(response.errors[0].message, "Query is too complex.");
+                    continue;
+                }
+                assert!(response.errors.is_empty(), "{field}: {:?}", response.errors);
+                assert_eq!(
+                    response
+                        .data
+                        .into_json()
+                        .unwrap()
+                        .as_object()
+                        .unwrap()
+                        .len(),
+                    2
+                );
+            }
+        }
     }
 
     #[tokio::test]
--- a/crates/gitlawb-node/src/graphql/mod.rs
+++ b/crates/gitlawb-node/src/graphql/mod.rs
@@ -177,12 +177,72 @@ mod tests {
         assert!(!err.message.contains("failed to open"));
     }
 
+    fn production_test_schema() -> GitlawbSchema {
+        let pool = sqlx::postgres::PgPoolOptions::new()
+            .connect_lazy("postgres://localhost/unused")
+            .unwrap();
+        build_schema(
+            Arc::new(Db::for_testing(pool)),
+            tokio::sync::broadcast::channel(1).0,
+            tokio::sync::broadcast::channel(1).0,
+        )
+    }
+
+    #[tokio::test]
+    async fn production_limits_reject_mutation_aliases_and_large_lists() {
+        let schema = production_test_schema();
+        for field in [
+            "claimTask(id: \"missing\", assigneeDid: \"did:key:test\") { id }",
+            "refUpdates(limit: 200) { repo }",
+            "tasks(limit: 200) { id }",
+        ] {
+            let count = if field.starts_with("claim") { 8 } else { 2 };
+            let fields = (0..count)
+                .map(|n| format!("r{n}: {field}"))
+                .collect::<Vec<_>>()
+                .join(" ");
+            let prefix = if field.starts_with("claim") {
+                "mutation"
+            } else {
+                "query"
+            };
+            let response = schema.execute(format!("{prefix} {{ {fields} }}")).await;
+            assert_eq!(response.errors.len(), 1, "{:?}", response.errors);
+            assert_eq!(response.errors[0].message, "Query is too complex.");
+        }
+    }
+
+    #[tokio::test]
+    async fn seven_root_aliases_are_accepted() {
+        struct Root(Arc<AtomicUsize>);
+        #[Object]
+        impl Root {
+            #[graphql(complexity = "50 + child_complexity")]
+            async fn repos(&self) -> Vec<Nested> {
+                self.0.fetch_add(1, Ordering::Relaxed);
+                vec![Nested]
+            }
+        }
+        let calls = Arc::new(AtomicUsize::new(0));
+        let schema = apply_query_limits(Schema::build(
+            Root(calls.clone()),
+            EmptyMutation,
+            EmptySubscription,
+        ))
+        .finish();
+        let fields = (0..7)
+            .map(|n| format!("r{n}: repos {{ value }}"))
+            .collect::<Vec<_>>()
+            .join(" ");
+        let response = schema.execute(format!("{{ {fields} }}")).await;
+        assert!(response.errors.is_empty(), "{:?}", response.errors);
+        assert_eq!(calls.load(Ordering::Relaxed), 7);
+    }
+
     #[tokio::test]
     async fn expensive_root_aliases_are_rejected_before_database_access() {
-        // QueryRoot expects a Db in the schema data. Deliberately omit it: if
-        // validation ever lets this document reach a resolver, the test fails.
-        let schema =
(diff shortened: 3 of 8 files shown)


beardthelion (changes requested):
The core mechanism is sound: `limit_complexity(400)` and `limit_depth(12)` are installed on the schema builder, and async-graphql 7.2.1 rejects documents exceeding either before any resolver runs. I verified both guards are load-bearing by replacing the limits with 1000000 in a mutation worktree: `expensive_root_aliases_are_rejected_before_database_access` and `query_depth_limit_accepts_twelve_and_rejects_thirteen` both went red. The PR's test-plan command (`cargo test -p gitlawb-node graphql::tests::`) passes with 9 tests green, matching the claim.

The complexity model has two gaps that leave the PR's stated goal, "bound GraphQL query cost," partially unmet, and the test suite doesn't prove the production builder is wired. Details below.

## Findings

- [P2] Add complexity costs to DB-backed mutation roots
  `crates/gitlawb-node/src/graphql/mutation.rs`
  The PR annotates four query roots with `#[graphql(complexity = "50 + child_complexity")]` but leaves `MutationRoot` fields at the default cost of 1. I verified with a probe that 200 aliases of a mutation field all execute: resolver count was 200, no complexity rejection, complexity score 200 (under 400). Each mutation does real DB work (inserts, updates, broadcast sends). Mutations run serially per the GraphQL spec (async-graphql calls `resolve_container` with `serial=true` for mutations), so this is not a parallel fan-out, but a single authenticated request still amplifies into 200 sequential DB writes. DIDs are permissionless, so the `require_signer` gate does not prevent amplification. Adding the same `50 + child_complexity` annotation to `create_task`, `claim_task`, `complete_task`, and `fail_task` would cap mutation aliases at 8 per request, matching the query-root budget.

- [P2] Scale list-root complexity by the limit argument
  `crates/gitlawb-node/src/graphql/query.rs:51`
  `ref_updates` and `tasks` accept a `limit: i64` argument (clamped to 200 in `tasks`) but the complexity formula is flat: `50 + child_complexity`. A query `refUpdates(limit: 200) { repo }` scores 51, the same as `limit: 1`, despite returning up to 200 rows. Seven aliases of `refUpdates(limit: 200) { repo }` score 357 (under 400) but return up to 1400 rows. async-graphql's complexity expression can reference field arguments directly, as shown in the library's own test suite (`count * child_complexity + 2`). A formula like `50 + limit * child_complexity` would charge proportionally to requested row count.

- [P2] Exercise `build_schema` in at least one limit test
  `crates/gitlawb-node/src/graphql/mod.rs:181`
  All three PR tests call `apply_query_limits(Schema::build(...))` directly, not `build_schema(...)`. I verified that removing `apply_query_limits` from `build_schema` (line 99) leaves all three tests green. The tests prove `apply_query_limits` works but don't prove the production builder calls it. A test that constructs a schema through `build_schema` with a minimal `Db` fixture, or that asserts the production schema's `SchemaInner` carries the configured limits, would close this gap.

- [P3] Add an accepted-boundary test at seven aliases
  `crates/gitlawb-node/src/graphql/mod.rs:181`
  The PR tests that eight aliases of `repos { name }` are rejected (408 > 400) but doesn't test that seven are accepted (357 < 400). I verified both directions with a probe using a cost-50 root: seven aliases passed and ran 7 resolvers, eight were rejected with "Query is too complex." and 0 resolvers. An accepted-boundary test confirms the limit isn't too aggressive and that legitimate aliased queries still work.

The depth limit test uses a synthetic recursive `Nested` type. The current public schema is flat (no recursive types), so depth greater than 12 isn't reachable today. The test proves the depth guard works but doesn't prove production schema behavior. Not an ask because the synthetic test is sufficient for the current schema shape and the depth limit is forward-looking protection.

Subscriptions (`ref_updates`, `task_events`) have default complexity (1 + child_complexity). A subscription document with many aliases is complexity-scored, but the score applies only to the initial document, not the event stream. The `ref_updates` subscription is unauthenticated by design (documented in `subscription.rs`). This is a residual architectural concern outside the PR's query-root scope, not a gap introduced by this PR.

The introspection test uses a simplified query. I have not verified whether a full GraphiQL or Apollo client introspection query would exceed complexity 400. If the project ships a GraphiQL playground, this could block legitimate introspection. Not an ask without evidence of the actual client query in use.

beardthelion (changes requested):
The prior round asked for complexity costs on mutation roots, a limit-scaled formula on list roots, a test exercising `build_schema`, and an accepted-boundary test at seven aliases. All four are addressed on this head. I verified the production wiring, the complexity limit, the depth limit, and the per-field complexity annotations are each load-bearing by removing them one at a time and running the test suite: each removal makes the matching test fail with the expected error. The full test suite passes on CI.

The remaining gap is the `repos` resolver. It materializes every visible repository with no server-side cap, yet its complexity cost is flat (50 + child_complexity), the same as a single primary-key lookup. A single `repos { name }` query scores 51 and loads the entire deduped repos table plus a batched visibility-rules query. The complexity limit bounds alias amplification but not row-count amplification, which is the dominant cost for this field.

## Findings

- [P2] Scale the `repos` complexity by a limit argument or cap the result set
  `crates/gitlawb-node/src/graphql/query.rs:14`
  The `repos` resolver calls `list_all_repos_deduped()` (a SELECT with no LIMIT) and `list_visibility_rules_for_repos` for every returned ID, but its complexity annotation is `50 + child_complexity`, the same as `task(id)` which does a single PK lookup. A single `repos { name }` query scores 51 and materializes the entire repos table. The `tasks` and `ref_updates` resolvers both clamp to 200 rows and scale complexity with the limit; `repos` does neither. Add a `limit` argument with the same clamp-and-scale pattern, or push a LIMIT into the SQL query.

- [P3] Add complexity annotations to subscription fields
  `crates/gitlawb-node/src/graphql/subscription.rs:23`
  `ref_updates` and `task_events` have no `#[graphql(complexity = ...)]`, so they default to a low per-field cost. A subscription document can alias many `refUpdates` roots inside the 400 budget, each creating a separate broadcast receiver that amplifies memory and per-event fan-out. The event stream is relay-only (no per-event DB work), which is why this is P3, but the cost model should charge for the fan-out it creates. Annotate both with `50 + child_complexity` to match every other root.

- [P3] Add an accepted-boundary test for mutation aliases
  `crates/gitlawb-node/src/graphql/mod.rs:192`
  `production_limits_reject_mutation_aliases_and_large_lists` asserts 8 `claimTask` aliases are rejected but does not assert 7 are accepted. The `seven_root_aliases_are_accepted` test uses a synthetic schema, not the production one, so no test proves the production mutation root accepts 7 aliases and rejects 8. Add the accepted side to the production test.

- [P3] Pin the scaled complexity formula with a limit comparison
  `crates/gitlawb-node/src/graphql/mod.rs:192`
  The test uses `limit: 200` with 2 aliases to force rejection, but it does not compare `limit: 1` against `limit: 200`, so the `limit.clamp(0, 200) * child_complexity` multiplier in `ref_updates` and `tasks` is not directly demonstrated. Add a case where the same alias count is accepted at `limit: 1` and rejected at `limit: 200`.

Not an ask, recorded only: the complexity and depth limits apply to the initial subscription document but not to the ongoing event stream. A single low-complexity subscription opens a stream that relays every broadcast event for the socket's lifetime. This is inherent to GraphQL subscriptions and not fixable by complexity tuning; a concurrent-connection cap or per-connection event rate limit at `/graphql/ws` would address it, but that is a separate change.

beardthelion (changes requested):
The complexity limit (400), depth limit (12), repos overflow guard, reposPage limit validation, and subscription complexity annotations are all load-bearing. Mutation-tested each by neutering the guard and running the relevant tests: all five went red, all five restored green. The SQL visibility predicate in `list_visible_repos_page` mirrors `visibility::listable_at_root` (owner check, no-rule fallback to `is_public`, root-rule reader membership, malformed array deny), pinned by the differential test across six callers and eight rule shapes. Cursor authority is position-only: the cursor is base64 JSON `(owner_did, name)` used solely in the `> ($4, $5)` keyset predicate, with the visibility predicate ANDed regardless of cursor content. CI is green, fmt and clippy clean, the full GraphQL suite passes.

Three test gaps remain, one required by the contributor rules.

## Findings

- [P2] Add a route-level GraphQL test for reposPage with an unauthorized authenticated caller
  `crates/gitlawb-node/src/graphql/query.rs:388`
  The contributor rules require a new gated handler to exercise an unauthorized authenticated caller and an anonymous caller, assert the denial each way, and assert the body leaks nothing. `repos_page_rechecks_cursor_authority_and_exact_boundary` tests the owner and an anonymous caller. `repos_page_visibility_matches_the_shared_gate` tests the DB helper for six callers but only sends a GraphQL `reposPage` request for anonymous. No test drives the resolver with an authenticated caller who is neither the owner nor in `reader_dids`. A bug in the context threading (the resolver reads `AuthenticatedDid` from context and passes it to the DB helper) would not be caught. Add a test that calls `authed(&schema, query, "did:key:zReader")` for a non-owner, non-reader caller and asserts the returned page excludes private, quarantined, and root-deny repos.

- [P3] Assert the hidden repo is excluded in the legacy repos boundary test
  `crates/gitlawb-node/src/graphql/query.rs:274`
  `repos_legacy_accepts_exactly_the_visible_bound` creates 200 public repos plus one private repo named "hidden", then asserts the response has exactly 200 entries. It does not assert that "hidden" is absent. A bug that included the private repo and dropped a public one would pass with count 200. Collect the returned names and assert "hidden" is not among them.

- [P3] Add a depth-limit test through the production build_schema path
  `crates/gitlawb-node/src/graphql/mod.rs:506`
  `query_depth_limit_accepts_twelve_and_rejects_thirteen` uses a synthetic `CountingQuery` schema with `apply_query_limits` applied directly. The complexity limit is proven through `production_test_schema()` (which calls `build_schema`), but the depth limit is not. `build_schema` calls `apply_query_limits`, so the depth limit is wired through production, but a test that sends a depth-13 query through `production_test_schema()` would close the gap and prove both limits through the same production entry point.

One process note, not a finding: `reposPage` rejects out-of-range limits with an error while `refUpdates` and `tasks` clamp them. The reject behavior is arguably better for a new field with explicit bounds, but the inconsistency is worth being aware of for client expectations.

beardthelion (changes requested):
All three asks from last round are in and verified on this head: `reposPage` now drives an authenticated non-reader through the resolver and asserts the hidden names are absent, the legacy `repos` boundary test asserts `hidden` is not among the 200, and `production_schema_enforces_depth_limit` pushes a depth-13 introspection document through `build_schema`. I re-ran the load-bearing checks on this head rather than carrying them forward: gutting `apply_query_limits` in `build_schema` turns seven tests red, and the complexity constant, depth constant, `repos` overflow guard, `reposPage` limit validation, subscription annotation, caller threading, and the SQL root-rule selection each go red under their own mutation. The GraphQL suite is 36/36 green on the head and CI is 12/12.

One contract defect remains on the new field, plus two small test pins.

## Findings

- [P2] Make the advertised 1-200 reposPage range servable
  `crates/gitlawb-node/src/graphql/query.rs:70`
  `50 + limit.clamp(1,200) * child_complexity` makes `reposPage(limit: 200)` unreachable: the cheapest selection `{ nodes { name } }` scores 450, and the doc example's shape (`nodes { name ownerDid } hasNextPage endCursor`, child cost 5) already exceeds the budget at limit 71. I ran both against the production schema: `limit: 200` returns "Query is too complex.", `limit: 70` passes validation, `limit: 71` does not. A caller following `docs/graphql-pagination.md` at the documented maximum gets a generic complexity error rather than the range error. Charging rows once, `(limit.clamp(1, 200) as usize) + child_complexity`, makes limit 200 servable and still rejects two aliases at 404; I applied it in a scratch tree and the suite stays green. Lowering the validated and documented maximum works too. Either way, alias amplification has to stay rejected.

- [P3] Extend the alias-rejection tests to every annotated root
  `crates/gitlawb-node/src/graphql/mod.rs:380`
  The production rejection tests prove the class on `claimTask`, but `task`, `create_task`, `complete_task`, and `fail_task` carry annotations nothing exercises. I deleted the attribute from `task` and from `create_task` in a scratch tree and the full GraphQL suite stayed green both times. Fold the other annotated roots into the existing rejection loops so a dropped annotation goes red.

- [P3] Pin the cursor length cap with a decodable oversized input
  `crates/gitlawb-node/src/graphql/query.rs:29`
  `repos_page_rejects_invalid_inputs_before_database_access` asserts `parse_repo_cursor` errors on `"a".repeat(4097)`, but 4097 is 1 mod 4 and fails base64 decoding before the 4096-byte check can fire; I removed the check and the test stayed green. A 4100-character input, a valid base64 length, distinguishes the cap from the decode error.

Not an ask, recorded only: a `reader_dids` value that is not valid JSON makes the `::jsonb` cast fail the whole listing (fail-loud) where the Rust gate would deny just that repo, and each accepted list call still scans and sorts the deduped table, so the budget bounds requests per document rather than work per table size. The first is reachable only through direct DB writes; the second is pre-existing shape.

One process note, not a finding: this PR and #396 touch overlapping hunks in `graphql/query.rs` and `graphql/mutation.rs`; whichever lands second will need a rebase.

beardthelion (changes requested):
The three asks from last round are in and each is now load-bearing. Reverting the `reposPage` formula to `limit * child` fails the new maximum-page test, deleting each complexity annotation fails a production-schema rejection test, and the reworked oversized-cursor test uses a valid base64+JSON payload so decoding can no longer mask the length guard. The `repos` overflow error, the `reposPage` limit range, and the subscription alias budgets all go red when removed.

## Findings

- [P2] Raise the depth cap above the canonical introspection document
  `crates/gitlawb-node/src/graphql/mod.rs:22`
  The standard getIntrospectionQuery document, the FullType/InputValue/TypeRef fragment shape that GraphiQL, graphql-js clients, and codegen send on connect, is rejected at depth 13 against the production wiring ("Query is nested too deep."). The in-tree introspection test uses a hand-trimmed document about six levels deep, so the suite stays green. At depth 14 the canonical document executes with no errors; please raise the cap and pin it with the real document rather than a trimmed subset.

- [P3] Pin the ordering contracts the pagination doc promises
  `crates/gitlawb-node/src/graphql/query.rs:64`
  `docs/graphql-pagination.md` says small legacy `repos` queries retain activity ordering, but deleting the `updated_at` sort leaves every test green; nothing asserts order. Every multi-page `reposPage` fixture also uses a single owner, so the `(owner, name)` keyset comparison is never exercised across an owner boundary. Seed repos with distinct `updated_at` values and assert the order, and add one two-owner fixture where `after` crosses the boundary.

- [P3] Pin the max-limit contract for `refUpdates` and `tasks`
  `crates/gitlawb-node/src/graphql/query.rs:110`
  `reposPage` now charges `limit` additively so the advertised 200-row page serves a multi-field selection, while the sibling list roots still multiply: `refUpdates(limit: 200) { repo refName }` scores 450 and is rejected, and only a single-field selection serves the documented max 200. Whichever contract is intended, pin it with a single-alias test so a future formula change cannot silently flip it.

- [P3] Open an issue for the declared breaking change
  `crates/gitlawb-node/src/graphql/query.rs:58`
  Commit c9757473 declares `BREAKING CHANGE: repos returns an error above 200 visible repositories`, and CONTRIBUTING.md requires an issue before breaking changes and a reference from the PR. The body states no matching issue exists; file it and link it.

Not an ask, recorded only: the budget bounds documents per request, not work per table size. Each `refUpdates` call loads the full deduped repo set, the quarantined set, and all matching visibility rules, then scans up to 2048 rows, and the `repos`/`reposPage` dedup CTE materializes the whole non-quarantined table per call; up to seven anonymous aliases fit the budget, so one request can still force several table-sized reads. Also recorded only: `reader_dids` is stored as TEXT, so a malformed value (reachable only through direct SQL, since the write path always serializes JSON) would fail the whole page where the Rust gate denies the one repo; and nothing prevents a future DB-backed root from shipping without a complexity annotation.

beardthelion (changes requested):
Verified on 9d53e130: the delta lands every ask from the last round. The canonical getIntrospectionQuery document now passes through the production schema wiring under the raised depth cap, and the boundary is pinned on both sides (mutating the cap down by one reddens the production test). The refUpdates/tasks complexity contract is load-bearing under a formula mutation. Legacy repos errors above the visible-entry bound, reposPage serves the documented maximum, and #446 covers the breaking-change ask.

Two of the new pins do not pin what they claim to. I checked each by mutating the production code and watching the suite stay green.

## Findings

- [P2] Make the owner-boundary fixture able to fail when owner ordering is dropped
  `crates/gitlawb-node/src/graphql/query.rs:319`
  Every owner-two name sorts after the owner-one cursor name (a-repo, b-repo, then c-repo, d-repo) and the only page boundary lands exactly at the owner transition. Mutating the keyset predicate and ORDER BY in list_visible_repos_page to name-only keeps this test green, so a regression dropping the owner-key component of the (owner_key, name) tuple ships silently while the documented order is "ordered by owner and name". Seed a repository under the second owner whose name sorts before the cursor name (or place a page boundary inside one owner with interleaved names) so the tuple ordering is required to produce the asserted pages.

- [P3] Anti-correlate the legacy ordering fixture so the sort key itself is pinned
  `crates/gitlawb-node/src/graphql/query.rs:282`
  updated_at, name, and creation index all ascend together, so sorting by name descending (or any correlated key) produces the asserted repo-199..repo-000 order. Removing the sort entirely does go red, but the pinned property is "most recently updated first" and the fixture cannot distinguish updated_at from a wrong key. Reversing the seed (repo-000 newest, or names descending against ascending updated_at) makes a wrong-key regression visible.

One process note, not a finding: the PR body still says the depth cap is 12 and that no matching issue exists; the head commit raises it to 14 and references #446, so the description could use a refresh.

Not an ask, recorded only: the additive reposPage cost versus the multiplicative refUpdates/tasks cost is a documented asymmetry and both are capped at the same page-size bound. The exact complexity overflow boundary for a two-child selection near the cap is unpinned, but the max-limit contract test catches ceiling changes.

beardthelion (changes requested):
Verified the round-6 asks against the new head by re-running the mutations rather than reading the diff. The name-only keyset rewrite now fails `repos_page_pagination_crosses_owner_boundary`, and a wrong legacy sort key now fails `repos_legacy_accepts_exactly_the_visible_bound`, so both fixture rewrites are load-bearing. The rest of the guard sweep still holds: depth, list-cost formulas, and production wiring all go red when reverted. One residual gap in the same fixture below.

## Findings

- [P2] Seed an earlier-keyed owner whose repo name sorts past the cursor name
  `crates/gitlawb-node/src/graphql/query.rs:320`
  The cross-owner fixture now catches name-only ordering but cannot see the two rewrites most likely to break the tuple contract it pins. Expanding the row-value predicate at `crates/gitlawb-node/src/db/mod.rs:1579` to `({key}) > $4 OR d.name > $5` (missing the `key = $4 AND` conjunct, which silently re-serves earlier owners whose names sort after the cursor name) keeps all six reposPage tests green. Ordering by `d.owner_did` instead of the normalized `({key})` is also green because every seeded owner is a did:key DID where the two orderings coincide; bare and did:web stored forms diverge. I verified the fix both directions: seeding a did:web owner with a repo named past the cursor name makes the OR-expansion re-serve that row on page 2 and shifts raw ordering off page 1, so both go red while the correct predicate and ordering stay green.

One process note, not a finding: the PR description still says the depth cap is 12 and that no matching issue exists; the head sets 14 and #446 is now linked.

beardthelion (changes requested):
Re-verified on `1de5eb9b`. The round-7 fixture fix is load-bearing: both rewrites it names now fail `repos_page_pagination_crosses_owner_boundary` by execution. The unqualified `({key}) > $4 OR d.name > $5` predicate re-serves `zz-repo` on page 2, and raw `ORDER BY owner_did` returns `[b-repo, z-repo]` instead of `[zz-repo, b-repo]` on page 1. The native git shim is a faithful port of the shell fixtures (per-test config isolation, batch-check stdin, bare-oid argv matching), the `#[cfg(unix)]` gating is symmetric with its callers, and the 503 assertions now discriminate `overloaded` from `db_unavailable` on the body code rather than the shared status.

The `test (stable)` failure on this head is a pre-existing test-isolation issue on main, not this diff. Three sweep tests advance a shared process-wide run counter without the rearm lock, this change touches none of that code, and the full suite passes locally on this head alongside a green `test (beta)` lane. It needs a re-run; the durable fix belongs in a separate change.

## Findings

- [P2] Seed the pagination fixture so row ids and insertion order do not track the asserted (owner, name) order
  `crates/gitlawb-node/src/graphql/query.rs:328`
  Mutating `ORDER BY ({key}), d.name` to `ORDER BY ({key}), d.id` at `crates/gitlawb-node/src/db/mod.rs:1580` leaves the test green: the seeded ids `r0..r4` sort exactly like the expected pages, and `created_at` is insertion-ordered the same way. Ids that do not track name order (or rows inserted out of order) turn that mutation red. Same touch: size the data set so a full-size page is also the last page, since `hasNextPage = nodes.len() == limit` currently produces identical output on every page.

- [P3] Split the Windows promisor-filter fix out of this GraphQL-scoped PR, or restate the scope
  `crates/gitlawb-node/src/sync.rs:697`
  Commit `1de5eb9b` changes production clone/fetch behavior (Refs #95) inside a PR scoped to bounding GraphQL query cost. The value is right (`4294967295` is u32::MAX, matching the "4 GiB minus one byte" comment) and the platform divergence is dormant against gitlawb remotes, which build packs server-side and ignore client filters. But nothing pins either platform value: swapping the two cfg arms stays green because every fixture uses small files. Wherever it lands, pin the constant per platform.

One process note, not a finding: the PR body still says "nesting depth at 12" and "No directly matching issue or pull request was found"; the code sets 14 and #446 is the tracker item this references. Worth an edit alongside the next push.

beardthelion (changes requested):
Both round-8 asks landed and I re-verified by execution rather than inspection. The reseeded fixture now carries inverted ids and insertion order with a full terminal page, and the promisor revert is byte-exact against main for sync.rs production code and README. The remaining test-only portability machinery (native git shim, cfg(unix) gates) is accepted as-is. On this head, mutating the keyset to id ordering, created_at ordering, raw owner_did ordering, a name-only predicate, an unqualified OR, a post-truncate hasNextPage, depth 13, an additive refUpdates formula, or removing the query limits from the production schema builder all go red against the suite. Two ordering degrees of freedom remain unpinned, below.

## Findings

- [P2] Seed the fixture so folded and recency orderings cannot reproduce it
  `crates/gitlawb-node/src/graphql/query.rs:334`
  Every seeded name is lowercase `[a-z]+-repo`, so `ORDER BY ({key}), lower(d.name)` in `list_visible_repos_page` passes all six pagination tests while the keyset predicate still compares raw `d.name`. A probe seeding one owner with `B-repo` and `a-repo` serves `a-repo` then an empty page: the raw predicate `name > 'a-repo'` excludes `B-repo` byte-wise, permanently skipping the row. Mixed-case names are storable through the API (`is_alphanumeric` admits uppercase) and the mirror path (`[A-Za-z0-9._-]+`). Separately, insertion is the exact reverse of the asserted order and `repo()` stamps `Utc::now()` per call, so `ORDER BY ({key}), d.updated_at DESC` reproduces all three asserted pages and is caught only incidentally by the shared-gate fixture. Permute the insertion order so it is not the reverse, and seed one name whose folded order differs from its byte order.

- [P3] Exercise the documented maximum against a full table
  `crates/gitlawb-node/src/db/mod.rs:1591`
  The 200-row ceiling is proven in pieces: `limit:200` acceptance on an empty table, the `MAX+1` bind clamp via `usize::MAX`, and a full terminal page only at `limit:2`. A clamp slip to `min(limit, 200)` would silently truncate the last page and stay green. Seed exactly 200 visible rows, assert `reposPage(limit:200)` returns 200 nodes with `hasNextPage=false`, then add row 201 and assert it flips true with identical page content.

One process note, not a finding: the reverted Windows promisor-filter change now has no landing spot. #95 is still open and only asks for documentation; if the u32 filter fix lands in a follow-up PR, reference it from the body here so it does not get lost.

beardthelion (changes requested):
Re-verified on `4e6488d8` by execution rather than inspection. Both round-9 asks landed: the pagination fixture is permuted with non-tracking ids and a mixed-case pair, and the documented maximum now runs against a full 200-row table. Mutating the keyset predicate to name-only, ordering by raw `d.name` or `lower(d.name)`, or slipping the fetch clamp to `min(limit, 200)` each turns a named test red, and `reposPage(limit: $l)` with a variable prices the same as a literal.

The head is red on its own new test, though, and the cause is the fixture's premise rather than the code under test.

## Findings

- [P2] Pin the collation the ordering contract depends on
  `crates/gitlawb-node/src/db/mod.rs:1580`
  `ORDER BY ({key}), d.name` carries no COLLATE, so page order follows the database default collation. The reseeded fixture at `crates/gitlawb-node/src/graphql/query.rs:341` asserts byte order (B-repo before a-repo), which holds under C.UTF-8 but not under the en_US.utf8 that the CI postgres:16 service initializes with: stable and beta both fail at the page-2 assertion with `a-repo` emitted before `B-repo`. Under en_US the pair also cannot do the job it was seeded for, since `d.name` and `lower(d.name)` order it identically there. Pin `COLLATE "C"` on both ORDER BY terms and on the keyset tuple `(({key}), d.name) > ($4, $5)` so order and resume agree on every database; as a side effect, production page order stops depending on a server locale that a collation refresh can silently change.

Not an ask, recorded only: the cargo audit lane fails on a repo-wide rustls advisory (RUSTSEC-2026-0285) unrelated to this diff, which touches no lockfile. Also recorded only: dropping `d.name` from the outer ORDER BY stays green because the dedup CTE emits `(owner key, name)`-sorted rows already, so the suite cannot see that clause. Keep the full tuple regardless; Postgres does not formally guarantee subquery order propagates.

One process note, not a finding: the reverted Windows promisor-filter change still has no landing spot. #95 remains open and documentation-scoped, so if that fix lands in a follow-up PR, reference it from the body here so it does not get lost.

euxaristia:
Addressed in f778793: pinned `COLLATE "C"` on both `ORDER BY` terms and the keyset tuple in `list_visible_repos_page` so page order and resume agree on byte order across all database server locales, making `repos_page_pagination_crosses_owner_boundary` pass consistently under `en_US.utf8` and `C.UTF-8` alike. Updated PR description with references to #446 and follow-up note for #95.

beardthelion (dismissed):
Verified on f778793f: the COLLATE "C" pins satisfy the last review ask, and each is load-bearing. I ran the graphql suite green on this head (40/40), then broke the ordering guarantees one at a time against Postgres 16: a wrong-collation substitution flips the mixed-case fixture's page order, a name-only keyset re-serves rows across the owner boundary, raw owner_did ordering drops the normalized-key contract, and a folded name ordering inverts byte order. The production wiring is still load-bearing; gutting the limit application fails the nine wiring tests.

Two small findings on this head, both reproduced against the database rather than read.

## Findings

- [P3] Reject decoded cursor strings containing NUL before they reach the query bind.
  `crates/gitlawb-node/src/graphql/query.rs:35`
  serde_json materializes \u0000 escapes into real NUL bytes in the (owner, name) tuple, and Postgres text cannot hold 0x00. The $4/$5 bind fails with 22021 and routes through graphql_db_err as an error-level "graphql database error" log plus the opaque "a database error occurred", instead of the "invalid repository cursor" this function returns for every other malformed input. Any anonymous caller can emit that log line per request. Reject a tuple element containing '\u{0}' and return the existing invalid-cursor error; stored owner and name values are ASCII-validated, so no issued cursor can contain NUL.

- [P3] Fail closed per repo on malformed reader_dids instead of aborting the whole page query.
  `crates/gitlawb-node/src/db/mod.rs:1571`
  reader_dids is TEXT with no CHECK constraint, and each ::jsonb cast raises 22P02 on non-JSON text: one visibility_rules row with malformed reader_dids fails the entire repos and reposPage listing for every caller. I reproduced it end-to-end by seeding a row and running the shipped query. The Rust gate this CASE mirrors degrades the same row to a per-repo deny (the unwrap_or_default at db/mod.rs:4161), which is also the contract the existing ELSE arm and non-string-member guard encode for valid-but-wrong JSON. Add a validity arm ahead of the cast, for example `WHEN NOT pg_input_is_valid(root_rule.reader_dids, 'jsonb') THEN FALSE` (Postgres 16 is the floor in CI and docker-compose), and seed the differential fixture with raw SQL because set_visibility_rule cannot produce the trigger.

Not an ask, recorded only: the cargo audit failure is the repo-wide rustls advisory RUSTSEC-2026-0285 and this PR touches no lockfile. Two smaller notes: the page bound caps response size, not per-request work, since each call still scans and sorts the deduped set; and the subscription alias-budget test polls receiver registration for a fixed 30ms, a mild flake risk on a stalled executor. Neither is new in this diff.

beardthelion (approved):
Both round-10 asks landed on a80845e8 and each is load-bearing, verified by execution rather than inspection. Reverting the NUL rejection in `parse_repo_cursor` turns the new cursor test red with the opaque "a database error occurred", and dropping the `pg_input_is_valid` arm makes the malformed-readers fixture abort the page query with a 22P02 cast error. The CASE ordering is right: the validity arm sits ahead of every `reader_dids::jsonb` cast, valid-but-non-array JSON falls through to the deny arm, and the owner disjunct still admits an owner ahead of rule evaluation, matching the shared gate.

The graphql suite is green on this head (41/41), and the earlier certifications still bite on it: gutting the query-limit wiring fails the nine production-schema tests, and substituting a different collation on the name ordering flips the mixed-case fixture's page order.

Not an ask, recorded only: the cargo audit failure is the repo-wide rustls advisory RUSTSEC-2026-0285; this PR touches no lockfile and the bump is already in flight on #455. Two smaller notes: `pg_input_is_valid` makes Postgres 16 an implicit floor (CI and the compose files already pin 16, but no doc says so), and a stored reader list containing a `\u0000` escape would deny on this path while the Rust gate still grants the valid sibling entries, an over-deny on pathological input rather than a leak.

PR #410, opened 2026-09-07
fix(node): bound Identify-derived Kademlia addresses

### LINKED ISSUE #429: p2p(kademlia): Unbounded address accumulation from Identify Push messages
### Summary
Every Identify event copies advertised listen addresses into Kademlia without a per-peer maximum, TTL, or replacement policy (`crates/gitlawb-node/src/p2p/mod.rs:365-371`).

### Impact
A peer advertising fresh valid multiaddrs over repeated Identify Push messages continuously inflates retained address state.

### Remediation
1. Enforce a small per-peer address maximum with eviction.
2. Apply TTL to addresses.
3. Rate-limit Identify Push processing.

### Fix
Addressed in #410.

euxaristia:
## Summary

Identify events currently copy every advertised listen address into Kademlia with no application-level lifetime or cardinality bound. This change bounds Identify-derived address state while preserving explicitly configured peer addresses.

Fixes #429

## Changes

- cap Identify-derived addresses at 8 per peer and 1,024 globally with deterministic FIFO eviction
- admit at most 8 new addresses per peer per minute and expire unrefreshed entries after 30 minutes
- canonicalize the peer suffix and reject addresses naming a different peer
- retain explicit `AddKnownPeer` addresses when an overlapping Identify lease expires
- cover per-peer and global bounds, cumulative rate limiting, expiry, and address normalization

## Test plan

- `cargo test -p gitlawb-node p2p::tests::`
- `cargo clippy -p gitlawb-node --bin gitlawb-node -- -D warnings`

## Summary by CodeRabbit

- **New Features**
  - Peer-discovered network addresses are now retained temporarily across disconnects.
  - Address tracking includes expiration, rate limits, per-peer and global capacity limits, duplicate normalization, and predictable eviction.
  - Address updates refresh existing entries and retain a stable subset when reports exceed capacity.
  - Expired or evicted discovered addresses are automatically removed, while explicitly configured addresses remain available.

- **Bug Fixes**
  - Improved validation and cleanup of peer-discovered addresses to prevent stale or invalid entries.
  - Invalid manually added peer addresses now generate a warning before being discarded.

### DIFF
--- a/crates/gitlawb-node/src/p2p/mod.rs
+++ b/crates/gitlawb-node/src/p2p/mod.rs
@@ -662,7 +662,8 @@ pub async fn start(
                             }
                         }
                         P2pCommand::AddKnownPeer { peer_id, addr } => {
-                            let Ok(addr) = addr.with_p2p(peer_id) else {
+                            let Ok(addr) = addr.clone().with_p2p(peer_id) else {
+                                warn!(%peer_id, %addr, "dropping known-peer address with a foreign peer suffix");
                                 continue;
                             };
                             explicit_addresses
--- a/crates/gitlawb-node/src/p2p/mod.rs
+++ b/crates/gitlawb-node/src/p2p/mod.rs
@@ -216,48 +216,84 @@ impl IdentifyAddressBook {
             }
         }
 
-        for address in addresses {
-            let Some(address) = address.clone().with_p2p(peer_id).ok() else {
-                continue;
-            };
+        let reported = addresses
+            .iter()
+            .filter_map(|address| address.clone().with_p2p(peer_id).ok())
+            .collect::<HashSet<_>>();
 
-            if let Some(existing) = self.peers.get_mut(&peer_id).and_then(|state| {
-                state
-                    .addresses
-                    .iter_mut()
-                    .find(|existing| existing.address == address)
-            }) {
-                existing.expires_at = now + IDENTIFY_ADDRESS_TTL;
-                continue;
+        // Refresh every address in the report before admitting anything new.
+        // This prevents an oversized report from evicting an address merely
+        // because it appeared later in the input.
+        if let Some(state) = self.peers.get_mut(&peer_id) {
+            for existing in &mut state.addresses {
+                if reported.contains(&existing.address) {
+                    existing.expires_at = now + IDENTIFY_ADDRESS_TTL;
+                }
             }
+        }
 
+        let mut new_addresses = self
+            .peers
+            .get(&peer_id)
+            .map(|state| {
+                reported
+                    .iter()
+                    .filter(|address| {
+                        !state
+                            .addresses
+                            .iter()
+                            .any(|existing| &existing.address == *address)
+                    })
+                    .cloned()
+                    .collect::<Vec<_>>()
+            })
+            .unwrap_or_else(|| reported.iter().cloned().collect());
+        // HashSet iteration is deliberately unordered. Sort new candidates so
+        // an oversized Identify report has deterministic admission behavior.
+        new_addresses.sort();
+
+        for address in new_addresses {
             if self
                 .peers
                 .get(&peer_id)
                 .is_some_and(|state| state.new_addresses >= IDENTIFY_NEW_ADDRESS_LIMIT)
             {
-                continue;
-            }
-
-            if self.address_count >= IDENTIFY_GLOBAL_ADDRESS_LIMIT {
-                let Some(evicted) = self.evict_oldest(peer_id) else {
-                    continue;
-                };
-                changes.removed.push(evicted);
+                break;
             }
 
             let evicted = self.peers.get_mut(&peer_id).and_then(|state| {
-                state.new_addresses += 1;
-                if state.addresses.len() >= IDENTIFY_ADDRESS_LIMIT {
-                    state.addresses.pop_front()
-                } else {
-                    None
+                if state.addresses.len() < IDENTIFY_ADDRESS_LIMIT {
+                    return None;
                 }
+                let index = state
+                    .addresses
+                    .iter()
+                    .position(|existing| !reported.contains(&existing.address))?;
+                state.addresses.remove(index)
             });
             if let Some(evicted) = evicted {
                 self.remove_insertion_token(peer_id, &evicted.address);
                 self.address_count -= 1;
                 changes.removed.push((peer_id, evicted.address));
+            } else if self
+                .peers
+                .get(&peer_id)
+                .is_some_and(|state| state.addresses.len() >= IDENTIFY_ADDRESS_LIMIT)
+            {
+                // Every retained address was refreshed in this report. Keep
+                // that stable subset and drop surplus new candidates.
+                break;
+            }
+
+            if self.address_count >= IDENTIFY_GLOBAL_ADDRESS_LIMIT {
+                let Some(evicted) = self.evict_oldest(peer_id) else {
+                    break;
+                };
+                changes.removed.push(evicted);
+            }
+
+            if let Some(state) = self.peers.get_mut(&peer_id) {
+                state.new_addresses += 1;
             }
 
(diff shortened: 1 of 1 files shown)


beardthelion (changes requested):
The core address-book logic is sound. I verified all six named guards are load-bearing by gutting each one in turn and running its test: per-peer cap, rate limit, global cap, TTL expiry, foreign-peer-suffix rejection, and refresh-preserving eviction all go RED when their guard is removed and GREEN when restored. CI is green (13 checks, 0 bad). The one inline thread about FIFO eviction discarding refreshed addresses is resolved in the code itself: the eviction only targets entries not in the current report, and if all entries were refreshed, it breaks instead of evicting.

This PR shares `crates/gitlawb-node/src/p2p/mod.rs` with #324, #325, and #334, so a rebase onto any of them will be mechanical.

## Findings

- [P3] Log the dropped address in AddKnownPeer
  `crates/gitlawb-node/src/p2p/mod.rs:665`
  The new `with_p2p` call on the `AddKnownPeer` path silently drops an address whose `/p2p/` suffix names a different peer. The `continue` gives no signal. A `warn!` before it would make the drop observable when this handler gains callers.

Not an ask, recorded only: `update` processes the full `listen_addrs` slice (clone, canonicalize, hash, sort) before the 8-address admission cap applies. libp2p-identify 0.47.0 caps an Identify message at 4096 bytes, so the practical input is bounded to roughly 130 addresses, but pre-truncating the slice before processing would be defense in depth.

Not an ask, recorded only: `update` inserts an empty `IdentifyPeerAddresses` entry before filtering addresses, so a peer whose report is empty or all-foreign-suffix leaves an uncounted row. These are cleaned on the 60-second expiry tick but are not bounded by the 1,024 global address cap in the meantime. Each entry is small and requires a distinct PeerId with an active connection, so the practical risk is low.

beardthelion (changes requested):
Re-reviewed at 0ffdbdb4. The previous P3 (log rejected foreign-suffix known-peer addresses) is in at p2p/mod.rs:665-667. The targeted p2p tests pass on this head. I re-ran each named guard as a vacuity sweep: rate limit, global cap, TTL expiry, foreign-suffix rejection, and the refresh-preserving break all go RED when removed. The per-peer cap eviction trigger at line 265 is vacuous in the test suite (removing it leaves the cap test green), but the cap itself is still enforced by the break at line 279.

Three bound-completeness gaps survive on this head. All are in IdentifyAddressBook::update or its eviction helpers, all are confirmed by direct reading, and all have simple fixes.

## Findings

- [P2] Bound peer entries independently of the address count
  crates/gitlawb-node/src/p2p/mod.rs:204
  update() inserts a peers entry at lines 204-210 before the with_p2p filter at line 219 drops foreign-suffix addresses. A peer sending an empty or all-foreign-suffix report creates a zero-address row that never increments address_count, so the 1,024 global cap at line 288 does not constrain the number of such rows. The 60-second cleanup tick at line 315 clones every PeerId into a Vec before iterating, so unbounded row growth also amplifies cleanup CPU. Filter the report first and only insert the peer row when reported is non-empty or the peer already exists, or add an independent cap on self.peers.len().

- [P3] Bound per-event processing before the admission cap
  crates/gitlawb-node/src/p2p/mod.rs:219
  update() clones and canonicalizes the full addresses slice into a HashSet at lines 219-222, then sorts the full new-candidate vector at line 253, before the 8-address admission loop at line 255. Per-event work is O(n log n) in the peer-controlled report size, not bounded by the 8-address retained-state limit. libp2p's 4,096-byte Identify message limit bounds the practical input to roughly 130 addresses, but the code does not truncate the input itself. Cap addresses to a small constant after the with_p2p filter, or select the first 8 sorted candidates without sorting the full set.

- [P3] Preserve refreshed addresses in global eviction
  crates/gitlawb-node/src/p2p/mod.rs:289
  The per-peer eviction at line 264-273 skips addresses in the current reported set, but the global eviction at line 288-293 calls evict_oldest (line 359), which evicts by insertion order without consulting reported. At the 1,024 global boundary, a just-refreshed address can be evicted to make room for a new one, removing it from Kademlia until the next report. Move the insertion token to the back of insertion_order when refreshing an address at line 230, so evict_oldest evicts the least-recently-refreshed address instead of the oldest-admitted one.

Not an ask, recorded only: the per-peer cap eviction trigger at line 265 (state.addresses.len() < IDENTIFY_ADDRESS_LIMIT) is vacuous in the test suite. Removing it leaves identify_addresses_are_capped_with_fifo_eviction green because the rate limit (also 8) binds first and the break at line 279 still enforces the cap. The cap works, but no test proves the eviction trigger is load-bearing. A test that starts below the cap and adds a new address (expecting growth, not eviction) would close the gap.

Not an ask, recorded only: the AddKnownPeer foreign-suffix rejection at line 665-667 is tested only in the IdentifyAddressBook path, not at the command-handler level. The code path is simple and verified by reading, but a handler-level test would match the Identify path's coverage.

beardthelion (approved):
All three requested changes from the last round are in. The early return at line 211 prevents zero-address peer rows for empty or all-foreign-suffix reports. The `take(64)` at line 208 bounds canonicalization and HashSet collection before the admission loop. The refresh logic at lines 241-243 moves each refreshed address to the back of `insertion_order` so `evict_oldest` spares it. The `warn!` at line 677 for dropped AddKnownPeer foreign-suffix addresses is present. 14/14 p2p tests pass on a fresh private target dir, and CI is green (12 checks, 0 pending).

I mutated every named guard and ran its test. The report limit, empty-report early return, refresh move-to-back, global cap, rate limit, foreign-suffix rejection, and TTL expiry each went RED when the guard was removed and GREEN after restore. The per-peer cap at line 276 is still vacuous in `identify_addresses_are_capped_with_fifo_eviction` (the rate limit, also 8, binds first), but the new `identify_addresses_grow_without_eviction_below_the_peer_cap` test is load-bearing for it: mutating the cap to `if false` makes that test fail on `changes.removed.is_empty()`.

Not an ask, recorded only: `remove_insertion_token` at line 392 does a linear scan of `insertion_order` (up to 1024 entries) per removal, called up to 24 times per update (8 refresh, 8 per-peer evict, 8 expire). This is O(24K) comparisons per event, bounded by the PR's own global cap. Not a security issue since the attacker's QUIC handshake cost exceeds the node's scan cost. An `IndexSet` or HashMap-backed index would make removal O(1) if the constants grow.

Not an ask, recorded only: the `explicit_addresses` guard that prevents Kademlia removal of AddKnownPeer addresses (lines 540-547 and 639-648) is correct by reading but has no unit test at the event-loop level. A regression removing it would pass all current tests. The AddKnownPeer foreign-suffix rejection at line 676 has the `warn!` log (the prior P3 ask) but also lacks a handler-level test.

Not an ask, recorded only: global eviction at line 299 can remove another peer's Identify-derived address from Kademlia when the global cap is full. This is by design for any bounded DHT with unbounded membership; a Sybil attacker with 128+ identities (1024/8) can displace legitimate entries. Explicit AddKnownPeer addresses are protected, and the per-peer cap and rate limit bound a single attacker.

One process note, not a finding: the `needs-issue` label is still on this PR. Link a closing issue if one exists.

euxaristia:
@beardthelion lack of issue is intentional and my standard practice for security PRs; so as to publish minimum sensitive information

beardthelion:
> @beardthelion lack of issue is intentional and my standard practice for security PRs; so as to publish minimum sensitive information

I understand the intent, but the same principle applies to the PR itself. You can limit the sensitive information in the issue just as you’ve limited it in the PR. The existence of an issue doesn’t inherently require publishing any additional details.

In this case, the issue is mainly useful for tracking: giving the finding a persistent place to reference, discuss, and track its status. The sensitive technical details can remain limited to the PR (or private communication) just as they already are.

jatmn (approved):
LGTM

PR #434, opened 2026-09-10
fix(gl): pass the mirror temp path to git as an OS string

### LINKED ISSUE #417: fix(gl): preserve non-UTF-8 temporary paths in mirror
## Problem

`gl mirror` converts its temporary clone destination through `mirror_path.to_str().unwrap()`. Unix paths can contain non-UTF-8 bytes, so a valid temporary directory selected by `TMPDIR` can make the command panic before Git is started.

Source on main at bfc44f926d08c0bf774e2c05dd76b245871294f1:
- [temporary directory and clone invocation](https://github.com/Gitlawb/node/blob/bfc44f926d08c0bf774e2c05dd76b245871294f1/crates/gl/src/mirror.rs#L69-L87)

## Reproduction to add (Unix)

1. Create a writable directory whose name includes a non-UTF-8 byte, such as an `OsString::from_vec` containing `0xff`.
2. Start `gl mirror` as a child process with `TMPDIR` set to that directory and a valid existing identity.
3. Use an ordinary valid repository name and a local Git fixture as the source.

The temporary directory creation succeeds, but converting the destination to `&str` returns `None` and the unwrap panics. Expected: pass the filesystem path to Git without requiring UTF-8.

## Suggested fix

Pass the destination as an OS-native argument:

```rust
Command::new("git")
    .args(["clone", "--mirror", &source])
    .arg(&mirror_path)
```

Keep this fix focused on local path handling. Add a Unix-only regression using a non-UTF-8 temporary directory; inject environment settings into a child command rather than mutating process-global environment in parallel tests.

## Validation status

Source-confirmed; this Unix-specific end-to-end reproduction was not executed on the Windows audit host. Severity: Low (localized CLI robustness).

beardthelion:
## Summary

`gl mirror` converted the temp clone destination through `mirror_path.to_str().unwrap()`, so a TMPDIR containing non-UTF-8 bytes panicked before git started. The destination now goes to git as an OS-native `Path`.

## Motivation & context

Closes #417

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

- `gl`: the clone destination is passed via `.arg(&mirror_path)` instead of `to_str().unwrap()`, through a small `clone_mirror` helper so a test can drive the real argv.
- Added a Unix-only regression test cloning into a `mirror-\xffdest` directory.

## How a reviewer can verify

```sh
cargo test -p gl test_clone_mirror_non_utf8_dest
```

Panics on the old `to_str().unwrap()` form (revert-checked), passes with the `Path` arg.

## Before you request review

## Protocol & signing impact

None: client-side argv construction only.

## Notes for reviewers

Overlaps #376, which rewrites the same clone call to add a `--` delimiter in front of the source. The fixes compose (OS-string destination + delimited source); whichever lands second needs a small rebase. The `// allow-unbounded-git:` marker documents that this spawn is the CLI's foreground clone, not a node request handler; it is the same spawn as before, relocated into the helper.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mirror cloning support for destinations with non-standard or non-UTF-8 paths.
  * Preserved existing error handling when mirror cloning fails.
  * Improved reliability when temporary directories use non-UTF-8 paths on Unix systems.
  * Added coverage for cloning to non-UTF-8 destinations and temporary-directory configurations.

### DIFF
--- a/crates/gl/src/mirror.rs
+++ b/crates/gl/src/mirror.rs
@@ -255,6 +255,58 @@ mod tests {
         assert!(dest.join("HEAD").exists());
     }
 
+    // #417 end-to-end: the real trigger is `std::env::temp_dir()` itself
+    // returning a non-UTF-8 path (a TMPDIR with non-UTF-8 bytes), because the
+    // mirror dest is built from it. Env is process-global, so this test
+    // re-execs the test binary in a child with a non-UTF-8 TMPDIR rather than
+    // mutating this process's env.
+    #[cfg(unix)]
+    #[test]
+    fn test_clone_mirror_under_non_utf8_tmpdir() {
+        use std::ffi::OsStr;
+        use std::os::unix::ffi::OsStrExt;
+
+        if std::env::var_os("GL_MIRROR_NONUTF8_TMPDIR").is_none() {
+            // Parent: a source repo under a normal path, a TMPDIR whose name
+            // is not UTF-8, and a child run of this same test under that env.
+            let parent = tempfile::TempDir::new().unwrap();
+            let src = parent.path().join("src.git");
+            let st = Command::new("git")
+                .args(["init", "--bare"])
+                .arg(&src)
+                .status()
+                .unwrap();
+            assert!(st.success());
+            let bad_tmp = parent.path().join(OsStr::from_bytes(b"tmp-\xff"));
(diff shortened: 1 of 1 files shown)


jatmn (approved):
lgtm

PR #437, opened 2026-09-11
fix(node): split absent ref from unreadable object in issue reads

### LINKED ISSUE #426: close_issue expects a resolved ref to stay readable, but get_issue folds every git cat-file failure into None
`close_issue` (`git/issues.rs:158`) resolves an issue ref, then asserts that reading it back cannot fail:

```rust
let full_id = match resolve_issue_id(repo_path, issue_id)? {
    Some(id) => id,
    None => return Ok(None),
};

let raw = get_issue(repo_path, &full_id)?
    .expect("ref existed in resolve but not in get — should be impossible");
```

The expectation is not guaranteed by the code it depends on.

## Why the invariant does not hold

`get_issue` returns `Ok(None)` for two different situations, and only one of them is "the ref is absent":

```rust
let cat_output = Command::new("git")
    .args(["cat-file", "blob", &ref_name])
    .current_dir(repo_path)
    .output()
    .context("failed to run git cat-file")?;

if !cat_output.status.success() {
    return Ok(None);
}
```

Any non-zero exit from `git cat-file` becomes `None`. A missing ref produces that, but so does an unreadable object, a corrupt or truncated object file, a permissions problem on the object store, a repository that has been repacked or `gc`'d between the two calls, or an ENOSPC-class failure. `resolve_issue_id` runs `cat-file -e` / `for-each-ref` at an earlier instant against the same store, so the two calls are not atomic with respect to any of that.

The result is that an operational Git failure is converted into a panic rather than an error, on a path that has a perfectly good `Result` to return it through.

## Reachability and severity

Low, and worth stating plainly rather than inflating.

The trigger is a Git or filesystem fault, not attacker-controlled input — the `issue_id` reaching `create_issue` is a server-minted UUID, and `resolve_issue_id` interpolates into a ref name and a `for-each-ref` pattern that always begins `refs/gitlawb/issues/`, so neither argument injection nor a traversal outside that namespace is available. I did not find a route that lets a caller induce the fault on demand.

What makes it worth fixing anyway is where it sits. `close_issue` runs from `POST /api/v1/repos/{owner}/{repo}/issues/{id}/close`, inside a request handler, while a write guard is held — so the failure mode for a degraded repository is a panicking task rather than a 500, and unwinding out of that path is a worse way to discover a corrupt object store than an error is.

## Fix direction

Preserve the distinction `get_issue` currently collapses. Separate "ref absent" from "read failed": inspect the `cat-file` exit status and stderr, return `Ok(None)` only for the absent case, and propagate everything else as an `Err`. Then `close_issue` uses `?` and the `expect` disappears rather than being replaced by a different assertion.

`list_issues` has the same shape at `git/issues.rs:96` — a failed `cat-file` silently drops the issue from the listing — so the same change is worth applying there, where the current behavior is to under-report rather than to panic.

## Validation status

Verified by reading the code and tracing both callers of `get_issue`. Not reproduced against a corrupt repository; the claim is that the invariant is unenforced, not that a specific fault has been observed to trigger it.

## Disclosure note

Originally submitted through private vulnerability reporting and filed publicly at the maintainer's direction. Public filing is appropriate given the trigger is an operational fault rather than attacker-controlled input, and no exploitation route was demonstrated.

Related but distinct: #350 covers the panic class in `gl` (fixed-width slicing of node-supplied strings) and #417 covers the non-UTF-8 mirror path. This is the node-side sibling.

---

Found during an external audit pass.

beardthelion:
## Summary

`get_issue` folded every `cat-file` failure into `Ok(None)`, so a corrupt or unreadable object store looked identical to a missing issue. A new `read_issue_blob` helper re-resolves the ref with `rev-parse` (which does not read the object) when `cat-file` fails: unresolvable means genuinely absent, resolvable means a real read failure that propagates as an error.

## Motivation & context

Closes #426

The fold produced three bad outcomes: `get_issue` answered a fake 404 for a corrupt store, `list_issues` returned a 200 that silently dropped the unreadable issue, and `close_issue` could reach an `expect` that panicked while the handler held the write guard.

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

- `gitlawb-node`: `read_issue_blob` splits absent-ref from unreadable-object; `get_issue` and `list_issues` propagate read errors; `close_issue` drops the `expect` and returns `Ok(None)` when the ref vanished between resolve and read.
- fn-level tests: missing issue, corrupt object (errors on get/list/close instead of panicking), normal listing.
- handler-level tests (`corrupt_store_tests`): real bare repo + deleted object through the actual handlers. Corrupt store: get -> 500 not 404, close -> 500 not panic/404, list -> error not a partial 200. Absent issue still 404.

## How a reviewer can verify

```sh
cargo test -p gitlawb-node git::issues::
cargo test -p gitlawb-node corrupt_store_tests
```

The handler tests fail on the unfixed fold (404s and a 200 with `{\"issues\": []}`); verified by temporarily restoring `Ok(None)` on the failure path.

## Before you request review

## Protocol & signing impact

None: error-classification change inside the node; no wire or signature surface.

## Notes for reviewers

Open-PR overlap: #285 and #196 both rewrite the `close_issue` handler and append tests in `api/issues.rs`; #196 also inserts after `create_issue` in `git/issues.rs`, next to where `read_issue_blob` lands. Semantically independent, but whichever lands second will want a rebase. #79 touches the same handler region; #193 is a mode-only touch.

The two new git spawns in `read_issue_blob` carry `// allow-unbounded-git:` markers: they extend this module's pre-existing `Command::new(\"git\")` convention, the second runs only on an already-failed read, and callers hold the repo guard. Happy to route through the bounded runner if a maintainer prefers, but that is a wider refactor of the module's six existing spawns.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of corrupted or unreadable Git issue data.
  * Issue retrieval and listing now report internal Git errors instead of incorrectly returning 404 responses or silently omitting issues.
  * Closing an issue no longer panics when its reference is invalid or unreadable.
  * Missing issue references continue to return 404 responses.
  * Invalid or unsafe issue references are handled safely.

### DIFF
--- a/crates/gitlawb-node/src/api/issues.rs
+++ b/crates/gitlawb-node/src/api/issues.rs
@@ -449,3 +449,198 @@ mod lock_pool_shed_tests {
         );
     }
 }
+
+/// #426: end-to-end at the handler seam. A corrupt object store behind a
+/// resolving issue ref used to look identical to an absent issue: get returned
+/// a fake 404, close returned a fake 404 after releasing the write guard, and
+/// list silently under-reported. All three must now surface a git error.
+#[cfg(test)]
+mod corrupt_store_tests {
+    use super::*;
+    use axum::response::IntoResponse;
+    use sqlx::PgPool;
+    use tempfile::TempDir;
+
+    fn seed_repo(owner_did: &str, name: &str) -> crate::db::RepoRecord {
+        let now = Utc::now();
+        crate::db::RepoRecord {
+            id: Uuid::new_v4().to_string(),
+            name: name.to_string(),
+            owner_did: owner_did.to_string(),
+            description: None,
+            is_public: true,
+            default_branch: "main".to_string(),
+            created_at: now,
+            updated_at: now,
+            disk_path: format!("/tmp/{name}"),
+            forked_from: None,
+            machine_id: None,
+        }
+    }
+
+    /// A state whose repo store roots in `repos_dir`, so the test can lay down
+    /// a real repo and then corrupt it without touching /tmp.
+    async fn state_with_repos_dir(pool: &PgPool, repos_dir: &std::path::Path) -> AppState {
+        let mut state = crate::test_support::test_state(pool.clone()).await;
+        state.repo_store =
+            crate::git::repo_store::RepoStore::for_testing(repos_dir.to_path_buf(), pool.clone());
+        state
+    }
+
+    /// Bare repo at the store's expected path, holding one issue ref.
+    fn bare_repo_with_issue(
+        repos_dir: &std::path::Path,
+        owner: &str,
+        name: &str,
+    ) -> (std::path::PathBuf, String) {
+        let repo_path = repos_dir
+            .join(owner.replace([':', '/'], "_"))
+            .join(format!("{name}.git"));
+        std::fs::create_dir_all(&repo_path).unwrap();
+        let out = std::process::Command::new("git")
+            .args(["init", "--bare"])
+            .arg(&repo_path)
+            .output()
+            .unwrap();
+        assert!(out.status.success());
+        let issue_id = "aaaabbbb-0000-0000-0000-000000000000".to_string();
+        git_issues::create_issue(
+            &repo_path,
+            &issue_id,
+            &serde_json::json!({
+                "id": issue_id,
+                "title": "t",
+                "status": "open",
+            })
+            .to_string(),
+        )
+        .unwrap();
+        (repo_path, issue_id)
+    }
+
+    /// Delete the blob object behind the issue ref and put garbage in its
+    /// place: `cat-file` fails while `rev-parse` still resolves the ref.
+    fn corrupt_issue_object(repo_path: &std::path::Path, issue_id: &str) {
+        let out = std::process::Command::new("git")
+            .args(["rev-parse", &format!("refs/gitlawb/issues/{issue_id}")])
+            .current_dir(repo_path)
+            .output()
+            .unwrap();
+        let oid = String::from_utf8(out.stdout).unwrap().trim().to_string();
+        let obj = repo_path.join("objects").join(&oid[..2]).join(&oid[2..]);
+        std::fs::remove_file(&obj).unwrap();
+        std::fs::write(&obj, b"garbage").unwrap();
+    }
+
+    /// The regression: closing over a corrupt store used to return a fake 404
+    /// (or panic deeper in). It must reach the caller as a git 500, and the
+    /// handler must return rather than unwind.
+    #[sqlx::test]
+    async fn close_issue_on_corrupt_store_returns_git_error_not_panic(pool: PgPool) {
+        let owner = "did:key:zISSUECORRUPTCCCCCCCCCCCCCCCCCCCCCCCCC";
+        let dir = TempDir::new().unwrap();
+        let state = state_with_repos_dir(&pool, dir.path()).await;
+        state
+            .db
+            .create_repo(&seed_repo(owner, "corrupt-close"))
+            .await
+            .expect("seed repo");
+        let (repo_path, issue_id) = bare_repo_with_issue(dir.path(), owner, "corrupt-close");
+        corrupt_issue_object(&repo_path, &issue_id);
+
+        let err = close_issue(
+            State(state),
+            Extension(AuthenticatedDid(owner.to_string())),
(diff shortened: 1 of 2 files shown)


jatmn (changes requested):
I found one integration issue that needs to be addressed before this is ready. The node-side distinction between an absent issue and an unreadable object is useful; the listing failure also needs to reach the terminal user correctly.

## Finding

### [P2] Preserve the new listing failure through `gl issue list`

Changed producer: `crates/gitlawb-node/src/git/issues.rs:209`  
API boundary: `crates/gitlawb-node/src/api/issues.rs:108-109`  
Affected consumer: `crates/gl/src/issue.rs:219-237`

**Trigger and user-visible result**

Consider a repository containing two issue refs: one points to a healthy JSON blob, and the other points to a missing or unreadable blob.

| Stage | Before this PR | With this PR |
|---|---|---|
| Git issue listing | Skips the unreadable object and returns the healthy issue | Propagates the read failure |
| Node API | Returns a partial HTTP 200 listing | Returns HTTP 500 with an error body |
| `gl issue list` | Displays the healthy issue | Prints `No issues for <owner>/<repo>` and exits successfully |

The new server response is the intended correction. The problem is how the shipped CLI consumes it: the user receives an authoritative-looking empty result instead of a failed listing. This also gives scripts a successful exit status when the repository could not be listed.

**Root cause and attribution**

`cmd_list` calls `NodeClient::get_authed`, immediately parses the response as JSON, and then evaluates:

```rust
let issues = resp["issues"].as_array().cloned().unwrap_or_default();
```

`get_authed` and its underlying `get`/`get_signed` methods return HTTP responses without rejecting non-success statuses. The node's HTTP 500 body contains `error` and `message`, not `issues`, so the expression above becomes an empty array. The following branch prints `No issues` and returns `Ok(())`.

The unchecked-status code predates this PR. The concrete connection to this change is the new fail-whole-list response: for the same mixed healthy/unreadable repository, it changes the terminal result from a visible healthy issue to a false empty success. This is a consumer integration gap exposed by the server correction.

**Requested outcome and implementation guidance**

Please make `gl issue list` distinguish a failed HTTP request from a successful issue collection before entering the empty-list/display logic. A failed response should produce a useful command error and a non-success exit; it must never reach the `No issues` success branch.

A status check in `cmd_list`, following the local approach already used by `cmd_show` and `cmd_close`, is one small way to address the cause. Check the response status independently of whether its body parses as the expected JSON. Preserve useful failure context when available, and let a non-JSON error body fail rather than defaulting it to an empty collection. Avoid matching one particular Git error message: the distinction being enforced is unsuccessful HTTP response versus successful listing.

Keep the node's new error propagation. The successful `200 {"issues":[]}` case should continue to mean that the repository has no issues, and a successful nonempty listing should render normally. A change confined to this command can satisfy the finding; changing shared `NodeClient` behavior would require checking its other callers and is not necessary here.

**Focused regression coverage**

The existing `test_cmd_list_empty` and `test_cmd_list_with_issues` cover successful responses. Add failure coverage at the same command seam:

- HTTP 500 with the node's `{ "error": "git_error", "message": "..." }` shape must return a command error. If exercising the CLI process, assert nonzero exit and absence of the `No issues` success output.
- A non-success response with a non-JSON body must also fail, so body parsing cannot convert an HTTP failure into success.
- Retain the successful empty and nonempty listing tests to prove the fix preserves their behavior.

On the node side, a mixed healthy/unreadable fixture should assert that listing fails as a whole. The existing corrupt-store tests establish the intended failure response; pairing that response with the command-level regression covers the boundary where it is currently lost. A single full-stack test is not required if these focused tests establish both sides of the same contract.

## Guidance for closing this out

The gap here spans Git storage, HTTP error mapping, and CLI interpretation. The server tests can correctly prove “unreadable object becomes HTTP 500” while the client tests independently pass using only HTTP 200 fixtures. Neither proves that the user is told the listing failed.

To complete this fix, carry the same failure scenario through the affected consumer and test the final distinction: a failed listing is visibly unsuccessful, while a successfully empty repository is visibly empty. That gives this PR a concrete completion criterion and addresses the cause of this finding without accumulating special cases for individual Git diagnostics.

This review has one actionable finding. The guidance above explains how to resolve that finding completely; it does not request a broader client refactor or additional changes to unrelated commands or storage behavior.

PR #438, opened 2026-09-11
fix: warn or error instead of returning silent empty results

### LINKED ISSUE #400: Silent empty results on clone recovery and changelog fetch mask failures as success

beardthelion:
## Summary

Four sites folded failures into empty results indistinguishable from genuine emptiness: two silent recovery degrades in `gl clone` and two folds in the changelog endpoint that turned a degraded repo or DB into a 200 with an empty timeline.

## Motivation & context

Closes #400

A caller cannot tell "nothing recovered" from "recovery failed" or "no events" from "the store could not be read". The clone paths keep best-effort behavior but now warn; the endpoint now errors.

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

- `gl`: the encrypted-blobs list fetch warns on a non-2xx or transport error instead of `_ => return Ok(vec![])`; both `run()` recovery arms go through a new `warn_recovery_failed` helper on `emit_warning` (the first arm's raw `eprintln!` moves onto the test-mirrored write).
- `gitlawb-node` changelog: `store::log` failures map to a git error and `list_prs` failures propagate instead of `unwrap_or_default`.
- `store::log` itself only returns empty when the ref genuinely does not resolve: a resolving ref whose `git log` fails (corrupt object, mid-read gc) is now an error, checked via `rev-parse` which resolves the name without reading objects.
- Tests: mockito-500 list fetch warns and returns empty; `warn_recovery_failed` warns and returns empty; `store::log` empty-vs-corrupt unit test; three endpoint tests (corrupt store -> 500, dropped `pull_requests` table -> 5xx, healthy commit -> 200 with the event, empty repo -> 200 empty).

## How a reviewer can verify

```sh
cargo test -p gl encrypted_blobs_list_failure warn_recovery_failed
cargo test -p gitlawb-node log_empty changelog
```

Both endpoint failure tests returned a 200-empty on the unfixed folds (verified by restoring `unwrap_or_default` / the plain `Ok(vec![])`).

## Before you request review

## Protocol & signing impact

None: error-propagation and warning changes only.

## Notes for reviewers

- `repos.rs` keeps its own `unwrap_or_default` on `store::log`; the same Err still folds to an empty commits list there, so that endpoint's behavior is unchanged. Flagging in case a maintainer wants it hardened too.
- Open-PR overlap: #376 edits `recover_encrypted_blobs` / `recover_from_arweave` bodies in `clone.rs` (different lines than the match arms changed here); several PRs append to the `store.rs` tests mod. `changelog.rs` has only #193's file-mode flip.
- The `rev-parse` recheck in `store::log` carries an `// allow-unbounded-git:` marker: it extends the module's existing `Command::new("git")` convention and runs at most once per already-failed `git log`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Changelog retrieval now reports Git and database failures instead of showing an empty history.
  * Git history errors, including repository and configuration issues, are surfaced correctly.
  * Empty repositories and missing references continue to return an empty history as expected.
  * Encrypted-blob listing failures now provide warnings while preserving graceful empty results.
  * Recovery failures now emit warnings instead of failing silently.

### DIFF
--- a/crates/gitlawb-node/src/api/changelog.rs
+++ b/crates/gitlawb-node/src/api/changelog.rs
@@ -43,7 +43,10 @@ pub async fn get_changelog(
         .await
         .map_err(|e| AppError::Git(e.to_string()))?;
     let head_ref = store::resolve_head(&disk_path, &record.default_branch);
-    let commits = store::log(&disk_path, &head_ref, limit).unwrap_or_default();
+    // A read failure is not an empty history: returning a bare 200 with no
+    // events makes a degraded repo look identical to a brand-new one (#400).
+    let commits =
+        store::log(&disk_path, &head_ref, limit).map_err(|e| AppError::Git(e.to_string()))?;
 
     let mut events: Vec<serde_json::Value> = commits
         .into_iter()
@@ -60,7 +63,9 @@ pub async fn get_changelog(
         .collect();
 
     // ── Merged PRs ───────────────────────────────────────────────────────
-    let prs = state.db.list_prs(&record.id).await.unwrap_or_default();
+    // Same for the DB half: an outage must surface as an error (503 when the
+    // pool is unreachable), not an empty timeline (#400).
+    let prs = state.db.list_prs(&record.id).await?;
     for pr in prs.iter().filter(|p| p.status == "merged") {
         events.push(serde_json::json!({
             "type": "pr_merged",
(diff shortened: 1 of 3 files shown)


jatmn (changes requested):
I found one incomplete-fix case that needs to be addressed before this is ready.

## Findings

### [P2] Distinguish malformed branch refs from genuinely empty history

`crates/gitlawb-node/src/git/store.rs:155–158`

This behavior also exists on the base branch. It belongs within #400 because that issue targets damaged repositories being reported as successful empty changelogs, and this PR changes the failure classifier responsible for distinguishing those outcomes.

**Failure path and impact**

Start with a populated bare repository whose HEAD points to `refs/heads/main`, verify its commit history is readable, then replace that branch ref's contents with `garbage\n`. The commit objects remain on disk, but the ref metadata is damaged:

1. `git log HEAD` exits 128 with “your current branch appears to be broken”.
2. `git rev-parse --verify --quiet HEAD` exits **1**, with empty stderr.
3. The new `Some(1)` arm returns `Ok([])`, discarding the failed history read.

The endpoint's resolver does not prevent this: HEAD and preferred-branch resolution fail, `for-each-ref` ignores the broken ref, and `resolve_head` falls back to HEAD. Consequently, `get_changelog` receives a successful empty commit list. With a healthy database and no merged PRs, it returns 200 with `events: []` and `count: 0`. If merged PRs exist, it can present a successful partial timeline with the commits omitted. Callers cannot distinguish this damaged history from genuine emptiness.

**Root cause and requested correction**

The classifier equates “the ref did not resolve” with “the ref is genuinely absent/unborn”. Quiet `rev-parse` gives the same exit code for both absence and malformed existing refs; an empty diagnostic stream does not distinguish them either. The added object-corruption test does not exercise this ambiguity because its ref remains resolvable. The non-repository test exercises a different probe exit status.

Please correct this distinction in the `store::log` failure path. Return an empty history only when the check establishes a genuinely missing/unborn ref; preserve a read failure for damaged ref metadata. Choose a check that retains that distinction rather than treating another unsuccessful lookup as proof of absence. The existing changelog `map_err` can then propagate the error through its current response handling.

The requested outcome is reliable failure classification and reporting. Keep normal ref-selection behavior, healthy-history output, and successful empty/unborn responses compatible. This finding does not require repairing repository data, changing other endpoints, or introducing a new API response format.

**Regression coverage**

- Add a populated-repository test that corrupts the branch ref, exercises `resolve_head` followed by `store::log`, and requires an error. Keep the commit objects intact so the test specifically covers ref metadata.
- Extend the existing changelog endpoint tests with that fixture and a healthy PR database; require the existing Git-error response rather than 200/empty. This verifies the failure reaches the caller.
- Retain the healthy/unborn, corrupt-object, and non-repository cases so correcting this ambiguity preserves the behaviors already covered by this PR.

PR #440, opened 2026-09-11
fix(core): cap the blinded recipient scan in open_blob (#419)

### LINKED ISSUE #419: open_blob trial-decrypts every recipient entry with no cap, so envelope authors set the work
`open_blob` (`crates/gitlawb-core/src/encrypt.rs:135`) scans the envelope's recipient list and attempts one unwrap per entry. The list length comes from the envelope and nothing bounds it.

## Why the scan is necessary, and why that makes the bound necessary

The scan is not a defect. Recipient identities are deliberately blinded — the module's own comment says so:

> Identities are blinded: no entry says which recipient it belongs to, so try each one. The ChaChaBox AEAD tag authenticates, so exactly the reader's own entry unwraps; every other entry fails cleanly.

That is the right design for the confidentiality property it buys. The consequence is that the work is linear in a field the envelope author chooses, and each iteration is an X25519 scalar multiplication rather than something cheap. Because the cost cannot be avoided without giving up the blinding, it has to be capped instead.

## Where the length comes from

`decode` reads a `u32` header length, parses the header as JSON, and deserializes `recipients` into a `Vec<Recipient>` with no length check (`encrypt.rs:121-131`). Each entry is a few hundred bytes of base64, so the entry count reachable within a given envelope size is large.

The loop then runs before any other validation:

```rust
for entry in &header.recipients {
    // base64 decode eph / nonce / wrap, then:
    let abox = ChaChaBox::new(&eph, &my_x);
    if let Ok(ck) = abox.decrypt(/* … */) { content_key = Some(ck); break; }
}
```

A non-recipient runs the full list before returning "not a recipient of this envelope", so the worst case is the common case for anyone who is not the intended reader.

## Reachability

Envelopes move between nodes over IPFS during replication, and the mirror is explicitly not expected to validate them — `sync.rs` notes that "the mirror only ever handles ciphertext and never decrypts". So an envelope reaching an opener has not been vouched for by anything upstream.

I have not traced a specific end-to-end route from an unauthenticated remote to a particular `open_blob` call, so treat the reachability claim as argued rather than demonstrated. The bound is worth having regardless of which callers turn out to be reachable, since it costs nothing.

## Distinct from #291

#291 bounds the recipient loop in `git/visibility_pack.rs`, which is the **sealing** side and whose cost the repo owner controls. This is the **opening** side in `gitlawb-core`, and its cost is controlled by whoever produced the envelope. Same shape, different loop, different trust boundary. Fixing one does not fix the other.

## Fix direction

Reject an oversized recipient list when the header is decoded, before any scalar multiplication runs. The ceiling should come from the largest recipient set the write path can legitimately produce — `encrypted_pin` / `visibility_pack` know that number — with headroom, rather than from a guess.

Note the interaction with the fixed-width framing rationale in `scan_token.rs`: rejecting on count leaks nothing that the envelope's own length does not already leak, since the entries are fixed-size.

## Validation status

Verified by reading the code. The per-iteration cost is asserted from what an X25519 scalar multiplication costs in general, not measured against this build. A measurement would sharpen the argument for where to set the ceiling.

---

Found during an external audit pass. Duplicate-checked against open and closed issues; no existing coverage found, but the search used a limited keyword set.

beardthelion:
## Summary

Closes #419. `open_blob` trial-decrypts every entry in the envelope's blinded recipient list, one X25519 exchange each, and the envelope author controls the count. The scan itself is by design (the AEAD tag authenticates the reader's own entry), so the bound has to be a cap: a header carrying more than `MAX_RECIPIENTS = 256` entries is rejected right after decode, before any crypto. Real envelopes carry a handful of authorized readers, so 256 is far above legitimate use.

## Test plan

- `cargo test -p gitlawb-core` — 100/100 pass.
- New `over_cap_recipient_list_is_rejected_before_the_scan` reframes a valid envelope with 257 junk recipient entries and asserts the cap error; an at-cap (256) envelope reaches the scan and fails as "not a recipient" rather than tripping the cap.
- Revert check: with the cap removed, the 257-entry envelope runs the full scan and returns the plain not-a-recipient error, so the test is load-bearing.
- `cargo test --workspace` green; `cargo clippy --workspace --all-targets -- -D warnings` clean via the pre-push gate.

## Notes

- Overlap: #314 edits the `my_x` setup a few lines below this insertion and appends tests at the same mod anchor; the changes compose, and whichever lands second rebases.
- Seal side untouched: `seal_blob` takes the caller's own recipient set.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a safety limit for recipients when sealing data.
  * Oversized recipient lists are rejected before encryption or serialization.

### DIFF
--- a/crates/gitlawb-core/src/encrypt.rs
+++ b/crates/gitlawb-core/src/encrypt.rs
@@ -75,6 +75,14 @@ pub fn seal_blob(plaintext: &[u8], recipients: &[VerifyingKey]) -> Result<Vec<u8
     if recipients.is_empty() {
         return Err(anyhow::anyhow!("seal_blob: no recipients"));
     }
+    // Same bound open_blob enforces: an envelope over the cap could never be
+    // opened, so refuse to seal it (#419).
+    if recipients.len() > MAX_RECIPIENTS {
+        return Err(anyhow::anyhow!(
+            "seal_blob: {} recipients, over the {MAX_RECIPIENTS} cap",
+            recipients.len()
+        ));
+    }
     let mut content_key = [0u8; 32];
     OsRng.fill_bytes(&mut content_key);
     let body_cipher = XChaCha20Poly1305::new_from_slice(&content_key)
@@ -358,6 +366,20 @@ mod tests {
         assert!(
             !err.to_string().contains("over the"),
             "at-cap envelope must reach the scan, got: {err}"
+        );
+    }
+
+    #[test]
+    fn seal_refuses_a_recipient_list_over_the_cap() {
+        // An envelope over the cap can never be opened, so seal must reject it
+        // instead of minting one.
+        let keys: Vec<VerifyingKey> = (0..MAX_RECIPIENTS + 1)
+            .map(|_| Keypair::generate().verifying_key())
+            .collect();
+        let err = seal_blob(b"blob", &keys).unwrap_err();
+        assert!(
+            err.to_string().contains("recipients"),
+            "expected the recipient-cap error, got: {err}"
         );
     }
 }
--- a/crates/gitlawb-core/src/encrypt.rs
+++ b/crates/gitlawb-core/src/encrypt.rs
@@ -49,6 +49,13 @@ use serde::{Deserialize, Serialize};
 const MAGIC: &[u8] = b"GLENC";
 const VERSION: u8 = 2;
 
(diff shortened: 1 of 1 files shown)


jatmn (approved):
LGTM

PR #441, opened 2026-09-11
test(node): cover the component walk in validated_repo_disk_path (#411)

### LINKED ISSUE #411: Only the allowlist layer of validated_repo_disk_path is covered: the component walk has no test that reaches it
`validated_repo_disk_path` has three layers and only the first is covered. The
component walk, which is the layer written specifically so a static analyser can
see the sanitisation, has no test that reaches it.

The function (`crates/gitlawb-node/src/git/repo_store.rs`) is:

1. `validate_path_components`, the allowlist on `owner_did` and `repo_name`
2. a `starts_with(repos_dir)` containment check
3. a `Component` walk rejecting `ParentDir` and `CurDir`

The three existing tests (`local_path_resolves_safe_inputs`,
`local_path_rejects_traversal_in_repo_name`, `local_path_rejects_traversal_in_owner_did`)
all drive traversal through the arguments, so layer 1 refuses every one of them
and layers 2 and 3 never run.

## Evidence

Replacing the `ParentDir` arm with an empty body leaves the whole suite green.
Removing the `validate_path_components` call, by contrast, turns those tests red
immediately, which is what tells the two apart: layer 1 is load-bearing, and the
walk is currently unfalsifiable rather than proven.

That does not make the walk useless. It is real depth against a future caller
that reaches the join by another route. It does mean nobody would notice if it
were deleted or inverted, and it is the layer the comment above it singles out as
the one CodeQL is meant to recognise, so it is the one most likely to be
"tidied" by someone who cannot see what it is for.

## Not a vulnerability

Layer 1 rejects `..`, a leading `.` or `-`, an empty string, and anything outside
`[A-Za-z0-9._:-]` before either later layer is reached. I could not construct an
input that gets past it, and I am not claiming one exists. This is a coverage
gap, not a live traversal.

## What would close it

A test that reaches layers 2 and 3 has to come in through `repos_dir` rather than
through the two validated arguments, since that is the one input the allowlist
never sees. Call `validated_repo_disk_path` with a `repos_dir` that itself carries
a `..` component and a legitimate owner and name, and assert it errors. That
input is not reachable from a request today, which is exactly why it is the input
that exercises the depth.

## Related

#302 is the nearest neighbour and a different property. Same shape though: a
layer verified by reading while the tests on either side of it stay green.

beardthelion:
## Summary

Closes #411. `validated_repo_disk_path` has three layers and the existing tests only exercised the first: every traversal input goes through `owner_did`/`repo_name`, which `validate_path_components` refuses before the `starts_with` check and the `Component` walk run. The walk was unfalsifiable; the issue demonstrated that emptying the `ParentDir` arm left the suite green.

The walk is reachable through `repos_dir` itself: the allowlist never sees it, `PathBuf::join` does not normalize it, and the component-wise `starts_with` still matches the unnormalized prefix, so `ParentDir`/`CurDir` land on the walk. Three tests: `repos_dir` containing `..` is rejected by the `ParentDir` arm, a relative `repos_dir` starting with `.` is rejected by the `CurDir` arm (interior `.` segments are dropped by `components()` and never reach the walk, so the leading-dot relative path is the shape that covers it), and a clean `repos_dir` resolves.

## Test plan

- `cargo test -p gitlawb-node repo_store` passes, including the three new tests.
- Load-bearing check: with both walk arms emptied, the two reject tests fail while the clean-dir control still passes; restored, all three pass.
- `cargo test --workspace` green (one unrelated coalesced-push announce test flaked under concurrent-suite load and passed on an isolated rerun). `cargo clippy --workspace --all-targets -- -D warnings` clean via the pre-push gate.

## Notes

- Overlap: #285 changes `make_store`'s signature and touches this tests module at different anchors; these tests call `validated_repo_disk_path` directly, so no compile dependency and no shared lines. The test-module addition in #196 sits on a different base at a different anchor; adjacency risk only.

## Summary by CodeRabbit

- **Tests**
  - Added coverage verifying that repository directory paths with unsafe path components are rejected.
  - Added coverage confirming that clean directory paths are accepted.

### DIFF
--- a/crates/gitlawb-node/src/git/repo_store.rs
+++ b/crates/gitlawb-node/src/git/repo_store.rs
@@ -1832,6 +1832,39 @@ mod tests {
         }
     }
 
+    // #411: layers 1-3 of validated_repo_disk_path were only reachable through
+    // arguments, and layer 1 refused them all, so the Component walk ran on no
+    // test. Drive it through repos_dir: the allowlist never sees it, join does
+    // not normalize it, and the component-wise starts_with still matches the
+    // unnormalized prefix, so ParentDir/CurDir land on the walk itself.
+    #[test]
+    fn component_walk_rejects_parent_dir_in_repos_dir() {
+        let dir = TempDir::new().unwrap();
+        let repos_dir = dir.path().join("a").join("..");
+        assert!(
+            validated_repo_disk_path(&repos_dir, "did:key:z6MkAlice", "hello").is_err(),
+            "repos_dir containing '..' must be rejected by the component walk"
+        );
+    }
+
+    #[test]
+    fn component_walk_rejects_cur_dir_in_repos_dir() {
+        // Interior '.' segments never reach the walk: components() drops them.
+        // A leading '.' on a relative repos_dir is the one that survives.
+        let repos_dir = Path::new(".").join("repos");
+        assert!(
+            validated_repo_disk_path(&repos_dir, "did:key:z6MkAlice", "hello").is_err(),
+            "relative repos_dir starting with '.' must be rejected by the component walk"
+        );
+    }
+
+    #[test]
+    fn component_walk_accepts_clean_repos_dir() {
+        let dir = TempDir::new().unwrap();
+        let path = validated_repo_disk_path(dir.path(), "did:key:z6MkAlice", "hello").unwrap();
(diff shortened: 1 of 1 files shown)


jatmn (approved):
LGTM

PR #442, opened 2026-09-11
fix(git-remote): stop waiting on EOF for a delete-only push

### LINKED ISSUE #369: git push --delete hangs forever: the helper reads receive-pack to EOF, but a delete-only push sends no pack
## What happens

`git push origin --delete <branch>` against a gitlawb remote never returns. I killed it at five minutes; there is no error, no timeout, no output — it simply sits there.

Reproduced against `https://origin.gitlawb.com` with `git-remote-gitlawb` built from `feat/ucan-push-authorization`, git 2.54.0.windows.1:

```sh
git push origin probe/slash-test:refs/heads/probe/slash-test   # succeeds
git push origin --delete probe/slash-test                      # hangs forever
```

The branch creation immediately before it worked, so this is specific to deletion rather than to the remote, the auth, or the ref name.

## Why (likely)

`handle_connect` reads the entire receive-pack request with `read_to_end` on stdin, and the comment above it states the assumption directly (`crates/git-remote-gitlawb/src/main.rs:294-296`):

> git-receive-pack (push): Git sends ref-update commands + the complete PACK blob, then closes its write pipe. read_to_end is safe and correct here, a single POST.

That assumption holds for a push that carries objects. **A delete-only push carries none** — git sends the ref-update command (`<old-oid> 0000000…0 refs/heads/x`) plus a flush, and no pack follows, because there is nothing to transfer. If git then holds its write pipe open waiting for the report-status response before closing it, `read_to_end` waits for an EOF that only arrives after a response that is never sent, and the two deadlock.

I have not instrumented this to prove it, so treat the mechanism as a strong hypothesis rather than a finding. The hang itself is reproducible.

## Why it matters

Deleting a remote branch is routine — merged feature branches, cleaning up after CI, tidying a fork. Today it does not fail, which would at least be honest; it hangs, so a script doing it has no natural recovery and a person has to notice and interrupt it.

## Suggested direction

Stop depending on EOF to know the request is complete: parse the pkt-line stream and stop at the flush packet that ends the command list, then decide whether a pack follows based on whether any command has a non-zero new-oid. `negotiate_upload_pack` already does per-round pkt-line framing rather than reading to EOF (`read_exact` on the 4-byte length, flush handling at `main.rs:671-701`), so the machinery exists in this file.

A test that drives `handle_connect` with a delete-only command stream and asserts it returns would pin it. The existing harness at `main.rs:1367` already feeds a synthetic stdin cursor.

## Environment

- `git-remote-gitlawb` @ `fad75a6` (also present on `main` — this code is untouched by open PRs)
- git 2.54.0.windows.1, Windows 11
- node `origin.gitlawb.com`, v0.7.1

beardthelion:
## Summary

`git push --delete` against a `gitlawb://` remote deadlocked because the helper read stdin to EOF for the receive-pack request, while git holds the pipe open for the report-status response after a delete-only command list. The request is now delimited by the protocol's own boundary: the flush after the command list.

## Motivation & context

Closes #369

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

- git-remote-gitlawb: `read_receive_pack_request` reads the command list's pkt-lines to its terminating flush, then reads the pack to EOF only when a command has a non-zero new-oid. A delete-only push (all-zero new-oids) posts immediately instead of waiting for an EOF that never comes. Push-cert blocks read past their own flush since only a command-bearing block ends the request. Malformed pkt-line lengths bail, same strictness as `read_upload_pack_round`.
- The integration test shim now serves receive-pack (advertise + POST piped to real `git receive-pack --stateless-rpc`).

## How a reviewer can verify

```sh
cargo test -p git-remote-gitlawb
```

The regression proof is `real_git_push_delete_completes`: a real `git push --delete` through the built helper against the shim completes in one POST and deletes the ref server-side. On the old `read_to_end` shape it hangs for the full 30s deadline (verified by reverting the function body).

## Before you request review

## Protocol & signing impact

Notes: this changes how the helper FRAMES the receive-pack request it reads from git, not the wire bytes it sends: the POST body is byte-identical (commands, flush, pack when present). The request boundary moved from EOF to the command flush, which is what the protocol specifies.

## Notes for reviewers

`BlockAfterSeed` (the live-pipe reader from the earlier multi-round-fetch work) reproduces the hang in a unit test; a Cursor would EOF and hide it. EOF is still honored so git versions that close the pipe early keep working.

## Summary by CodeRabbit

* **Bug Fixes**
  * Delete-only pushes now complete without waiting for unnecessary pack data.
  * Pushes that include updates continue to process pack data correctly.
  * Push certificates and push options are now handled correctly during receive-pack operations.

* **Tests**
  * Added end-to-end coverage for delete-only pushes, pack processing, push certificates, push-option framing, and stateless RPC handling.
  * Verified branch deletion using a real Git push.

### DIFF
--- a/crates/git-remote-gitlawb/src/main.rs
+++ b/crates/git-remote-gitlawb/src/main.rs
@@ -535,40 +535,39 @@ fn read_upload_pack_round<R: Read>(stdin: &mut R) -> Result<(Vec<u8>, RoundEnd)>
 }
 
 /// Read one git-receive-pack request: the command list's pkt-lines (each
-/// re-emitted into the body, up to and including the terminating flush), then
-/// the pack ONLY when a command carries a non-zero new-oid.
+/// re-emitted into the body, up to and including the terminating flush), the
+/// negotiated extra sections, then the pack ONLY when a command carries a
+/// non-zero new-oid.
 ///
 /// A delete-only push sends no pack, and git holds its write pipe open waiting
 /// for the report-status response, so a plain read_to_end deadlocks the two
-/// (#369). A push-cert block, when advertised, precedes the command list as its
-/// own flush-terminated pkt-lines, so a flush only ends the request after at
-/// least one command line has been seen. EOF at any point also ends the read,
-/// covering git versions that close the pipe early.
+/// (#369). When the negotiated capabilities include them, two more
+/// pkt-line sections sit between the command flush and the pack: a
+/// push-options block (terminated by its own flush) and a push-cert block
+/// (terminated by a "push-cert-end" pkt-line, not a flush). Both are consumed
+/// here, since git writes them before it ever blocks on the response. EOF at
+/// any point also ends the read, covering git versions that close the pipe
+/// early.
 fn read_receive_pack_request<R: Read>(stdin: &mut R) -> Result<Vec<u8>> {
-    let mut body = Vec::new();
-    let mut saw_command = false;
-    let mut pack_expected = false;
-    loop {
+    /// Read one pkt-line into `body` (length prefix + data), returning its data
+    /// or `None` on flush or EOF. Malformed lengths bail rather than collapse
+    /// into a synthesized flush (#192 F5).
+    fn read_pkt<R: Read>(stdin: &mut R, body: &mut Vec<u8>) -> Result<Option<Vec<u8>>> {
         let mut len_bytes = [0u8; 4];
         match stdin.read_exact(&mut len_bytes) {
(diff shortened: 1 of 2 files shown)


beardthelion:
@greptileai review

jatmn (approved):
LGTM

jatmn (approved):
LGTM

PR #443, opened 2026-09-11
fix(gl): refuse signed requests to a plaintext remote

### LINKED ISSUE #413: gl signs requests to any GITLAWB_NODE with no scheme check, so a plaintext remote node sends the signature in cleartext
`gl` sends RFC 9421 signed requests to whatever `GITLAWB_NODE` names, with no
check on the scheme. Pointed at a remote node over `http://`, the signature and
the request body both go out in cleartext.

## Verified by execution

Against `origin/main` at bfc44f92, with the built binary:

```
$ GITLAWB_NODE=http://10.0.0.36:7777 RUST_LOG=debug gl whoami
DEBUG reqwest::connect: starting new connection: http://10.0.0.36:7777/
DEBUG hyper_util::client::legacy::connect::http: connecting to 10.0.0.36:7777
Error: agent lookup failed: GET http://10.0.0.36:7777/api/v1/agents/did:key:z6Mks...
```

`10.0.0.36` is this machine's LAN address, so the hop is off-loopback. It failed
only because nothing was listening. No warning, no refusal.

## Why it is reachable

Every `--node` argument is declared with the same shared environment fallback,
for example `crates/gl/src/agent.rs:20`:

```rust
#[arg(long, default_value = "https://node.gitlawb.com", env = "GITLAWB_NODE")]
```

`NodeClient::get_authed` (`crates/gl/src/http.rs:102`) routes to `get_signed`
whenever a keypair is present, and `get_signed` builds its URL as
`format!("{}{}", self.node_url, path)` with no scheme inspection anywhere in
between.

The default is `https://`, so a stock install is not affected. What makes this
worth fixing is that `GITLAWB_NODE` is shared with `git-remote-gitlawb`, whose
own module docs and `--help` teach `GITLAWB_NODE=http://my-node:7545` as the way
to point at a self-hosted node. An operator following that instruction downgrades
every `gl` command at the same time, and nothing says so.

## Why cleartext matters here specifically

RFC 9421 authenticates a request; it does not encrypt one. On a plaintext hop the
body is readable and so is the `Signature` header. #253 records that a signed
request is replayable for roughly 600s and is not bound to a target host, which
turns a captured header into a usable credential for that window. Plaintext moves
that from a compromised-endpoint problem to a passive-observer one.

## Fix direction

The chokepoint is `NodeClient`, not the ~40 call sites: every path already goes
through `new` / `with_timeout`. The obstacle is that `new` returns `Self` rather
than `Result`, so refusing there is a signature change across its callers.

Two shapes worth weighing before anyone writes code:

1. Refuse a non-loopback `http://` node, with an explicit opt-in environment
   variable for a trusted private network. Fails closed, and matches what the git
   transport should do. Breaking for anyone already pointing `gl` at a plaintext
   remote node.
2. Warn once per invocation and proceed. No breakage, weaker guarantee.

Whichever is chosen, the loopback test should read the parsed address rather than
compare strings. `is_loopback_url` in `crates/gl/src/doctor.rs:345` compares
against four literals (`localhost`, `127.0.0.1`, `[::1]`, `::1`) and so misses
`127.0.0.2`, the IPv4-mapped `[::ffff:127.0.0.1]`, and the numeric spellings that
`Url` normalises at parse time.

## Related

#253 (replay window, no host binding) is the issue that sets the severity here.
#336 covers signed writes across a redirect, a neighbouring property on the same
surface. #367 and its PR #372 specify RFC 9421 for reimplementers and say nothing
about transport security at all, which is the same gap one level up.

beardthelion:
## Summary

`gl` sent RFC 9421-signed requests to whatever `GITLAWB_NODE` named, including a remote `http://` host, where the signature and body travel in cleartext and a captured signature is replayable inside its freshness window. Signed calls now refuse a plaintext remote unless the operator opts in, matching the guard the remote helper already carries for the same hop.

## Motivation & context

Closes #413

## Kind of change

- [x] Bug fix
- [ ] Feature
- [x] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

- gl: `is_insecure_remote` flags `http://` to a non-loopback host, deciding loopback from the parsed address so `127.x`, `[::1]` and IPv4-mapped literals are all recognized as this machine.
- `ensure_signing_transport` fails closed at the three signing sites (`get_signed`, `get_maybe_signed`, `send_once`), before the request is built or sent. `GITLAWB_ALLOW_INSECURE_HTTP` (presence) is the opt-in, the same variable and semantics as the remote helper's guard.
- Unsigned calls are unaffected: with no keypair there is no signature to leak and the request proceeds.

## How a reviewer can verify

```sh
cargo test -p gl
```

End to end with the built binary:

```sh
gl identity new --dir /tmp/glid --force
GITLAWB_NODE=http://node.example.com gl star add o/r --dir /tmp/glid   # refuses
GITLAWB_ALLOW_INSECURE_HTTP=1 GITLAWB_NODE=http://node.example.com gl star add o/r --dir /tmp/glid   # attempts the request
GITLAWB_NODE=http://127.0.0.1:9 gl star add o/r --dir /tmp/glid        # attempts the request
```

## Before you request review

## Protocol & signing impact

Notes: signatures are not modified; the change is whether a signed request may leave for a plaintext hop. Existing dev flows against `http://localhost` are unaffected.

## Notes for reviewers

The refusal reads the env var at request time, so the opt-in works per invocation. The env-mutating tests run under the suite's existing env lock.

## Summary by CodeRabbit

- **Security**
  - Signed requests to HTTP destinations routed through configured proxies are protected against unintended off-machine loopback access.
  - `NO_PROXY` entries are matched consistently regardless of letter casing, including exact and suffix matches.
  - Loopback HTTP requests remain allowed when the destination is covered by `NO_PROXY`.
  - Secure HTTPS requests continue to work.
  - Insecure HTTP access can be explicitly enabled through the `GITLAWB_ALLOW_INSECURE_HTTP` setting.
  - Unsigned requests remain unaffected.

### DIFF
--- a/crates/gl/src/http.rs
+++ b/crates/gl/src/http.rs
@@ -116,9 +116,10 @@ fn loopback_http_host(url: &str) -> Option<String> {
 /// Refuse to sign a request destined for a cleartext hop off this machine
 /// unless the operator has opted in. `GITLAWB_ALLOW_INSECURE_HTTP` exists for a
 /// private LAN where the operator has decided that is acceptable; its presence
-/// alone opts in, matching git-remote-gitlawb's guard for the same hop. The
+/// alone opts in, matching the variable git-remote-gitlawb's guard uses for
+/// the same hop. This guard is stricter than the helper's in one respect: the
 /// loopback exemption does not apply when a proxy would carry the request
-/// off-machine anyway; `proxy_env` is the proxy state the client was built
+/// off-machine anyway. `proxy_env` is the proxy state the client was built
 /// with, so the guard judges the same routing the client will use.
 fn ensure_signing_transport(
     node_base: &str,
--- a/crates/gl/src/http.rs
+++ b/crates/gl/src/http.rs
@@ -271,9 +271,11 @@ impl NodeClient {
     /// to be authenticated. Mirrors the conditional signing of post/put/delete.
     pub async fn get_maybe_signed(&self, path: &str) -> Result<reqwest::Response> {
         let url = format!("{}{}", self.node_url, path);
+        if self.keypair.is_some() {
+            ensure_signing_transport(&self.node_url, insecure_http_allowed(), &self.proxy_env)?;
+        }
         let mut req = self.inner.get(&url);
         if let Some(kp) = &self.keypair {
-            ensure_signing_transport(&self.node_url, insecure_http_allowed(), &self.proxy_env)?;
             let signed = sign_request(kp, "GET", path, b"");
             req = req
                 .header("Content-Digest", signed.content_digest)
@@ -348,6 +350,9 @@ impl NodeClient {
         proof: Option<&str>,
     ) -> Result<reqwest::Response> {
         let url = format!("{}{}", self.node_url, path);
+        if self.keypair.is_some() {
+            ensure_signing_transport(&self.node_url, insecure_http_allowed(), &self.proxy_env)?;
+        }
         let mut req = self
             .inner
             .request(method.parse().expect("valid HTTP method"), &url)
@@ -355,7 +360,6 @@ impl NodeClient {
             .body(body.to_vec());
 
         if let Some(kp) = &self.keypair {
-            ensure_signing_transport(&self.node_url, insecure_http_allowed(), &self.proxy_env)?;
             let signed = sign_request(kp, method, path, body);
             req = req
                 .header("Content-Digest", signed.content_digest)
--- a/crates/gl/src/http.rs
+++ b/crates/gl/src/http.rs
@@ -64,32 +64,23 @@ fn is_insecure_remote(url: &str) -> bool {
 }
 
 /// True when `no_proxy` covers `host`: a `*` entry, an exact match, or a
-/// domain suffix match (`example.com` covers `a.example.com`).
+/// domain suffix match (`example.com` covers `a.example.com`). Entries match
+/// case-insensitively, as DNS names are.
 fn no_proxy_covers(host: &str, no_proxy: &str) -> bool {
     no_proxy
         .split(',')
         .map(str::trim)
-        .map(|e| e.trim_start_matches('.'))
+        .map(|e| e.trim_start_matches('.').to_ascii_lowercase())
         .filter(|e| !e.is_empty())
         .any(|e| e == "*" || host == e || host.ends_with(&format!(".{e}")))
 }
 
 /// True when a plaintext request to a loopback `host` would still leave this
-/// machine. reqwest reads HTTP_PROXY/ALL_PROXY (and lowercase) at client
-/// build; a configured proxy without a NO_PROXY entry for the host carries the
-/// signed request off-machine in cleartext, which is the leak this module
-/// guards against.
-fn loopback_goes_off_machine(host: &str) -> bool {
-    let proxied = ["HTTP_PROXY", "http_proxy", "ALL_PROXY", "all_proxy"]
-        .iter()
-        .any(|k| std::env::var_os(k).is_some_and(|v| !v.is_empty()));
-    if !proxied {
-        return false;
-    }
-    let no_proxy = std::env::var("NO_PROXY")
-        .or_else(|_| std::env::var("no_proxy"))
-        .unwrap_or_default();
-    !no_proxy_covers(host, &no_proxy)
+/// machine under the proxy env the client was built with. A configured proxy
+/// without a NO_PROXY entry for the host carries the signed request
+/// off-machine in cleartext, which is the leak this module guards against.
+fn loopback_goes_off_machine(host: &str, proxy_env: &(bool, String)) -> bool {
+    proxy_env.0 && !no_proxy_covers(host, &proxy_env.1)
 }
 
 /// The normalized host when `url` is an `http://` loopback address, else None.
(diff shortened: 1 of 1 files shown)


jatmn (dismissed):
LGTM

beardthelion:
@greptileai review

jatmn (changes requested):
I found one security issue that needs to be addressed before this is ready, plus a nonblocking documentation correction.

## Merge readiness

The reviewed head is `d6ffeaf9`, based on current `main` at `bfc44f92`. GitHub reports no merge conflicts and all reported checks pass. Approval is required; the earlier review is dismissed. There is no current rebase requirement. #414 is complementary work for `git-remote-gitlawb`, remains open, and does not replace this `gl` fix.

## Findings

### [P1] Make the signing guard agree with the actual proxy route

`crates/gl/src/http.rs:69-75`, with the related host transformation at line 95 and proxy-state capture at lines 217-223.

**Failure and impact**

With an identity loaded, `HTTP_PROXY=http://<remote-proxy>:3128`, `NO_PROXY=*`, `GITLAWB_NODE=http://127.0.0.1:7545`, and `GITLAWB_ALLOW_INSECURE_HTTP` unset, the new guard permits signing. The locked HTTP stack nevertheless routes that request through the proxy. A controlled proxy received the `Signature` header from an actual `NodeClient::get_signed` call under this configuration. A remote plaintext proxy on that route can observe the signature and any request body. The signature is not bound to the target host, so the exposure also retains the replay risk within its freshness window.

This is an **incomplete fix for the existing cleartext exposure**, not a claim that this PR introduced the original leak. The PR also introduces the reverse failure: some supported exclusions that keep requests on loopback are now rejected.

**Root cause**

There are two independent interpretations of routing. `reqwest::Client` builds its proxy policy, while `NodeClient` stores a proxy-presence boolean and raw `NO_PROXY` string and interprets them through a new string matcher. Capturing the environment at construction fixes the earlier request-time snapshot problem, but it does not make the two matchers equivalent.

In locked reqwest 0.12.28, the system matcher delegates to hyper-util 0.1.20. That implementation parses IP addresses and networks separately from domains, and numeric destinations are checked only against its IP matcher. The wildcard is a domain entry there. The new `no_proxy_covers` instead applies wildcard and domain-suffix matching to every host. Separately, `loopback_http_host` removes a trailing dot from the host used by the guard, while the request sent to reqwest retains it.

With `HTTP_PROXY` configured and the override unset:

| Node host | `NO_PROXY` | New guard | Actual transport decision |
| --- | --- | --- | --- |
| `127.0.0.1` | `*` | Allows signing | Uses proxy; signature captured |
| `localhost.` | `localhost` | Allows signing | Uses proxy; signature captured |
| `127.0.0.1` | `.127.0.0.1` | Allows signing | Uses proxy; signature captured |
| `127.0.0.1` | `127.0.0.0/8` | Refuses | Bypasses proxy |
| `[::1]` | `0:0:0:0:0:0:0:1` | Refuses | Bypasses proxy |
| `localhost.` | `localhost.` | Refuses | Bypasses proxy |
| `127.0.0.1` | `127.0.0.1` | Allows signing | Bypasses proxy |

The leaking rows were exercised through actual signed requests. The direct-routing cases were checked against the locked dependency's matching behavior; the guard's false refusals were also reproduced. The earlier CIDR and trailing-dot comments therefore remain valid. These cases belong to one finding because they share the same independent-routing-model defect.

**Requested root-cause correction**

Please make the safety decision and the client’s effective routing agree. Without an override, a signed plaintext request must not reach an off-machine proxy, and supported exclusions that actually route directly to loopback must continue to work. A change that only handles `*`, lowercases another string, or adds CIDR recognition would leave the other demonstrated disagreements unresolved.

Prefer one authoritative routing decision used consistently by transport configuration and the guard. The implementation mechanism is yours to choose; the requirement is the observable result, not a new abstraction or a particular dependency. If considering explicit direct routing for signed loopback traffic, account for the existing proxy contract rather than applying `.no_proxy()` to every request indiscriminately. A blanket refusal of all proxied environments would also leave the demonstrated safe-local regressions unresolved.

**Validation that binds the fix to routing**

Use a controlled proxy and local server so the test observes where the request actually goes, rather than asserting only what `no_proxy_covers` returns. For the leaking cases, the proxy must receive no signed request; either a clear refusal or an intentionally configured direct local request can satisfy the safety requirement. For supported direct exclusions, assert that the local server receives the signed request successfully. Include exact IP, wildcard/IP, dotted localhost, CIDR, and equivalent IPv6 cases from the table.

Exercise `get_signed`, keyed `get_maybe_signed`, and signed writes through `send_once`; all three consume the same guard. Keep proxy environment cases in isolated child processes, and retain the existing iCaptcha retry tests so retries cannot escape the transport decision. Preserve positive controls for HTTPS, ordinary direct loopback, unsigned calls, and the explicitly enabled insecure override.

The scope is the `NodeClient` routing/guard disagreement. This finding does not request changes to signature format, server authorization, redirect policy, unrelated CLI error handling, or override-value parsing. It also does not require implementing or merging #414.

### [P3, nonblocking] Correct the claim that the remote helper already has this protection

`crates/gl/src/http.rs:118-122`.

The new comment says that `git-remote-gitlawb` already uses this variable and guard, then describes `gl` as stricter for proxied loopback traffic. The helper in this head and current `main` has neither the transport guard nor `GITLAWB_ALLOW_INSECURE_HTTP` handling: its startup constructs `repo_base` and enters the helper, whose signed request paths have no such check. The protection lives in open PR #414. Moreover, #414's current implementation disables proxying for loopback HTTP, making the strictness comparison stale relative to that proposed implementation too.

The root cause is describing another branch's proposed behavior as an existing contract. The earlier helper-parity comment remains valid despite the wording change. Please state what this PR actually protects and avoid an unsupported cross-client comparison. For example, describe the `gl` rule directly and, if useful, identify helper protection as separate pending work in #414. Verify any retained comparison against the code that will actually be present at merge time.

This is a correction to newly added security documentation, not an additional runtime vulnerability attributed to this PR. It should not independently block merge and does not justify expanding the implementation into the remote helper or making #414 a prerequisite.

## Needs maintainer decision

**Override-value semantics are a separate policy question, not a third finding.** At `http.rs:147`, this PR deliberately uses presence-based parsing: `0`, `false`, and an empty value all enable the override. My review on #414 requests affirmative-value parsing for the same variable. We should settle that shared expectation explicitly. Do not change this PR's parser merely to satisfy the proxy finding; its reproduction leaves the override unset, and its fix does not depend on a different value policy.

## Validation

`cargo fmt --check` passes. All 36 HTTP-client tests and all 372 `gl` tests pass. The controlled proxy reproduction nevertheless confirms the security finding, which is why routing-observation tests are necessary here.

PR #444, opened 2026-09-11
fix(icaptcha): return unsolved instead of blocking or overflowing

### LINKED ISSUE #345: icaptcha-client: interactive_prompt hangs on non-interactive stdin, and the solvers overflow i64 on service-supplied prompts
Two small defects in the same crate, both reachable from the iCaptcha origin, which the crate's own
header treats as untrusted. Filing together since they share a fix surface and a test file.

## 1. interactive_prompt blocks forever on open-but-silent stdin

The docstring at `lib.rs:353-355` states the contract: "Returns None when stdin isn't a usable
interactive source (e.g. an agent), so the caller surfaces a clear 'couldn't auto-solve' error
instead." The implementation does not honor it. `lib.rs:366-368` is a bare read with no TTY test and
no timeout:

```rust
match stdin().read_line(&mut line) { Ok(0) | Err(_) => None,
```

`Ok(0)` is EOF only. An open pipe with a live writer and no data returns neither, so it hangs. There is
no TTY facility anywhere in the workspace: `grep -rn "is-terminal\|atty\|IsTerminal"` over `crates/`
returns nothing.

Measured, three stdin shapes against the read step compiled verbatim:

```
stdin = /dev/null (EOF)        -> returned None, exit 0
stdin = open pipe, no data     -> exit 124 (timed out, blocked)
stdin = pipe with data         -> returned Some("42"), exit 0
```

`/dev/null`, the typical CI shape, is fine. The bad case is an agent, wrapper script, or supervisor
handing the child an inherited pipe it keeps open. Under `gl` this also pins a tokio blocking thread
(`crates/gl/src/http.rs:200-202`).

Reachable because nothing enforces that the service's reply is one of `SOLVABLE_TYPES` (`lib.rs:179`):
an unexpected `type`, or an unparseable prompt inside a known one, makes `solvers::solve` return `None`
and drops straight to this branch (`lib.rs:265-267`), with `solver: None` on the production path.

Fix: check for a TTY before prompting, and return `None` when there is not one, which is what the
docstring already promises.

## 2. i64 overflow in the deterministic solvers

Plain `i64` arithmetic with no `checked_`/`saturating_`/`wrapping_` anywhere in `solvers.rs`:
`:32-34` (`acc += n`, `acc -= n`), `:131-136` (`n[1] - n[0]`, `last + d`), `:180-183` (`isqrt_exact`,
where `cand * cand` overflows for `v` near `i64::MAX`), plus `:48`, `:53`, `:83`, `:104`, `:107`, and
`:159` (`v.abs()` at `i64::MIN`).

The workspace release profile (`Cargo.toml:59-61`) sets `lto` and `strip` but not `overflow-checks`, so
release wraps and debug panics. Both halves confirmed by running a standalone copy both ways:

```
DEBUG    PANIC  arithmetic  "What is 9223372036854775807 + 1?"
DEBUG    PANIC  sequence    "... 1, 4, 9223372036854775807, ?"
RELEASE  OK     arithmetic  -> Some("-9223372036854775808")
RELEASE  OK     algebra     -> Some("0")
```

Every panicking input is well-formed for the parser: each literal is in range and parses fine, and the
overflow happens during evaluation, so no malformed-input guard stands in the way. Source is
`challenge.prompt` (`lib.rs:234`), reaching `solvers::solve` at `lib.rs:265`.

Consequences are mild. In release a wrapped answer just fails grading; in debug the panic happens
inside `spawn_blocking` and `gl` reports "iCaptcha solver task panicked". Worth fixing because it is
cheap and the function's contract is already to return `None` when a prompt cannot be handled
(`solvers.rs:10-11`), so overflow should join that path via `checked_*`.

## Not filed: the PoW iteration cap

For the record, since it came up in the same pass and I checked it: `pow::solve` burning its full
iteration cap on an unsolvable difficulty is not worth an issue. The bound already exists
(`pow.rs:31`, `MAX_ITERS = 1 << 26`) with a comment explaining the choice, and an early exit at an
"impossible" threshold buys nothing, since a difficulty of 28 is equally unsolvable in practice and
sits below any such threshold. Measured worst case is 121s of one core in a release build, which is
bounded, client-side, and self-inflicted. The module header's "well under a second" framing is
optimistic and could use a correcting comment; that is the whole of it.

beardthelion:
## Summary

`obtain_proof` could block forever on `stdin().read_line` when stdin was an open, silent non-TTY (a pipe an orchestrator or agent harness holds open), and the deterministic iCaptcha solvers evaluated service-controlled prompts with unchecked `i64` arithmetic, which panics in debug builds and wraps to a wrong answer in release.

## Motivation & context

Closes #345. The interactive prompt is the documented last resort (returns `None` when stdin isn't a usable interactive source), but it never checked whether stdin was a terminal, so the fallback itself was the hang. Separately, every literal in a challenge prompt can individually parse as `i64` while the evaluation still overflows.

## Kind of change

- [x] Bug fix
- [ ] Feature
- [ ] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

- `icaptcha-client`: `interactive_prompt` checks `IsTerminal` before prompting and returns `None` on non-TTY stdin, so the caller surfaces the existing cannot-solve error.
- `icaptcha-client`: `solvers.rs` uses `checked_*` arithmetic throughout (add/sub chains, algebra coefficient/constant/denominator/numerator including `i64::MIN / -1`, sequence diffs, geometric multiply, `abs()` at `i64::MIN`, and the `isqrt_exact` candidate square). Overflow returns `None` (unsolvable) rather than panicking or submitting a wrapped answer.
- New end-to-end test (`tests/icaptcha_stdin_e2e.rs`) runs the real `obtain_proof` in a child process over real HTTP against a mocked iCaptcha service with stdin held open as a silent pipe, for both an unsolvable type and an overflowing arithmetic prompt; `/v1/answer` is pinned at zero calls so a wrapped answer would fail the test.
- Solver overflow regression tests cover arithmetic, algebra, and sequence overflow classes plus boundary-adjacent positive cases.

## How a reviewer can verify

```sh
cargo test -p icaptcha-client
```

Reverting the `IsTerminal` check makes the e2e block for the full 15s deadline; reverting `checked_add` to `+=` panics the overflow test.

## Before you request review

- [x] `cargo test -p icaptcha-client` passes locally
- [x] Checked existing PRs so this isn't a duplicate (#246 touches lib.rs hunks ending ~line 226; disjoint)

## Protocol & signing impact

None: client-side challenge solving only; no wire format, signature, or API change.

## Notes for reviewers

The `gl`-binary path through `send_signed` is not exercised end-to-end locally because `resolve_solver_url` deliberately refuses non-https advertised solver URLs (the SSRF guard); the child-process e2e drives the same `obtain_proof` `gl` calls via `spawn_blocking`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented interactive prompts from hanging when standard input is open but non-interactive.
  * Added safe handling for arithmetic, algebra, and sequence calculations that exceed supported numeric limits.
  * Overflowing or otherwise unsolvable challenges now fail cleanly instead of producing incorrect results or causing a crash.
  * Improved reliability when processing challenges through silent or redirected input streams.

### DIFF
--- a/crates/icaptcha-client/src/lib.rs
+++ b/crates/icaptcha-client/src/lib.rs
@@ -355,7 +355,13 @@ fn submit_answer(
 /// Returns `None` when stdin isn't a usable interactive source (e.g. an agent),
 /// so the caller surfaces a clear "couldn't auto-solve" error instead.
 fn interactive_prompt(challenge: &Challenge) -> Option<String> {
-    use std::io::{stderr, stdin, Write};
+    use std::io::{stderr, stdin, IsTerminal, Write};
+    // An open but silent stdin (a pipe an orchestrator holds open, an agent
+    // harness) has no one to answer: `read_line` would block forever. Only a
+    // terminal can be interactive.
+    if !stdin().is_terminal() {
+        return None;
+    }
     let mut err = stderr();
     let _ = writeln!(
         err,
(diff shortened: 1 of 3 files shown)


jatmn (approved):
LGMT

PR #456, opened 2026-09-15
fix(attest): bound did:key method-id length before base58 decode (#360)

### LINKED ISSUE #360: gitlawb-attest's did:key parser is missing the length bound gitlawb-core already added, so base58 decode is unbounded
Two copies of the same parser, one hardened and one not.

`crates/gitlawb-core/src/did.rs:83-97` carries the guard, added by commit `73bf132c` ("resolve the key
on the unproven insert path, and bound the decode"), with the reasoning in the comment:

```rust
// Refuse an oversized id before decoding. base58 decoding is quadratic
// in its input, and both callers of this function take the string from
// an untrusted request...
const MAX_METHOD_ID_LEN: usize = 64;
if self.method_id().len() > MAX_METHOD_ID_LEN {
    return Err(Error::InvalidDid("did:key method-specific id too long".to_string()));
}
```

`crates/gitlawb-attest/src/attestation.rs:196-206` is the second copy and has no such check:

```rust
let method_id = did.strip_prefix("did:key:")...?;
if !method_id.starts_with('z') { ... }
let (base, bytes) = multibase::decode(method_id).map_err(...)?;
```

The length checks at `:212-216` (`bytes.len() != 34`) run after the decode, so they cannot help.

## The cost is real

Measured against the pinned versions (bs58 0.5.1 via multibase 0.9.2, `Cargo.lock` lines 1978-1979 and
4698-4699), release build:

| method-id length | dense body | leading-zero body |
|---|---|---|
| 64 chars (a real did:key) | 2.2 µs | 920 ns |
| 1 KB | 95.8 µs | 19.0 µs |
| 10 KB | 9.64 ms | 1.16 ms |
| 100 KB | 941 ms | 131 ms |
| 1 MB | 95.7 s | 17.0 s |

Ten times the length gives roughly a hundred times the work, which is the quadratic accumulator loop
in `bs58-0.5.1/src/decode.rs:411-442`. `multibase-0.9.2/src/lib.rs:33-39` passes the input straight
through with no size check of its own.

The decode is reached from `verify_signature` (`attestation.rs:98-107`) after only a hex compare and
the type-grammar check, so nothing expensive gates it.

## Why this is low and not a DoS

`gitlawb-attest` has zero reverse dependencies. It appears once in `Cargo.lock` as a package and in no
other crate's dependency list; `gitlawb-node`, `gl`, `git-remote-gitlawb`, and `icaptcha-client` each
depend on `gitlawb-core` only. Every caller of `verifying_key_from_did_key` today is one of the crate's
own tests.

So there is no handler, no body-size question, and no attestation-count question. The crate is a
release-please-versioned publishable member (`release-please-config.json:15`, version 0.7.1), so the
exposure is to a future or external consumer that wires it to a request path, which is precisely the
scenario `did.rs` documents.

## Distinct from the small-order key work

#313, #315, and PR #314 concern accepting small-order Ed25519 keys, which is a property of
`VerifyingKey::from_bytes` and the X25519 conversion after a successful decode. This is decode cost
before any key material exists. Different function, different failure mode.

## Fix direction

Apply core's bound to the attest copy: reject a method-id longer than `MAX_METHOD_ID_LEN` before
calling `multibase::decode`. Better still, have `gitlawb-attest` call `gitlawb-core`'s parser rather
than keeping a second one, since the two have already drifted once and nothing would catch the next
drift. The crate does depend on `gitlawb-core` in dev-dependencies today, so promoting that is the
larger but more durable change.

## Wider point worth recording

Both findings filed against this crate today (this and the sibling on `verify_all`) are the same
shape: a rule that exists elsewhere in the workspace is absent here, and nothing notices because the
crate has no consumers exercising it. If `gitlawb-attest` is meant to stay a published library ahead
of its in-tree use, it is worth deciding what keeps it in step with `gitlawb-core`, because right now
nothing does.

beardthelion:
## Summary

`gitlawb-attest`'s `verifying_key_from_did_key` passed the method-specific id to `multibase::decode` with no length bound, and base58 decode is quadratic in input length. `gitlawb-core` added the same guard in 73bf132c; this ports it to the second copy of the parser.

## Motivation & context

Closes #360

## Kind of change

- [x] Bug fix
- [ ] Feature
- [x] Security fix
- [ ] Docs
- [ ] Tests / CI

## What changed

- `crates/gitlawb-attest/src/attestation.rs`: reject a method-id over 64 chars before decoding. A real ed25519 did:key method-id is a fixed 48 chars, so the bound is slack, not a behavior change.
- New test `verify_rejects_oversized_method_id_before_decoding` asserts the bound's own error fires, not a downstream decode or length failure, so it cannot pass on the unbounded code.

## How a reviewer can verify

```sh
cargo test -p gitlawb-attest
```

The new test goes red if the length check is removed: without it the oversized id decodes and fails later with a different `Error::Did` message (verified by deletion).

## Before you request review

- [x] `cargo test -p gitlawb-attest` passes locally
- [x] `cargo fmt --all` and `cargo clippy -p gitlawb-attest --all-targets -- -D warnings` are clean

## Protocol & signing impact

Notes: the bound only rejects method-ids no valid ed25519 did:key can produce; every previously valid signer string still verifies.

## Notes for reviewers

`attestation.rs` is also touched by open #314, #244, and #193; this change is confined to `verifying_key_from_did_key` plus one test, so a rebase should be mechanical whichever order they land in.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to reject oversized `did:key` identifiers during verification.
  * Oversized identifiers now return a clear validation error instead of being processed.


beardthelion:
> Anchor the verification key before signature verification. (`crates/gitlawb-attest/src/attestation.rs:107`)

Fair point, and consistent with the repo's own guidance that a verifying key should come from outside the artifact being checked. It is pre-existing on main though: this diff only adds the pre-decode length bound in `verifying_key_from_did_key` and does not touch `verify_signature` or `Registry::verify`. Plumbing an expected-signer parameter through those callers is a separate, larger change; I'd rather track it as its own issue than grow this PR.

